PR #66, opened 2013-08-19
Allow disabling drop targets from within drag event.

miracle2k:
http://bugs.qooxdoo.org/show_bug.cgi?id=7218


tobiasoberrauch:
Thank you

PR #67, opened 2013-09-06
Allow decorator properties to be repeated for multiple values for certain CSS properties

sabberworm:
As I have I’ve previously [mentioned](http://qooxdoo.678.n2.nabble.com/Qooxdoo-3-0-decorator-questions-td7584521.html), I’ve patched MLinearBackgroundGradient and MLinearBackgroundGradient to allow for multiple background images and MBoxShadow to allow for multiple shadow definitions.

I’ve run the framework test suite of qx.test.ui namespace with the changes with no failures (but some test runs did have 133 failures but I did not get consistent results) though I’ve not added any new test cases for the new features.

### DIFF
--- a/framework/source/class/qx/lang/Array.js
+++ b/framework/source/class/qx/lang/Array.js
@@ -453,7 +453,7 @@ qx.Bootstrap.define("qx.lang.Array",
     /**
      * Extends an array up to the given length by repeating the elements already present.
      * @param arr {Array} Incoming array. Has to contain at least one element.
-     * @param arr {Integer} Desired length. Must be greater than or equal to the the length of arr.
+     * @param to {Integer} Desired length. Must be greater than or equal to the the length of arr.
      */
     prolong : function(arr, to)
     {
--- a/framework/source/class/qx/lang/Array.js
+++ b/framework/source/class/qx/lang/Array.js
@@ -450,6 +450,18 @@ qx.Bootstrap.define("qx.lang.Array",
       return result === undefined ? null : result;
     },
 
+    /**
+     * Extends an array up to the given length by repeating the elements already present.
+     * @param arr {Array} Incoming array. Has to contain at least one element.
+     * @param arr {Integer} Desired length. Must be greater than or equal to the the length of arr.
+     */
+    prolong : function(arr, to)
+    {
+      var initial = arr.length;
+      while(arr.length < to) {
+        arr.push(arr[arr.length%initial]);
+      }
+    },
 
     /**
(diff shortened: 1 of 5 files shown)


tobiasoberrauch:
Hi @sabberworm 
Sorry for the late answer. I just saw your PR and want to ask you if you can update a few things
- It's not common to define methods like prolong in qx.lang.Array. Please define a local util method or class.
- It's not clear why you removed all the checks in MBoxShadow.js, MBackgroundImage.js and MLinearBackgroundGradient.js
- It would be nice if you write some documentation to clear why you solved this problem like this

I hope you are still the will to change this.
Thanks for supporting qooxdoo :+1: 

Cheers Tobias

sabberworm:
Hi. I’m not anymore with the company that I needed this for. However, I saw that they [updated this](https://github.com/Vertec/qooxdoo/tree/decorators-for-multiple-css-properties-4_0_1) to work with qooxdoo 4.0.1.

I removed the checks because I did not know how to express the following constraints:
* [1](https://github.com/qooxdoo/qooxdoo/pull/67/files#diff-b2182ed1f61b1b4b6cd5f8ab3590b59cL30): The value should be either be a string or an array of strings.
* [2](https://github.com/qooxdoo/qooxdoo/pull/67/files#diff-b2182ed1f61b1b4b6cd5f8ab3590b59cL39): The value should be either one of `"repeat", "repeat-x", "repeat-y", "no-repeat", "scale"` or an array of strings (with each one being an item of the aforementioned set).
* [3](https://github.com/qooxdoo/qooxdoo/pull/67/files#diff-3126db7e9e08e4a615806ea5a0ec4c9fL38): An integer or an array of integers.
* [4](https://github.com/qooxdoo/qooxdoo/pull/67/files#diff-3126db7e9e08e4a615806ea5a0ec4c9fL70): A color or an array of colors.
* [5](https://github.com/qooxdoo/qooxdoo/pull/67/files#diff-3126db7e9e08e4a615806ea5a0ec4c9fL78): A boolean or an array of booleans.

…and so on.

Also, I’m not sure checks are particularly useful here. If a new keyword (lets say as a value for `background-repeat`) gets to be valid CSS, I should be able to use it without having to wait for a new version of qooxdoo to come out. Someone using these decorators will already have to know how the corresponding CSS property works. With a decorator, detecting failure is quite simple: If the usage was wrong, the output won’t look as intended, I don’t think an additional error during compilation or in the browser console is going to do much good.

wittemann:
Thanks for the feedback. So I guess you don't plan to update the pull requests so we can close this one and wait for another one based on the fork you linked.

PR #102, opened 2014-10-31
Initial commit. Assert for interfaces homogenized.

stefanand:
There is no assert function checking types of an array, no function argument assert and there is no multi-condition assert f.i. with chaining:

this.assertBoolean(<var>, <condition>);

Where condition could be the previous assert result, where assert is not fired and postponed till all conditions have been fulfilled such as:

@param date {Date|null} Here the parameter can have two states...an actual date and null

I have not been able to test the proposed changes.

### DIFF
--- a/framework/source/class/qx/bom/request/IRequest.js
+++ b/framework/source/class/qx/bom/request/IRequest.js
@@ -89,7 +89,13 @@ qx.Interface.define("qx.bom.request.IRequest",
      * @param async {Boolean?true}
      *  Whether or not to perform the operation asynchronously.
      */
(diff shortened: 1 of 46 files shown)


tobiasoberrauch:
Hi Stefan,

thank you for helping us. I had a look on it and found some errors.
I fixed many things (https://github.com/qooxdoo/qooxdoo/commit/d73f27ee1f8c0b30914ab0e5281f117c43a74713) in a branch (https://github.com/qooxdoo/qooxdoo/tree/pr/102). But there are still some errors.
Would you please have a look?

Cheers Tobias

PR #114, opened 2014-12-16
[Bug 8824] Add support for dimensions provided in SVG files

cajus:
This pull request allows us to use SVG images without the need to tell qx about the desired size. It extends the resource scanning to read the dimensions from the SVG if applicable. Only "px" dimensions are supported.


rsternagel:
Looks good to me. Nice that you've collaborated, guys!

PR #116, opened 2015-01-07
[Bug 8824] Add support for dimensions provided in SVG files  - with different width/height units

ofilippov:
Add support for dimensions provided in SVG files
Support different units for height and width (px, in, cm, mm, pt, pc)

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -174,6 +174,10 @@ qx.Class.define("qx.ui.basic.Image",
   },
 
 
+  statics:
+  {
+    PLACEHOLDER_IMAGE: "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"
+  },
 
   /*
   *****************************************************************************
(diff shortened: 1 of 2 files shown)


cajus:
Cool! I was a bit lazy, because our SVG only were in px ;-) Didn't you need to patch the Image class to get proper sizes for SVG images?

ofilippov:
@cajus I tested it a little (on checkbox component with my svg icons) and it works good - css width and height are specified correctly.

cajus:
Hmm. Does not work here. The images are of zero width. If I add the explicit sizing it works. Strange.

ofilippov:
@cajus do you have property "scale" enabled for your image?

ofilippov:
I tried this file: https://www.iconfinder.com/icons/358219/download/svg/32
It works good for me - in both modes (scaled - when img node is used and non-scaled - when div)

cajus:
The problem is that you don't want to set scale for all images. If you i.e. use the SVG in a qx.ui.form.Button, you've to set the "icon" childcontrol to "scale". That's why I've edited the Images class. So - yes, it works if you use scale, but it's not practical at all.

ofilippov:
@cajus could you please share with me your example which doesn't work? Thanks!

cajus:
Every SVG I try that comes right from Illustrator isn't recognized by your code. I'm allways getting

    Unable to get image info from file: /home/cajus/Projekte/svgtest/source/resource/svgtest/test.svg

when generating the source. When I use my code, it is detected correctly. Maybe I'll just add your DPI conversation method there.

If I go past this - without modifying Image.js, all SVGs I try with the basic demo generated by create_application.py just with a replaced image (test.png -> test.svg) produce a zero size Image in the button. If I patch Image.js with the stuff from my pull request it works. I'm using Google Chrome for the tests.

cajus:
You seem to have mixed in your high resolution patch in this pull request.

cajus:
Ok. Got a step further. After patching my code with your unit conversation function, it seems to work without a patched Image.js. That's interesting, because I don't see any difference in the generated data. What so ever. I've just modified my pull request to do the unit detection. So it works with your provided SVG and our designers SVGs.

ofilippov:
@cajus I'm going to close my pull-request cause yours contains unit conversion part now (and this one is corrupted by another commit anyway). Thanks!

PR #119, opened 2015-02-05
Added null check to Spinner.js which fixes the following bug:

mbgonicus:
The property numberFormat of qx.ui.form.Spinner is nullable. If you do something like

spinner.setNumberFormat(new qx.util.format.NumberFormat());
spinner.setNumberFormat(null);

it will produce an error in the shape of:

/framework/source/class/qx/ui/form/Spinner.js:502 Uncaught TypeError: Cannot read property 'addListener' of null

because it tries

this.getNumberFormat().addListener(...);

although this.getNumberFormat() is null.

Fixed it with an "if (this.getNumberFormat() !== null)".

### DIFF
--- a/framework/source/class/qx/ui/form/Spinner.js
+++ b/framework/source/class/qx/ui/form/Spinner.js
@@ -499,8 +499,10 @@ qx.Class.define("qx.ui.form.Spinner",
       var textfield = this.getChildControl("textfield");
       textfield.setFilter(this._getFilterRegExp());
 
-      this.getNumberFormat().addListener("changeNumberFormat",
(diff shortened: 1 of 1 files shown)


wittemann:
Thanks for the pull request. We usually have for all commits a bug report in our bugzilla. Could you open a bug report for that change?

mbgonicus:
Shure, it's bug #8917: http://bugzilla.qooxdoo.org/show_bug.cgi?id=8917

wittemann:
Thanks. :)

tobiasoberrauch:
Done. Thanks for your PR @mbgonicus

PR #120, opened 2015-02-11
Manually remove $$widget in case of qx.ui.basic.Image

cajus:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=8700


wittemann:
Thanks. :+1:

PR #121, opened 2015-02-13
qx.ui.table.Table: handle contextmenu event via _onTapPane event handler to get selection of the pointed row again back.

level420:
See bug http://bugzilla.qooxdoo.org/show_bug.cgi?id=8920
Compensates changes made to solve http://bugzilla.qooxdoo.org/show_bug.cgi?id=8873


level420:
Thank you Martin.

PR #122, opened 2015-02-18
[BUG #8952] FSM does not handle constant TERMINATE

mbgonicus:
fixes http://bugs.qooxdoo.org/show_bug.cgi?id=8952

For that purpose I added the private flag _terminated and for convenience the public function isTerminated().

### DIFF
--- a/framework/source/class/qx/util/fsm/FiniteStateMachine.js
+++ b/framework/source/class/qx/util/fsm/FiniteStateMachine.js
@@ -244,7 +244,17 @@ qx.Class.define("qx.util.fsm.FiniteStateMachine",
     __groupToFriendly            : null,
     __friendlyToGroups           : null,
     __bEventProcessingInProgress : false,
+    __terminated                 : true,
 
+    /**
+     * Checks whether the finite state machine is terminated or not.
+     *
+     * @return {Boolean} If the finite state machine is terminated.
+     */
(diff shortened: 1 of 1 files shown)


mbgonicus:
[BUG #8953] terminate event for FSM

I added this feature, but it requires the first fix (for #8952), so it is inside this PR.

mbgonicus:
check. Sorry, didn't see it

mbgonicus:
As far as the convention goes, no problem. I'll change that in the next commit.

I get your point to automatically restart the fsm and not fatal error the fsm. On the other hand that means one can use the fsm, although it is terminated. I don't know if that's expected behaviour. I expect a terminated fsm to not run unless the user starts it.

I think, that's a decision you must make. However, the result must be documented, so a user knows what to expect.

derrell:
See my line note regarding line 745 in https://github.com/mbgonicus/qooxdoo/commit/820477fa2bc5ac11ac69eb71f43e79dac2ef4c74

mbgonicus:
Ok, changed __terminated to __bTerminated and changed the fatal error to a debug message. Latest commit is 2b20d5a, see also https://github.com/qooxdoo/qooxdoo/pull/122/files

I'm sorry, but I cannot provide you with unit tests - workload to high.

PR #128, opened 2015-03-26
[Bug #7522] Added support for different group item sizes

cajus:
In some usecases ordinary list items are relative huge (i.e. when containing a
specific icon, etc.) and the group header only consists of a headline that
groups a couple of list items.

Currently it is not possible to set the group height to a different value from
defaultItemSize - although the infrastructure that is used supports different
heights for layer items.

Sadly there's no way to easily inherit from qx.ui.list.List because the parts
that need access are private.

Please find a patch that adds a few lines of code to support a
"groupItemHeight" property and adjusts the row config item sizes if it's
specified.

See: http://bugzilla.qooxdoo.org/show_bug.cgi?id=7522

### DIFF
--- a/framework/source/class/qx/ui/list/List.js
+++ b/framework/source/class/qx/ui/list/List.js
@@ -145,6 +145,17 @@ qx.Class.define("qx.ui.list.List",
     },
 
 
+    /** Group item height */
+    groupItemHeight :
+    {
+      check : "Integer",
+      init : null,
+      nullable : true,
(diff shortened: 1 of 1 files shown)


wittemann:
Great, thanks. Will check that soon. :)

wittemann:
Looks like Daniel fixed the bug with the use of the patch file. Anyway, it's fixed now. :)

PR #132, opened 2015-04-14
[BUG #9056] Call directly _applyValue() if value is the same than before

woprandi:
Hi,
This is a fix for the bug reported here : http://bugs.qooxdoo.org/show_bug.cgi?id=9056
I added a new unit test to the spinner.


danielwagner:
Thanks!

PR #133, opened 2015-04-23
[BUG #9105] Fix text selection after moving a Window

zpo:
I am not sure my proposal is correct. Nevertheless I encountered a problem when moving the object of qx.ui.window.Window class. After move action a user is not able to select the text within form text widgets (I mean select by mouse). In my case the "roll" event handling was not removed.

### DIFF
--- a/framework/source/class/qx/ui/core/MMovable.js
+++ b/framework/source/class/qx/ui/core/MMovable.js
@@ -285,7 +285,7 @@ qx.Mixin.define("qx.ui.core.MMovable",
      */
     _onMovePointerUp : function(e)
     {
-      if (this.hasListener("roll", this._onMoveRoll, this)) {
+      if (this.hasListener("roll")) {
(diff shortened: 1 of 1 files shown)


wittemann:
Thanks for the pull request. Could you maybe open up a bug report at http://bugzilla.qooxdoo.org/ and describe the problem and how to reproduce? That would help a lot to verify your fix is the proper way of fixing it.

zpo:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9105

I created a short example in the playground. I hope it will help to find out what is the issue :)

rsternagel:
I'm closing this pull request because the solution has no effect for me (when using Firefox) but feel free to add more context/details/ideas within the bug report!

ecker:
Hi Zbigniew, your PR was correct, as it fixed a wrong API signature to hasListener().

Applied, thanks again!

PR #134, opened 2015-05-12
[BUG #9135] '+' is allowed in an email address

woprandi:
This PR for the bug : http://bugs.qooxdoo.org/show_bug.cgi?id=9135

### DIFF
--- a/framework/source/class/qx/test/util/Validate.js
+++ b/framework/source/class/qx/test/util/Validate.js
@@ -57,6 +57,9 @@ qx.Class.define("qx.test.util.Validate",
 
       //Valid since new domain extensions
       qx.util.Validate.email()("foo@bar.qooxdoo");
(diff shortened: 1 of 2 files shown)


wittemann:
Thanks for that. It would be great it would have a unit test to verify the change. This could be added here: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/util/Validate.js#L43

woprandi:
Yes, I will add a test in a new commit

wittemann:
Great, this will get the PR in quickly.

woprandi:
Added : https://github.com/woprandi/qooxdoo/commit/a98c56f352ca2f1825a9f621c9a538a56ed4b67c

wittemann:
Thanks. :)

PR #135, opened 2015-05-12
Bug 9119 - Web fonts integration does not allow to use multiple files…

vvandens:
… for the same family


wittemann:
Thanks, we will check it soon and give you feedback.

oetiker:
I'd like to merge that ... could you add some example usage in the javadoc ? 

Looking at the interface for google fonts, they come in various weights. Not just bold or not. How would this be handled ?

vvandens:
Sure, I can update the doc. You talk about the jsdoc (not javadoc), right ? BTW, do you think that the example I gave in the [bug](http://bugzilla.qooxdoo.org/show_bug.cgi?id=9119) would be enough ? And I'd put it in the `qx.bom.webfont.Manager` class ?

BTW, you are right. If we want to go beyond the bold / italic feature (I didn't know about the various weights), then it might be more versatile to implement `fontWeight` and `fontStyle` attributes that directly translate to their css peers. Then the example I gave would be re-written to :
```javascript
        sources: [
            {
              family: "Sansation",
              source: [
                "fonts/Sansation-Regular.ttf"
              ]
            },
            {
              family: "Sansation",
              fontWeight: "bold",
              source: [
                "fonts/Sansation-Bold.ttf",
              ]
            },
            {
              family: "Sansation",
              fontStyle: "italic",
              source: [
                "fonts/Sansation-Italic.ttf",
              ]
            },
            {
              family: "Sansation",
              fontWeight: "bold",
              fontStyle: "italic",
              source: [
                "fonts/Sansation-BoldItalic.ttf",
              ]
            }
          ]
        }
```

It should'nt be a big deal and I can take care of that. What do you think ?

oetiker:
yes jsdoc :) and integrating it into the class, yes. The idea is to have as much good documentation in the api viewer, since this is the simplest way to keep documentation up to date ... when it is close to the source.

fontWeight and fontStyle make sense for me as they keep with the official structure.

vvandens:
Ok. I go for it.
Then you reject the PR and I submit a new one.

oetiker:
ok :)

PR #137, opened 2015-06-11
[BUG #9149] Improve SVG handling in IE

axjpc:
PR for: http://bugzilla.qooxdoo.org/show_bug.cgi?id=9149

I had to replace `new Image()` with `document.createElement('img')`.
Using the former, the element wouldn't have the width/height set even
after appending it to the DOM.
The rest is explained in the bug ticket.

Let me know if there is something missing or if improvements are
necessary.

### DIFF
--- a/framework/source/class/qx/io/ImageLoader.js
+++ b/framework/source/class/qx/io/ImageLoader.js
@@ -218,7 +218,7 @@ qx.Bootstrap.define("qx.io.ImageLoader",
         }
 
         // Create image element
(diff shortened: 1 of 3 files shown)


wittemann:
Thanks for the pull request. It would be great if it had a unit test as well.

nanexo:
I expanded the ImageLoader unit test to cover the loading of SVGs.

wittemann:
Great, thanks! Will check that soon.

PR #139, opened 2015-06-24
[BUG #7715] Allow decorator properties to be repeated for multiple values for certain CSS properties

axjpc:
This is an updated PR based on https://github.com/qooxdoo/qooxdoo/pull/67
Bugreport here: http://bugzilla.qooxdoo.org/show_bug.cgi?id=7715
- I removed the `prolong` method defined in `qx.lang.Array` and defined a helper method in the `Decorator` class. I would have liked to put it in a helper class, but I'm not sure where such a class would belong.
- I left the checks out, because the reasoning [here](https://github.com/qooxdoo/qooxdoo/pull/67#issuecomment-75430195) made sense to me. If the checks are an absolute must, should I define custom check functions for every property to cover the constraints explained in the linked comment?
- I added some comments. If it's still lacking, let me know where and I'll add further documentation.

### DIFF
--- a/framework/source/class/qx/ui/decoration/Decorator.js
+++ b/framework/source/class/qx/ui/decoration/Decorator.js
@@ -111,6 +111,13 @@ qx.Class.define("qx.ui.decoration.Decorator", {
           this[name](styles);
         }
       }
+      for(var name in styles)
+      {
+        if(qx.lang.Type.isArray(styles[name])) {
+          styles[name] = styles[name].join(', ');
+        }
+      }
+
 
       this.__initialized = true;
       return styles;
--- a/framework/source/class/qx/ui/decoration/Decorator.js
+++ b/framework/source/class/qx/ui/decoration/Decorator.js
@@ -159,6 +159,19 @@ qx.Class.define("qx.ui.decoration.Decorator", {
     // overridden
     _isInitialized: function() {
       return this.__initialized;
+    },
+
+
+    /**
+    * Extends an array up to the given length by repeating the elements already present.
+    * @param arr {Array} Incoming array. Has to contain at least one element.
+    * @param to {Integer} Desired length. Must be greater than or equal to the the length of arr.
+    */
+    _prolongArray: function(array, to) {
+      var initial = array.length;
+      while(array.length < to) {
+        array.push(array[array.length%initial]);
+      }
     }
   }
 });
--- a/framework/source/class/qx/ui/decoration/Decorator.js
+++ b/framework/source/class/qx/ui/decoration/Decorator.js
@@ -161,13 +161,40 @@ qx.Class.define("qx.ui.decoration.Decorator", {
       return this.__initialized;
     },
 
+    /**
+    * Ensures that every propertyValue specified in propertyNames is an array.
+    * The value arrays are prolonged to match in length.
+    * @param arr {Array} Array containing the propertyNames.
+    * @returns {Array} Array containing the prolonged value arrays.
+    */
+    _getProlongedPropertyValueArrays: function(propertyNames) {
+      // transform non-array values to an array containing that value
+      var propertyValues = propertyNames.map(function(propName) {
+        var value = this.get(propName);
+        if(!qx.lang.Type.isArray(value)) {
+          value = [value];
+        }
+        return value;
+      }, this);
+
+      // Because it's possible to set multiple values for a property there's 
+      // a chance that not all properties have the same amount of values set.
+      // Prolong the value arrays by repeating existing values until all
+      // arrays match in length.
+      var items = Math.max.apply(Math, propertyValues.map(function(prop) { return prop.length; }));
+      for(var i = 0; i < propertyValues.length; i++) {
+        this.__prolongArray(propertyValues[i], items);
+      }
+
+      return propertyValues;
+    },
 
     /**
     * Extends an array up to the given length by repeating the elements already present.
     * @param arr {Array} Incoming array. Has to contain at least one element.
     * @param to {Integer} Desired length. Must be greater than or equal to the the length of arr.
     */
-    _prolongArray: function(array, to) {
+    __prolongArray: function(array, to) {
       var initial = array.length;
       while(array.length < to) {
         array.push(array[array.length%initial]);
--- a/framework/source/class/qx/ui/decoration/Decorator.js
+++ b/framework/source/class/qx/ui/decoration/Decorator.js
@@ -163,11 +163,11 @@ qx.Class.define("qx.ui.decoration.Decorator", {
 
     /**
     * Ensures that every propertyValue specified in propertyNames is an array.
-    * The value arrays are prolonged to match in length.
+    * The value arrays are extended and repeated to match in length.
     * @param arr {Array} Array containing the propertyNames.
-    * @returns {Array} Array containing the prolonged value arrays.
+    * @returns {Array} Array containing the extended value arrays.
     */
-    _getProlongedPropertyValueArrays: function(propertyNames) {
+    _getExtendedPropertyValueArrays: function(propertyNames) {
       // transform non-array values to an array containing that value
       var propertyValues = propertyNames.map(function(propName) {
         var value = this.get(propName);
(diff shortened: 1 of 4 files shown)


v3ss0n:
I can't wait this to be merged. I needed that and patched decorator for doing it.

oetiker:
This seems to be a pretty hefty change ... some of the coding is pretty WET. I would like to merge this, but in its present state I have a hard time seeing through all the side effects ... you seem to completely remove the canvas code for example.

nanexo:
Any pointers on how to improve the code?
Regarding the canvas code: It wasn't removed. The code moved to the __styleLinearBackgroundGradientWithCanvas method.

framework/source/class/qx/ui/decoration/MBackgroundImage.js
```diff
@@ -90,48 +110,85 @@ qx.Mixin.define("qx.ui.decoration.MBackgroundImage",
...
+      if(!qx.lang.Type.isArray(images)) images = [images];
+      var repeats = this.getBackgroundRepeat();
+      if(!qx.lang.Type.isArray(repeats)) repeats = [repeats];
+      var tops = this.getBackgroundPositionY();
+      if(!qx.lang.Type.isArray(tops)) tops = [tops];
+      var lefts = this.getBackgroundPositionX();
+      if(!qx.lang.Type.isArray(lefts)) lefts = [lefts];
+      var origins = this.getBackgroundOrigin();
+      if(!qx.lang.Type.isArray(origins)) origins = [origins];
+
```
oetiker: If you want to make sure you get an array back ... you should have a function that takes care of this and not just keep copy pasting the code ... the only real difference here is the property you are looking at, so I would expect to see something where you have a list of properties you are interested in and some code that takes that list as input. DRY

oetiker:
can you provide a sample on how to use the new interface ... the whole approach of having all the properties in an array and making sure that they align seems rather fragile to me.

nanexo:
In qooxdoo you can't define multiple BoxShadows/BackgroundImages/Gradients. We are stuck with something like this:
```javascript
'some-decorator': {
	style: {
		shadowBlurRadius: 0,
		shadowSpreadRadius: 0,
		shadowVerticalLength: 1,
		shadowHorizontalLength: 0,
		shadowColor: 'rgba(0, 0, 0, 0.14)',
		inset: true
	}
}
```
With the changes made here, the above still works. Additionally we are able to define multiple values for a property:
```javascript
'some-decorator': {
	style: {
		shadowBlurRadius: 0,
		shadowSpreadRadius: 0,
		shadowVerticalLength: 1,
		shadowHorizontalLength: 0,
		shadowColor: ['rgba(0, 0, 0, 0.14)', 'rgba(255, 255, 255, 0.45)'],
		inset: [true, false]
	}
}
```
The non-arrays values will be put in arrays. All needed value arrays get extended so that all match in size. The values will be repeated. For BoxShadows this is more a convenience than anything else. The above is the same as the following:
```javascript
'some-decorator': {
	style: {
		shadowBlurRadius: [0, 0],
		shadowSpreadRadius: [0, 0],
		shadowVerticalLength: [1, 1],
		shadowHorizontalLength: [0, 0],
		shadowColor: ['rgba(0, 0, 0, 0.14)', 'rgba(255, 255, 255, 0.45)'],
		inset: [true, false]
	}
}
```

BackgroundImage and LinearGradient are a little bit different.
Previously the implementations set the css properties: `background-image`, `background-size` etc. These properties do the same. e.g.:
```
background-image: url('foo.png'), url('bar.png');
background-position: bottom right;
```
places both images at the bottom right.

The problem with that and multiple images/gradients is, that both write to `background-image`, which means all background related properties apply. You have to be careful how you write your decorators, otherwise the css properties might get applied to `background-image`s you didn't want to.
To prevent this BackgroundImage and LinearBackgroundGradient now write to `background` with all the information condensed into one. To provide a similar interface as the css properties, the values have to be repeated manually, because we aren't setting the `background-*` properties anymore.

framework/source/class/qx/ui/decoration/Decorator.js
```diff
@@ -152,6 +159,46 @@ qx.Class.define("qx.ui.decoration.Decorator", {
...
+      var propertyValues = propertyNames.map(function(propName) {
+        var value = this.get(propName);
+        if(!qx.lang.Type.isArray(value)) {
+          value = [value];
+        }
+        return value;
+      }, this);
+
+      // Because it's possible to set multiple values for a property there's 
+      // a chance that not all properties have the same amount of values set.
```
oetiker: amount -> number

framework/source/class/qx/ui/decoration/Decorator.js
```diff
@@ -152,6 +159,46 @@ qx.Class.define("qx.ui.decoration.Decorator", {
...
       return this.__initialized;
+    },
+
+    /**
+    * Ensures that every propertyValue specified in propertyNames is an array.
+    * The value arrays are prolonged to match in length.
+    * @param arr {Array} Array containing the propertyNames.
+    * @returns {Array} Array containing the prolonged value arrays.
+    */
+    _getProlongedPropertyValueArrays: function(propertyNames) {
```
oetiker: Prolong -> Extend

oetiker:
your array extension repeats the values already in the array ... I can see why you are doing that when starting with a single element. but what is the use case for repeating the values if there is more than one value ... would't it make more sense to just copy the last entry ?

nanexo:
It's done this way to stay spec compliant, see example 1: https://www.w3.org/TR/css3-background/#layering

oetiker:
since you are using the background property I guess this here  https://www.w3.org/TR/css3-background/#background applies ... reading this, I think it is the initial value which gets substituted if a values are missing. And also if values are missing 'at the end of the string'  you can leave the replacement to the browser ..

nanexo:
Yes for the background property the missing values are filled in with the initial value, which is different from when you set background-repeat for example.
The goal was for qooxdoo to have a similiar interface.
Doing something like:
```javascript
'some-decorator': {
    style: {
        backgroundImage: ['foo.png', 'bar.png'],
        backgroundRepeat: 'no-repeat'
    }
}
```
I'd expect this to be the css equivalent of
```
background-image: url(foo.png), url(bar.png);
background-repeat: no-repeat
```
Which means both images won't get repeated, because the background-repeat value will be repeated as per my earlier comment.
If the values don't get repeated manually, we'd end up with:
```
background: url(foo.png) no-repeat, url(bar.png)
```
The second image will be repeated as the inital value is repeat.
Repeating the values ourselves leaves us with
```
background: url(foo.png) no-repeat, url(bar.png) no-repeat;
```
which produces the desired result.

Maybe this isn't as important. I'd be more than happy to remove the repeating as it definitely adds complexity.

framework/source/class/qx/ui/decoration/Decorator.js
```diff
@@ -152,6 +159,46 @@ qx.Class.define("qx.ui.decoration.Decorator", {
...
+
+    /**
+    * Extends an array up to the given length by repeating the elements already present.
+    * @param arr {Array} Incoming array. Has to contain at least one element.
+    * @param to {Integer} Desired length. Must be greater than or equal to the the length of arr.
+    */
+    __extendArray: function(array, to) {
+      var initial = array.length;
+      while(array.length < to) {
+        array.push(array[array.length%initial]);
```
oetiker: as I read the standard, you should be doing this:

    array.push(array[0]);

oetiker:
It is the initial value of the property not the initial value of the list .... ok ( have not tried it).

framework/source/class/qx/ui/decoration/Decorator.js
nanexo: Here's the relevant excerpt: "If a property doesn't have enough comma-separated values to match the number of layers, the UA must calculate its used value by repeating **the list of values until there are enough**"
Quick codepen for verification: http://codepen.io/anon/pen/grYxQd

nanexo:
Yes, the initial value of the property will be applied.
Another codepen: http://codepen.io/anon/pen/oxveQG

oetiker:
ok :-) thanks for your patience!

oetiker:
except for one thing ... please create another pull request to add some example code into the javascriptdoc ... showing how to use your new feature

nanexo:
Thank you for your time and will do.

PR #140, opened 2015-07-31
[Bug #8915] - Dragging over/onto an iFrame no longer works

qooalt:
Fix for the Bug #8915 This commit fixes in my case the odd behaviour of "blocking" the dragover and drop events when crossing an iframe while dragging elements. See links below to the playground.

3.5.1 (ok)
http://tinyurl.com/og6l97e

5.0 (wrong)
http://tinyurl.com/nw3ur95

See:  https://github.com/qooxdoo/qooxdoo/issues/8915

### DIFF
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -460,6 +460,7 @@ qx.Class.define("qx.event.handler.DragDrop",
     clearSession : function()
     {
       // Deregister from root events
+      this.__manager.removeListener(this.__root, "track", this._onPointermove, this);
       this.__manager.removeListener(this.__root, "pointermove", this._onPointermove, this);
       this.__manager.removeListener(this.__root, "pointerup", this._onPointerup, this, true);
 
@@ -575,6 +576,7 @@ qx.Class.define("qx.event.handler.DragDrop",
 
         this.__manager.addListener(this.__root, "pointermove", this._onPointermove, this);
         this.__manager.addListener(this.__root, "pointerup", this._onPointerup, this, true);
+        this.__manager.addListener(this.__root, "track", this._onPointermove, this);
       }
     },
 
@@ -613,7 +615,9 @@ qx.Class.define("qx.event.handler.DragDrop",
       }
 
       // find current hovered droppable
-      var el = e.getTarget();
+      // TODO: Review this (before the fix it was "e.getTarget()", but I feel this to be better)
+      var doc = this.__manager.getWindow().document;
+      var el = doc.elementFromPoint(e.getDocumentLeft(), e.getDocumentTop());
       var cursor = this.getCursor();
       if (!cursor) {
         cursor = qx.ui.core.DragDropCursor.getInstance();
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -575,6 +576,7 @@ qx.Class.define("qx.event.handler.DragDrop",
 
         this.__manager.addListener(this.__root, "pointermove", this._onPointermove, this);
         this.__manager.addListener(this.__root, "pointerup", this._onPointerup, this, true);
+        this.__manager.addListener(this.__root, "track", this._onPointermove, this);
       }
     },
```
oetiker: please elaborate on adding "track"

framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -665,6 +669,10 @@ qx.Class.define("qx.event.handler.DragDrop",
      */
     _getDelta : function(e)
     {
+      if (e.getDelta) {
+        return e.getDelta();
+      }
+      
       if (!this.__startConfig) {
```
oetiker: who supports getDelta ? this also seems to be unrelated to the iframe fix

framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -613,7 +615,9 @@ qx.Class.define("qx.event.handler.DragDrop",
       }
 
       // find current hovered droppable
-      var el = e.getTarget();
+      // TODO: Review this (before the fix it was "e.getTarget()", but I feel this to be better)
+      var doc = this.__manager.getWindow().document;
+      var el = doc.elementFromPoint(e.getDocumentLeft(), e.getDocumentTop());
       var cursor = this.getCursor();
```
oetiker: does it only not work when the iframe content is coming from another domain ... that would explain why your solution works as it does not look at the content of the iframe

oetiker:
ping ...

level420:
@qooalt Please have a look into this PR again. There are some open questions.

level420:
As @qooalt does not react: I think we should close the PR

oetiker:
is the bug not reproducible ?

level420:
I don't know, have currently no use case for that.

level420:
closing due to inactivity. feel free to reopen. The bugreport wil remain open.

level420:
@oetiker why did you re-open this PR?

oetiker:
I re-opened it because we ran into this but in osparc

level420:
@oetiker does this PR solve that bug? the corresponding bug is here https://github.com/qooxdoo/qooxdoo/issues/8915 and still open

framework/source/class/qx/event/handler/DragDrop.js
level420: Its http://www.qooxdoo.org/devel/api/#qx.event.type.Track where track is derived from http://www.qooxdoo.org/devel/api/#qx.event.type.Pointer only adding one additional method ```getDelta``` which is documented as:
Returns a map with the calculated delta coordinates and axis, relative to the position on trackstart event.

framework/source/class/qx/event/handler/DragDrop.js
level420: this is supporte by the http://www.qooxdoo.org/devel/api/#qx.event.type.Track event, which was registerd above.

oetiker:
it seems that @qooalt is a throwaway account created only to generate this PR ... hence no response.

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo/pull/9838

PR #141, opened 2015-08-08
[BUG #9188] Groupbox setLegend is now a property

rlexa:
[BUG #9188] Creating a pull request following (Andreas Ecker)'s request: Changed Groupox's "legend" from set/get methods to a real binded property.


ecker:
Resolved acc. to associated bug report.

PR #142, opened 2015-08-12
Fix target detection and usage

cajus:
In certain cases it is not possible to drag'n drop without running into errors. Looks like the mixin does stuff with an Element instead of a Widget.


cajus:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9192

oetiker:
to quote martin ... The patch looks a bit too easy. What if a widget will be returned? The
'getWidgetByElement' can not handle widgets as input so there should be a check
if its a widget or DOM element.

cajus:
At the time of creating the PR I did not find any location where the method gets called with a DOM element. Will take a deeper look at it.

oetiker:
resolved via #200

PR #144, opened 2015-08-21
MProperty.set always return instance

woprandi:
This fix the bugs http://bugs.qooxdoo.org/show_bug.cgi?id=9199

now MProperty.set will always return current instance (for chaining).


ecker:
@William, closing this PR: resolved bug #9199 by correcting documentation, created general task for consistent return values as bug #9202. Thanks again.

PR #145, opened 2015-10-16
fix scroll bug with floats; 

johnspackman:
the content element can return scrollX/Y values which are fractional, and that causes rounding errors which given the right sizes cause scrollbars to constantly scroll until they cannot scroll any more (without user input).  This fix forces the scroll values to integers and avoids the problem


johnspackman:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9223

oetiker:
makes sense

PR #146, opened 2015-10-16
fix to allow contenteditable fields to get the spacebar event

johnspackman:


framework/source/class/qx/ui/root/Abstract.js
```diff
@@ -252,8 +252,9 @@ qx.Class.define("qx.ui.root.Abstract",
       }
 
       // Require that widget does not accept text input
-      var nodeName = target.getContentElement().getNodeName();
-      if (nodeName === "input" || nodeName === "textarea") {
+      var el = target.getContentElement();
+      var nodeName = el.getNodeName();
+      if (nodeName === "input" || nodeName === "textarea" || el.getDomElement().contentEditable === "true") {
         return;
```
oetiker: Will getDomElement always return an object ?

johnspackman: Yes, I think so - because this method is only called in response to a keypress event, it must have a DOM element; the code is referring to e.getTarget() because it can be bubbling up, which would imply that that the target was on screen too.

OTOH its not an unreasonable check :)  I'll patch it just now

oetiker:
thanks

PR #147, opened 2015-10-16
Fix dispose bugs

johnspackman:


level420:
Related to https://github.com/qooxdoo/qooxdoo/pull/168

framework/source/class/qx/ui/form/AbstractField.js
```diff
@@ -687,8 +690,7 @@ qx.Class.define("qx.ui.form.AbstractField",
      * @return {String|null} The current value
      */
     getValue : function() {
-      var value = this.getContentElement().getValue();
-      return this.__nullValue ? null : value;
+      return this.isDisposed() || this.__nullValue ? null : this.getContentElement().getValue();
     },
```
oetiker: if it is disposed the value is 'true'.
wouldn't returning null make more sense?

johnspackman: oops - yes, very much an unintended result

oetiker:
yep like it!

PR #148, opened 2015-10-16
fix for rhino (Json is broken)

johnspackman:
The qx.lang.Json polyfill does not work under Rhino because it cannot convert a POJO to object; the polyfill sees the objects getClass method and ends up trying to serialise the entire Class heirarchy too.  Rhino's implementation of JSON.stringify knows how to avoid this, so the fix is to disable the polyfil.  Arguably a bug in Rhino, but as Rhino doesn't get a huge amount of development any more (except individual patches) this is an essential work around.

### DIFF
--- a/framework/source/class/qx/lang/Json.js
+++ b/framework/source/class/qx/lang/Json.js
@@ -230,6 +230,11 @@ qx.Bootstrap.define("qx.lang.Json",
 
 /*! JSON v3.2.5 | http://bestiejs.github.io/json3 | Copyright 2012-2013, Kit Cambridge | http://kit.mit-license.org */
 (function (window) {
+  // This polyfill does not work under Rhino because it cannot convert POJO to object (it tries
+  //  to serialise the class)
+  if (qx.core.Environment.get("runtime.name") === "rhino" || qx.core.Environment.get("runtime.name") === undefined)
+    return;
+  
   // Convenience aliases.
   var getClass = {}.toString, isProperty, forEach, undef;
 


framework/source/class/qx/lang/Json.js
```diff
@@ -230,6 +230,11 @@ qx.Bootstrap.define("qx.lang.Json",
 
 /*! JSON v3.2.5 | http://bestiejs.github.io/json3 | Copyright 2012-2013, Kit Cambridge | http://kit.mit-license.org */
 (function (window) {
+  // This polyfill does not work under Rhino because it cannot convert POJO to object (it tries
+  //  to serialise the class)
+  if (qx.core.Environment.get("runtime.name") === "rhino" || qx.core.Environment.get("runtime.name") === undefined)
+    return;
```
oetiker: You are not only changing the behavior for rhino but also for the case the runtime.name is not set ... please elaborate.

johnspackman:
It's some time ago now but I think that the test for undefined came about because I was finding that the runtime.name is not always defined at times when JSON.stringify is used (ie it is a recursive dependency issue).

In a wider viewpoint, this issue is that the polyfil breaks on platforms where it does not fully understand the platform; in the case of embedded systems like Rhino (or possibly Nashorn) which integrate Javascript objects with objects from the host language (in this case Java) it is essential to understand the platform when writing code like this, so resorting to default platform behaviour when the platform cannot be detected is a safer option.

Rhino's implementation of JSON does accommodate it's own implementation and is better than the polyfill if only because it actually works on Rhino; I think it's reasonable to have caution on other non-web platforms too.

AIUI this polyfil is implemented to support JSON3, which is supposed to be a "drop in replacement" for JSON 2 implementations, but with detection for recursive objects and a fix to date serialisation on some platforms.  The README on https://github.com/bestiejs/json3 says this can be fixed with custom Date.prototype.toJSON implementation, so it seems that the justification for having the poyfill in the first place is pretty thin.

oetiker:
ok makes sense

PR #149, opened 2015-10-16
Fix label oversizing in grids

johnspackman:
This pull request fixes a bug where the label is able to outgrow the maxWidth setting on a grid; the first commit fixes the grid so that it does not allow its own maxWidth setting to be exceeded, the second commit fixes label so that overflow is hidden otherwise the label still visually exceeds the maxWidth


johnspackman:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9222

oetiker:
thanks

PR #150, opened 2015-10-16
fixes a bug in qx.data.controller.List where changes to the model in

johnspackman:
which the length of the array stays the same prevents the bindings from
being updated properly.  Without this fix, the Controller will update
the ListItem label but not it's model.

### DIFF
--- a/framework/source/class/qx/data/controller/List.js
+++ b/framework/source/class/qx/data/controller/List.js
@@ -409,9 +409,7 @@ qx.Class.define("qx.data.controller.List",
       qx.ui.core.queue.Widget.add(this);
 
       // update on filtered lists... (bindings need to be renewed)
-      if (this.__lookupTable.length != this.getModel().getLength()) {
-        this.update();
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/data/controller/List.js
```diff
@@ -409,9 +409,7 @@ qx.Class.define("qx.data.controller.List",
       qx.ui.core.queue.Widget.add(this);
 
       // update on filtered lists... (bindings need to be renewed)
-      if (this.__lookupTable.length != this.getModel().getLength()) {
-        this.update();
-      }
+      this.update();
     },
```
oetiker: if the new model just happens to have the same length, the update will not happen

johnspackman: yes, that was the bug that this fixes?

oetiker: oops :) sorry

PR #151, opened 2015-10-16
Fix null form field values

johnspackman:
Fixes two bugs where reseting form field values to null causes an exception


oetiker:
looks good to me

PR #152, opened 2015-10-16
Logger add filters

johnspackman:
This enhancement allows filters to be specified to qx.log.Logger at runtime which control what log messages got to what Appenders; the filters are based on a regular expression matched against the class name of the object which is making the log.

This patch also adds "trace" as a seperate logging level so that each individual appender can decide whether to pass stack information on.

If no filters are applied, then the behaviour is unchanged from previous versions.

### DIFF
--- a/framework/source/class/qx/log/Logger.js
+++ b/framework/source/class/qx/log/Logger.js
@@ -509,18 +509,28 @@ qx.Bootstrap.define("qx.log.Logger",
      */
     __getAppenders: function(className, level) {
       var levels = this.__levels;
-      if (levels[level] < levels[this.__level])
-        return [];
       
-      if (!this.__filters.length)
+      // If no filters, then all appenders apply
+      if (!this.__filters.length) {
+        // Check the default level
+        if (levels[level] < levels[this.__level])
+          return [];
         return this.__appenders;
+      }
+      
       var names = [];
       for (var i = 0; i < this.__filters.length; i++) {
         var filter = this.__filters[i];
+        
+        // Filters only apply to certain levels
         if (levels[level] < levels[filter.level])
           continue;
+        
+        // No duplicates
         if (filter.appenderName && qx.lang.Array.contains(names, filter.appenderName))
           continue;
+        
+        // Test
         if (filter.loggerMatch.test(className)) {
           if (filter.appenderName)
             names.push(filter.appenderName);
@@ -528,12 +538,12 @@ qx.Bootstrap.define("qx.log.Logger",
             return this.__appenders;
         }
       }
+      
+      // Map appender names into appenders
       var appenders = [];
       var appendersByName = this.__appendersByName;
-      names.forEach(function(name, index) {
-        var appender = appendersByName[name];
-        if (appender)
-          appenders.push(appender);
+      appenders = names.map(function(name) {
+        return appendersByName[name];
       });
       return appenders;
     },
--- a/framework/source/class/qx/log/Logger.js
+++ b/framework/source/class/qx/log/Logger.js
@@ -97,7 +97,15 @@ qx.Bootstrap.define("qx.log.Logger",
     */
 
     /** @type {Map} Map of all known appenders by ID */
-    __appender : {},
+    __appenders : [],
+    
+    
+    /** @type {Map} Map of all known appenders by name */
+    __appendersByName: {},
+    
+    
+    /** @type {Array} Array of filters to apply when selecting appenders to append to */
+    __filters: [],
 
 
     /** @type {Integer} Last free appender ID */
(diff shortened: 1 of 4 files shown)


johnspackman:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9221

framework/source/class/qx/log/Logger.js
```diff
@@ -118,16 +126,19 @@ qx.Bootstrap.define("qx.log.Logger",
...
       for (var i=0, l=entries.length; i<l; i++) {
-        if (levels[entries[i].level] >= levels[this.__level]) {
-          appender.process(entries[i]);
-        }
+        var entry = entries[i];
+
+        var appenders = this.__getAppenders(entry.loggerName, entry.level);
+        if (qx.lang.Array.contains(appenders, appender))
+          appender.process(entry);
       }
```
oetiker: Wouldn't it make more sense to use a hash here instead of doing a linear search every time ?

framework/source/class/qx/log/Logger.js
```diff
@@ -457,12 +519,64 @@ qx.Bootstrap.define("qx.log.Logger",
...
+        var filter = this.__filters[i];
+        
+        // Filters only apply to certain levels
+        if (levels[level] < levels[filter.level])
+          continue;
+        
+        // No duplicates
+        if (filter.appenderName && qx.lang.Array.contains(names, filter.appenderName))
+          continue;
+
```
oetiker: array search again

framework/source/class/qx/log/Logger.js
```diff
@@ -144,11 +155,26 @@ qx.Bootstrap.define("qx.log.Logger",
...
       delete appender.$$id;
     },
 
 
+    /**
+     * Adds a filter that specifies the appenders to use for a given logger name (classname)
+     * @param logger {String|RegExp} the pattern to match in the logger name
+     * @param appenderName {String?} the name of the appender class, if undefined then all appenders
+     * @param level {String?} the minimum logging level to use the appender, if undefined the default level is used
+     */
```
oetiker: having an example in the javadoc would be very helpful in understanding the intention of this.

framework/source/class/qx/log/Logger.js
```diff
@@ -214,14 +240,20 @@ qx.Bootstrap.define("qx.log.Logger",
      * Prints the current stack trace at level "info"
      *
      * @param object {Object?} Contextual object (either instance or static class)
+     * @param message {var} Any number of arguments supported. An argument may
+     *   have any JavaScript data type. All data is serialized immediately and
+     *   does not keep references to other objects.
      */
```
oetiker: again please add an example

framework/source/class/qx/test/log/Filters.js
```diff
@@ -0,0 +1,64 @@
...
+     2004-2009 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     LGPL: http://www.gnu.org/licenses/lgpl.html
+     EPL: http://www.eclipse.org/org/documents/epl-v10.php
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+ * Fabian Jakobs (fjakobs)
+
```
oetiker: author would be you, right ?

zaucker: Wouldn't it make sense to start using the newly suggested MIT license?

See http://news.qooxdoo.org/license-update

On Thu, 18 Feb 2016, Tobias Oetiker wrote:

>> +
>> +   qooxdoo - the new era of web development
>> +
>> +   http://qooxdoo.org
>> +
>> +   Copyright:
>> +     2004-2009 1&1 Internet AG, Germany, http://www.1und1.de
>> +
>> +   License:
>> +     LGPL: http://www.gnu.org/licenses/lgpl.html
>> +     EPL: http://www.eclipse.org/org/documents/epl-v10.php
>> +     See the LICENSE file in the project's top-level directory for details.
>> +
>> +   Authors:
>> + * Fabian Jakobs (fjakobs)
>> +
>
> author would be you, right ?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/qooxdoo/qooxdoo/pull/152/files#r53308158

-- 

johnspackman: author is me :)

I'm changing the Filter to use the MIT license, does it make sense to do the same on files we edit also, ie qx.log.Logger?

oetiker: hmmm with existing files, I am not so sure ... there should be some input from @ecker on this ... he is working with 1&1 legal ... but for new files ... please use MIT.

ecker: Hi all, of course you should take care not to alter the license terms of existing files. That's to be done in the proposed, official license change. For newly added files on the other hand, I don't see a problem to use the liberate MIT license, but doubt it'll make much of a difference to continuing to go with the existing dual license for now.

oetiker:
like it!

PR #153, opened 2015-10-16
Making debugging easier - prevent unnecessary exceptions

johnspackman:
When debugging an app and "Pause on all exceptions" _has_ to be turned on, unnecessary exceptions can be a real pain and make it much harder to find the one exception you are really searching for.  The patches prevented many dozens of exceptions in my application startup (particularly DateField).


oetiker:
ok

PR #154, opened 2015-10-16
added qx.core.Object.isDisposing() 

johnspackman:
When disposing an object, the $$disposed flag is set immediately but during dispose methods can be called which trigger warnings because isDisposed() returns true; however, the object is in the process of being disposed which means that these warnings should be suppressed because it is normal (for example) to set a property to null during dispose, and therefore changeXxxx events can legitimately fire during dispose.  

This fix allows objects to determine that the dispose() method is in progress (as opposed to having completed); also in this PR is a fix to qx.event.dispatch.Direct which was generating a number of warnings in my use case, which could be legitimately suppressed.  The number of warnings generated can be significant and make it harder to identify other issues during disposal

### DIFF
--- a/framework/source/class/qx/core/Object.js
+++ b/framework/source/class/qx/core/Object.js
@@ -280,6 +280,7 @@ qx.Class.define("qx.core.Object",
 
       // Mark as disposed (directly, not at end, to omit recursions)
       this.$$disposed = true;
+      this.$$disposing = true;
       this.$$instance = null;
       this.$$allowconstruct = null;
 
@@ -319,6 +320,8 @@ qx.Class.define("qx.core.Object",
         clazz = clazz.superclass;
       }
 
+      this.$$disposing = false;
+      
       // Additional checks
       if (qx.core.Environment.get("qx.debug"))
       {
--- a/framework/source/class/qx/core/Object.js
+++ b/framework/source/class/qx/core/Object.js
@@ -257,6 +257,17 @@ qx.Class.define("qx.core.Object",
 
 
     /**
+     * Returns true if the object is being disposed, ie this.dispose() has started but 
+     * not finished
+     *
+     * @return {Boolean} Whether the object is being disposed
+     */
+    isDisposing : function() {
+      return this.$$disposing || false;
+    },
+
+
+    /**
      * Dispose this object
      *
      */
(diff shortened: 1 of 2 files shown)


johnspackman:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9220

cajus:
$$dispose is set in the very beginning of qx.core.Object.dispose(). Why does isDisposed() not return the correct value before the dispose is complete in this case?

johnspackman:
Sorry that pull request was incomplete and not accurate in its description; I've updated it now

cajus:
No need to sorry. I can't merge anyway. Was just tracking down some dispose problems on our side and saw your pull request(s) ;-)

johnspackman:
Well thanks for pointing it out anyway, I obviously rushed that PR! 

I'm rapidly coming to the opinion that having to dispose() is a real problem area that needs to be removed entirely - I have a theory (which needs investigating) that the object registry is only really needed to provide loose coupling between DOM and Qx objects, and in other cases can be removed.  My guess is that hash codes _are_ needed elsewhere, but in many/most cases the hash codes are used to allow other objects to have their own lookup tables and the central qx.core.ObjectRegistry is not actually a requirement.  

Having all objects stored in qx.core.ObjectRegistry means that the garbage collector cannot kick in until the objects are dispose()d, which is the only reason why dispose() is so important.  Effectively, memory management of all Qx objects is manual because there needs to be loose coupling between DOM and UI widgets.

Another problem is the correct sequencing of disposal - this has become a real pain, and I've actually stubbed out qx.core.ObjectRegistry.shutdown() because it cannot guarantee the order of disposal and this leads to side effects triggered by dispose() (as well as a potentially huge delay when refreshing the page).  I suspect that this is why destructors are not available in garbage collected languages, eg Java & Javascript, IE there must not be side effects triggered by disposal, only memory release.

If this theory is true, then only UI widgets need to be registered, and that would allow the browser's garbage collection to do most of the tricky memory management that dispose() is designed to implement.  Most Qx objects would not need to implement any kind of destructor at all.

level420:
@ecker This pull request seems very useful. Would you mind applying it?
@johnspackman Thank you for the pull request

oetiker:
makes sense

PR #155, opened 2015-10-16
removed unnecessary dependencies from qx.dev.Debug

johnspackman:
This removes dependencies made by qx.dev.Debug onto UI classes because when used in a server environment, this allows a huge number of not relevant classes to be excluded into the build.  

In Rhino, there is a limited amount of space for compiled code and the volume of extra code these dependencies bring in can break the server implementation.

This patch uses runtime detection so is backward compatible.


oetiker:
makes sense

PR #157, opened 2015-10-16
optimises qx.event.Idle so that it only creates a timer if a listener

johnspackman:
has been attached; this is useful for profiling because qx.event.Idle
will otherwise pollute the timing data


johnspackman:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9224

oetiker:
makes sense

PR #158, opened 2015-10-29
[BUG #9228] Fix image theme switch

cajus:
If you've an image which is styled with "source" in one theme and without "source" in the other, it gets "display: none" when switching the theme to the one without "source".

Switching back to the one with source restores the image+path, but does not remove the "display: none" - so the image is optically gone while it's in the DOM.

The PR fixes a typo in the usage of __getMode and takes care about initially restoring the values correctly.


cajus:
http://bugzilla.qooxdoo.org/show_bug.cgi?id=9228

peuter:
Bug ID is 9228 not 9224

ecker:
Bug ID fixed, thanks for pointing out! ;-)

PR #159, opened 2015-11-18
* Add error event to media

RangelReale:


oetiker:
accepting this, but I would rather like to a see a more complete event coverage ... the code here definitely starts to get WET. https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events

oetiker:
please make this a pull request for master!

PR #160, opened 2015-11-20
Added dispose check for debounced functions

cajus:
Sync the behaviour of debounce function calls with the one used in event, defere, etc.: add a warning instead of indirectly rising exceptions.


level420:
@ecker This pull request seems very useful. Would you mind applying it?
@cajus Thank you for the pull request

oetiker:
again acting on disposed stuff ... a warning definitely seem preferable to dying ...

PR #161, opened 2015-11-25
[BUG #9235] Change some form validation manager methods from private to protected to allow overrides

level420:
It is desirable to allow overriding the Methods __validateRequired and __validateItem to be able to intercept the validation. In our use case we want to allow required fields to only be required if they are enabled.

This could be simply achieved by overriding those methods and return true if the form items are isEnabled() === false


level420:
See qooxdoo bugzilla bug http://bugzilla.qooxdoo.org/show_bug.cgi?id=9235

level420:
@Andreas: any chance to get this soon applied?

PR #162, opened 2015-12-04
[BUG #9237] qx.data.controller.Form: show exceptions while setting up binding

level420:
Currently only a warning is displayed if something is going wrong during initial binding, without really knowing the cause. Added the exception text to the warning message.


level420:
See bugzilla bug http://bugzilla.qooxdoo.org/show_bug.cgi?id=9237

@ecker very handy to have this commited, because the original code hides bugs in the form items

PR #164, opened 2016-01-07
[BUG #9238] qx.util.ResourceManager: add method getIds which retrieves an array of existing IDs

level420:
For some situations it is very handy to know the ids registered by the manager.

### DIFF
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -61,6 +61,30 @@ qx.Class.define("qx.util.ResourceManager",
 
   members :
   {
+    /**
+     * Get all known resource IDs.
+     *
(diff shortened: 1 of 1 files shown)


level420:
See corresponding bugzilla issue http://bugzilla.qooxdoo.org/show_bug.cgi?id=9238

framework/source/class/qx/util/ResourceManager.js
```diff
@@ -61,6 +61,30 @@ qx.Class.define("qx.util.ResourceManager",
 
   members :
   {
+    /**
+     * Get all known resource IDs.
+     *
+     * @param pathfragment{String|null|undefined} an options path fragment to check against with id.indexOf(pathfragment)
```
oetiker: an **optional**

oetiker:
that might indeed be useful ... thanks!

PR #165, opened 2016-01-07
[Bug #9243] drag&drop handler: attach the pointerdown event to the capturing phase

level420:
allowing widgets to stop event propagation of the pointerdown event.

This makes drag and drop working again for form buttons.


oetiker:
makes sense. thanks.

PR #166, opened 2016-01-07
[Bug #9244] qx.ui.decoration.Abstract never assigns anything to __insets, though it seems the intention was to cache them

level420:
Proposed patch for bug http://bugzilla.qooxdoo.org/show_bug.cgi?id=9244 reported by Veronika Silvesterova.


oetiker:
makes sense to me. thanks

PR #168, opened 2016-02-04
[Bug #9252] Do not render layout items if they are already disposed

level420:
This avoids exceptions in qx.ui.core.queue.Layout when disposing widgets

### DIFF
--- a/framework/source/class/qx/ui/core/LayoutItem.js
+++ b/framework/source/class/qx/ui/core/LayoutItem.js
@@ -440,6 +440,11 @@ qx.Class.define("qx.ui.core.LayoutItem",
      */
     renderLayout : function(left, top, width, height)
     {
+      // do not render if the layout item is already disposed
+      if (this.isDisposed()) {
+        return null;
+      }
+
       if (qx.core.Environment.get("qx.debug"))
       {
         var msg = "Something went wrong with the layout of " + this.toString() + "!";


level420:
See bugzilla bug http://bugzilla.qooxdoo.org/show_bug.cgi?id=9252

oetiker:
why is renderLayout being called on disposed widgets in the first place?

level420:
@oetiker I'ts mainly an observation within the testrunner. My tests also include disposal of complex widgets where I manually dispose the widget and test if it really is disposed by calling method isDisposed():

```javascript
testEditorClose : function() {
  var d = this.getEditor();
  qx.ui.core.queue.Dispose.add(d);
  qx.ui.core.queue.Dispose.flush();

  this.assertTrue(d.isDisposed());
}
```

The test runner then encounters an exception while the rendering queue tries to render the already disposed widget, because there is no DOM element anymore connected to the widget.

This addition inhibits rendering of disposed widgets.

level420:
@oetiker To answer you primary question: I don't know why the disposal triggers the render queue.

oetiker:
I did a little search (github is great) it seem it might make more sense to just not render disposed widgets by adding approriate code here.

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/queue/Layout.js#L97

level420:
OK. I'll try to catch this within the layout queue and see if this solves the issues with testrunner

oetiker:
I would expect that this might catch even more cases, as `renderLayout` is being overwritten in many places.

level420:
Placing

```javascript
if (widget.isDisposed()) {
  continue;
}
```
at position https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/queue/Layout.js#L97
does not solve this issue, as renderLayout is also directly called e.g. from layout classes (canvas, grid etc.) for each child:

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/layout/Canvas.js#L315

This case maybe occurs if the children of the layout are already disposed, but the layout itself is not.

It also seems that qx.ui.core.Widget overrides the renderLayout, but in the override first calls the superclass method, omitting all further actions if the result of the call is null:

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/Widget.js#L1015

So against your assumption catching this in qx.ui.core.queue.Layout handles fever cases.

An alternative PR would be to walk through all direct calls of renderLayout and inhibit the call by testing against isDisposed.

But in the end I don't really know if this is something which will happen in real life and not only in test runner.

oetiker:
ok, let's roll with this one then

PR #170, opened 2016-02-18
QxCompiler

johnspackman:
The changes in the branch are needed to enable the QxCompiler contrib with ES6 compilation; they consist of (a) fixes so that Qooxdoo is strict mode compatible, and (b) some changes to qx.Bootstrap and the loader to make it drastically simpler to calculate the load order dependency for scripts.  

These changes are backwards compatible with generate.py and only take effect when QxCompiler is producing the build; the dependency order is calculated slightly differently to the generate.py, and so there are a couple of places where explicit dependencies have been added in.

The key to dependency calculation changes are that qx.Bootystrap/qx.Class would call a class .defer() method immediately after it has been created; this forced generate.py to statically interpret the code in .defer() and all code that .defer() called indirectly.  The changes in this branch allow .defer() to be deferred until all code has loaded (except in a few specific circumstances) - this allows Javascript VM to do the heavy lifting for us.  It also makes recursive dependencies much less likely.


oetiker:
hmmm that pull request seems to drag in quite a lot of other stuff ...

johnspackman:
so that's not what I was expecting - I created a new branch from today's master, and merged my other branch (which was based on the 5.0.1 release) into the new branch; there were some conflicts but I resolved them in favour of qooxdoo/master.

The files that are changed in this pull request are only the ones that I've changed, but github says that it includes changes from a load of other commits (presumably the ones that had conflicts, even though they were resolved from qooxdoo/master).  My git GUI (SourceTree) does not show those extra commits either.

Any idea how I can fix this?

oetiker:
I think the problem is that the fixes that went into the 5_x branch also went into master, thus the best you can probably do, is

update your master with qooxdoo master

create a new branch

cherry pick the qxcompiler changes into this branch

and then create a pull request from this new branch.

johnspackman:
Great, thanks - I think i've got it now; I'm closing this and will reissue

PR #173, opened 2016-02-19
Changes to add ES6 (strict mode) and qxcompiler dependencies into master

johnspackman:
The changes in the branch are needed to enable the QxCompiler contrib with ES6 compilation; they consist of (a) fixes so that Qooxdoo is strict mode compatible, and (b) some changes to qx.Bootstrap and the loader to make it drastically simpler to calculate the load order dependency for scripts.

These changes are backwards compatible with generate.py and only take effect when QxCompiler is producing the build; the dependency order is calculated slightly differently to the generate.py, and so there are a couple of places where explicit dependencies have had to be added in - but explicit dependency overrides (via @use and @require) is a normal feature and is commonly used, especially in core classes that provide values at runtime to qx.core.Environment.

The key to dependency calculation changes are that qx.Bootstrap/qx.Class would call a class .defer() method immediately after it has been created; this forced generate.py to statically interpret the code in .defer() and all code that .defer() called indirectly, recursively.  This is inherently fragile as well as difficult to calculate - for example, qx.ui.form.AbstractField.defer() calls qx.ui.style.Stylesheet.getInstance(), the constructor for which calls qx.bom.Stylesheet.createElement(); that function in turns relies on the "html.stylesheet.createstylesheet" environment check which is implemented by qx.bom.client.Stylesheet. So qx.ui.form.AbstractField.defer() has a load time dependency on qx.bom.client.Stylesheet but this is not actually a requirement for loading and it would be easy for the author of (EG) qx.ui.style.Stylesheet to introduce unresolvable dependencies into other classes.

The changes in this branch allow .defer() to be delayed until all code has loaded (except in a few specific circumstances) - this allows Javascript VM to do the heavy lifting for us, and means that each class is in charge of it's own dependencies.


level420:
@johnspackman As I understood this PR does not affect compatibility for "traditional" generator.py based apps. Is this right?

johnspackman:
Yes, exactly - the changes for delayed defer have to be explicitly switched on in the boot loader (which is only done by QxCompiler) and the strict mode stuff is by feature detection or just suppressing exceptions etc

framework/source/class/qx/Bootstrap.js
```diff
@@ -401,6 +413,99 @@ qx.Bootstrap.define("qx.Bootstrap",
...
+      var t = this;
+      if (!dbClassInfo) {
+        var pendingDefers = this.__pendingDefers;
+        this.__pendingDefers = [];
+        pendingDefers.forEach(execute);
+        return;
+      }
+
+      function executeForDbClassInfo(dbClassInfo) {
+        if (dbClassInfo.environment) {
```
oetiker: I am rather fond of using 

`var executeForDbClassInfo = function(dbClassInfo) { `

for localized code references as it makes the locality of the definition more obvious.

what do you think?

johnspackman:
I think it's a bit 50/50 but personally I'm not so convinced about that form because it means that the function symbol is hoisted but the implementation is not, which leads to a strange situation where the function exists but it's code does not.

OTOH I found out the other day that it is a strict mode requirement that functions are defined at the top of a function or file only, but that is something which is clearly not enforced uniformly; AIUI the form you've used would be an effective way (the only way) to work around moving the function to the top of the file or function anyway.

I need to double-check this and get back to you in a bit

oetiker:
I did a quick search through the qooxdoo source ... and I think they are using the var x = function(..) pattern everywhere ... the only named functions are in qx/dev/unit/Sinon.js

johnspackman:
no worries, I'll switch it over.  I wanted to understand why strict mode wasn't rejecting my function, but i'll get onto this before the end of the week

oetiker:
looking good ... except for the isStrictMode function which is still there ...

oetiker:
still 95 tests failing ... so no worse than before :) merging now. thanks!

PR #175, opened 2016-02-22
source-server-nodejs

adrelino:
Added grunt task source-server-nodejs to the grunt templates.
Now you have the choice between
grunt source-server               -> python SimpleHttpServer
grunt source-server-nodejs   -> node grunt-http-server


oetiker:
looks good

PR #176, opened 2016-02-24
inherit styles from current element. 

kreimer:
related to this thread in forum:
http://qooxdoo.678.n2.nabble.com/cursor-on-images-with-scale-true-td7588002.html

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -590,8 +590,10 @@ qx.Class.define("qx.ui.basic.Image",
           
           //inherit styles from current element
           var currentStyles = currentContentElement.getAllStyles();
-          for(var prop in currentStyles) {
-              styles[prop] = currentStyles[prop];
+          if( (currentStyles!==null) && (currentStyles!==undefined) ) {
+            for(var prop in currentStyles) {
+                styles[prop] = currentStyles[prop];
+            }
           }
 
           // Copy dimension and location of the current content element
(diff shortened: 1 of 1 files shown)


oetiker:
looks fine to me

framework/source/class/qx/ui/basic/Image.js
```diff
@@ -587,6 +587,12 @@ qx.Class.define("qx.ui.basic.Image",
         {
           var pixel = "px";
           var styles = {};
+          
+          //inherit styles from current element
+          var currentStyles = currentContentElement.getAllStyles();
```
oetiker: to be on the save side I would check if `currentStyles` is not `null`

framework/source/class/qx/ui/basic/Image.js
```diff
@@ -587,6 +587,14 @@ qx.Class.define("qx.ui.basic.Image",
...
+          if( (currentStyles!==null) && (currentStyles!==undefined) ) {
```
oetiker: how about just a simple

`if (currentStyles) { ...`

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/undefined

kreimer: Hi thanks for telling me.. I always had doubts with this .. it this exactly
the same behaviour? cheers!

oetiker: no not exactly ... it would also be false if `currentStyles` is 0 or if it is an empty string ... but both of these cases would not be good news either here

kreimer:
I agree 100%, just commited, thanks!

oetiker:
thanks

PR #179, opened 2016-02-29
Added braces, removed unused variables, updated config.json

cajus:
Makes the built in _lint_ job happy with the framework subdirectory.


oetiker:
thanks

PR #180, opened 2016-02-29
added travic-ci link

oetiker:
lets see it travis starts testing


cajus:
The build stuck directly after initializing the page. I had this two times out of ~30 builds. Not sure what happened in these cases. Can you re-trigger the build for this PR to see if it runs completely?

PR #185, opened 2016-03-03
Replace some equality tests by identity tests when the behavior is the same

woprandi:
Replace some '==' by '===' when comparing strings


oetiker:
thanks

PR #187, opened 2016-03-04
Eliminate discrepancy of the locale property between qx.locale.Manager and qx.util.format.NumberFormat

level420:
qx.locale.Manager defines property locale nullable : true, but
qx.util.format.NumberFormat does not, but the properties get bound by
single value binding, which causes null value exceptions on
qx.locale.Manager.getInstance().resetLocale().

This PR changes the locale property of qx.locale.Manager being
initialized with the current client locale, or if not available with the
default locale "C" which is now availavle as a variable via
qx.core.Environment.get("locale.default")

### DIFF
--- a/framework/source/class/qx/bom/client/Locale.js
+++ b/framework/source/class/qx/bom/client/Locale.js
@@ -102,5 +102,6 @@ qx.Bootstrap.define("qx.bom.client.Locale",
   defer : function(statics) {
     qx.core.Environment.add("locale", statics.getLocale);
     qx.core.Environment.add("locale.variant", statics.getVariant);
+    qx.core.Environment.add("locale.default", "C");
   }
 });
--- a/framework/source/class/qx/core/Environment.js
+++ b/framework/source/class/qx/core/Environment.js
@@ -559,6 +559,10 @@
  *       <td>locale.variant</td><td><i>String</i></td><td><code>de</code></td>
  *       <td>{@link qx.bom.client.Locale#getVariant}</td>
  *     </tr>
+ *     <tr>
+ *       <td>locale.default</td><td><i>String</i></td><td><code>C</code></td>
+ *       <td>default locale C as in good tradition of unix</td>
+ *     </tr>
 
  *     <tr>
  *       <td colspan="4"><b>os</b></td>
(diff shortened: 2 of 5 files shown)


oetiker:
I guess the locale.default should then be listed in  the api docs in qooxdoo/framework/source/class/qx/core/Environment.js  

what would the effect be if: in

https://github.com/qooxdoo/qooxdoo/pull/187/files#diff-0c948c087f48502fce031c1bc95dc8d0R85

the "" was set to "C".

wouldn't that make the whole thing way simpler ?

level420:
@oetiker But what if we want t know if some locale ist set at all? Wouldn't we prevent the user for detecting this via qx.bom.client.Locale?

But anyway: I've missed a failed test in qx.test.util.DateFormat. So please don't merge this PR atm.

oetiker:
well if no locale is set, we fall back to C ...  that is the unix way of things I think.

level420:
yes! I agree. But we implement the fallback in qx.locale.Manager and not in qx.bom.client.Locale

oetiker:
Hmmm I may have missed some part of the discussion. But isn't the problem that phantom-js is not setting a default locale and thus we get an empty string as a locale, which breaks tests. Now you are introducing code which allows one to figure out that the browser is not actually setting a locale, but is there any application for this ?

cajus:
FYI: I added the locale to phantomjs  in the accept-language custom header but it had no effect. So I didn't commit that one to the runner.

level420:
No, this all also happens in the normal testrunner in Chrome. The tests did the following:

1. create and instance of qx.util.format.NumberFormat and don't set a locale in the constructor: var nf = qx.util.format.NumberFormate(); If you omit the locale, the instance get bound to the locale managers property locale

2. call qx.locale.Manager.getInstance().resetLocale(). You'll get an exception, because on resetLocale the property is set to null which then is propagated to your NumberFormat instance where null is not allowed: you get an null value exception

cajus:
I tried this:

`
        page.customHeaders = {
          "Accept-Language": "en-DE,en;q=0.5"
        };
`

cajus:
@level420: ah ok, I didn't check that.

level420:
That's why I called this PR "discrepancy of the locale property between qx.locale.Manager and qx.util.format.NumberFormat"

oetiker:
ok ... lets go with this :)

PR #188, opened 2016-03-04
Fix qx.test.uti.DateFormat.testChangingLocales

level420:
The test assumes that on start the locale managers locale property is at
init state, but this is not shure if we had run other tests before which
also set a locale. So just reset the locale manager before starting.


oetiker:
thanks

PR #189, opened 2016-03-07
Add missing semicolons

woprandi:


oetiker:
are semicolons before the closing bracket required ?

woprandi:
With Javascript, almost nothing is required :)
But it is advisable.

PR #190, opened 2016-03-07
30 separate travis runs php output redirected output of php to /dev/null

level420:
A lot of tests do not clean up instances and other garbage at test end.
This makes other tests fail, even if calling the testclasses one by one
does not fail. We try to split up the runs to avoid these interferences.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,35 @@ before_script:
   - export PATH=$PWD/phantomjs-2.1.1-linux-x86_64/bin/:$PATH
 script:
   - tool/test/travis/setup
-  - php -S 127.0.0.1:31323 1>&2 &
-  - tool/test/travis/run http://127.0.0.1:31323/framework/test-console/index-source.html
+  - php -S 127.0.0.1:31323 1>/dev/null 2>/dev/null &
+  - tool/test/travis/run "http://127.0.0.1:31323/framework/test-console/index-source.html?testclass=qx.test.application"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -11,6 +11,35 @@ before_script:
   - export PATH=$PWD/phantomjs-2.1.1-linux-x86_64/bin/:$PATH
 script:
   - tool/test/travis/setup
-  - php -S 127.0.0.1:31323 1>&2 &
-  - tool/test/travis/run http://127.0.0.1:31323/framework/test-console/index-source.html
+  - php -S 127.0.0.1:31323 1>/dev/null 2>/dev/null &
+  - tool/test/travis/run "http://127.0.0.1:31323/framework/test-console/index-source.html?testclass=qx.test.application"
```
oetiker: the classic way to write this is:

```sh
php -S 127.0.0.1:31323 1>/dev/null 2>&1 &
```

oetiker:
```
tool/test/travis/run "http://127.0.0.1:31323/framework/test-console/index-source.html?testclass=qx.test.dev
```

unfortunately, if it is done in this way, we can not see how many tests pass.

maybe altering the runner code in phantom to start a new 'browser object' per test would be a good solution.

level420:
@oetiker do you mean that travis does not report success or failure for each "tool/test/travis/run ....." call?

level420:
Giving up on this, as the test now hangs on tool/test/travis/run "http://127.0.0.1:31323/framework/test-console/index-source.html?testclass=qx.test.application".

Maybe someone else with more travis insight can have a look into this  ( @oetiker @cajus )

PR #191, opened 2016-03-07
[Bug 9253] get rid of arguments.callee calls in table stuff

patrick-fischer-hirschstein:


oetiker:
patch would alter the api for the arguments expected in `setSortMethods` and thus breaking existing code ... 

I see two solutions

1. use different argument names for the new type of sorting function
2. use a different method call

the old sorting type can then be marked as deprecated ... but old code will still run.

level420:
I don't really have a use case in our apps currently, but if I read https://github.com/qooxdoo/qooxdoo/pull/191/files#diff-205182a795755cc03885e6cae2461dccL162 correctly, then we will loose the possibility to do combined sorts with multiple columns.

Let's say I have column year and column month and the user clicks performs a click on the "year" column for sorting. With the current method I can simply implement sort first on year, then on month.

But maybe I'm wrong.

level420:
If the arguments.callee.columnIndex is the ugly part, then we can think about adding an additional parameter columnIndex as third parameter to the sort method. As for current code arguments.callee.columnIndex stil works and does not break any existing code.

In the default sort methods and for the future we can use comparison between row1[columnIndex] and row2[columnIndex].

stahl-riesa:
I understand. As more as you think over it's getting even more difficult. The whole current sorting stuff is not good (enough). Maybe you are right and I should work a new second sort implementation.
As a starting point I prefer simple algorithms sorting only one column. So you have to define additional sorting columns, e.g. for a column year something like additionalSorting = [ { columnName : ''month', sortDirection : 'desc|asc|main|reverse' }] for more complex sorting stuff. The Keywords 'main' and 'reverse' refer to the current sorting direction of the column year.

PR #192, opened 2016-03-09
Call the applyer for the locale property in the constructor of qx.locale.Manager

level420:
To be shure the provite attribute __locale is initialized we have to invoke the applyer for the locale property by calling initLocale() in the constructor.


oetiker:
thanks

level420:
@oetiker Please appy this PR

PR #193, opened 2016-03-14
Use 'Indigo' theme as the default theme for new apps

woprandi:


oetiker:
that's just the default themes for the templates, right ? what about the api viewer, the demo browser and the other sample apps ?

woprandi:
only tutorials have old theme. I can update them also

oetiker:
If we do it, I think it should be consistant

woprandi:
Yeah, but we should also update tutorial screenshots

woprandi:
Updated PR

oetiker:
the apps don't need any changes ?

woprandi:
Apps already use the indigo theme

http://demo.qooxdoo.org/current/showcase
http://demo.qooxdoo.org/current/widgetbrowser/
http://demo.qooxdoo.org/current/feedreader/
http://demo.qooxdoo.org/current/demobrowser/
...

oetiker:
ok ... in that case I merge

PR #194, opened 2016-03-16
Travis ci

cajus:
Tweak testing. Does not yet work on travis for unknown reasons. I'm at it.

tool/test/travis/run can now be started without having "php -S" before. It's done automatically. So no need for a URL. See usage.


oetiker:
thanks

Issue #196, opened 2016-03-23
Filtered item visible in List when model has been changed.


peuter:
See Playground example:
http://tinyurl.com/hftbw23

Press the shift button to see the problem

A Listmodel with 5 Items where the item on index 1 is filtered. When you shift the model array the filtered Item is shown in list.

level420:
I think as of http://demo.qooxdoo.org/current/apiviewer/#qx.data.controller.List you have to update the controller after manipulating the model:

http://tinyurl.com/hghqbac

peuter:
Shouldn't the controller update itself after the model has changed?

And 2nd. if you check the console in the playground you can see that the filtered model item gets applied to the list item. This problem also exists when you use virtual lists btw. Your workaround just hides the real problem IMHO.

level420:
You're right. This should happen automatically.It may be, that the fix in https://github.com/qooxdoo/qooxdoo/commit/6570b3d1df776548d9d5af4f82ee2d1229df59fc addresses that issue.

Which version of qooxdoo do you use? Playground is still on qooxdoo 5.0.1 where the fix is not in yet.

peuter:
I started a new project with the current master-branch to investigate the problem (without any success yet). so the problem also exists in the current master branch.

level420:
@johnspackman: This is your fix: https://github.com/qooxdoo/qooxdoo/commit/6570b3d1df776548d9d5af4f82ee2d1229df59fc

Could you please have a look at this?
Thanks!

level420:
OK then let's see: 

If the model.shift() is called in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/data/Array.js#L301 the change event should be fired. 

Because the list controller registers an event listner for the model change event here https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/data/controller/List.js#L328 the controllers method __changeModel should be fired: 

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/data/controller/List.js#L404 which calls the list controllers update method at https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/data/controller/List.js#L412

johnspackman:
i can take a look at this if you want, i've got some hours put aside for Qooxdoo stuff this afternoon?

oetiker:
you write 'should' ? does the event not get fired?

level420:
Sorry, I meant it is fired.

[...]

PR #197, opened 2016-04-01
Adds qx.data.controller.ListOfObjects 

johnspackman:
In response to issue #196, this PR adds qx.data.controller.ListOfObjects which is an optional upgrade to qx.data.controller.List; the advantages over controller.List are bug fixes for changing array lengths and abstraction of the array handling and filtering.


framework/source/class/qx/data/controller/ListOfObjects.js
```diff
@@ -0,0 +1,425 @@
...
+ * Especially the model is nice to bind to another selection for example.
+ * The controller itself can only work if it has a model and a target set. The
+ * rest of the properties may be empty.
+ *
+ * *Update from List*
+ *
+ * This controller is a replacement for {@link qx.data.controller.List} which is now
+ * deprecated; it is a drop in replacement for List with the exception of the delegate
+ * which no longer provides the filter or configureItem methods.  Filtering is
+ * achieved by using {@link qx.data.controller.ArrayFilter}.
```
derrell: How does one now configure an item?

johnspackman: fixed - the delegate is now IArrayDelegate, and configureItem is replaced with [un]bindTargetItem

framework/source/class/qx/data/controller/List.js
```diff
@@ -409,7 +415,12 @@ qx.Class.define("qx.data.controller.List",
       qx.ui.core.queue.Widget.add(this);
 
       // update on filtered lists... (bindings need to be renewed)
-      this.update();
+      if (!this.__warningIssued) {
+        this.warn(this.classname + " can cause item model properties to change expectedly; please consider using qx.data.controller.ListOfObjects as a soltions.  See https://github.com/qooxdoo/qooxdoo/issues/196 for more details");
```
oetiker: change **expectedly** ?

framework/source/class/qx/data/controller/List.js
```diff
@@ -409,7 +415,12 @@ qx.Class.define("qx.data.controller.List",
...
+        this.__warningIssued = true;
+      }
+
+      qx.event.Timer.once(this.update, this, 0);
```
oetiker: this will actually fix the problem seen in the original bugreport, right ?

derrell: Although that simple one-liner is a bit of a kludge, maybe we should just
document why that's been put there (with a link to the discussion), and
*not* add or deprecate a class? I recall from earlier in the discussion
that there was stuff going on that really wasn't needed (adding and then
removing a listener, I think) but retaining a long-standing API has pretty
high value...

Derrell

johnspackman: yes, I've been using @peuter & @level420 's test case.  We can wait for them to try it?  there's no rush to merge

framework/source/class/qx/data/controller/List.js
johnspackman: oops, fixed

framework/source/class/qx/data/controller/List.js
johnspackman: fixed

framework/source/class/qx/data/controller/ArrayFilter.js
```diff
@@ -0,0 +1,191 @@
...
+    sortMethod: {
+      init: null,
+      nullable: false,
+      apply: "_applyXxx"
+    },
+    
+    /** Method called to filter */
+    filterMethod: {
+      init: null,
+      nullable: true
```
peuter: the filter method doesn`t get applied on change, I guess an apply: "_applyXxx" is missing here

framework/source/class/qx/data/controller/ListOfObjects.js
```diff
@@ -0,0 +1,444 @@
...
+    
+    
+    /*
+    ---------------------------------------------------------------------------
+       ITEMS ARRAYMIRROR DELEGATE METHODS
+    ---------------------------------------------------------------------------
+    */
+    _onCreateTargetItem: function() {
+      var delegate = this.getDelegate();
+      if (delegate && typeof delegate.createItem == "function")
```
peuter: referring to IArrayDelegate I guess this must be delegate.createTargetItem, right?

framework/source/class/qx/data/controller/ArrayFilter.js
johnspackman: Yes - fixed

framework/source/class/qx/data/controller/ListOfObjects.js
johnspackman: Also yes & fixed :)

framework/source/class/qx/data/controller/List.js
```diff
@@ -409,7 +407,15 @@ qx.Class.define("qx.data.controller.List",
       qx.ui.core.queue.Widget.add(this);
 
       // update on filtered lists... (bindings need to be renewed)
-      this.update();
+      if (!this.__warningIssued) {
+        this.warn(this.classname + " can cause item model properties to change unexpectedly; please consider using qx.data.controller.ListOfObjects as a soltion.  See https://github.com/qooxdoo/qooxdoo/issues/196 for more details");
```
peuter: Typo: ... please consider using qx.data.controller.ListOfObjects as a solution.

cajus:
This one was resolved in another pull request, wasn't it? Or at least it was not needed anymore?

johnspackman:
Yes, this can be closed; there are some classes (eg ArrayFilter) that i'll submit in another PR later, but the main point of this one is no longer needed

PR #198, opened 2016-04-05
Introduce blacklist in event manager...

### LINKED ISSUE #196: Filtered item visible in List when model has been changed.

peuter:
... to prevent dispatching of removed events + remove bindings in virtual lists on every model change

fixes #196

### DIFF
--- a/framework/source/class/qx/event/Manager.js
+++ b/framework/source/class/qx/event/Manager.js
@@ -70,6 +70,12 @@ qx.Class.define("qx.event.Manager",
     this.__dispatchers = {};
 
     this.__handlerCache = {};
+
+    this.__blacklist = [];
+
+    this.__clearBlackList = new qx.util.DeferredCall(function() {
+      this.__blacklist = [];
+    }, this);
   },
 
 
@@ -119,6 +125,8 @@ qx.Class.define("qx.event.Manager",
     __handlerCache : null,
     __window : null,
     __windowId : null,
+
+    __blacklist : null,
 
 
     /*
@@ -657,6 +665,7 @@ qx.Class.define("qx.event.Manager",
         if (entry.handler === listener && entry.context === self)
         {
           qx.lang.Array.removeAt(entryList, i);
+          this.__addToBlacklist(entry.unique);
 
           if (entryList.length == 0) {
             this.__unregisterAtHandler(target, type, capture);
@@ -716,6 +725,7 @@ qx.Class.define("qx.event.Manager",
         if (entry.unique === unique)
         {
           qx.lang.Array.removeAt(entryList, i);
+          this.__addToBlacklist(entry.unique);
 
           if (entryList.length == 0) {
             this.__unregisterAtHandler(target, type, capture);
@@ -751,6 +761,10 @@ qx.Class.define("qx.event.Manager",
         {
           // This is quite expensive, see bug #1283
           split = entryKey.split("|");
+
+          targetMap[entryKey].forEach(function(entry) {
+            this.__addToBlacklist(entry.unique);
+          }, this);
 
           type = split[0];
           capture = split[1] === "capture";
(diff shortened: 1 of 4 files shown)


johnspackman:
+1

level420:
I have two questions:

1. If I understand the code right the black list array is never emptied and grows for every removed event listener. Or am I wrong?

2. Wouldn't it be better to use a hash to speed things up? I'm bothered about the isBlacklisted method and the performance which may degrade if we have a lot of listeners. 

```javascript
// init the blacklist
this.__blacklist = {};

// add the listener uid to the blacklist
this.__blacklist[uid] = true;

// check if the listener is blacklisted
if(this.__blacklist[uid] !== true) {
  // call listener
}
```

peuter:
There is a DeferredCall which cleans the blacklist aafter everything is done here: https://github.com/peuter/qooxdoo/blob/fix/196/framework/source/class/qx/event/Manager.js#L77

The call is scheduled when the first item has been added to the blacklist. Because of that I don't think that a hash is necessary there.

level420:
Thank you for the explanation @peuter . 

so the code in https://github.com/peuter/qooxdoo/blob/fix/196/framework/source/class/qx/event/Manager.js#L928-L930
is meant to schedule the emptying of the blacklist until the current event listener loop has finished. Right?

I'm still not convinced about the implementation of the isBlacklisted method in https://github.com/peuter/qooxdoo/blob/fix/196/framework/source/class/qx/event/Manager.js#L941 findind blacklisted uids with indexOf, as it is called on every listener call.

But I think the idea is great and we should go wit a commit. Thank you for the PR.

level420:
@peuter I wanted to ask if you would consider changeing the implementaion to a hash based blacklist and especially take care of performance for situations where there are no blacklisted listeners at all.
That would be really great! Thank you.

peuter:
Ok, done.

peuter:
> so the code in https://github.com/peuter/qooxdoo/blob/fix/196/framework/source/class/qx/event/Manager.js#L928-L930
is meant to schedule the emptying of the blacklist until the current event listener loop has finished. Right?

Yes thats right.

level420:
Thank you @peuter ! Great!

peuter:
One question:

Should I add an this.__blacklist.hasOwnProperty(uid) before this.__blacklist[uid] here:
https://github.com/peuter/qooxdoo/blob/fix/196/framework/source/class/qx/event/Manager.js#L945

How would that affect the performance?

oetiker:
since bla[x] returnes undefined if x is not an existing key, I don't think testing is necessary ...

level420:
I don't think it is needed. I'd even go with a null for this.__blacklist in the empty case and test against:

```javascript
return (this.__blacklist !== null && this.__blacklist[uid] === true)
```
without the using this.__blacklistLength.

level420:
and in __addToBlacklist:

```javascript
if (this.__blacklist === null) {
  this.__blacklist = {};
  this.__clarBlacklist.schedule();
}
```
and clearing the blacklist would be simply setting it to null.

peuter:
@level420 Good point, I changed it to use null checks.

oetiker:
so it sems everyone is happy with this one :-)

level420:
Perfect now! Let's go with this PR.

And also let's hope, that our applications do not somehow rely on the side effect that removed listeners were executed after removal. Let's see what my monster form with approx. 800 fields is doing after this commit.

level420:
Small typo in

https://github.com/qooxdoo/qooxdoo/pull/198/files#diff-c9f3678e3bed8e09ff7b8d0919566f4dR934

'remaved' vs 'removed'

peuter:
typo is fixed

oetiker:
with everyone happy I merge ... thanks @peuter !

PR #199, opened 2016-04-11
Clean some listeners

cajus:
Spinner and form renderer left a couple of open listeners on external objects. Lets clean them.


oetiker:
thanks

PR #200, opened 2016-04-11
Convert target if required

cajus:
This is a follow up to #142. I needed to remove the original fork because I accidently pushed some non easy revertable stuff to master. Sigh.

As requested, there's a check for qx.ui.core.Widget now.


oetiker:
looks good ... so I'll close #142

PR #201, opened 2016-04-14
Work around exception when pressing Enter in non focusable TextField

cajus:
In widgets where TextField is a child control, TextField is mostly added
using setFocusable(false). Focus is handeled in the parent widget.
Pressing "Enter" in such widgets (Spinner, etc.) results in an exception
in these cases, because the virtual keyboard is hidden with blur() -
which is not supported on non focusable TextField.


oetiker:
neat

PR #203, opened 2016-04-22
Fix travis tests

cajus:
This PR does a bit more than the title suggests.
- removes old travis integration based on phantomjs
- adds support for karma testing + its travis integration
- fixes some outdated tests that were based on browser implementation facts that are not valid anymore
- fixes a couple of issues in qooxdoo itself

The framework tests can be simply run on the console like this:

```
cd framework
npm install
npm test
```

Test supports setting of the browser to be used and the tests to be run:

```
npm test -- --browsers=Chrome --client.testClass=qx.test.bom.\(Stylesheet\|element.Style\).*
```


oetiker:
wow ... I AM IMPRESSED

PR #208, opened 2016-04-25
Disable failing tests which work for themselves in chrome.

cajus:


oetiker:
yep ... that is helpful so that we can see if  new code break something ...

PR #216, opened 2016-04-30
Fix BZ#9213 - Table select all rows w/ Ctrl+A

petitben:
Using keyup rather than keypress fix the problem with Chrome as with CH4+, some "reserved browser key combinations" aren't anymore transmitted through keypress event:https://bugs.chromium.org/p/chromium/issues/detail?id=33056

Note I leave the related method name as _onKeyPress for backward compatibility. It can be adapted it necessary.

### DIFF
--- a/framework/source/class/qx/ui/table/Table.js
+++ b/framework/source/class/qx/ui/table/Table.js
@@ -189,7 +189,7 @@ qx.Class.define("qx.ui.table.Table",
 
     // Make focusable
     this.setTabIndex(1);
-    this.addListener("keypress", this._onKeyPress);
+    this.addListener("keyup", this._onKeyPress);
     this.addListener("focus", this._onFocusChanged);
     this.addListener("blur", this._onFocusChanged);
 
--- a/framework/source/class/qx/ui/table/Table.js
+++ b/framework/source/class/qx/ui/table/Table.js
@@ -189,7 +189,7 @@ qx.Class.define("qx.ui.table.Table",
 
     // Make focusable
     this.setTabIndex(1);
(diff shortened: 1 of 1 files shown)


cajus:
Thanks for that one. Since it works in all other browsers and the fix changes the feel a little bit, what about adding a `qx.core.Environment.get("browser.name") == "chrome"` check, so that it stays like it is for other browsers - plus works (very) slightly different in chrome?

oetiker:
Would it not also work on `keydown`?

petitben:
Yes it will work too with keydown. Actually my first attempt was using keydown then I changed my mind. The UX would be a bit different but I think the final user shouldn't notice it. So both solutions look equivalent to me.

Regarding the qx.core.Environment.select version... Well I don't know... I guess it's better to have the same behaviour across browsers.

Thoughts ?

oetiker:
what made you change your mind? I just did a little test here on my ipad, here the letters only appear ehen releasing the key, but I am pretty sure they do appear on a normal computer when pressing the key ..

petitben:
Basically it was just regarding the fix as I consider the select-all should occur when the user releases keypress.

All things considered, though, it would be finally better to use `keydown` because it won't fit some part of the usecase.

Indeed, navigating with arrows in the table won't be easy because the user will be forced to release and then press "down" again to go further for instance. That's even more obvious using this basic example page as unlike keydown, keyup event can't of course be repeated: http://unixpapa.com/js/testkey.html

I'll review my commit and change it to `keydown.

cajus:
Just tried it in Chrome and it works perfectly. So lets take that one. Thanks ;-)

PR #217, opened 2016-04-30
Spellcheck run for comments / API docs

cajus:
This commit contains spell fixes for API relevant docstrings, normal comments, and in one case a spell fix with a variable name. No check for grammatical or context errors have been made.

Sure, it is a large commit, but in a spell correction case it does not make sense to commit it in various smaller chunks.


oetiker:
looks good to me!

PR #2121, opened 2016-05-02
Fix gecko engine detection for Firefox 47beta+

level420:
To detect the gecko rendering window.navigator.mozApps was used, which was (obviously) removed in Firefox47beta. But there still exists the firefox specific mozContacts API which we additionally use now for detection

### DIFF
--- a/framework/source/class/qx/bom/client/Engine.js
+++ b/framework/source/class/qx/bom/client/Engine.js
@@ -180,7 +180,8 @@ qx.Bootstrap.define("qx.bom.client.Engine",
      *
      * Note:
      *  "window.controllers" is gone/hidden with Firefox 30+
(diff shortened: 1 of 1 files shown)


oetiker:
is there any documentation on mozContacts?

level420:
https://wiki.mozilla.org/WebAPI/ContactsAPI

level420:
But wait with the merge. I've found the following article:

https://www.fxsitecompat.com/en-CA/docs/2016/firefox-os-apis-have-been-hidden-from-web-content/

level420:
OK. It seems that navigator.mozContacts will desappear in Firefox 48 :-(

https://developer.mozilla.org/en-US/Firefox/Releases/48#Others

Closing, as it is just a temporary fix until FF48 is released.

PR #8984, opened 2016-05-02
Fix gecko engine detection for Firefox 47beta+

level420:
To detect the gecko rendering window.navigator.mozApps was used, which was (obviously) removed in Firefox47beta. We additionally test against https://developer.mozilla.org/en-US/docs/Web/API/Navigator/buildID which is still present

### DIFF
--- a/framework/source/class/qx/bom/client/Engine.js
+++ b/framework/source/class/qx/bom/client/Engine.js
@@ -180,18 +180,20 @@ qx.Bootstrap.define("qx.bom.client.Engine",
      *
      * Note:
      *  "window.controllers" is gone/hidden with Firefox 30+
-     *  "window.navigator.mozApps" is supported since Firefox 11+
+     *  "window.navigator.mozApps" is supported since Firefox 11+ and is gone/hidden with Firefox 47 beta
+     *  "window.navigator.buildID" is supported since Firefox 2+
      *  "window.navigator.product" is actually useless cause the HTML5 spec
      *    states it should be the constant "Gecko".
      *
      *  - https://developer.mozilla.org/en-US/docs/Web/API/Window.controllers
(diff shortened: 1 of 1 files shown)


oetiker:
ok let's take that

petitben:
Please note there is a discussion here regarding buildID: https://bugzilla.mozilla.org/show_bug.cgi?id=583181

So unless nobody removes it in the near future or set this firefox bug as WON'T FIX, your commit might also be a temporary solution too... :-(

level420:
@petitben I'm aware of that bug in the mozilla bugzilla. In the end it may be better to re think on how the firefox version is detected and parsed and how the gecko engine is detected reliable. Have a look into https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/client/Engine.js#L181-L191
You can see that the history allready tells us about several changes in the gecko engine which made the detection fail.
Additionally we currently do not detect the blink engine on current chrome, opera etc., which is currently reported as webkit.
Maybe it would be worth creating an issue for this problem.

woprandi:
What is the need to identify Gecko ?

oetiker:
@woprandi qooxdoo provides an interface to detect what rendering engine the website is running on ... in the code this facility is used to workaround bugs in a particular engine ... 

The problem with gecko is that everyone and their dog claims to be gecko and sometimes go to great length to fake this ... which makes detecting the real gecko a bit of a challenge. Especially since some people say that javascript in the browser should not be able to detect what engine and version it is running on as this could allow the js to then exploit bugs in the particular version it knows about ... 
Or also that it allows for better fingerprinting the more details about the local environment are available ...

level420:
@woprandi, @oetiker, @petitben Please change to https://github.com/qooxdoo/qooxdoo/issues/8994

Issue #215, opened 2016-04-29
Automatic Memory Management


johnspackman:
Memory management in Qooxdoo is currently done manually - it is essential to track and `.dispose()` objects properly to avoid leaks.  This is a big hindrance to writing large, complex applications which are long running.  

I originally suggested this in [a proposal on the mailing list](http://qooxdoo.678.n2.nabble.com/Qooxdoo-Automatic-Memory-Management-a-proposal-td7587901.html) 3 or 4 months ago, and since then I've been using my patch to deprecate .dispose() on live applications.

My observations are that this is a good thing, does not introduce any problems that I've noticed, and has allowed me to quickly fix some tricky bugs just by getting rid of calls to `.dispose()`.

As demonstration that this is a good thing, the Qooxdoo framework test suite (in Jan 2016) had only 413 objects that were not disposed when using this patch, compared to 4,500 objects unpatched.  (OK it’s fair to say that unit tests are not required to free up resources that they consume, but it is striking how easy it is to have significant resource leaks that can be so easily countered by garbage collection)

The key points of the proposal are that:

(1) it is impossible to know if you should free up an object that you pass in to a property, for example `qx.ui.form.SelectBox.setSelected` will copy the array that you give it, while other objects might take it over and keep it

(2) Methods often allocate objects that cause hard to spot leaks, for example `qx.data.Array` methods such as splice and slice return an array of the removed items, but the returned object is an instance of `qx.data.Array` so you *must* dispose it.  This is counter intuitive (and error prone) partly because `qx.data.Array` goes to some lengths to act like a native array, but native array methods don't need their side effects to be disposed

(3) These memory issues are only there because every `qx.core.Object` is registered with `qx.core.ObjectRegistry` and this is not necessary in almost all cases.  We can remove this registry with almost no backward compatibility issues, and in doing so make it much easier to code Qooxdoo apps.

(4) The only parts of the Qooxdoo framework that depend on `qx.core.ObjectRegistry.fromHashCode` are `qx.ui.core.Widget` and `qx.data.controller.Tree`; Widget uses the hash code to avoid having DOM objects that refer to Javascript objects, which was to avoid IE6 bugs - so this can be dropped.  Tree can implement its own lookup if necessary

The backwards compatibility issues that this will introduce are:

(A) `qx.core.ObjectRegistry.fromHashCode()` will no longer work, unless the object is manually registered with `qx.core.ObjectRegistry.register()`.

[...]

PR #8986, opened 2016-05-02
Automatic memory management v2

johnspackman:
See issue https://github.com/qooxdoo/qooxdoo/issues/215


oetiker:
@johnspackman I guess that some parts of the manual need to be changed accordingly too ?

johnspackman:
Yes - how do we do that BTW?

Ill go through the code again this afternoon, looking for destructors/dispose which are required and update the API docs to match

oetiker:
just patch the relevant parts of https://github.com/qooxdoo/qooxdoo/tree/master/documentation along with the code.

level420:
@johnspackman beside taking care of calls to qx.core.ObjectRegistry.fromHashCode(): what else is needed to migrate existing code? Does it break compatibility somehow?

johnspackman:
@level420 no its limited to fromHashCode, the entire trick is just about not storing references to objects, ie not calling `qx.core.ObjectRegistry.register`.  The big impact internally in the Qooxdoo source is changing `qx.ui.core.Widget` and qx.html.Element so that they don't try and use weak references (ie don't require the `.register` call)

oetiker:
looked at the docs ... is it the case, that any class which has a 'dispose' method also must be disposed ?

johnspackman:
not necessarily - any destructor or overridden dispose may do things which need to be called to clean up; but in a lot of cases all they do is set member variables to null, or call dispose on other objects.

So whether something has to be disposed becomes an API issue, rather than an qx.core.Object issue

The main things are event listeners for native events, and locale changes.  This means that qx.ui.* is pretty much always having to be disposed.

BTW I have to go back through and check the overridden dispose, I realise I only checked destructors

johnspackman:
Making it so that you have to "know" that this object requires dispose is not great though, nor is relying on API documentation especially when you might have to manually traverse the class hierarchy to check it.

How about if the class implements a marker interface that indicates that dispose is required; this would also mean that the object will be automatically stored in ObjectRegistry, which means that it is tracked and can be inspected, leaks can be checked for, and there is an explicit means of identifying classes at runtime and compile time which have this requirement.  

Also, to mitigate the effects of breaking change for users of fromHashCode(), qx.core.Object.toHashCode() could issue a warning, once per class, if it is not automatically registered.

johnspackman:
The principal of using garbage collection for managing objects instead of explicitly .dispose()ing them is that destructors and dispose() methods are not implemented.  This is important for many reasons, the principal being that the Javascript garbage collector does not notify us when an object is about to be collected.

But this does not mean that objects can’t have startup and shutdown procedures - many do in fact, it’s just that cycle is (or should be) separate to garbage collection.  

Historically, a common pattern in Qooxdoo classes is to use dispose/destructor as a general purpose shutdown, so if we want to move to automatic garbage collection we need to identify those classes which have a special shutdown procedure so that the calling code can respect that and free up resources.  We could take the view that all destructor/dispose methods mean that the class has special shutdown procedures but this is unnecessarily harsh because there are a lot of classes which have trivial destructor implementations, or can be easily refactored to not have a destructor.

The difficulty here is in making it clear and obvious what needs to be disposed, and I’ve taken two approaches: first, each class which needs to be disposed implements the marker interface qx.core.IDisposable.  Every Object which implements that interface will be registered with ObjectRegistry exactly as before, so that it can be identified for debugging purposes.  This has the side effect that fromHashCode will also work for those objects, and that the classes can be identified reliably via reflection and (potentially) in the API viewer.

Secondly, many destructors simply set member variables to null and these have been removed.  This is because they are unnecessary and by not providing one we make it clear that it is not necessary.

Thirdly, in some rare cases the code has been refactored to avoid the need for a destructor where previously there was one - for example, only adding a listener for the duration of the task and removing at the end of the task, instead of keeping the listener until dispose is called.  Except in one or two places I have avoided this approach because the risk of introducing bugs with rash, unfocused changes is high.

In many cases, the destructors are for classes which are typically used as singletons and need not be tracked - for example the various qx.event.handler.* and qx.event.dispatch.* classes.

There is one remaining global list of objects which could benefit from refactoring - qx.data.SingleValueBinding keeps a global lookup of every bound object, so any listeners to or from an object property will prevent that object from being garbage collected.  This could be modified so that the bindings are stored with the object, rather than in a global list indexed by hash code, but note that even when binding a single property, the target object also has links back to the source object; this means that unless both source and target are unbound, the binding will prevent garbage collection of both objects.


The net effect of this patch is that automatic garbage collection is now possible, with the proviso that (a) if you must release any bindings manually, and (b) you must observe any “shutdown” requirements of classes.

johnspackman:
apologies, this PR is not ready - I pushed too early yesterday having forgotten to return to the unit tests first :(

level420:
@johnspackman do you use this already in production?

johnspackman:
Yes; the new changes to do with what is or is not kept in the ObjectRegistry (ie IDisposable etc) is very new, but the base has been in use on live sites since January.

cajus:
What is the state of the PR? I guess some more automatic memory management is a "must have" and would simplify a couple of tasks. As mentioned in the preceding PR, it would be cool to erase existing listeners targeting to objects to be disposed, too.

level420:
I've did some short tests with the PR against current master with my app which has a real monster form. I opened/loaded with data/closed the form window about 20 times and had a look into the windows task manager and the reported memory usage there. This is what users do constantly in my app.

So far memory usage walked up and down but in sum stabilized at some value.

Additionally I opened nearly all forms and windows of the app and no problem showed up so far. The app worked as expected.

I've used current chrome on windows 7 for the tests.

level420:
I added my vote whereby I can not say having really checked the changes. I'm trusting here in @johnspackman and travis and others.
I'm still undecided when we should merge as it breaks compatibility regarding fromHashCode(), qx.core.Object.toHashCode()

zaucker:
What is holding up a merge?

level420:
@zaucker : your vote ;-) and the votes of other core members

zaucker:

level420:
I think the main risk of having this PR merged, is that we loose the object registry functionality. I've tested the PR with my largest app and had no problems so far. Therefore I gave the PR my vote.

zaucker:
No risk, no fun ...

oetiker:
OK waiting to give my vote ...

johnspackman:
@level420 it's true that the `ObjectRegistry.fromHashCode` functionality is lost by default for non-widgets, it's easily switched on and imho there is an argument for localised, purpose-specific registries.  Poor consolation for anyone who has a bug caused by this, but there is the warning logged on the console when `fromHashCode()` can't find an object.

zaucker:
Just tried this branch with a larger application and found no problem so far.

woprandi:
I'll try to test this PR. Should I check something special ?

level420:
@woprandi You should simply test your app and as much of its functionality as possible and have a look into memory consumption if it grows over time (wait a bit to allow the automatic garbage collector to work).

woprandi:
After few minutes, I don't see any strange behavior or memory behavior. I'll work with these branch all that day

cajus:
Will test it here with a bigger block of code. Please hold the line ;-)

gnikolaidis:
I am impatient to use this PR in production. Whatever the risks of breaking compatibility with existing code, the benefits of NOT registering all qx.core.Object's are too important to ignore.

hkollmann:
I tested this patch with our vzaweb application without any problems

cajus:
I can't build the project with the PR activated:

```
>>> Collapsing parts  
>>> Verifying parts \<type 'exceptions.RuntimeError'> :
Unfullfilled dependency of class 'qx.event.Manager'[31,44]: 'qx.core.IDisposable'
```

cajus:
Hmm. Looks like it is not in the patch. Verifying...

cajus:
Ok. The bigger application loads and I don't see any issues going thru the most common usecases. Testing it using selenium would last some hours, so I'm skipping this for now.

@johnspackman: One build issue with the currently external upload manager:

`Uncaught Error: Overwriting private member "__widget" of Class "com.zenesis.qx.upload.InputElement" is not allowed!
`

That one has been added in the `qx.html.Element`.

cajus:
At least two business days have passed since the call for votes, and two assenting votes have been cast with no dissenting votes. Going to merge this one.

PR #8987, opened 2016-05-03
qx.bom.FileReader#readAsArrayBuffer added to File API

mbgonicus:
As readAsArrayBuffer is part of the File API standard, it should be part of qx.bom.FileReader as well.


oetiker:
that failure was transient it seems ... I ran the tests again and now they are fine

PR #8988, opened 2016-05-04
Include all sources into test-source job...

peuter:
... in order ot get realistic coverage data, for evenrything and not just for the sources that are used by the tests,


oetiker:
merging since the failing tests seem to have a different origin

PR #8989, opened 2016-05-04
#8977

hlefebvre:
When a form is validated with errors and then the model of the form controller changes then the validator should be reset to clear previous errors or they will remain displayed erronously.

### DIFF
--- a/framework/source/class/qx/data/controller/Form.js
+++ b/framework/source/class/qx/data/controller/Form.js
@@ -278,6 +278,11 @@ qx.Class.define("qx.data.controller.Form",
       if (this.getTarget() == null) {
         return;
       }
+      else {
+        // if form was validated with errors and model changes
+        // the errors should be cleared see #8977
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/data/controller/Form.js
```diff
@@ -278,6 +278,11 @@ qx.Class.define("qx.data.controller.Form",
       if (this.getTarget() == null) {
         return;
       }
+      else {
+        // if form was validated with errors and model changes
+        // the errors should be cleared see #8977
+        this.getTarget().getValidationManager().reset();
+      }
```
oetiker: is there always a validationmanger present, or can it be null too ?

hlefebvre: As far as I see the validation manager is always created for the qx.ui.form.Form in the construct method, thus should be ok not to test it

oetiker:
ok

Issue #8976, opened 2016-01-29
Wrong behaviour of scroll pane when the count of elements in the model changes (BZ#9249)



qx-bug-importer:
_Hugues Lefebvre (samael) wrote:_

If a qx.ui.list.List is populated with more element than it&#39;s height it will then display scrollbars.
Scroll to the bottom of the list and select last item.
Put a button somewhere that removes the last element in the model.
Try to select an item without scrolling =&gt; selection is not working as expected.

In the qx.ui.virtual.Pane when the rowCount changes the __scrollTop member can have an out of range value, it&#39;s not set to getScrollMaxX in this case and it make the computing of the widget under mouse wrong.

The issue can be reproduced here :

[...]

PR #8991, opened 2016-05-04
#8976

hlefebvre:
When an item is removed from the List's model and the scrollbar was positionned to the bottom, reducing the list size will also reduce the maxScrollY making the current scrollTop out of range. When it happen then all click to item are selecting another one making the component unusable until scollbar is manualy scrolled up (see #8976).


oetiker:
I have reopened the issue :)

oetiker:
I think we should merge this ... any other  comments ?

level420:
Seems fine for me. merged.

Issue #8990, opened 2016-05-04
VirtualDropDownList internal list selection keeps receiving bubbling events after being disposed


RemiHeugue:
I found a bug in the `qx.ui.form.core.VirtualDropDownList` class.

If you use a Widget with a `VirtualDropDownList` (like `VirtualSelectBox`) with items that include `qx.data.marshal.MEventBubbling` then dispose it while any element other than the first is selected and then change a property, the bubblingEvent will be fired and use the `VirtualDropDownList` internal `List` selection as context, which is now disposed.

This is due to the `__synchronizeSelection` method in [VirtualDropDownList.js](https://github.com/qooxdoo/qooxdoo/blob/dbd14ab643bd7fafc2a88e928f9e649567322ccc/framework/source/class/qx/ui/form/core/VirtualDropDownList.js#L385) that changes the native array of the selection directly which prevent removal of the bubbling event listeners.

Here is a [playground example of the bug](http://tinyurl.com/zodlqov).

1. Open the log panel
2. Select the second item
3. Click the "Destroy" Button
4. Click the "Change Item2 value" Button

You will get something like

    154941 qx.event.dispatch.Direct: The context object '' for the event 'changeBubble' of 'crash.Item[548-0]'is already disposed.
[...]

PR #8992, opened 2016-05-04
Fix VirtualDropDownList internal list selection keeps receiving bubbling events after being disposed

RemiHeugue:
Fix for #8990 

I use splice on the `qx.data.Array` which remove bubbling event listeners unlike manual modification of the native array.

### DIFF
--- a/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
+++ b/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
@@ -393,15 +393,7 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
       }
       else
       {
-        var nativeArray = target.toArray();
-        qx.lang.Array.removeAll(nativeArray);
-        for (var i = 0; i < source.getLength(); i++) {
-          nativeArray.push(source.getItem(i));
-        }
-        target.length = nativeArray.length;
-
-        // necessary for keeping preselected items in sync
-        target.fireDataEvent("change", {});
+        target.splice.apply(target, [0, target.length].concat(source.toArray())).dispose();
       }
     },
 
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -393,15 +393,7 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
...
-        qx.lang.Array.removeAll(nativeArray);
-        for (var i = 0; i < source.getLength(); i++) {
-          nativeArray.push(source.getItem(i));
-        }
-        target.length = nativeArray.length;
-
-        // necessary for keeping preselected items in sync
-        target.fireDataEvent("change", {});
+        target.splice.apply(target, [0, target.length].concat(source.toArray())).dispose();
       }
```
oetiker: so this calls `dispose` on all the old entries of target ?

johnspackman: no, target is `qx.data.Array` and `.splice` returns a new instance of `qx.data.Array` which needs to be disposed

RemiHeugue: Sorry, it's the first time I contribute to an open source project. :sweat_smile:
I've made a new commit with the line splitted and comments.

I use splice because this is whats I think is the closest to the previous code (remove all from target, add all from source and send an event).

I don't use it directly but with apply because if you try to do `target.splice(0, target.length, source.toArray()` you will get an array of array (`[[item1, item2, ...]]`) instead of an array containing all the items of source(`[item1, item2, ...]`).

The problem was that when you add objects that include the mixin `qx.data.marshal.MEventBubbling` to a `qx.data.Array`, the `_registerEventChaining` add a listener on it to propagate bubbling events but when you remove items manually the listeners aren't removed whereas splice does.

oetiker:
I think we should merge this

level420:
I think the PR is good. merging it.

PR #8998, opened 2016-05-09
Fix qx.test.bom.Location tests in non headless firefox and chrome

level420:
6 tests failed in qx.test.bom.Location because a testnode is created and added as a direct child to the document.body. This is at least true for chrome 50 and firefox 46 on windows 7.

The tests create additional elements within the testnode, awaiting the testnode being the only child of the body. With this assumption the testnodes assume to be positioned at the top of the body. But this is not true, as the iframe html code contains text which creates a text node as the first child, making the tests fail.

The fix deletes the unnecessary text within the body tag of the html templates making all test succeed within qx.test.bom.Location.


oetiker:
makes sense :)

PR #9003, opened 2016-05-11
allow asynchronous setUp testCases by using wait/resumeSetUp

peuter:


oetiker:
since this seems to pass all the tests, I guess we are good.

level420:
OK. merging.

PR #9004, opened 2016-05-11
Add missing unit tests for string handling

woprandi:
Add the last unit tests to have qx.lang.String class 100% covered.
I had to patch qx.lang.Function.globalEval to check if it's correctly called but I don't know if it is the best way to do.

### DIFF
--- a/framework/source/class/qx/test/lang/String.js
+++ b/framework/source/class/qx/test/lang/String.js
@@ -20,6 +20,7 @@
 qx.Class.define("qx.test.lang.String",
 {
   extend : qx.dev.unit.TestCase,
+  include : [qx.dev.unit.MMock],
 
   members :
   {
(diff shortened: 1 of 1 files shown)


oetiker:
you could add a comment to the test why you are patching globalEval ... since the code is running synchronously and you are putting things back in order after the test I don't see a problem.

woprandi:
done

framework/source/class/qx/test/lang/String.js
```diff
@@ -213,6 +213,37 @@ qx.Class.define("qx.test.lang.String",
...
+
+      var nativeGlobalEval = qx.lang.Function.globalEval;
+      var called = false;
+      
+      // Patch to check if the function has been called by the stripScripts function
+      qx.lang.Function.globalEval = function(str) {
+          nativeGlobalEval(str);
+          called = true;
+      };
+
```
peuter: You can use the build-in spy functionality to cover this use case, like:

`var spy = this.spy(qx.lang.Function, "globalEval');`

and check if it has been called with:

`this.assertCalledOnce(spy);`

and restore everything at the end:

`spy.restore()`

You have to include the qx.dev.unit.MMock Mixin to have the spy + assertion availabe in your test case.

woprandi:
I used MMock.spy for my need regarding @peuter comment here https://github.com/qooxdoo/qooxdoo/pull/9004/commits/94fa21bcabc8139772347460b95720701c84bab4

level420:
Looks good!
merging.

PR #9006, opened 2016-05-11
Remove warning about "require is beeing autodetected"

cajus:


level420:
This is a simple one.
Merging.

PR #9007, opened 2016-05-11
Bumped sizzle version to 2.3.0

cajus:
Update is done like explained in the techincal documentation.


level420:
@cajus one test fails What is the reason?

cajus:
I guess that's a false positive. But please wait wit with merging. I've just recognized that there are a couple of lint errors that get introduced by the change.

level420:
OK the test is gone green now.
Merging.

PR #9010, opened 2016-05-11
Let the build fail if the linter has a problem

cajus:
The first test run will fail currently. At least until #9009 is merged.


oetiker:
I like it!

oetiker:
#9009 is merged, but I think it still fails ... relaunching the tests

level420:
@oetiker, @cajus: what is the state of this PR. I'd like to have lint run in as well.

cajus:
From my point of view, it's ok to give it a go.

oetiker:
ok I merge

PR #9012, opened 2016-05-12
Start to write some unit test for appearances

woprandi:


oetiker:
hmmm more unittests and less coverage ... hmmm

woprandi:
I don't understand

oetiker:
according to https://coveralls.io/builds/6151054/source?filename=source%2Fclass%2Fqx%2Ftheme%2Fsimple%2FAppearance.js nothing is covered

woprandi:
Have you any explanation ?

oetiker:
I have not figured out what is going on ... but I think that coverage testing is not yet working as it should. maybe @peuter has an idea.

peuter:
Not really. @woprandi Can you check if it is a general problem or a coveralls problem, by generating the coverage data locally and checking if your test is executed there and if the Appearance is coverered?

@oetiker I suggest to change the coverage decrease threshold to a higher value (like -0.5%) to prevent the coveralls check to fail in cases like this when only -0.009% changed. Must be possible somewhere in the coveralls settings.

level420:
@peuter Yes I think raising the threshhold would be good, because over time nobody will give attention to the coverage tests anymore as everybody is used to them failing for nothing (like -0.009%).

oetiker:
sure, nothing against raising the threshold ... BUT this thing here should INCREASE the coverage ...

cajus:
I'm not sure if it really increases coverage, because it just sets an appearance and checks if the change happens. If you change the appearance due to a bugfix, these tests will fail.

IMO this is not the right way to test if appearences work in general.

level420:
Still, for me, the covarege percentage value seems like sort of a random number generator. Every time looking into this, I see coverage changes which seem to be unrelated to the commit.

level420:
@cajus: even if it does not raise coverage, why does it decrease?!?!?!?

cajus:
Maybe just because the total amount of line numbers changed? Even though it's in the test directory. I don't know how that is implemented, but these 0.000x% changes just feel like a little bit of jitter.

level420:
OK let's have a look into coverage run 154 (https://coveralls.io/builds/6150802) which has the following changes: https://github.com/qooxdoo/qooxdoo/commit/e58898639b5d8bde09d776efca993af39fb90aa5 
which is not related to an javascript file and should not decrease coverage. But it does.

Additionally compare runs 154 and 155 (which is the run for this PR) in https://coveralls.io/github/qooxdoo/qooxdoo?page=3

run 154: covarage 55.45%
run 155: coverage 55.46%

Coveralls reports a DECREASE from run 154 to run 155 which is by simple mathematics not true! It shoud be increased!

peuter:
Coveralls shows source changes in 5 files for run 155. @woprandi Could you rebase your local branch onto upstream/master just to make sure that there are no other differences between the two branches  than the additional test?

woprandi:
@peuter rebased !

peuter:
@woprandi Something went wrong:

`ecmascript.frontend.SyntaxException.SyntaxException: Syntax error: Illegal map key: var (pos (29, 7))
/home/travis/build/qooxdoo/qooxdoo/framework/source/class/qx/test/theme/simple/Appearance.js`

The build is broken.

framework/source/class/qx/test/theme/simple/Appearance.js
```diff
@@ -0,0 +1,159 @@
...
+************************************************************************ */
+qx.Class.define("qx.test.theme.simple.Appearance",
+{
+  extend : qx.test.unit.TestCase,
+
+  members :
+  {
+    __obj : qx.theme.simple.Appearance.appearances,
+
+    testLabel :
```
level420: I think ```function()``` is missing here

level420:
@woprandi please have a look at https://github.com/qooxdoo/qooxdoo/pull/9012/files#diff-ddcf092421e105799c77133d5663518eR27

woprandi:
Sorry guys,

`push -f` is evil

framework/source/class/qx/test/theme/simple/Appearance.js
```diff
@@ -0,0 +1,159 @@
...
+
+    testSlideBarButtonBackward : function()
+    {
+      var style = this.__obj["slidebar/button-backward"];
+
+      this.assertIdentical(qx.test.simple.Image.URLS["arrow-up"], style({vertical: true}).icon);
+      this.assertIdentical(qx.test.simple.Image.URLS["arrow-left"], style({vertical: false}).icon);
+    }
+  }
+};
```
peuter: ) missing

peuter:
@woprandi still failing due to a missing ) at the end.

framework/source/class/qx/test/theme/simple/Appearance.js
```diff
@@ -0,0 +1,159 @@
...
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * William Opandi (woprandi)
+
+************************************************************************ */
+qx.Class.define("qx.test.theme.simple.Appearance",
+{
+  extend : qx.test.unit.TestCase,
+
```
peuter: This class does not exist, use qx.dev.unit.TestCase

peuter:
@woprandi There are more errors in your test class. Please run ./generate test-source, fix the warnings and run `npm run coverage` locally to check if the tests have been executed.

framework/source/class/qx/test/theme/simple/Appearance.js
oetiker: does such an error not show when building the test suite  on travis ?

peuter: The `generate test-source` only showed warnings, no errors, so travis did not stop.

oetiker:
and we have a winner!

level420:
@woprandi The coverage is raised, but the tests are still failing

oetiker:
what remains odd, is that travis did not fail when you were using a non-existing parent class

oetiker:
congratulations!

woprandi:
8)

level420:
@woprandi congratulations and thank you for your contribution! :+1:

PR #9013, opened 2016-05-12
Enable icon for qx.bom.Notification for newer FF versions

level420:
At least FF 46 supports icons in notifications.


cajus:
Go for it ;-)

PR #9015, opened 2016-05-12
statics optimisation breaks the websitewidgetbrowser

johnspackman:
component/standalone/website/module.json, line 90 in “min-options” job - the statics optimisation is breaking it

### DIFF
--- a/component/standalone/website/module.json
+++ b/component/standalone/website/module.json
@@ -87,7 +87,6 @@
             "strings",
             "variables",
             "variants",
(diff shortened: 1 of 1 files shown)


oetiker:
I guess we have to love with the 0.002 decrease ... merging

level420:
@oetiker: where does the decrease come from? @johnspackman only deleted a line in the non javascript code. This should not affect coverage at all. 
What is the reason for the coverage value change?

PR #9016, opened 2016-05-12
Set the translated value if it is translateable.

### LINKED ISSUE #9002: qx.ui.form.VirtualSelectBox does not respond well to changeLocale
given the following (tested in current playground, chrome and ff):

`
var localeManager = qx.locale.Manager.getInstance();
localeManager.setLocale("en");
var locale = localeManager.getLocale();
this.debug("locale: " + locale);

// Create a button
var button1 = new qx.ui.form.Button(localeManager.getLocale());

var doc = this.getRoot();

doc.add(button1,
{
  left : 100,
  top  : 50
});

// Add an event listener
  if(localeManager.getLocale()==="en") {
    localeManager.setLocale("de");
  } else {
    localeManager.setLocale("en");
  }
  e.getCurrentTarget().setLabel(localeManager.getLocale());
}, this);

var translations_en = {
  "PUBLISHED_DESC" : "published (newest first)",
  "RENTAL_PRICE_ASC" : "rental price (cheaper first)",
  "RENTAL_PRICE_DESC" : "rental price (more expesive first)",
  "BUILT_AREA_ASC" : "built area (smaller first)",
  "BUILT_AREA_DESC" : "built area (bigger first)"
};

var translations_de = {
  "PUBLISHED_DESC" : "fecha de publicacion (nuevos primero)",
  "RENTAL_PRICE_ASC" : "precio de alquiler (mas baratos primero)",
  "RENTAL_PRICE_DESC" : "precio de alquiler (mas caros primero)",
  "BUILT_AREA_ASC" : "m2 (mas pequeños primero)",
  "BUILT_AREA_DESC" : "m2 (mas grandes primero)"
};

localeManager.addTranslation("en", translations_en);
localeManager.addTranslation("de", translations_de);

var items = new qx.data.Array();
items.push(this.tr("PUBLISHED_DESC"));
items.push(this.tr("RENTAL_PRICE_ASC"));
items.push(this.tr("RENTAL_PRICE_DESC"));
items.push(this.tr("BUILT_AREA_ASC"));
items.push(this.tr("BUILT_AREA_DESC"));

var selectBox = new qx.ui.form.VirtualSelectBox(items);

selectBox.getSelection().addListener("change", function(e) {
  this.debug(e.getData().added[0]+"");
  this.debug(e.getData().added[0].getMessageId());
}, this);

doc.add(selectBox, {
  left : 100,
  top  : 150
});
`

the button is for alternating the locale ["en", "de"]
the dropdown responds to the changeLocale... but the virtualSelectBox button seems not to..
the button of the virtualSelectBox translates ok on changeLocale, but as soon as you change selection, if backs to the previous lang

cajus:
Please review. Fixes #9002 but I'm not 100% sure if there isn't a simplier way.

### DIFF
--- a/framework/source/class/qx/test/ui/basic/Label.js
+++ b/framework/source/class/qx/test/ui/basic/Label.js
@@ -200,6 +200,20 @@ qx.Class.define("qx.test.ui.basic.Label",
       this.assertTrue(qx.lang.Array.equals(label.getBindings()[0], textfield2.getBindings()[0]), "label and textfield1 must have the same binding");
 
       label.dispose();
+    },
+
+    testLocaleInitialization : function() {
+      var label = new qx.ui.basic.Label();
+      var localeManager = qx.locale.Manager.getInstance();
+      localeManager.addTranslation("en", {"TEST" : "false"});
+      localeManager.addTranslation("de", {"TEST" : "true"});
+      localeManager.setLocale("en");
+
+      var test = qx.locale.Manager.tr("TEST");
+      localeManager.setLocale("de");
+      label.setValue(test);
+
+      this.assertEquals("true", label.getContentElement().getValue(), "label must have the current locale set");
     }
   }
 });
--- a/framework/source/class/qx/test/ui/basic/Label.js
+++ b/framework/source/class/qx/test/ui/basic/Label.js
@@ -205,15 +205,15 @@ qx.Class.define("qx.test.ui.basic.Label",
     testLocaleInitialization : function() {
       var label = new qx.ui.basic.Label();
       var localeManager = qx.locale.Manager.getInstance();
-      localeManager.addTranslation("en", {"TEST" : "false"});
-      localeManager.addTranslation("de", {"TEST" : "true"});
+      localeManager.addTranslation("en", {"TEST" : "EN"});
+      localeManager.addTranslation("de", {"TEST" : "DE"});
       localeManager.setLocale("en");
 
       var test = qx.locale.Manager.tr("TEST");
       localeManager.setLocale("de");
       label.setValue(test);
 
-      this.assertEquals("true", label.getContentElement().getValue(), "label must have the current locale set");
+      this.assertEquals("DE", label.getContentElement().getValue(), "label must have the current locale set");
     }
   }
 });
--- a/framework/source/class/qx/ui/basic/Label.js
+++ b/framework/source/class/qx/ui/basic/Label.js
@@ -532,22 +532,35 @@ qx.Class.define("qx.ui.basic.Label",
 
 
     // property apply
-    _applyValue : function(value, old)
-    {
-      // Sync with content element
-      if (value && qx.core.Environment.get("qx.dynlocale") && value.translate) {
-        this.getContentElement().setValue(value.translate());
-      }
-      else {
+    _applyValue : qx.core.Environment.select("qx.dynlocale", {
+      "true" : function(value, old)
+      {
+        // Sync with content element
+        if (value && value.translate) {
+          this.getContentElement().setValue(value.translate());
+        }
+        else {
+          this.getContentElement().setValue(value);
+        }
+
+        // Mark text size cache as invalid
+        this.__invalidContentSize = true;
+
+        // Update layout
+        qx.ui.core.queue.Layout.add(this);
+      },
+
+      "false" : function(value, old)
+      {
         this.getContentElement().setValue(value);
+
+        // Mark text size cache as invalid
+        this.__invalidContentSize = true;
+
+        // Update layout
+        qx.ui.core.queue.Layout.add(this);
       }
-
-      // Mark text size cache as invalid
-      this.__invalidContentSize = true;
-
-      // Update layout
-      qx.ui.core.queue.Layout.add(this);
-    }
+    })
   },
 
 
--- a/framework/source/class/qx/ui/basic/Label.js
+++ b/framework/source/class/qx/ui/basic/Label.js
@@ -535,7 +535,12 @@ qx.Class.define("qx.ui.basic.Label",
     _applyValue : function(value, old)
     {
       // Sync with content element
-      this.getContentElement().setValue(value);
+      if (value.translate) {
+        this.getContentElement().setValue(value.translate());
+      }
+      else {
+        this.getContentElement().setValue(value);
+      }
 
       // Mark text size cache as invalid
       this.__invalidContentSize = true;
--- a/framework/source/class/qx/ui/basic/Label.js
+++ b/framework/source/class/qx/ui/basic/Label.js
@@ -535,7 +535,7 @@ qx.Class.define("qx.ui.basic.Label",
     _applyValue : function(value, old)
     {
       // Sync with content element
-      if (value.translate) {
+      if (value && qx.core.Environment.get("qx.dynlocale") && value.translate) {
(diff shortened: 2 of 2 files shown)


cajus:
Looks like this needs investigation since it breaks other tests.

oetiker:
is `qx.dynlocale` active? looking through the code it st seem pretty magic

cajus:
dynlocale is on by default. If you have it off, locale objects do not react on locale changes.

framework/source/class/qx/ui/basic/Label.js
```diff
@@ -535,7 +535,12 @@ qx.Class.define("qx.ui.basic.Label",
     _applyValue : function(value, old)
     {
       // Sync with content element
-      this.getContentElement().setValue(value);
+      if (value.translate) {
```
level420: Does this mean, that if I have something like
```javascript
value = this.tr('the real value');
```
that then value isn't a string anymore?!?!?

oetiker: the result of this is a LocalizedString ... which will react to changing locales ... often :+1:

cajus:
You're getting a `qx.locale.LocalizedString` object. That knows about the translations. If you've a widget, it listens on `changeLocale` in case of `qx.dynlocale` set to *true*. Which is the default.

The problem above is, that the LocalizedString itself does not act on locale changes. Which is OK for performance reasons. The bug is: if you assign a variable with a LocalizedString and change the locale afterwards, the object does not reflect the current value in its __txt property. If you assign this one to a label after you've changed the locale, the Label does not call `.translate()`. It just passes it over.

If you change the locale again, everything starts to be correct again.

level420:
@cajus thank you, I was not aware of that.

What about the failing travis tests?

cajus:
I had no time to inspect them yet. So merging would be a bad idea in the moment ;-)

Maybe there's a more general solution for that. Not sure if there's time for more investigation today. Lets see...

level420:
It seems related to the PR: https://travis-ci.org/qooxdoo/qooxdoo/jobs/129851356#L455

level420:
I think this is because it may be that value is null

cajus:
Looks like it passed with the hint from @level420

kreimer:
IMHO, widgets as Label and other that show text, could test if the string is a native string or a LocalizedString (Label does that already) but instead of calling str.translate() (wich returns a new instance of localizedString) it should call str.toString(). That would be possible if __txt in baseString were protected instead of private, and LocalizedString, listen to changeLocale to set itself to 'dirty' in order to update itself in the next toString call.

cajus:
@kreimer no, you're not right in this case. It needs to be like it is done in the PR. `toString()` will return a string - and that is not what we want, because it will break dynlocale.

From my point of view we can merge the PR.

kreimer:
oh I see, sorry, . I've not taken into consideration dynlocale...

kreimer:
but given that now translate() will be called in the _applyValue

will this be more efficient? :

 _onChangeLocale : qx.core.Environment.select("qx.dynlocale",
    {
      "true" : function(e)
      {
        this._applyValue(this.getValue());
      },

      "false" : null
    }),

cajus:
`_onChangeLocale()´ is only active is dynlocale is enabled. So there's no need to do so.

The only thing that can be done to save one `if` would be to use env.select for _applyValue.

kreimer:
my point is when dynlocale is true,  translate() is called twice, inside the the changeLocale handler... and subsequently to setValue.. within the _applyValue

framework/source/class/qx/ui/basic/Label.js
```diff
@@ -535,7 +535,12 @@ qx.Class.define("qx.ui.basic.Label",
     _applyValue : function(value, old)
     {
       // Sync with content element
-      this.getContentElement().setValue(value);
+      if (value && qx.core.Environment.get("qx.dynlocale") && value.translate) {
```
oetiker: so if we are talking performance, I would call

````javascript
if (value && value.translate && qx.core.Environment.get("qx.dynlocale")) {
````
as this would make the most expensive operation be called only if both the cheep operations are ok.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Short-Circuit_Evaluation

File: framework/source/class/qx/test/ui/basic/Label.js [hunk 3]
```diff
@@ -200,6 +200,20 @@ qx.Class.define("qx.test.ui.basic.Label",
...
 
       label.dispose();
+    },
+
+    testLocaleInitialization : function() {
+      var label = new qx.ui.basic.Label();
+      var localeManager = qx.locale.Manager.getInstance();
+      localeManager.addTranslation("en", {"TEST" : "false"});
+      localeManager.addTranslation("de", {"TEST" : "true"});
+      localeManager.setLocale("en");
```
oetiker:
````javascript
     localeManager.addTranslation("en", {"TEST" : "EN"});
     localeManager.addTranslation("de", {"TEST" : "DE"});
````

File: framework/source/class/qx/test/ui/basic/Label.js [hunk 4]
```diff
@@ -200,6 +200,20 @@ qx.Class.define("qx.test.ui.basic.Label",
...
+
+      var test = qx.locale.Manager.tr("TEST");
+      localeManager.setLocale("de");
+      label.setValue(test);
+
+      this.assertEquals("true", label.getContentElement().getValue(), "label must have the current locale set");
```
oetiker:
````javascript
 this.assertEquals("DE", label.getContentElement().getValue(), "label must have the current locale set");
````
would seem a more obvious choice of translations.

framework/source/class/qx/ui/basic/Label.js
cajus: No. I'd then go the *select* way to get rid of *get("qx.dynlocale")* to optimize the non dynlocale path. The other one isn't really optimizable at all, because you need the *if*. See updated PR.

framework/source/class/qx/test/ui/basic/Label.js
cajus: Whatever. I've changed it.

cajus:
@kreimer: *translate()* basically updates the __txt value. So you need it in _onChangeLocale, otherwise the string will not be updated in DOM. A fact is that the *qx.Environment.select* for _onChangeLocale is only to save space.

Whatever - I think the way it is now, is the way to go. Remember that you've just exploited a small corner case. Normally, applications have their locale set from the beginning. In this case, the bug will not occur. It's just because all LocalizedStrings get pre-initialized.

If you want performance, disable *qx.dynlocale* and let the user reload the page after he requests an active locale change. Personally, I think that having qx.dynlocale is fancy, but not really needed in most cases.

That's why I tend to look at other bugreports now ;-)

kreimer:
I don't know if anyone else is using dynlocale. I certainly do. Thank you very much for fixing this!

cajus:
We have only one special case, where I need *qx.dynlocale* set to *true*: It's a wizard where you can choose the locale - where each string is translated into one of the 18 target languages to make it more intuitive for users.

The remaining parts of the application need a reload after a locale switch anyway: too much stuff is requested from the server which contains strings in the target locale. So it doesn't make a big difference if I reload the model or just reload the page.

oetiker:
I like it ... lets merge

Issue #9017, opened 2016-05-15
Enhance qx.bom.request.Xhr


hkollmann:
Hello,

there are some missing items of qx.bom.request.Xhr:

-  Support of native Response
- Support of setting the responseType 

With this you can handle binary data with qx.bom.request.Xhr direct.

best regards,
Henner Kollmann

oetiker:
Hi @hkollmann do you have a Pull Request for us in store ?

hkollmann:
Not yet - i should start learning to work with git. I just tried to use qx.bom.request.Xhr with binary data. It's possible to use the request object with getRequest - but that's not the qooxdoo way.

One question before i start the implemtation:
In native Xhr the responseType ist just a variable of the object. Qooxdoo uses propertys with getter and setter.

So what's to use in qx.bom.request.Xhr? Because responseText, responseXML are handeled the native way i would prefer this.

oetiker:
Look at the implementation in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/request/Xhr.js it does not use the qooxdoo setters and getters mechanism ... qooxdoo classes must hide away any differences occurring due to browser differences and provide a stable interfaces ... in that way, if there are browser changes, we often just have to fix the relevant classes which makes all the applications built on top of qooxdoo work again.

hkollmann:
Tried a pull request. My first work with git. Hopefully everything is correct.... #9018

level420:
fixed with https://github.com/qooxdoo/qooxdoo/pull/9018

hkollmann:
Changes are commited! Thank you!

PR #9018, opened 2016-05-15
Enhance qx.bom.request.Xhr #9017 : Add support for native reponse and…

hkollmann:
Here is my first try....

### DIFF
--- a/framework/source/class/qx/bom/request/Xhr.js
+++ b/framework/source/class/qx/bom/request/Xhr.js
@@ -37,6 +37,22 @@
  *  req.open("GET", url);
  *  req.send();
  * </pre>
+ *
+ * Example for binary data:
+ *
+ * <pre class="javascript">
+ *  var req = new qx.bom.request.Xhr();
+ *  req.onload = function() {
+ *    // Handle data received
+ *    var blob = req.response;
+ *    img.src = URL.createObjectURL(blob);
+ *  }
+ *
+ *  req.open("GET", url);
+ *  req.responseType = "blob";
+ *  req.send();
+ * </pre>
+ 
  * </div>
  *
  * @ignore(XDomainRequest)
@@ -226,7 +242,6 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
         async = true;
       }
       this.__async = async;
-      
       // Default values according to spec.
       this.status = 0;
       this.statusText = this.responseText = "";
@@ -414,8 +429,6 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
       // behaviour of all other browsers (Chrome, IE and Safari)
       var dataType = qx.Bootstrap.getClass(data);
       data = (data !== null && this.__dataTypeWhiteList.indexOf(dataType) === -1) ? data.toString() : data;
-
-
 
       // Some browsers may throw an error when sending of async request fails.
       // This violates the spec which states only sync requests should.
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/bom/request/Xhr.js
```diff
@@ -922,15 +934,19 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
...
       if (this.readyState >= qx.bom.request.Xhr.HEADERS_RECEIVED) {
         // In some browsers, XHR properties are not readable
         // while request is in progress.
         try {
           this.status = nxhr.status;
           this.statusText = nxhr.statusText;
-          this.responseText = nxhr.responseText;
-          this.responseXML = nxhr.responseXML;
+          this.response = nxhr.response;
+          if (typeof nxhr.response === "string") {
```
oetiker: are responseXML and responseText not defined if typeof response is not string ?
is this the case for all browsers?

oetiker:
what would be really cool, would be if you could add another example to the top of the file demonstrating how to use the new functionality.

oetiker:
as you can see from the travis-report above, the tests are not happy with your change ... click on 'details' to see where the problem ist ... also, you may want to add additional tests to https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/bom/request/Xhr.js to check the new functionality you just added.

oetiker:
oh and a note on pull requests ... best is to do

```bash
git checkout -b xhr-pull
```

before committing and pushing ... in that way your new code will be in a separate branch and your pull request will not be from 'master'

when you have fixes for this one. Just push again, the PR in github will automatically update with your next push ... you don't have to modify the PR. But since you are working in your 'master' branch you should make sure to only push changes to the PR and nothing else (hence the hint with the checkout and a separate branch for future PRs, as you may be working on multiple separate PRs).

framework/source/class/qx/bom/request/Xhr.js
hkollmann: Yes, if responseTyp is not empty or not "text" an object will be returned.
https://xhr.spec.whatwg.org/#the-responsetype-attribute

hkollmann:
Thanks for the hints, it's my first day with git. So forgive me....

hkollmann:
How can i run the tests locally? I am using an Windows enviroment...

hkollmann:
Is it possible to run the test's with Debugging?

hkollmann:
OK, i looked into the Tests. The poblem is that the tests only work with responseText.
Not with  response. This is ok if you only use responseType="text", In this case the response is copied into the responseText. 
In the test you should set fakeReq.response = "Affe". This could be compared with responseText.

oetiker:
the easiest is just to push an update to the PR, then the travis tests run automatically ... :-) 

you can also use the test runner in your browser http://qooxdoo.github.io/devel/pages/development/frame_apps_testrunner.html

oetiker:
note that your changes should not break existing code, but on the other hand, the tests in qx/test/bom/request/Xhr.js may also bee too simple so your PR should update them too.

hkollmann:
OK, test are running. I added an example in the header.
Writing a new test not possible because FakeXMLHttpRequest only supports string data in his respond method. Sorry for that

oetiker:
I think we should merge this!

level420:
OK merging.
Thank you @hkollmann !

level420:
Merge also approved by @johnspackman

PR #9020, opened 2016-05-17
Bump cldr from v25 to v29 as suggested in the technical documentation

cajus:


oetiker:
a current unicode database ... certainly a good thing :)

RemiHeugue:
I think than since this was merged, the decimal separator in fields like spinners is the wrong one for local fr (french).

It uses the "٫" (arabic decimal separator (U+066B)) (which I can't even type with my keyboard) instead of "," (comma (U+002C))

cajus:
@RemiHeugue that is just converted from the official CLDR. Looks like that there are different *numberSystem* definitions for french. I can see "arabic", "latn", etc.

I need to verify the transformation process in order to see if that's *new behaviour* or if there's something else broken.

cajus:
Ok. Need to modify the generator. It does not expect multiple number/symbol/decimal paths.

cajus:
As a quick fix, can you check whether this helps:

```
diff --git a/tool/pylib/misc/cldr.py b/tool/pylib/misc/cldr.py
index da39720..9472b88 100755
--- a/tool/pylib/misc/cldr.py
+++ b/tool/pylib/misc/cldr.py
@@ -134,7 +134,7 @@ def extractDelimiter(tree):
 def extractNumber(tree):
     data = {}
 
-    decimalSeparatorNode = tree.find("numbers/symbols/decimal")
+    decimalSeparatorNode = tree.find("numbers/symbols/decimal[@numberSystem='latn']")
     if decimalSeparatorNode != None:
         data['cldr_number_decimal_separator'] = decimalSeparatorNode.text
 
```

You need to run *generate.py* to make it happen.

cajus:
Please see #9224. Maybe @RemiHeugue can check whether the PR makes french work again for him.

PR #9021, opened 2016-05-17
Expose qx.ui.tabview.TabView radio group manager instance creation override and access

level420:
This PR allows the radio group manager to be accessed and the instance creation to be overriden by deriving classes.


derrell:
LGTM

oetiker:
then lets merge

PR #9024, opened 2016-05-25
Fix a serious error when maintaining addListenerOnce listener removal.

cajus:
When you add two _addListenerOnce_ calls with a different context, the second one gets removed after the first one was executed.

There's one place where it is even visible. Open the [current devel playground](http://qooxdoo.github.io/devel/playground) and take a look at the "Playground" string on the top left. It is cut off by an ellipsis.

### DIFF
--- a/framework/package.json
+++ b/framework/package.json
@@ -24,6 +24,6 @@
     "karma-coverage": "^0.5.5",
     "karma-coveralls": "^1.1.2",
     "karma-firefox-launcher": "^0.1.7",
-    "karma-qooxdoo": "^0.4.4"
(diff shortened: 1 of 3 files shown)


cajus:
This makes a couple of tests fail. Most likely because they're not cleaned up properly. Will update the PR soon in order to fix them.

cajus:
Thanks @peuter ;-)

peuter:
You`re welcome ;-)

level420:
Sounds all reasonable. Thank you @cajus 
I think we should merge.

johnspackman:
looks good to me :)

oetiker:
yep :)

PR #9025, opened 2016-05-25
Tweak utf-8 decoding in generator to allow unicode paths.

### LINKED ISSUE #8959: Encoding problem related to the path in which the Qooxdoo SDK resides (BZ#9232)
_Julie BOURBEILLON (julie.bourbeillon) wrote:_

When the path under which the Qooxdoo SDK has been installed contains non-ASCII characters, attempts to re-encode it in create-application.py fails with a UnicodeDecodeError.

Example scenario on a linux system:

mkdir héhé
cd héhé
unzip qooxdoo-5.0.1-sdk.zip
./qooxdoo-5.0.1-sdk/create-application.py --name=test

Will cause the following error:

Traceback (most recent call last):
  File &quot;../qooxdoo-5.0.1-sdk/tool/bin/create-application.py&quot;, line 37, in &lt;module&gt;
    SKELETON_DIR  = unicode(os.path.normpath(os.path.join(FRAMEWORK_DIR, &quot;component&quot;, &quot;skeleton&quot;)))
UnicodeDecodeError: &#39;ascii&#39; codec can&#39;t decode byte 0xc3 in position 20: ordinal not in range(128)

It is specifically the path to the qooxdoo SDK which causes the problem and not the path to the application (both cases have been tested).

OS : Debian linux
Python : 2.7

cajus:
I've located a couple of places that were problematic when utf-8 characters in path. At least it fixes the report from the ticket below. As I'm not 100% familiar with the generator code, there may be more cases. The mentioned case and usual generator calls look fine, though.

fixes #8959

### DIFF
--- a/tool/bin/create-application.py
+++ b/tool/bin/create-application.py
@@ -33,7 +33,7 @@ from misc import Path
 
 
 SCRIPT_DIR    = qxenviron.scriptDir
-FRAMEWORK_DIR = os.path.normpath(os.path.join(SCRIPT_DIR, os.pardir, os.pardir))
+FRAMEWORK_DIR = os.path.normpath(os.path.join(SCRIPT_DIR, os.pardir, os.pardir).decode('utf-8'))
 SKELETON_DIR  = unicode(os.path.normpath(os.path.join(FRAMEWORK_DIR, "component", "skeleton")))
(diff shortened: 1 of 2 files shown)


oetiker:
looks good

PR #9026, opened 2016-05-25
User getBoundingClientRect() for SVGElement

### LINKED ISSUE #195: Chrome v 50 deprecation of SVGElement.offset(Parent|Top|Left|Width|Height)
Chrome ist dropping support of SVGElement.offset(Parent|Top|Left|Width|Height) with upcoming version 50:

See https://www.chromestatus.com/features/5724912467574784

This means that trying e.g. to retreive the width and height of a inline svg element within qooxdoo does not work anymore, if this is done via elem.offsetWidth and elem.offsetHeight.

Playground example created by Alexander Kirsanov which shows the problem: http://tinyurl.com/h89s2pk

See discussion on the qooxdoo-devel mailing list:

http://qooxdoo.678.n2.nabble.com/Upgrade-for-offsetWidth-td7588186.html

cajus:
Propsal to replace the deprecated SVGElement properties. Tested in Firefox/Chrome. Maybe we should check this in other browsers, too. Anyone with IE/Edge/Safari? Comments?

fixes #195

### DIFF
--- a/framework/source/class/qx/bom/element/Location.js
+++ b/framework/source/class/qx/bom/element/Location.js
@@ -43,6 +43,8 @@
  * * Internet Explorer 6.0 + 7.0 (both standard & quirks mode)
  * * Opera 9.2
  * * Safari 3.0 beta
+ *
+ * @ignore(SVGElement)
  */
 qx.Bootstrap.define("qx.bom.element.Location",
 {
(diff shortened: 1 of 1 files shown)


level420:
Looks good to me.

oetiker:
could we have a test for this fix too, so that we catch further changes in this area?

cajus:
Not sure if that makes sense in this case. qx is not built to be SVG aware. There are several more places that have to be adjusted. What happens now, is that we're just using the SVG based API instead of the accidently working standard DOM way.

So what exactly do you want to test here? It's more a general issue.

level420:
I'm also not shure about the test. One possibility would be to test exactly what failed before: testing if the created SVG element has a reasonable height and width (or right or bottom position) using qx.bom.element.Location.

cajus:
As said, it's more a general issue and would imply adding a qx.bom.element.SVG and tests for it - which we do not have or support currently. Adding one which tests the browser SVG implementation is beyond the scope of this ticket. And it will be very time consuming, because SVG is a big bucket.

The PR just adds the SVG way of getting dimensions. But that's not the only place where you can create problems when injecting SVG tags in the DOM. I'd still recommend using the SVG contribution instead. That one captures (most) of the problems.

level420:
OK then. As qx.bom.element.Location is generic and can be used with SVG elements which may not be related to qooxdoo widgets I'm in favour of merging this PR.

oetiker:
ok second

PR #9032, opened 2016-06-01
Return removed children in _removeAll

### LINKED ISSUE #9030: qx.ui.toolbar.Toolbar.removeAll() doesn't return children array
qx.ui.toolbar.Toolbar.removeAll() doesn't return children array,
returns undefined instead

quick fix:

ToolBar.js

    // overridden
    _removeAll : function() {
      var ret = this.base(arguments);
      this._recalculateOverflow(null, 0);
      return ret;
    },

Regards, Johannes

level420:
as documented in https://github.com/qooxdoo/qooxdoo/blob/4bfad7f/framework/source/class/qx/ui/core/MChildrenHandling.js#L173

fixes #9030


oetiker:
great ... more coverage, fewer bugs ... way to go!

PR #9033, opened 2016-06-01
Let qx.ui.toolbar.ToolBar method removeAt return the removed child

level420:
As documented in the api.
Including unit test


oetiker:
more correctness and more coverage .. I like it!

Issue #9034, opened 2016-06-01
Decorator styles applied inconsistently


johnspackman:
When several mixins apply styles that overlap, the generated CSS needs to be applied in a suitable order; for example, the MBackgroundImage mixin applies the `background` style and MBackgroundColor applies the `background-color` style.

I've found that in recent versions of Qx on Chrome 50/Mac, the CSS string has `background-color` *before* `background` and Chrome compiles this into rules where `background-color` is completely ignored.  This can be seen by debugging the code and inspecting the CSSRule immediately after it has been added to the Stylesheet.

The same version of Chrome running an app with an older version of Qx 5.0.1 does not exhibit this behaviour, because it sets the `background` style before the `background-color`.

AIUI Chrome is working correctly, but Qx is not enforcing an order so is unpredictable in its results.

There is a quick fix by enforcing the ordering of style keys, please see this commit https://github.com/johnspackman/qooxdoo/commit/b9a5ed2ed4d66387b49e367a39b9979114f603a9 for a working example

It could be argued that the correct implementation is to merge MBackgroundImage and MBackgroundColor into a single mixin that understands how to completely and accurately generate a single `background` style.

As this is a fundamental change, we should consider which is the safest approach - is there a situation where simplistic alphabetical ordering of CSS style names could simply cause a different problem?

oetiker:
that does look like a neat solution to me. especially since it solves the problem in a rather generic way.

level420:
Fixed with https://github.com/qooxdoo/qooxdoo/commit/1ddf54d11b8c66bb1ba363cac532bd1cddac0c2f

PR #9035, opened 2016-06-01
Workaround for an issue where decorator styles are not applied in any specific order

johnspackman:
This fixes a bug where CSS styles are applied in a random order, but order is important eg for `background` must come before `background-color`.

Please see also https://github.com/qooxdoo/qooxdoo/issues/9034

### DIFF
--- a/framework/source/class/qx/theme/manager/Decoration.js
+++ b/framework/source/class/qx/theme/manager/Decoration.js
@@ -122,8 +122,11 @@ qx.Class.define("qx.theme.manager.Decoration",
       // create and add a CSS rule
       var css = "";
       var styles = instance.getStyles(true);
-      for (var key in styles) {
-
+      
+      // Sort the styles so that more specific styles come after the group styles, 
+      //  eg background-color comes after background
+      var t = this;
+      Object.keys(styles).sort().forEach(function(key) {
         // if we find a map value, use it as pseudo class
         if (qx.Bootstrap.isObject(styles[key])) {
           var innerCss = "";
@@ -133,14 +136,14 @@ qx.Class.define("qx.theme.manager.Decoration",
             inner = true;
             innerCss += innerKey + ":" + innerStyles[innerKey] + ";";
           }
-          var innerSelector = this.__legacyIe ? selector :
+          var innerSelector = t.__legacyIe ? selector :
             selector + (inner ? ":" : "");
-          this.__rules.push(innerSelector + key);
+          t.__rules.push(innerSelector + key);
           sheet.addRule(innerSelector + key, innerCss);
-          continue;
+          return;
         }
         css += key + ":" + styles[key] + ";";
-      }
+      });
 
       if (css) {
         sheet.addRule(selector, css);
--- a/framework/source/class/qx/theme/manager/Decoration.js
+++ b/framework/source/class/qx/theme/manager/Decoration.js
@@ -127,6 +127,7 @@ qx.Class.define("qx.theme.manager.Decoration",
       //  eg background-color comes after background.  The reordering is only applied
       //  to rules which begin with the names in REORDER; the sort order is alphabetical
       //  so that short cut rules come before actual
+      var t = this;
       Object.keys(styles).sort().forEach(function(key) {
         // if we find a map value, use it as pseudo class
         if (qx.Bootstrap.isObject(styles[key])) {
--- a/framework/source/class/qx/theme/manager/Decoration.js
+++ b/framework/source/class/qx/theme/manager/Decoration.js
@@ -127,23 +127,7 @@ qx.Class.define("qx.theme.manager.Decoration",
       //  eg background-color comes after background.  The reordering is only applied
       //  to rules which begin with the names in REORDER; the sort order is alphabetical
       //  so that short cut rules come before actual
-      var t = this;
-      var orderedStyles = [];
-      var unorderedStyles = [];
-      var REORDER = [ "background" ];
-      var savedStyles = {};
-      for (var key in styles) {
-        var needsOrdering = REORDER.some(function(start) {
-          return key.length >= start.length && key.substring(0, start.length) == start;
-        });
-        if (needsOrdering) {
-          unorderedStyles.push(key);
-        } else {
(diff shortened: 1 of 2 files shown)


oetiker:
having a test for this would be way cool :)

framework/source/class/qx/theme/manager/Decoration.js
```diff
@@ -122,8 +122,11 @@ qx.Class.define("qx.theme.manager.Decoration",
       // create and add a CSS rule
       var css = "";
       var styles = instance.getStyles(true);
-      for (var key in styles) {
-
+      
+      // Sort the styles so that more specific styles come after the group styles, 
+      //  eg background-color comes after background
+      var t = this;
+      Object.keys(styles).sort().forEach(function(key) {
```
derrell: This bugs me (as does the original unspecified order). The fix just "happens" to work because the two styles we care about happen to be lexicographicaly in the correct order, but who knows what other changes this is making. I would much prefer to see the specific values that we know of that must be in a specific order being explicitly placed in that order. (I agree, though, that using a sorted order to start with is better than the browser-specific (or worse) order provided by "for (keys in styles)".

derrell: By the way, I would leave it with the original for loop and just reorder the specific broken ones for now, to fix this bug. We should then change it to the sorted list at the next major version.

oetiker: According to https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map the keys in the map get returned in insertion order ... But since I can not see how a lexical sorting should have a negative influence, I would urge to have the keys sorted as you suggest and add appropriate documentation ... that would certainly improve the current rather random situation.

derrell: That order is for the new Map type. Object iteration order is unspecified. See http://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

level420:
It shure would be better to have this problem sorted out by implementing die style additions not being held in a hash, but @johnspackman s PR is better than the current situation, so I'll go with this for now. As @oetiker said acccompanied by documentation and of course if possible with a test.

derrell:
Ok. I just hope it doesn't break anything, given that the new order will be used for everything.

oetiker:
since the breakage will be visual, It will get noticed quickly I guess ... but I still would like to see a test added

johnspackman:
@derrell IKWYM, it isn't right to rely on alphabetical sorting, IMHO the correct solution would be to reimplement the mixins; however, this would be quite a bit of change and carry its own risks and unless we can find situations where the sorting breaks the code this patch has the advantage of being low impact.

As far as breaking code - yes this is a risk, but what Im finding is that Qooxdoo is breaking code without this fix.  I have a long established app using the Aristo theme with an older version of Qooxdoo, and by simply upgrading Qooxdoo to master the visual styles are broken because the `key in styles` is returning it in a different order.  Same browser, one tab with the live app and one with the development app.

Very little has changed in decorator handling for years, so my guess is that the change in order is to do with these commits (Jun 2015 and Feb 2016) which altered how styles were merged together:

https://github.com/qooxdoo/qooxdoo/commit/5be90454744773ceae28c149d6dac9c018c050d5
https://github.com/qooxdoo/qooxdoo/commit/edca905ec8e60c6684619fc46ce66abb2e9d7659
https://github.com/qooxdoo/qooxdoo/commit/b4130576707032af025e3712c104700d8f308c43

@oetiker Re tests: no problem, I started yesterday and got waylaid.  I'll be finishing that this morning

framework/source/class/qx/theme/manager/Decoration.js
oetiker: all the more reason to apply the patch :-)

johnspackman:
done; I also changed the code to only selectively sort the rules background*

oetiker:
@johnspackman as @derrell points out the order of keys in objects (as opposed to Maps) is undetermined, so I would much better like to have all the keys sorted alphabetically than getting them in seemingly random order.

johnspackman:
ok, fair enough ill do that now

framework/source/class/qx/theme/manager/Decoration.js
```diff
@@ -122,8 +122,13 @@ qx.Class.define("qx.theme.manager.Decoration",
       // create and add a CSS rule
       var css = "";
       var styles = instance.getStyles(true);
-      for (var key in styles) {
-
+      
+      // Sort the styles so that more specific styles come after the group styles, 
+      //  eg background-color comes after background.  The reordering is only applied
+      //  to rules which begin with the names in REORDER; the sort order is alphabetical
```
oetiker: that REORDER bit does not apply anymore, right ?

framework/source/class/qx/theme/manager/Decoration.js
```diff
@@ -122,8 +122,13 @@ qx.Class.define("qx.theme.manager.Decoration",
...
+      //  so that short cut rules come before actual
+      var t = this;
```
oetiker: in other places in the code they use `self` instead of `t` I think

oetiker:
@johnspackman you still have the limited REORDER comment in the patch ... if that was gone, it would be good to merge I think ...

johnspackman:
okies, ill do that in a sec.

BTW just seen your comment about "self" vs "t" - I used to use self, but there is a this.self() from qx.core.Object that returns the current class. Im happy to use something more descriptive than just "t" but not convinced that "self" is the best choice.

How about "_this"?

oetiker:
isn't there some standard in the code ... if not it would make sense to have one ... I thought it was self

level420:
I'd prefer (and doing it in my code) _this

cajus:
Is there a reason for not adding the context to the forEach loop? So you don't have to talk about naming it _t_, _self_ or __this_ ;-)

http://demo.qooxdoo.org/current/apiviewer/#qx.lang.normalize.Array~forEach!method_public

Should support adding the context and I'm using it here in several places.

level420:
Merging

Issue #8648, opened 2014-11-12
VirtualSelectBox does not work with form controller (BZ#8694)



qx-bug-importer:
_Martijn (mevers47) wrote:_

If using VirtualSelectBox in a form with form controller I get an error on this line:

https://github.com/qooxdoo/qooxdoo/blob/c748395261531c26954705e1c131e48c25f0a547/framework/source/class/qx/data/controller/Form.js#L181

Because VirtualSelectBox does not implement {qx.ui.core.ISingleSelection}.

PR #9036, opened 2016-06-06
#8648 : VirtualSelectBox does not work with form controller

mever:
Adds the ability to validate a virtual selectbox. #8648


oetiker:
looks good to me

oetiker:
thanks!

Issue #9039, opened 2016-06-08
Enhance Property definition with "comparator"


schnepe2:
Extension of the property definitions to be able to "override" the equality-check of the generated property setters.

```
  properties :
  {
    value : {
      check    : "Integer",
      nullable : true,
      init     : null,
      event    : "changeValue",
      compare  : function (a, b) { return Object.is(a, b); } // <= instead of regular "a===b"
    }
  }

```

One rationale for this feature request is to be able to create a (numeric) property that can distinguish between zero (0) and minus-zero (-0) ...and is asked here:
http://stackoverflow.com/questions/37681623/qooxdoo-enhance-property-to-distinguish-0-from-0

level420:
@schnepe2 as I read from the stackoverflow post you already have some code which hooks into property creation in qx.core.Property. Please share your code so we can have a look into. Thank you.

oetiker:
I would call it something else than 'compare' since it has the purpose of detecting value change ...but I like the idea ... it could also be used to make the property less sensitive to value change instead of more sensitive ...

level420:
...then maybe comparator?

johnspackman:
👍  for comparator

@schnepe2 FWIW you might like to consider the transform feature of properties - it would allow you to normalise -0 into 0 so that your example would work, although obviously that only works if you don't care about -0.

As a general case for properties, could we change the default implementation of comparator for Integers to consider -0 and 0 as the same?  It seems pretty dumb to me that there is this kind of distinction, as it is an artifact of implementation and not something arithmetic.  

It's the sort of thing that I could imagine spending hours trying to reproduce a bug caused by this (even the code `+0 === -0` returns true), and I can't imagine that there are many situations where someone would *want* -0 to be distinct.

schnepe2:
I've done several approaches to this 'issue'/'feature':

1. Brutal search & replace in qx.core.Property
    (changing the framework source!)

2. Configurable option in qx.core.Property
    (still a change of the framework source needed!)
    See attachments for the changes. I've deliberately called the config option
    "foobar" to indicate that this is *not* my favorite choice ;)

3. Runtime search & replace in defer method
    (a bit nicer since no framework change needed, but still feels ugly)
    See attachments 2 & 3 (for usage example)


That's all I've got so far.
The feature request however asked for a bit more general approach by defining a 
"comperator" function instead (we can argue about the name ;) )

/Peter


[...]

PR #9040, opened 2016-06-09
"comparator" option for properties

schnepe2:
My proposal for the feature request #9039.
Allowed variations for "`comparator`" are
-  `comparator : "a==b" // String` (inline) parameter signature bool::(var a, var b)
-  `comparator : function (a,b) { return a===b; } // Function`
-  `comparator : "__checkIt" // String` (member function)

### DIFF
--- a/framework/source/class/qx/core/Property.js
+++ b/framework/source/class/qx/core/Property.js
@@ -272,7 +272,8 @@ qx.Bootstrap.define("qx.core.Property",
       check        : null,       // Array, String, Function
       transform    : "string",   // String
       deferredInit : "boolean",  // Boolean
-      validate     : null        // String, Function
+      validate     : null,       // String, Function
+      comparator   : null        // String, Function
     },
 
 
@@ -813,6 +814,8 @@ qx.Bootstrap.define("qx.core.Property",
 
       var store = this.__getStore(variant, name);
 
+      this.__emitComparatorFunction(code, clazz, config, name);
+
       this.__emitSetterPreConditions(code, config, name, variant, incomingValue);
 
       if (incomingValue) {
@@ -887,6 +890,48 @@ qx.Bootstrap.define("qx.core.Property",
       }
 
       return store;
+    },
+
+
+    /**
+     * Emit code for the equality check evaluation
+     *
+     * @param code {String[]} String array to append the code to
+     * @param clazz {Class} the class which originally defined the property
+     * @param config {Object} The property configuration map
+     * @param name {String} name of the property
+     */
+    __emitComparatorFunction : function (code, clazz, config, name)
+    {
+      code.push('var equ=');
+
+      if (typeof config.comparator === "function")
+      {
+        code.push('function(a,b){return !!', clazz.classname, '.$$properties.',
+                  name, '.comparator.call(this,a,b);};');
+      }
+      else if (typeof config.comparator === "string")
+      {
+        var members = clazz.prototype;
+        // Name of member?
+        if (members[config.comparator]!==undefined)
+        {
+          code.push('this.', config.comparator, ';');
+        }
+        else // 'inline' code
+        {
+          code.push('function(a,b){return !!(', config.comparator, ');};');
+        }
+      }
+      else if (typeof config.comparator === "undefined")
+      {
+        code.push('function(a,b){return a===b;};');
+      }
+      else
+      {
+        throw new Error( "Invalid type for 'comparator' attribute " +
+          "of property '" + name + "' in class '" + clazz.classname + "'" );
+      }
     },
 
 
@@ -991,7 +1036,7 @@ qx.Bootstrap.define("qx.core.Property",
       );
 
       if (incomingValue) {
-        code.push('if(this.', store, '===value)return value;');
+        code.push('if(equ(this.', store, ',value))return value;');
       } else if (resetValue) {
         code.push('if(this.', store, '===undefined)return;');
       }
@@ -1427,7 +1472,7 @@ qx.Bootstrap.define("qx.core.Property",
       code.push('}');
 
       // Compare old/new computed value
-      code.push('if(old===computed)return value;');
+      code.push('if(equ(old,computed))return value;');
 
       // Note: At this point computed can be "inherit" or "undefined".
 
@@ -1474,7 +1519,7 @@ qx.Bootstrap.define("qx.core.Property",
       }
 
       // Compare old/new computed value
-      code.push('if(old===computed)return value;');
+      code.push('if(equ(old,computed))return value;');
 
       // Normalize old value
       if (config.init !== undefined && variant !== "init") {
--- a/framework/source/class/qx/core/Property.js
+++ b/framework/source/class/qx/core/Property.js
@@ -273,7 +273,7 @@ qx.Bootstrap.define("qx.core.Property",
       transform    : "string",   // String
       deferredInit : "boolean",  // Boolean
       validate     : null,       // String, Function
-      comparator   : null        // String, Function
+      isEqual      : null        // String, Function
     },
 
 
@@ -814,7 +814,7 @@ qx.Bootstrap.define("qx.core.Property",
 
       var store = this.__getStore(variant, name);
 
-      this.__emitComparatorFunction(code, clazz, config, name);
+      this.__emitIsEqualFunction(code, clazz, config, name);
 
       this.__emitSetterPreConditions(code, config, name, variant, incomingValue);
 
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/core/Property.js
```diff
@@ -891,6 +894,48 @@ qx.Bootstrap.define("qx.core.Property",
...
+      if (typeof config.comparator === "function")
+      {
+        code.push('function(a,b){return !!', clazz.classname, '.$$properties.',
+                  name, '.comparator.call(this,a,b);};');
+      }
+      else if (typeof config.comparator === "string")
+      {
+        var members = clazz.prototype;
+        // Name of member?
+        if (members[config.comparator]!==undefined)
```
johnspackman: what if the member is in a base class?

level420:
@schnepe2 Thank you for the PR! It would be great if you also can create a unit test for the new functionality. At best one for the string and one for the function type of comparator. That would be great!

framework/source/class/qx/core/Property.js
level420: So in addition to the first idea the string type comparator does not only contain some simple code which is wrapped in a function, but may also contain a reference to a class method by its name.
I think we should ommit this part and only have the inline code,
If we we call the inline code wrapped function in the context of the instance, we have access to all instance methods via ```this.```

framework/source/class/qx/core/Property.js
```diff
@@ -991,7 +1036,7 @@ qx.Bootstrap.define("qx.core.Property",
       );
 
       if (incomingValue) {
-        code.push('if(this.', store, '===value)return value;');
+        code.push('if(equ(this.', store, ',value))return value;');
       } else if (resetValue) {
```
level420: Shouldn't this be called in the instance context?

```javascript
code.push('if(equ.call(this, this.', store, ',value))return value;');
```

schnepe2:
> @schnepe2 Thank you for the PR! It would be great if you also consider to create a unit test for the new functionality. At best one for the string and one for the function type of comparator. That would be very helpfull and would make your PR complete!

I am already on it (I would never commit any change without a test :smile: !), but as I said, I am inexperienced with git / git-hub / ...

framework/source/class/qx/test/core/Property.js
```diff
@@ -570,5 +570,170 @@ qx.Class.define("qx.test.core.Property",
 
       object.dispose();
     }
+
```
level420: @schnepe2 you missed a semicolon here

framework/source/class/qx/core/Property.js
schnepe2: ...not sure, should it? It worked for me.
But please write a test-case which shows the potential problem

schnepe2: Ah! I see! and **yes** it **should** be called with the instances context!
Thanks for the hint

framework/source/class/qx/core/Property.js
schnepe2: > does this work if the method is defined in a superclass?

It seem so (last context-changes might be necessary for it)
test case is in work...

schnepe2:
I have a general question regarding throwing the exception (around line 930):
 ```
else
{
  throw new Error( "Invalid type for 'comparator' attribute " +
                   "of property '" + name + "' in class '" + clazz.classname + "'" );
}
```
should this be encapsulated in "qx.debug" definition?
 ```
else if (qx.core.Environment.get("qx.debug"))
{
  throw new Error( "Invalid type for 'comparator' attribute " +
                   "of property '" + name + "' in class '" + clazz.classname + "'" );
}
```
I think it's enough to check only in source builds ("qx.debug").
Depending on your decision I will add test-cases for this code, too

framework/source/class/qx/core/Property.js
```diff
@@ -272,7 +272,8 @@ qx.Bootstrap.define("qx.core.Property",
       check        : null,       // Array, String, Function
       transform    : "string",   // String
       deferredInit : "boolean",  // Boolean
-      validate     : null        // String, Function
+      validate     : null,       // String, Function
+      comparator   : null        // String, Function
```
derrell: Can we change this to a name that implies the meaning of the return value? Instead of comparator, how about isEqual or something of the sort?

derrell:
Please change "comparator" to "isEqual" so that the property name implies the return value.

level420:
I think that comparator is a good choice. Of course something like ```compareLogikForTestingTheEqualityOfOldAndNewValue``` would be best, but that is too much ;-)
In the end you need to read the documentation to really know what the attribute does.
The attribute ```event``` does also not completely self explain its purpose by its name.

cajus:
I can't really vote it up or down as I don't understand - or never had - the usecase. As there's much discussion on that it seems to be there.

Just from the reading, I don't understand what `comparator` means and what it is intended for. I guess a word should raise an idea of what it does in the developers mind. From the wording it feels like something that helps comparing something with something else. But what?

In fact it is applied as a check if a value has changed. Maybe just the default way to compare values is not strict enough?

Tends to break stuff if we change that. What about naming it `valueChangedCheck`? Sure, it's a bit long, but it goes conform with the `check` naming and it expresses what it is used for. And: as I didn't had the need for it for the last 10 years, it might not be used so often to get angry about the property length ;-)

level420:
The use case is where a simple comparison ```a === b``` fails. But yes we can name it ```valueChangedCheck```.

level420:
vote for ```comparator``` by adding a thumbs up emoji to this comment

level420:
vote for ```isEqual``` by adding a thumbs up emoji to this comment

level420:
vote for ```valueChangedCheck``` by adding a thumbs up emoji to this comment

derrell:
I think maybe I didn't make the point I was trying to make, clearly. I am pushing for boolean function names that imply the meaning of the result. When you see the name *comparator* and it returns true, you don't know for sure what what that means. Similarly, if you see the name *valueChangedCheck*, you aren't sure.

On the other hand, if you see *isEqual* and it returns true, you know, without consulting the documentation, that the arguments are considered equal. The same could be accomplished with *valueChanged* (as opposed to *valueChangedCheck*). With *valueChanged*, you know that if it returns true, the value has changed.

My goal here is to lean towards an API in which one needn't always be consulting the docs to know the meaning of return values (or arguments, in other cases).

johnspackman:
I think we're spending way too much time debating this - any of those names, IMHO, are just as good as the other, and whether you're looking at `apply`, `check`, `transform` or `comparator`/`isEqual`/`valueChanged*` you would have to understand the API to know how to use it

What is important is that the name is concise, easily remembered, and doesn't have any obvious overlap with other concepts.

Lets just pick one and move on

oetiker:
Since we are developing a framework, I think it is worth to discuss these things ... it is the attention to detail that makes it a joy to work with qx ... 

And since no one seems to have any point AGAINST calling it something that makes sense, lets follow @derrell  s advise ...

zaucker:
On Mon, 13 Jun 2016, Tobias Oetiker wrote:

> Since we are developing a framework, I think it is worth to discuss these
> things ...  it is the attention to detail that makes it a joy to work with
> qx ...

+1

> And since no one seems to have any point AGAINST calling it something that
> makes sense, lets follow @derrell  s advise ...

+1

Cheers,
Fritz

-- 

level420:
Hmm. So do we have a consent?

slachtar:
upvote for isEqual()

On Mon, Jun 13, 2016 at 1:27 PM, Dietrich Streifert <
notifications@github.com> wrote:

> Hmm. So do we have a consent?
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/qooxdoo/qooxdoo/pull/9040#issuecomment-225566797>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ACaj6n8JPqbvs8trC2bnlwjyVnmk95-rks5qLUzEgaJpZM4Ix9Ud>
> .
>



-- 
Slah Lachtar
Tel. (+216) 98 221 575

level420:
We have the following vote:

- ```isEqual``` : 3
- ```valueChangedCheck``` : 5
- ```comparator``` : 1

where @oetiker voted double for isEqual and valueChangedCheck.

cajus:
Maybe it should be reset because @derrell 's arguments came in later? ;-)

level420:
OK then. here we go again:

level420:
vote for ```comparator``` by adding a thumbs up emoji to this comment

level420:
vote for ```isEqual``` by adding a thumbs up emoji to this comment

level420:
vote for ```valueChangedCheck``` by adding a thumbs up emoji to this comment

level420:
@schnepe2 By voting and by final decission of @derrell : Please rework the PR to use ```isEqual``` as the attributes name instead of ```comparator```. Thank you for your help and effort.

level420:
Voting ist closed ;-)

schnepe2:
> [...] Please rework the PR to use isEqual as the attributes name instead of comparator.

Done!

framework/source/class/qx/core/Property.js
```diff
@@ -891,6 +894,48 @@ qx.Bootstrap.define("qx.core.Property",
...
+      if (typeof config.isEqual === "function")
+      {
+        code.push('function(a,b){return !!', clazz.classname, '.$$properties.',
+                  name, '.isEqual.call(this,a,b);};');
+      }
+      else if (typeof config.isEqual === "string")
+      {
+        var members = clazz.prototype;
+        // Name of member?
+        if (members[config.isEqual]!==undefined)
```
derrell: Being picky here. For consistency with line 926, should this be changed to use typeof?
```
if (typeof members[config.isEqual] !== "undefined")
```

level420: What is the difference?

derrell: members[config.isEqual] !== "undefined"
vs
typeof members[config.isEqual] !== "undefined"

The two do the same thing, but both are used in the same code, just a few lines apart, that's all.

derrell: Y'know what... too picky. Forget about it. It's fine.

level420:
Merging now. 

@schnepe2 Thank you for your work!

PR #9042, opened 2016-06-09
Add check for NaN in isNumber. Fixes #9041

### LINKED ISSUE #9041: qx.lang.Type.isNumber(NaN) returns true
If we check if NaN is a Number with qx.lang.Type.isNumber it returns true because `typeof NaN` is number. Still, according to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) 

> The global NaN property is a value representing Not-A-Number.

So intuitively when someone checks qx.lang.Type.isNumber(NaN) expects false.

[Here is the relevant playground example](http://tinyurl.com/hlx35wl)

voger:


framework/source/class/qx/lang/Type.js
```diff
@@ -101,7 +101,9 @@ qx.Bootstrap.define("qx.lang.Type",
       // JavaScript null and a null returned from calling DOM.
       // e.q. by document.getElementById("ReturnedNull").
       return (
-        value !== null && (
+        value !== null &&
+        // Check if value is NaN
+        value === value && (
```
cajus: Shouldn't this contain something like isNaN()? I guess it is a bit more intuitive, or is there a reason to do so?

voger:
According to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) isNaN coerces to number and then checks if is NaN, 

Number.isNaN() is [not supported in ie](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN)

Equality checking seems to give the [less surprises ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN#Testing_against_NaN)

framework/source/class/qx/lang/Type.js
voger: According to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) isNaN coerces to number and then checks if is NaN, 

Number.isNaN() is [not supported in ie](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN)

Equality checking seems to give the [less surprises ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN#Testing_against_NaN)

derrell: Number.isNaN may not be supported since that's from recent Ecmascript, but isNaN is supported by all. I believe that the test should include: && ! isNan(value)

voger: The idea was that since this is a general purpose function the value could be even an empty string for which isNaN("") returns false. This would serve the purpose anyway and if something went wrong there was still the last part of the && clause to pick it up. Still it felt less predictable than equality check.

PR #9043, opened 2016-06-10
Update documentation for isNumber to state explicitly that it check f…

### LINKED ISSUE #9041: qx.lang.Type.isNumber(NaN) returns true

voger:
Update documentation for isNumber to state explicitly that it check for the type of value. Fixes #9041

### DIFF
--- a/framework/source/class/qx/lang/Type.js
+++ b/framework/source/class/qx/lang/Type.js
@@ -91,6 +91,11 @@ qx.Bootstrap.define("qx.lang.Type",
     /**
     * Whether the value is a number.
     *
+    * This function checks if the _type_ of the value is Number.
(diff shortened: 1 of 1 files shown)


level420:
Looks good! Merging.

level420:
the travis run for current chrome beta fails, but this is unrelated to the PR. I think we have new incompatibilites with chrome ahead!

PR #9044, opened 2016-06-13
Updated loop context

cajus:
Added loop context to #9034.


level420:
Checks passed. Looks good. Merging

Issue #9045, opened 2016-06-13
change bugreport link in mobile doc


hkollmann:
in http://manual.qooxdoo.org/current/pages/mobile/mobile_overview.html#how-to-contribute the link to how to send a bugreport should show to github.

Discussion/Feedback should link to stackoverflow or gitter

oetiker:
the source is here ... 

https://github.com/qooxdoo/qooxdoo/blob/master/documentation/manual/source/pages/mobile/mobile_overview.rst

could you make a PR ?

hkollmann:
I can try. Any documention about how to generate the docs?
What's about 
"Devices: If you have an old smartphone (Android, iPhone, Blackberry, Windows Phone, WebOS, etc.) that you don't need anymore, you could donate it to qooxdoo. We would be happy to test qooxdoo mobile on it." 
Should i delete the sentence?

oetiker:
yes that sentence does not make much sense any more :-)

the documentation gets build automatically ... https://github.com/qooxdoo/qooxdoo/blob/master/.travis/build-site.sh

once your PR is merged

hkollmann:
Andy Build Local to Check the result?

Von meinem iPhone gesendet

oetiker:
try

```bash
cd documentation/manual
make html
```

hkollmann:
After installing Sphinx it works. Thank you!

Next question: is the Mailing list still active should i change the hint to stack overflow?

hkollmann:
And when will the changes be visible?

oetiker:
they are visible ... it's just that we have not yet gotten the official qooxdoo domains signed over to the new qooxdoo association, so for now things are only visible to the initiated :( http://qooxdoo.github.io/devel/pages/mobile/mobile_overview.html

hkollmann:
Great!
May the force be with you!

PR #9046, opened 2016-06-13
Fix mobile doc

hkollmann:
issue  #9045: change bugreport link in mobile doc


hkollmann:
How can coverage decrease when i change the doc's?

oetiker:
thanks! the coverage does tend to fluctuate a bit ... nothing to worry about :)

Issue #9050, opened 2016-06-14
shift do not work in table editor


hkollmann:
In the current version you can not enter Shift Char, e.g. A. Only with caps lock on this works.
This is confirmed for Windows, OSX and Linux for all browsers.

See gitter talk from 14.06.2016 22:05

Code to reproduce from playground example:

function createRandomRows(rowCount) {
  var rowData = [];
  var now = new Date().getTime();
  var dateRange = 400 * 24 * 60 * 60 * 1000; // 400 days
  var nextId = 0;
  for (var row = 0; row < rowCount; row++) {
    var date = new Date(now + Math.random() * dateRange - dateRange / 2);
    rowData.push([ nextId++, Math.random() * 10000, date, (Math.random() > 0.5) ]);
  }
  return rowData;
}

// window
var win = new qx.ui.window.Window("Table").set({
  layout : new qx.ui.layout.Grow(),
  allowClose: false,
  allowMinimize: false,
  contentPadding: 0
});
this.getRoot().add(win);
win.moveTo(30, 40);
win.open();

// table model
var tableModel = new qx.ui.table.model.Simple();
tableModel.setColumns([ "ID", "A number", "A date", "Boolean" ]);
tableModel.setData(createRandomRows(1000));

// make second column editable
tableModel.setColumnEditable(1, true);

// table
var table = new qx.ui.table.Table(tableModel).set({
  decorator: null
});
win.add(table);

var tcm = table.getTableColumnModel();

// Display a checkbox in column 3
tcm.setDataCellRenderer(3, new qx.ui.table.cellrenderer.Boolean());

// use a different header renderer
tcm.setHeaderCellRenderer(2, new qx.ui.table.headerrenderer.Icon("icon/16/apps/office-calendar.png", "A date"));

hkollmann:
I try to debug down to browser events.
 The key down event is not generated with shift pressed.

level420:
@hkollmann you could use current master playground at http://qooxdoo.github.io/devel/playground/

level420:
but confirmed by @derrell

level420:
also confirmed on windows 7 with chrome 52, IE 1, FF 48beta

example by @hkollmann in current master playground: http://tinyurl.com/h3sblg3

hkollmann:
This is just the table demo in the playground!
http://qooxdoo.github.io/devel/playground/#Table-ria

hkollmann:
Just played a round with git bisect - This is the result:

[...]

PR #9051, opened 2016-06-15
fix for issue #9050 shift do not work in table editor

hkollmann:
fix for issue #9050 shift do not work in table editor


derrell:
LGTM. Sounds from the gitter discussions like the proper regression testing has been done here, too.

level420:
Second that!

johnspackman:
@hkollmann Thanks for the PR!

PR #9058, opened 2016-06-17
code clean up: deprecate _onKeyPress and introduce _onKeyDown instead

hkollmann:
To have a clean code base where method names match to the event i suggest to deprecate _onKeyPress for the next major release and use _onKeyDown instead.

This pull request add the new method an the deprecate hints.


oetiker:
makes sense ...

level420:
4 votes and more than 2 business days.
merging
Thank you @hkollmann

PR #9059, opened 2016-06-17
adds support for Array.find and Array.findIndex

johnspackman:


slachtar:
in the comments for the findIndex function, please change <code>find</code>
by <code>findIndex</code>
Regards

On Fri, Jun 17, 2016 at 3:05 PM, Coveralls <notifications@github.com> wrote:

> [image: Coverage Status] <https://coveralls.io/builds/6642798>
>
> Coverage decreased (-0.02%) to 55.663% when pulling *63b0083
> <https://github.com/qooxdoo/qooxdoo/commit/63b00838b86672d97fb1de1c4a4aee363a08ae03>
> on johnspackman:normalize-Array-methods* into *cfcd453
> <https://github.com/qooxdoo/qooxdoo/commit/cfcd4533cfcd9c6b48e1976c0d5176ed1b25854f>
> on qooxdoo:master*.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/qooxdoo/qooxdoo/pull/9059#issuecomment-226777601>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ACaj6qPWTVKwcp6meeqLIgUOEjT55Se9ks5qMqmcgaJpZM4I4YQJ>
> .
>



-- 
Slah Lachtar
Tel. (+216) 98 221 575

level420:
It would be nice to have tests for this PR

johnspackman:
there are tests ....

oetiker:
looking good to me

level420:
Oops! Yes Tests are there! Cool.

cajus:
@johnspackman as @slachtar notes the docstring for *findIndex()* contains *find()* in its text. The rest looks good to me ;-)

johnspackman:
@slachtar @cajus darn, didnt quite make a perfect PR there 😄  Thanks, updated.

framework/source/class/qx/bom/client/EcmaScript.js
```diff
@@ -117,6 +117,26 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
 
 
     /**
+     * Checks if 'find' is supported on the Array object.
+     * @internal
+     * @return {Boolean} <code>true</code>, if the method is available.
+     */
+    getArrayFind : function() {
```
derrell: Based on the "intuitive return value function names" discussion we had recently, I'd suggest that ascertaining if something exists or is available should include one of those words in the name. I'd suggest *arrayFindAvailable* or *arrayFindExists*. That way, a return value of true clearly indicates that it is available or does exist. (I like the *available* form better.)

framework/source/class/qx/lang/normalize/Array.js
```diff
@@ -251,6 +251,58 @@ qx.Bootstrap.define("qx.lang.normalize.Array", {
...
+     */
+    find : function(callback, obj) {
+      if (qx.core.Environment.get("qx.debug")) {
+        qx.core.Assert.assertFunction(callback);
+      }
+      
+      var l = this.length;
+      for (var i = 0; i < l; i++) {
+        var value = this[i];
+        if (callback.call(obj || window, value, i, this)) {
```
derrell: Let's be careful with the use of *window* since it exists only the browser environment. qooxdoo is also very useful for its OO features, in a backend app, where *window* does not exist. *global* is used as a replacement in nodejs. We may want to make this *obj || window || global*, but should check whether *window* is mapped in the start-up code when using the standalone qooxdoo variant.

derrell: https://github.com/qooxdoo/qooxdoo/blob/master/tool/data/generator/oo.loader.tmpl.js#L3

window is defined for the oo variant, but it's defined only through closure, not really as a global variable. Sigh. We should probably fix oo.loader.tmpl.js to use an existing global context if one exists, e.g., in NodeJS, use *global*. I think what you have here is fine. We can assume that the window variable is global (even though in some environments, currently, it is not).

framework/source/class/qx/bom/client/EcmaScript.js
johnspackman: I'd agree except that it follows the convention of all the other Array feature detection methods in that class of "getArray" + the name of the method, eg getArrayReduce, getArrayEvery, getArraySome, etc

level420: I'd leave the naming convention "getArray..." as is.

derrell: Yes, existing code trumps preferred naming. (Maybe at some point we fix all of 'em, but that's not part of this PR.)

oetiker:
LGTM

level420:
4 votes and more than 2 business days
merging
Thank you @johnspackman

Issue #9053, opened 2016-06-15
Cannot override a setter function like setWidth


johnspackman:
A question on SO http://stackoverflow.com/questions/37832341/cannot-override-a-setter-function-like-setwidth asks why setter functions cannot be overridden; this issue is to discuss the pros and cons of whether this is a bug or a feature.

My perspective is that any non-private function in a base class can be overridden by a derived class, and this should include auto generated functions which are created to implement properties.  The justification for this is that 

(1) although most properties are implemented via the ```properties``` section of a class, many Qooxdoo classes implement properties themselves - i.e. they have accessors and events, but there is no formal property defined.  They do this because sometimes they do not want the property value changed, ever, so they implement the property manually so that the setXxx method manipulates a value instead of replacing it.  For example `qx.ui.core.MSingleSelectionHandling` and `MultiSelectionHandling` have setSelection, which change the array behind the selection pseudo property rather than replacing it (this is a really cool and useful feature)

(2) when binding to properties, Qooxdoo judges whether a property is a bindable property by a kind of feature detection, rather than whether there is a formal property defined - EG is there a setXxx, getXxx, and a changeXxx event.

(3) given that it is perfectly valid to manually implement setXxx/getXxx/changeXxxx in order to have a property, it is valid to override those methods in derived classes.  If the implementation of the property changes (to be formally defined via properties) then derived code is broken, despite zero visible changes.  *How* the property is defined is not part of the specification - the “how” is effectively a private implementation detail, and the consumer of that class should not be broken by changes in private implementation.

(4) it is impossible to reliably distinguish functions which are property accessors from those which are explicitly defined - just because a function starts “set”, “get”, or “is” should not mean that it has special meaning to the OO concept of classes, e.g. it should definitely not mean that it cannot be overridden.

(5) what is the nature of an auto-generated function and what makes it special?  For example, If someone writes a babel plugin that generates methods to implement new features, are they overridable?  And how would the user, or the framework, know that they cannot be overridden?  What if that generated method overrides some other base class method that was previously manually defined?

I did a search of the Qooxdoo code for setXxx methods that are manually defined (reg ex of ```^\s*set[\w]+\s*:```) and found 591 instances; that reg ex might need improving but at a quick glance thats a lot of properties accessors that can be legitimately overridden.

I understand that overridding accessors suggests that there is a design flaw, but IMHO that doesn't mean it's always invalid - overriding a property accessor could generate a warning (definitely a lint message!).

[...]

PR #9060, opened 2016-06-17
fix to allow property accessors to be overridden

johnspackman:
See https://github.com/qooxdoo/qooxdoo/issues/9053

### DIFF
--- a/framework/source/class/qx/Class.js
+++ b/framework/source/class/qx/Class.js
@@ -1329,9 +1329,11 @@ qx.Bootstrap.define("qx.Class",
             throw new Error('Overwriting member "' + key + '" of Class "' + clazz.classname + '" is not allowed!');
           }
 
-          if (proto[key] != undefined && proto[key].$$propertyMethod) {
-            throw new Error('Overwriting generated property method "' + key + '" of Class "' + clazz.classname + '" is not allowed!');
-          }
+        }
+
+        // If it's a property accessor, we need to install it now so that this.base can refer to it
+        if (proto[key] != undefined && proto[key].$$install) {
+        	proto[key].$$install();
         }
 
         // Added helper stuff to functions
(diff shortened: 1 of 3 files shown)


oetiker:
oh :-) you have been burning the midnight oil :+1:

derrell:
LGTM.

johnspackman:
@oetiker yes, it was bugging me and I had to check it out! :)

johnspackman:
PS this PR is intended primarily as a proof of concept, ie to demonstrate that it is possible to make the property system work this way.

Can I just double check though: is this a change we want to make?  @derrell you had some concerns on gitter, is your LGTM that you're OK with the rationale in issue #9053?

derrell:
My concern on gitter was that the code was never intended to allow overriding the automatically-generated setter, so any attempt to do so was a hack. This PR alters the mechanism so that it's a feature to be able to override automatic setters. We should, at some point, go through and see how many of the existing manual setters and getters that were implemented that way because properties could be used previously, could be simplified based on this PR. So yes, I'm ok with the rationale.

derrell:
My one concern is that automatically-generated setters/getters don't now all get created up front, before being used. That would add, possibly significantly, to memory usage for functions that may never be called. It looks to me like they are still only created as needed (or, now, when overridden). John, can you confirm?

johnspackman:
yes, they're still on demand, it's just that previously the only way of triggering the demand was by actually calling the method; now when overriding a method, we can trigger the demand via the setXxx.$install method

derrell:
That's what I thought. Great.

oetiker:
LGTM

level420:
This PR has 4 votes and is older than 2 business days.
Merging.
Thank you @johnspackman

level420:
Fixes #9053

Issue #210, opened 2016-04-28
Methods from superinterface not recognized by API generator as inherited


dteleguin:
If there is more than one interface in the inheritance chain, API generator fails to recognize inherited  methods as such. For example:

* interface `IFoo` defines+documents method `foo(x)`
* interface `IBar` extends `IFoo` and defines+documents method `bar(x)`
* class `Clazz` implements `IBar` and both methods

Invoking `generate.py api-data` results in a warning (line 10 is exactly where `foo(x)` is implemented):
```
- Warning: foo.Clazz (10,14): Parameter is not documented: 'x'
- Warning: foo.Clazz (10,14): Documentation is missing.
```
In the API viewer, `Clazz#foo` will be shown as broken.

As a workaround, one can explicitly list all the interfaces implemented by the class; in our case, that would be `implement: [ IFoo, IBar ]`. This will make the warning go away, and will fix method display in the API viewer.

level420:
@thron7 would please be so kind to have a look into this? I think big parts of the generator are written by you. Thank you.

thron7:
@level420 Will have a look, but might take some time.

level420:
@thron7 thank you!

level420:
Fixed via https://github.com/qooxdoo/qooxdoo/commit/72388c7cdbcc6390a3ade19b41ba33eaf1525061

PR #9062, opened 2016-06-18
fix for issue #210

thron7:
This fixes issue #210 for a skeleton app as described in the issue (with two interfaces extending each other).

The underlying bug was that mixins and interfaces where treated like classes when climbing the inheritance hierarchy. Unfortunately, classes encode their super classes as a string attribute in the AST, whereas interfaces (and, I presume, mixins) encode their corresponding super interfaces as a list of dedicated child nodes. So I split the code to treat these different encodings individually.

I haven't tested this extensively, particularly not for mixins. But at least the new code doesn't seem to break things, running 'api-verify' in the framework yielded the same results before and after the change.

Obviously, "the right thing" would be to have the ASTs modelled more homogeniously for those kinds of meta objects, but that would be a major refactoring which I wouldn't dare at the moment.

### DIFF
--- a/tool/pylib/ecmascript/backend/api.py
+++ b/tool/pylib/ecmascript/backend/api.py
@@ -1412,10 +1412,15 @@ def dependendClassIterator(docTree, classNode):
     if superClassName:
         directDependencies.append(superClassName)
 
-    for list_ in ["mixins", "interfaces", "superMixins", "superInterfaces"]:
+    for list_ in ["mixins", "interfaces"]:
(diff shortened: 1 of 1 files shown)


oetiker:
@dteleguin can you please test the fix

cajus:
Just tested localy. Works fine.

LGTM

level420:
We have five thumbs up here. 4 in the head description and one in the LGTM comment of @cajus.
Merging

level420:
Thank you for the PR @thron7 !

Issue #9061, opened 2016-06-18
qx.bom.element.AnimationHandle does not report properly the state


voger:
Overview:
Calling isEnded() and isPlaying() in a qx.bom.elementAnimationHandle object
when the animation is over, doesn't report properly the state of the animation.

Steps to Reproduce:
Look at this playground example: http://tinyurl.com/oblnlot
Open the log. Press the button and let the animation finish. **After the
animation is over** hoover the mouse cursor over the button. Observe the log.

Actual Results:
Every time the cursor enters the button in the log is printed:
_animation.isEnded(): false
_animation.isPlaying(): true

Expected Results:
Every time the cursor enters the button in the log is printed:
_animation.isEnded(): true
_animation.isPlaying(): false

PR #9063, opened 2016-06-18
Fix animation report

voger:
See #9061
I would like to hear comments and suggested improvements.

### DIFF
--- a/framework/source/class/qx/bom/element/AnimationHandle.js
+++ b/framework/source/class/qx/bom/element/AnimationHandle.js
@@ -135,7 +135,7 @@ qx.Bootstrap.define("qx.bom.element.AnimationHandle",
     },
 
     /**
-     * set the animation state to ended
+     * Set the animation state to ended
      */
     __setEnded : function(){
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/bom/element/AnimationHandle.js
```diff
@@ -131,6 +132,11 @@ qx.Bootstrap.define("qx.bom.element.AnimationHandle",
         this.stopped = true;
         qx.bom.element.AnimationJs.stop(this);
       }
+    },
+
+    __setEnded : function(){
```
oetiker: documentation is missing here

voger: I though it was not needed since it was a private function. I will promptly fix it.

johnspackman:
LGTM

level420:
We have 4 thumbs up in the head description and the LGTM comment of @johnspackman. That makes 5 votes.
Merging.
Thank you @voger !

PR #9064, opened 2016-06-19
Add support to override the built in comparison string

cajus:
When we load a webfont, we use a comparison string to check whether the font has been loaded properly or not. This string is hard coded to some ASCII characters.

For fonts that mainly contain icons, it is very likely that these ASCII characters are not set at all. In this case the validation test fails and the font is "unloaded". That's unwanted, because we cannot load most icon fonts then.

The PR addresses it by introducing a font theme `comparisonString` property which is used instead of the default one. It injects it in the source map to avoid API changes. The source map is only used internally and the additional property is not used anywhere else.


oetiker:
hmmm have we lost travis ?

level420:
I think it would be helpful to have a note about the new property comparisonString in the parameter documentation here: https://github.com/qooxdoo/qooxdoo/pull/9064/files#diff-6e619d02c49683e8d4ba40b296d8f9c7R130

cajus:
@level420 what do you mean by "here"? It is already documented in the Font class.

cajus:
Ah. It scrolls the line outside the view when you click on the link. Verry irritating. I don't think it's needed, because there are just some "e.g." properties listed there. Its not everything which can be found in the source map.

level420:
OK then. LGTM

framework/source/class/qx/bom/webfonts/Validator.js
```diff
@@ -123,6 +128,14 @@ qx.Class.define("qx.bom.webfonts.Validator", {
       apply : "_applyFontFamily"
     },
 
+    /**
+     * Comparison string used to check whether the font has loaded or not.
+     */
+    comparisonString :
+    {
+      nullable : true,
+      init : null
```
derrell: Is there a reason not to just eliminate COMPARISON_STRING and set the init value here to what that string was? Then at line 310, the code needn't look at this as a separate default.

cajus: There was no special reason for that. Static storage is once, property for every instance. There's only one place in the code where it is or'ed to the default.

derrell: Agreed, except that if someone is looking at the code and sees the property definition, they won't know what its default is. Not a big deal, but putting it right in the property could make the code easier to read. (When I was reading it, I had to go find usages in order to figure out that there was a static value defined elsewhere.)

framework/source/class/qx/bom/webfonts/Validator.js
```diff
@@ -307,7 +306,7 @@ qx.Class.define("qx.bom.webfonts.Validator", {
       }
 
       var elem = document.createElement("span");
-      elem.innerHTML = comparisonString || qx.bom.webfonts.Validator.COMPARISON_STRING;
+      elem.innerHTML = this.getComparisonString();
```
cajus: What do you mean?

derrell: Line 296 still has comparisonString as a parameter, but it appears to be not used in the function. Change parameter list to be only (fontFamily) ?

cajus: Ok. I'll revert the last commit. The initial version was correct, because we need to distinguish between two types of comparison strings.

oetiker:
VOTE PLEASE

derrell:
Ship it!

level420:
4 votes and more than 2 business days.
Merging.
Thank you @cajus

PR #9067, opened 2016-06-22
Update demo

woprandi:
I updated one of the demo to add a button to show/hide a table column. I also added a small example in the Filtered class

### DIFF
--- a/framework/source/class/qx/ui/table/model/Filtered.js
+++ b/framework/source/class/qx/ui/table/model/Filtered.js
@@ -20,6 +20,51 @@
 /**
  * A filtered table model to provide support for hiding and filtering table
  * rows. Any rows that match any applied filters will be hidden.
+
+<pre class='javascript'>
+var model = new qx.ui.table.model.Filtered();
+model.setColumns(["Login", "Name", "Email"], ["login", "name", "email"]);
+
+var table = new qx.ui.table.Table(model);
+
+var data = [{
+  login : "darthvader",
+  name : "Darth Vader",
+  email : "darthvader@tatooine.org"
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/table/model/Filtered.js
```diff
@@ -20,6 +20,51 @@
...
+}, {
+  login : "luke",
+  name : "Luke Skywalker",
+  email : "luke@tatooine.org"
+}, {
+  login : "obi-wan",
+  name : "Obi-Wan Kenobi",
+  email : "obiwan@jedi.org"
+}];
+
```
oetiker: can't we have the address of Rey too ?

woprandi:
@oetiker just for you ;)

framework/source/class/qx/ui/table/model/Filtered.js
```diff
@@ -20,6 +20,55 @@
...
+  name : "Luke Skywalker",
+  email : "luke@tatooine.org"
+}, {
+  login : "obi-wan",
+  name : "Obi-Wan Kenobi",
+  email : "obiwan@jedi.org"
+}, {
+  login : "rey",
+  name : "Rey",
+  email : rey@jakku.sw
```
oetiker: quotes are missing

zaucker:
Wouldn't it make sense  to call the button variable "checkbox" instead? It is a checkbox and not a button after all ...

woprandi:
@zaucker You're right ;)

level420:
five votes!
merging
Thank you @woprandi

PR #9069, opened 2016-06-25
change map to ol3

hkollmann:
change the map example to the modern ol3 classes


oetiker:
thanks for your contributions

level420:
Didn't check that, but do we have tests for the openlayer integration?

oetiker:
note this is part of the showcase .... not part of the library

cajus:
@level420 I don't think that there are tests for the demos at all.

level420:
Ah! Now I understand! Thank you all

cajus:
Just gave it a try. Seems to work fine. Thanks ;-)

level420:
3 thumbs up and more than 2 business days.
merging.
Thany you @hkollmann

Issue #9071, opened 2016-06-27
"change" event fired too much on VirtualSelectBox


woprandi:
Hi,

Check this example : http://tinyurl.com/hjol7rx
The event "change" is fired 3 times instead of once

zaucker:
The three change events happen upon the call to setModel(). Once all is
setup, there is only on event per selection change.

If you add the listener after the setModel() call, the three initial events
are ignored (obviously).

I guess the setModel() generates an event for each element added to the
model. I am not sure if this is useful or not or a bug.

On Mon, 27 Jun 2016, William Oprandi wrote:

> Check this example : http://tinyurl.com/hjol7rx
> The event "change" is fired 3 times instead of once

-- 
Oetiker+Partner AG              tel: +41 62 775 9903 (direct)
Fritz Zaucker                        +41 62 775 9900 (switch board)
Aarweg 15                            +41 79 675 0630 (mobile)
CH-4600 Olten                   fax: +41 62 775 9905
Schweiz                         web: www.oetiker.ch

slachtar:
I did a lot of hacking to avoid such problem (event being generated with
every element in the model).
The change event should be generated only once.

On Mon, Jun 27, 2016 at 4:19 PM, Fritz Zaucker <notifications@github.com>
wrote:

> The three change events happen upon the call to setModel(). Once all is
> setup, there is only on event per selection change.
>
> If you add the listener after the setModel() call, the three initial events
> are ignored (obviously).
>
> I guess the setModel() generates an event for each element added to the
> model. I am not sure if this is useful or not or a bug.
>
> On Mon, 27 Jun 2016, William Oprandi wrote:
>
> > Check this example : http://tinyurl.com/hjol7rx
> > The event "change" is fired 3 times instead of once
>
> --
> Oetiker+Partner AG tel: +41 62 775 9903 (direct)
> Fritz Zaucker +41 62 775 9900 (switch board)
> Aarweg 15 +41 79 675 0630 (mobile)
> CH-4600 Olten fax: +41 62 775 9905
> Schweiz web: www.oetiker.ch
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/qooxdoo/qooxdoo/issues/9071#issuecomment-228778081>,
> or mute the thread
[...]

PR #9073, opened 2016-06-28
This fixes issue https://github.com/qooxdoo/qooxdoo/issues/9071 where events are fired incorrectly during startup

johnspackman:
Please see issue https://github.com/qooxdoo/qooxdoo/issues/9071.

The `getSelection().removeAll()` is not necessary because it is already being kept synchronised via change listener, and event ordering means that `removeAll()` undoes the work done by synchronisation

### DIFF
--- a/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
+++ b/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
@@ -316,7 +316,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @param event {qx.event.type.Data} The change event.
      */
     _onChangeModel : function(event) {
-      this.getSelection().removeAll();
     },
 
 


framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -316,7 +316,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @param event {qx.event.type.Data} The change event.
      */
     _onChangeModel : function(event) {
-      this.getSelection().removeAll();
     },
```
woprandi: I think you could remove entire method and the bound listener

johnspackman: I left it in because it's protected and could be overridden in derived classes

level420:
More than 2 business days and 3 votes.
Merging.
Thank you @johnspackman

PR #9076, opened 2016-07-01
String "startsWith()" and "endsWith()" polyfills ...ECMAScript 2015 (6th Edition, ECMA-262)

schnepe2:
Hi there,

here's my PR for the polyfills for the String prototype methods `startsWith` and `endsWith`.

I always found it irritating that there are still some clients (IE) missing this and I had to use
`qx.lang.String.startsWith(str, "foo")` instead of the easy to read `str.startsWith("foo")` :smile: 

Including tests & environment checks

References:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

### DIFF
--- a/framework/source/class/qx/lang/String.js
+++ b/framework/source/class/qx/lang/String.js
@@ -189,24 +189,28 @@ qx.Bootstrap.define("qx.lang.String",
     /**
      * Check whether the string starts with the given substring
      *
+     * @deprecated {6.0} Please use String instance startsWith method instead
+     *
      * @param fullstr {String} the string to search in
      * @param substr {String} the substring to look for
      * @return {Boolean} whether the string starts with the given substring
      */
     startsWith : function(fullstr, substr) {
-      return fullstr.indexOf(substr) === 0;
+      return fullstr.startsWith(substr);
     },
 
 
     /**
      * Check whether the string ends with the given substring
      *
+     * @deprecated {6.0} Please use String instance endsWith method instead
+     *
      * @param fullstr {String} the string to search in
      * @param substr {String} the substring to look for
      * @return {Boolean} whether the string ends with the given substring
      */
     endsWith : function(fullstr, substr) {
-      return fullstr.substring(fullstr.length - substr.length, fullstr.length) === substr;
+      return fullstr.endsWith(fullstr);
     },
 
 
--- a/framework/source/class/qx/lang/String.js
+++ b/framework/source/class/qx/lang/String.js
@@ -210,7 +210,7 @@ qx.Bootstrap.define("qx.lang.String",
      * @return {Boolean} whether the string ends with the given substring
      */
     endsWith : function(fullstr, substr) {
-      return fullstr.endsWith(fullstr);
+      return fullstr.endsWith(substr);
     },
 
 
--- a/framework/source/class/qx/bom/client/EcmaScript.js
+++ b/framework/source/class/qx/bom/client/EcmaScript.js
@@ -207,6 +207,27 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
       return !!Date.now;
     },
 
+
+    /**
+     * Checks if 'startsWith' is supported on the String object.
+     * @internal
+     * @return {Boolean} <code>true</code>, if the method is available.
+     */
+    getStringStartsWith : function() {
+      return typeof String.prototype.startsWith === "function";
+    },
+
+
+    /**
+     * Checks if 'endsWith' is supported on the String object.
+     * @internal
+     * @return {Boolean} <code>true</code>, if the method is available.
+     */
+    getStringEndsWith : function() {
+      return typeof String.prototype.endsWith === "function";
+    },
+
+
     /**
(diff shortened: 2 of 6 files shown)


cajus:
Are there now two different implementations of startsWith / endsWith - one using the polifill mechanism and one in qx.lang.String?

oetiker:
yes they are called the same, but the ECMA versions have an additional position parameter and they are string methods not static calls

cajus:
Ok. I guess the duplication can't be worked around without chaning the API. Or can the ECMA versions be used as a base for the existing ones? That's the only thing I could argument against it. If they're too different to have the same code base, lets go with them.

schnepe2:
Wouldn't "_mark them deprecated_" work for the next release?
I'm pretty sure not many users would miss those static functions. And a migration would be easy as
replace "qx.lang.String.startsWith(foo, bar)" by "foo.startsWith(bar)". So something that **is** manageable :smile: 
The static `qx.lang.String` methods could use the ECMA versions, though:

```
qx.Bootstrap.define("qx.lang.String",
{
  statics :
  {
    // ...
    startsWith : function(fullstr, substr) { return fullstr.startsWith(substr); },
    endsWith : function(fullstr, substr) { return fullstr.endsWith(substr); },
    // ...
  }
});

```

No magic, no API change (that I can think of), no problem.

cajus:
I like the idea of deprecating it (at least add a deprecation warning). It's enough to have the functionality once, I guess. And the migration job should be easy to add.

schnepe2:
I'll be happy to add the deprecation hint to my PR.
Since **when** should it be deprecated? Version 5.0.2?
In that case, here are my proposals (including deprecation warning information):
```
    /**
     * Check whether the string starts with the given substring
     *
     * @deprecated {5.0.2} Please use String instance startsWith method instead
     *
     * @param fullstr {String} the string to search in
     * @param substr {String} the substring to look for
     * @return {Boolean} whether the string starts with the given substring
     */
    startsWith : function(fullstr, substr) {
      return fullstr.startsWith(substr);
    },


    /**
     * Check whether the string ends with the given substring
     *
     * @deprecated {5.0.2} Please use String instance endsWith method instead
     *
     * @param fullstr {String} the string to search in
     * @param substr {String} the substring to look for
     * @return {Boolean} whether the string ends with the given substring
     */
    endsWith : function(fullstr, substr) {
      return fullstr.endsWith(substr)
    },
```

oetiker:
we have not yet written down the policy, but I guess deprecation would only happen on major version changes ... 6.0.0 in that case

schnepe2:
So, should I add these changes (with `5.0.2` changed to `6.0`) into the PR?

level420:
@schnepe2 yes but please use 6.0 as the deprecation version.

schnepe2:
@level420 done (fingers crossed that I did not forgot any commas or semicolons as I usually do :grinning: )

level420:
Let's see what travis says ;-)

framework/source/class/qx/lang/String.js
```diff
@@ -189,24 +189,28 @@ qx.Bootstrap.define("qx.lang.String",
...
      *
+     * @deprecated {6.0} Please use String instance endsWith method instead
+     *
      * @param fullstr {String} the string to search in
      * @param substr {String} the substring to look for
      * @return {Boolean} whether the string ends with the given substring
      */
     endsWith : function(fullstr, substr) {
-      return fullstr.substring(fullstr.length - substr.length, fullstr.length) === substr;
+      return fullstr.endsWith(fullstr);
```
level420: @schnepe2 this should be: ```fullstr.endsWith(substr);```

schnepe2: Sure! I'm really not a git guy (it's much to many steps to do things....) so therefore I use the web UI...sorry

schnepe2:
Once we are at this point: Shouldn't I replace all uses of `qx.lang.String.startsWith` & `qx.lang.String.endsWith` within the framework, too?
I think it would be kind of strange to deprecate methods but still use 'em all over the place.
But here comes a question: What to do at `qx.module.util.String` ? Depecate them, too? What's your recommendation here?

level420:
I think we should go wit this PR as is and use another PR to replace qx.lang.String.<startsWith|endsWith> in the framework.

johnspackman:
I agree we should go with the PR, but I'd be happy if the qx.lang.String is replaced thoughout the framework at the same time, provided that it is done as a separate commit to any other changes.

level420:
5 repo writers thums up.
Merging
Thank you @schnepe2

level420:
@schnepe2 It would be great if you create another PR which replaces ```qx.lang.String.startsWith``` and ```qx.lang.String.endsWith``` in the framework.

schnepe2:
...I'll do my very best

schnepe2:
FYI: #9089

PR #9075, opened 2016-07-03
Custom form fields, missing array and model field interfaces

mever:
Hello,

To make a new field for cooperation with the form component one needs to implement `{qx.ui.form.IForm}` and an interface for setting, getting and resetting the input value. Currently these interfaces have been defined for the latter (all from the `{qx.ui.form}` package):
`IRange`, `INumberForm`, `IStringForm`, `IDateForm`, `IColorForm`, `IBooleanForm` and the little more complicated; `IModelSelection`. Leaving out `{qx.ui.form.IExecutable}` and `{qx.ui.form.IModel}` as those are not meant for accessing the input value, the later only as model for the selection which is actually representing the input value.

The two interfaces I am missing now (from the `{qx.ui.form}` package) are: `IArrayForm` and `IModelForm`. Allowing access to {qx.data.Array} and {qx.core.Object} values, respectively.

I also think it shouldn't matter if the form knows if a field contains a selection or not. In some fields the input value may be a selection while in others the input value may be a list. Both input values represented by an `{qx.data.Array}`. While the former field has a model from which the selection is derived, the later has none (as the value is directly added by the end-user). But this is a side track I would like to discuss further before making a PR.

Using this perspective of the form component I already build a form manager which replaced the task of form controller in my application, allowing form models with nested values (i.e. order model with order lines array). This manager already uses these 'complex' fields using downstream constructs. But now I am confident enough to discuss the possibility of adding them as interfaces upstream.

If the community likes this approach I am happy to contribute a PR.

Regards,
Martijn

### DIFF
--- a/framework/source/class/qx/test/data/controller/FormWithArrayAndModel.js
+++ b/framework/source/class/qx/test/data/controller/FormWithArrayAndModel.js
@@ -8,8 +8,7 @@
      2016 Martijn Evers, The Netherlands
 
    License:
-     LGPL: http://www.gnu.org/licenses/lgpl.html
-     EPL: http://www.eclipse.org/org/documents/epl-v10.php
+     MIT: https://opensource.org/licenses/MIT
      See the LICENSE file in the project's top-level directory for details.
 
    Authors:
--- a/framework/source/class/qx/test/data/controller/FormWithArrayAndModel.js
+++ b/framework/source/class/qx/test/data/controller/FormWithArrayAndModel.js
@@ -1,0 +1,146 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016 Martijn Evers, The Netherlands
+
+   License:
+     LGPL: http://www.gnu.org/licenses/lgpl.html
+     EPL: http://www.eclipse.org/org/documents/epl-v10.php
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Martijn Evers (mever)
+
+************************************************************************ */
+
+/**
+ * @ignore(qx.test.data.controller.fixture.ArrayField)
+ */
+
+qx.Class.define("qx.test.data.controller.FormWithArrayAndModel",
+{
+  extend : qx.dev.unit.TestCase,
+
+  members :
+  {
+    /** @type {qx.test.data.controller.fixture.ArrayField} */
+    __arrayField : null,
+
+    /** @type {qx.ui.form.Form} */
+    __form : null,
+
+    /** @type {qx.core.Object} */
+    __model : null,
+
+
+    setUp : function() {
+
+      // imagine me being a table like widget containing two columns
+      qx.Class.define("qx.test.data.controller.fixture.ArrayField", {
+        extend : qx.ui.core.Widget,
+        implement : [ qx.ui.form.IArrayForm, qx.ui.form.IForm ],
+        include : [ qx.ui.form.MForm ],
+
+        events : {
+          changeValue : "qx.event.type.Data"
+        },
+
+        members : {
+          /** @type {qx.data.Array|null} */
+          __value : null,
+
+          /**
+           * @param value {qx.data.Array|null}
+           * @returns void
+           */
+          setValue : function(value) {
+            var oldValue = this.__value;
+            this.__value = value;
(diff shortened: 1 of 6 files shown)


mever:
Can anyone review my changes?

I am going to use this branch in my application. If I find and resolve any other issues I am going to send more pull requests.

oetiker:
I have seen them ... but not had a chance to walk through

level420:
LGTM 
Starting voting

framework/source/class/qx/test/data/controller/FormWithArrayAndModel.js
```diff
@@ -0,0 +1,146 @@
...
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016 Martijn Evers, The Netherlands
+
+   License:
+     LGPL: http://www.gnu.org/licenses/lgpl.html
+     EPL: http://www.eclipse.org/org/documents/epl-v10.php
```
oetiker: could you replace LGPL and EPG  with

`MIT: https://www.tldrlegal.com/l/mit`

since we are bound to go MIT rsn

framework/source/class/qx/ui/form/IArrayForm.js
```diff
@@ -0,0 +1,76 @@
...
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016 Martijn Evers, The Netherlands
+
+   License:
+     LGPL: http://www.gnu.org/licenses/lgpl.html
```
oetiker: same thing ... please go MIT

level420:
@mever : please set your contribution under the MIT license and use the following text passage in the header for this:

```
   License:
     MIT: https://opensource.org/licenses/MIT
```
qooxdoo will change to MIT license soon.
Thank you for your help.

framework/source/class/qx/ui/form/Form.js
```diff
@@ -291,6 +291,26 @@ qx.Class.define("qx.ui.form.Form",
...
+      for (var i = 0; i < this.__groups.length; i++) {
+        var group = this.__groups[i];
+        for (var j = 0; j < group.names.length; j++) {
+          if (group.names[j] === name) {
+            return group.items[j];
+          }
+        }
+      }
+
+      return null;
```
yybalam: Hello.

I think will be nice throw an `Error` instead of only return null:

    if(qx.core.Environment.get("qx.debug")) {
        throw new Error("No field named " + name + " was found.");
    }

This can be useful for find typos:

    var myField = someForm.getItem('aTipo'); // is a typo :P

level420: I wouldn't throw an error in debug mode and no error in the build version, as it changes behaviour between those two versions. I'd got with a ```this.error``` instead.

mever: Calling code should decide what to do when no item is found. IMHO exceptions should only be thrown in exceptional cases. Also, I do not think it is an error to not find an item. If one tries to access a property of null by supplying a wrong name, then the browser will notify you of your mistake. In the meantime generic code can iterate form items without worrying getItem to assume proper outcome:
```
var describeItem = function(item) {...};
var describeItems = function(form, names) {
names.forEach(function(name) {
  var item = form.getItem(name);
  if (null === item) {
    form.warn("Item " + name + " is not found in this form.");
  } else {
    describeItem(item);
  }
});
```

level420: @mever OK. agree with that.

oetiker: how do I decide if the item does not exist or has the value null ?

level420: @oetiker the method returns the item (widget implementing qx.ui.form.IForm) which was added using the given name to the form, not the value of the item.

For this you would do form.getItem('password').getValue()

oetiker: ah ok :)

level420:
3 and a half thumbs up and older than 2 business days.
merging
@mever Thank your for the PR

level420:
@mever It would be great to have some more documentation and maybe some example code on how the new interface is used.
Thank you.

level420:
@mever  What is missing is some more documentation on how it's intended to be used. most of the classes have some quick starter on the top, that gives newbies an idea on what to do with it.

PR #9077, opened 2016-07-04
Improve widget performance creating new widgets

johnspackman:
When adding widgets to an application, the time taken to add a widget increases proportionally to the number of widgets which are being created.  Once the program stops adding widgets and allows the UI to be drawn, the time it takes resets to “normal”.  This is because when a widget is created, it is added to a queue for UI rendering (`qx.ui.core.queue.Appearance`) and another for detection of visibility (`qx.ui.core.queue.Visibility`), and the larger these queues get the longer it takes to add to the queue because the queues are backed by arrays and `qx.lang.Array.contains` is used to prevent duplicates (i.e. a linear, unindexed search is used).  

Note that this slow down happens whether the parent widget is added to the UI or not - so for example, adding the widgets to a `qx.ui.tabview.Page` but not adding that page to a `TabView` does not prevent the slow down.

The demo app http://tinyurl.com/z24qx6r creates 10 `qx.ui.tabview.Page`’s with identical content; this is the time taken to create each page:

```
page 0: 164ms
page 1: 175ms
page 2: 268ms
page 3: 350ms
page 4: 384ms
page 5: 424ms
page 6: 483ms
page 7: 537ms
page 8: 590ms
page 9: 652ms
```

This PR patches the code to use a lookup via widget hash code instead of scanning the array; this changes the timings so that they are at least as good as before, and are not affected by the size of the queue:

```
page 0: 160ms
page 1: 137ms
page 2: 132ms
page 3: 135ms
page 4: 125ms
page 5: 133ms
page 6: 135ms
page 7: 163ms
page 8: 126ms
page 9: 128ms
```

The time taken actually reduces, presumably because the browser’s optimiser is able to help, in this case with a further 22% improvement in performance.


johnspackman:
damn - please ignore this, the fix is correct but the PR includes loads of old commits.  I'll fix it and resubmit

PR #9087, opened 2016-07-06
Improve performance on filtering table

woprandi:
This PR improve performance on filtering wihout any API break. It's still much lower than use `Array.filter` instead but much better than currently. Tips are welcome


framework/source/class/qx/ui/table/model/Simple.js
```diff
@@ -748,6 +810,362 @@ qx.Class.define("qx.ui.table.model.Simple",
...
+
+      for (var row = 0; row < rowLength; row++) {
+        filter_test = false;
+
+        for (i in this.Filters) {
+
+          if (qx.lang.Array.contains(this.numericAllowed, this.Filters[i][0]) && filter_test === false) {
+            compareValue = this.getValueById(this.Filters[i][2], row);
+
+            switch(this.Filters[i][0]) {
```
oetiker: since `this.Filters[i][0]` does not depend on the `row` the decision should be outside of the potentially long loop over the rows of the table

oetiker:
What is the advantage of putting this functionality into the parent class ?

woprandi:
@oetiker I thought the other way that there are not overidden method which justified to have a child class.

oetiker:
the overridden methods is one thing, the other is that you are now adding a ton of 'questionable' functionality to the `simple.js` forcing everyone who uses simple tables to include it in their code ...

woprandi:
@oetiker I could revert this changes. Anyway it should not be on the same PR

woprandi:
I'll open another PR

woprandi:
I'll need to access to `__rowArr` from the child class.

PR #9089, opened 2016-07-06
replace qx.lang.String.startsWith & qx.lang.String.endsWith by their prototype counterparts

schnepe2:
- replaced `qx.lang.String.startsWith(foo,bar)` by `foo.startsWith(bar)`
- replaced `qx.lang.String.endsWith(foo,bar)` by `foo.endsWith(bar)`

Except for `qx.module.util.String` which I don't know what to do with.

### DIFF
--- a/framework/source/class/qx/bom/element/Background.js
+++ b/framework/source/class/qx/bom/element/Background.js
@@ -93,7 +93,7 @@ qx.Class.define("qx.bom.element.Background",
       // only check the first 50 characters for performance, since we do not
       // know how long a base64 image url can be.
       var firstPartOfUrl = url.substr(0, 50);
-      return String.startsWith(firstPartOfUrl, "data:") && String.contains(firstPartOfUrl, "base64");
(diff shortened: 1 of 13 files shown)


level420:
regarding https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/module/util/String.js#L95

I'd wrap it via a function body like this:

```javascript
    /**
     * Check whether the string ends with the given substring.
     *
     * @attachStatic {qxWeb, string.endsWith}
     * @signature function(fullstr, substr)
     * @param fullstr {String} the string to search in
     * @param substr {String} the substring to look for
     * @return {Boolean} whether the string ends with the given substring
     */
    endsWith : function(fullstr, substr) {
        return fullstr.endsWith(substr);
    },
```
as this preserves API compatibility.

level420:
Looks good! 
Start voting.

level420:
5 thumbs up.
merging
Thank you @schnepe2

PR #9091, opened 2016-07-09
Fix #9090

### LINKED ISSUE #9090: «escape» and «unescape» Js function are deprecated
Actually `qx.bom.Cookie` class use `escape` and `unescape` but those functions are deprecated: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features

I don't know if is better to be changed by `encodeURI` and `decodeUIR` JS function or by analog methods of `qx.bom.String`.

yybalam:
Call to 'encodeURI' and 'decodeURI' instead of 'escape' and 'unescape' deprecated functions.

### DIFF
--- a/framework/source/class/qx/bom/Cookie.js
+++ b/framework/source/class/qx/bom/Cookie.js
@@ -63,7 +63,7 @@ qx.Bootstrap.define("qx.bom.Cookie",
         end = document.cookie.length;
       }
 
-      return unescape(document.cookie.substring(len, end));
+      return decodeURI(document.cookie.substring(len, end));
     },
 
 
@@ -81,7 +81,7 @@ qx.Bootstrap.define("qx.bom.Cookie",
     set : function(key, value, expires, path, domain, secure)
     {
       // Generate cookie
-      var cookie = [ key, "=", escape(value) ];
+      var cookie = [ key, "=", encodeURI(value) ];
 
       if (expires)
       {


framework/source/class/qx/bom/Cookie.js
```diff
@@ -63,7 +63,7 @@ qx.Bootstrap.define("qx.bom.Cookie",
         end = document.cookie.length;
       }
 
-      return unescape(document.cookie.substring(len, end));
+      return decodeURI(document.cookie.substring(len, end));
     },
```
derrell: I think this may break existing cookies. The escape() function encodes all ascii characters other than letters/numbers, including '+' and '&'. If those characters were escaped()'ed, then decodeURI() will not yield the original result. I believe, however, that decodeURIComponent() will, but we should confirm, in a test, that at least every ascii character, and preferably other latin characters as well, are properly decoded from existing cookies... or decide that we just don't care. (I think we do care, though, until a major release.)

yybalam:
I see.
Well, as this change may break some cookies I guess the best is wait for a major version release, also use `(en | de)codeURIComponent` instead.

I found an useful article: http://xkr.us/articles/javascript/encode-compare/

* escape() will not encode: `@*/+`
* encodeURI() will not encode: `~!@#$&*()=:/,;?+'`
* encodeURIComponent() will not encode: `~!*()'`

So, I'm not sure how to proceed. :sweat_smile:

johnspackman:
The PR can still be merged, but will not be included in minor releases

oetiker:
Also I guess escaping is less of a problem than un-escaping, as existing stored cookies should still be properly handled ... and this seems to be the case when using `decodeURIComponent`

```javascript
var test = escape('~!@#$%^&*(){}[]=:/,;?+\'"\\');
decodeURIComponent(test) == unescape(test);
```

yybalam:
I created a new branch on my fork to follow this issue, so, I will to close this pull request and open another one from that new branch. :sweat_smile:

PR #9093, opened 2016-07-12
Make it easier to inherit from the default row renderer

cajus:
If you want to use the Default row renderer as a base for slightly modified renderers, you've the problem that the __fontStyle\* members are not accessible. Copying the complete class over is the way to go then.

The mini patch moves them in scope, so that they can be accessed from child classes. Alternatively one could add a getter if we decide to not expose them to child classes.


oetiker:
fine with me

derrell:
I don't think that protected members require setter/getter. They're intended for modification only by subclasses, where presumably the developer knows exactly what they're doing. LGTM.

level420:
3 thumbs up and more than two business days.
merging
Thank you @cajus

Issue #9079, opened 2016-07-05
Christian Boulanger


oetiker:
Get permission from Christian Boulanger for MIT License Change

oetiker:
@cboulanger

cboulanger:
Permission granted. Keep up the great work!

oetiker:
thanks!

PR #9094, opened 2016-07-12
fix for #9079 - change license to MIT

oetiker:


framework/source/class/qx/event/message/Bus.js
```diff
@@ -8,8 +8,7 @@
      2007 Christian Boulanger
 
    License:
-     LGPL: http://www.gnu.org/licenses/lgpl.html
-     EPL: http://www.eclipse.org/org/documents/epl-v10.php
+     MIT: https://opensource.org/licenses/MIT
      See the LICENSE file in the project's top-level directory for details.
```
derrell: There is no LICENSE file. There is a license.txt file which includes the former licenses. That should be one of the first files changed, and either it should be called LICENSE, or all of these individual file license changes should reference license.txt. (I think the .txt is important for viewing it on the browser from github, so probably the latter option -- correcting this line in each of the individual files -- is the better one.)

oetiker: I will fix this with the next path which changes all the 1&1 files ... it seems that calling the file LICENSE is 'the right thing'

PR #9101, opened 2016-07-13
[RFC] Issue #9090 Update calls to '[en|de]codeURICompoonent()' functions

yybalam:
So, this PR change the `espace()` and `unescape()` calls of those deprecated functions to recommended `encodeURIComponent()` and `decodeURIComponent()` respectively.

Also I added a `testEncoding()` method for compare results. A string of special characters looks work fine `'~!@#$%^&*(){}[]=:/,;?+\'"\\'`, but some unicode does not `'äëíöü'`.

So, let's open the discussion. :sweat_smile:


johnspackman:
Please can you add some documentation about backwards compatibility?

oetiker:
since `decodeURIComponent(escape('ä'))` actually raises an exception, how about wrapping the `decodeURIComponent` call into a `try` and in the `catch` branch handle the `URIError` condition by returning `null`

yybalam:
Done. I'm not sure if my english and explanation are good enough. :sweat_smile:

oetiker:
@yybalam I think @johnspackman meant a comment in the jsdoc at the top of the two methods affected by the change

johnspackman:
yes 👍  that comment in changelog just needs to go where the API viewer will pick it up.

Your english is fine, although I think instead of "There no must be issues" you mean "There are no issues"

framework/source/class/qx/bom/Cookie.js
```diff
@@ -62,7 +62,13 @@ qx.Bootstrap.define("qx.bom.Cookie",
         end = document.cookie.length;
       }
 
-      return unescape(document.cookie.substring(len, end));
+      try {
+        return decodeURIComponent(document.cookie.substring(len, end));
+      }
+      catch (URIError) {
+        console.error(URIError.message);
```
cajus: That should be something like:

`qx.log.Logger.error("Error decoding URI components", URIError.message);`

level420:

oetiker:
@yybalam please add the entry to the Changelog to the class documentation so that it is visible to people looking at the api viewer. The english is fine

level420:
@oetiker should we restart with a new voting or are we good to go?

oetiker:
@level420 I guess we are ok :) ... I would have mentioned that it was versions before qx 6.0 ... but alas

level420:
5 Votes. Merging.
Thank you @yybalam

PR #9106, opened 2016-07-21
Mark Remote as abstract

woprandi:
The class `qx.ui.table.model.Remote` is supposed to be abstract (http://www.qooxdoo.org/current/apiviewer/#qx.ui.table.model.Remote) because you have to implements `_loadRowCount` and `_loadRowData` methods. But the class is not marked as abstract


level420:
More than 2 business days and 4 thumbs up.
Merging.
Thank you @woprandi

PR #9107, opened 2016-07-21
Add 'changeScale' event to scale property of qx.ui.basic.Image to allow binding

woprandi:


level420:
More than 2 business days and 4 thumbs up.
Merging.

level420:
Thank you @woprandi

PR #9108, opened 2016-07-21
Annotations

johnspackman:
This PR adds Annotations to the Qooxdoo class system; if you already know what Annotations are in Java then this will look pretty familiar, I've added a manual page that describes Annotations in general and how to use them

### DIFF
--- a/documentation/manual/source/pages/core/annotations.rst
+++ b/documentation/manual/source/pages/core/annotations.rst
@@ -1,0 +1,114 @@
+.. _pages/annotations#annotations:
+
+Annotations
+***********
+
+Annotations are the ability to add meta data to classes so that the meta data can be accessed at runtime; the meta data can be added to individual members, properties, statics, constructor, destructor, or the whole class.  
+
+Properties already include some meta data - the `check` attribute allows to specify data type or range of values, and Qooxdoo uses this meta data to enforce validation; you can access this meta data too, but it's pretty limited in its usefulness.  Annotations allow you to add your own, application specific meta data to describe anything you want (and not just to properties).
+
+Annotations can be anything you want - in Java, they have to be a specific type, but in this implementation they can be any object and (I find at the moment) a string is very handy.  Obviously it’s important to be clear about your naming convention (!), and it would be wise to prefix strings with your package name.
+
+One question about Annotations is, why can’t you use interfaces to do the same thing?  Interfaces have their (very important) place, but they require that you implement code in every class in order for that class to provide that interface’s given functionality; annotations allow you to separate the code which does the work from the class, but this can only work because the class is able to “annotate” itself with guidelines on how to that separate code should behave.  
+
+Example 1
+=========
+Suppose you create a form that allows the user to edit some object, and that object has lots of properties that can be edited; some of those properties are read only, some can only be edited by people with the correct permission, some are integers, some can only have a specific range of values, etc.  
+
+When you hand code your form you embed a lot of that business knowledge (about who can edit what, and what values are allowed) into your form editor, plus there is some basic validation by the property `check` attribute.
+
+But lets say you edit that object in more than one place - eg a second form editor UI, and maybe there are places where the object is edited by code as a result of server interaction or other event.  This means duplicating that validation logic in all form editors and in the code that manipulates and validates those properties.
+
+One possible solution using annotations would be to annotate each property with special information about what is allowed, and then let your form editors use that information to customise the controls; a separate validator could use the same annotations to check the values being set against the properties explicitly.
+
+Example 2
+=========
+Lets say you have an object model which replicates some data on the server, such as `Customer`; most of the properties are business data and need to be kept synchronised with the server when the user hits save.  This data is usually changed by the `CustomerEditor` widget, but admin users can edit the `Customer`s in a table, and there are a couple of UI wizards for manipulating the data in each `Customer` object too.
+
+You do not want the UI code to handle copying the data to the server and back because it happens in so many places, it would be much better is Customer just "knew" about the properties.
+
+To automate this, you could start by just synchronising *all* properties in Customer to the server, but the likelyhood is that there are properties which only exist in the UI, and anyway the naming conventions might mean that property names on the server are always 100% the same as that on the server.
+
+By adding annotations to each property that needs to be synchronised, you can describe which need to be transferred, what their name is at the server, whether they are one way, etc.  Once piece of generic code can then handle synchronising for Customer or any server object that works in this way.  
+
+.. _pages/annotations#declaration:
+
+Declaration
+===========
+
+Annotations are declared next to the method being annotated but with a leading "@" symbol, or in the case of properties and classes just an "@" within the definition.  For example:
+
+::
+
+    qx.Class.define("qx.test.Cat", {
+      "@": [ "class-annotation" ],
+      extend: qx.core.Object,
+      
+      "@construct": [ "constructor-annotation" ],
+      construct : function() { /* ... */ }
+    });
+
+In the above examples the two annotations are just strings, which is functional but you have to be careful to avoid name conflicts - if you use strings, make sure that it is prefixed with your library namespace.
+
+Perhaps a better way to annotate would be with an instance of a class that has well defined properties; for example:
+
+::
+
+    qx.Class.define("qx.test.Pet", {
+      extend: qx.core.Object,
+      
+      properties: {
+      	hasFur: {
+      		"@": [ new qx.serverio.SerialiseColumn("HAS_FUR") ],
+      		init: true,
+      		check: "Boolean"
+      	},
+      	
+      	color: [
+      		"@": [ new qx.serverio.SerialiseColumn("COLOR") ],
+      		nullable: false,
+      		check: [ "black", "brown" ]
+      	]
+      }
+    });
+  
+    qx.Class.define("qx.test.Cat", {
+      "@": [ new qx.serverio.SerialiseTable(true, "CAT") ],
+      extend: qx.test.Pet,
+      
+      properties: {
+      	hasFur: {
+      		"@": [ "qx.test.otherAnnotation" ],
+      		refine: true
+      	}
+      }
+    });
+
+.. _pages/annotations#reflection:
+
+Reflection
+==========
+
+Annotations are only useful if you can inspect them, and the ``qx.Annotation`` class provides methods for accessing them; for example:
+
+::
+
+    var pet = this.getPet();
+    var annos = qx.Annotation.getProperty(pet.constructor, "color");
+    qx.core.Assert.assertEquals(1, annos.length);
+    qx.core.Assert.assertEquals("qx.serverio.SerialiseColumn", annos[0].classname);
+
+Note that you can use annotations in super classes and add further annotations in derived classes; in the example above, the "hasFur" property has two annotations in the qx.test.Cat class, but only one for qx.test.Pet.
+
+::
+
+    var pet = this.getPet();
+    
+    // Only Cat's have the second annotation
+    qx.core.Assert.assertTrue(pet instanceof qx.test.Cat);
+    var annos = qx.Annotation.getProperty(pet.constructor, "hasFur");
+    qx.core.Assert.assertEquals(2, annos.length);
+    qx.core.Assert.assertEquals("qx.test.otherAnnotation", annos[0]);
+    qx.core.Assert.assertEquals("qx.serverio.SerialiseColumn", annos[1].classname);
+
+
--- a/documentation/manual/source/pages/core/annotations.rst
+++ b/documentation/manual/source/pages/core/annotations.rst
@@ -57,7 +57,7 @@ One possible solution using annotations would be to annotate each property with 
 		properties: {
 			allowNonUsa: {
 				init: false,
-				check: "Boolean
+				check: "Boolean"
 			}
 		},
 		
--- a/documentation/manual/source/pages/core/annotations.rst
+++ b/documentation/manual/source/pages/core/annotations.rst
@@ -21,15 +21,135 @@ But lets say you edit that object in more than one place - eg a second form edit
 
 One possible solution using annotations would be to annotate each property with special information about what is allowed, and then let your form editors use that information to customise the controls; a separate validator could use the same annotations to check the values being set against the properties explicitly.
 
+::
+	qx.Class.define("myapp.model.Customer", {
+		extend: qx.core.Object,
+		
+		properties: {
+			address: {
+				check: "String"
+			},
+			zipCode: {
+				check: "String",
+				"@": [ new myapp.forms.annotations.ZipCode().set({ allowNonUsa: true }) ]
+			}
+		}
+	});
+	
+	qx.Class.define("myapp.forms.annotations.ValidationAnnotation", {
+		extend: qx.core.Object,
+		type: "abstract",
+		
+		members: {
+			/**
+			 * Called to check whether a value is valid
+			 * @abstract
+			 * @param value {String} value to be interpretted
+			 * @return {Boolean}
+			 */
+			isValid: function(value) {
+			}
+		}
+	});
+	qx.Class.define("myapp.forms.annotations.ZipCode", {
+		extend: myapp.forms.annotations.ValidationAnnotation,
+		
+		properties: {
+			allowNonUsa: {
+				init: false,
+				check: "Boolean
+			}
+		},
+		
+		members: {
+			/*
+			 * @Override
+			 */
+			isValid: function(value) {
+				var ok = true;
+				/* ... zip code validation ... */
+				return ok;
+			}
+		}
+	});
+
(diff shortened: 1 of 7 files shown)


derrell:
Is there a more real-life example that I can look at, preferably with a "before and after" implementation of with and without annotations? I'm not seeing the big picture of how this will be used. I'd hate to add complexity without a clear benefit, and I don't yet see that benefit.

johnspackman:
Sure - I have an application for a national retailer that records lots of data points about their own brand products; there are 450+ "questions" that we ask the suppliers, and a question can be as simple as:

"What is the pallet barcode?" (enter a one line text field)
"What is the proposed title for the front of the pack?" (enter a multi-line text field)

Other "questions" are a little more complicated and have several fields, eg: for "What is the shelf life?" you must 6 fields: integer values for shelf life at depot, at store, audited values at depot and at store, upload validation data to prove the shelf life is correct, and an explanation if you are not able to provide validation data.  And some questions are a lot more complicated with arrays of other objects edited by tables, or recursive structures edited by trees.

There are over 100 different types of question, which is implemented as an abstract class of `Question`, and then 100+ classes derived from that.  Every Question-derived class has a corresponding editor widget derived from qx.ui.core.Widget, so you could do things like:

    var productSpec = this.getProductSpec();
    var ed = new package.ShelfLifeQuestionEditor();
    productSpec.bind("shelfLife", ed, "value");

One final complication is that every question must be approved by two different auditors, and can have comments attached by the supplier filling in the form, or the auditor when they check it.  This means that my Question class looks like this:

    qx.Class.define("mypackage.Question", {
      extend: qx.core.Object,
      type: "abstract",
      
      properties: {
        approvedByMicron: {
          init: false,
          nullable: false,
          check: "Boolean",
          event: "changeApprovedByMicron"
        },
        
        approvedByFbdc: {
          init: false,
          nullable: false,
          check: "Boolean",
          event: "changeApprovedByFbdc"
        },
        
        comments: {
          nullable: false,
          check: "qx.data.Array",
          event: "changeComments"
        }
      }
    });

And here's a skeleton for the shelf life question:

    qx.Class.define("mypackage.ShelfLifeQuestion", {
      extend: mypackage.Question,
      
      properties: {
        lifeAtDepot: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtDepot"
        },
        lifeAtStore: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtStore"
        },
        confirmedLifeAtDepot: { /* ... */ },
        confirmedLifeAtStore: { /* ... */ },
        uploadedValidationData: { /* ... */ },
        explain: { /* ... */}
      }
    });

We have to be able to copy product specifications, and when we do so only copy the business fields and not the approval or comments - so you could add an abstract method called "copyFrom" that does this one by one:

      members: {
        copyFrom: function(that) {
          this.set({
            lifeAtDepot: that.getLifeAtDepot(),
            lifeAtStore: that.getLifeAtStore(),
            confirmedLifeAtDepot: that.getConfirmedLifeAtDepot(),
            confirmedLifeAtStore: that.getConfirmedLifeAtStore(),
            uploadedValidationData: that.getUploadedValidationData(),
            explain: that.getExplain()
          });
        }
      }

This is pretty tedious to type, it's easy to make a few typos, and really easy to miss out some properties, especially if the number of properties change over time.  For complex types, like arrays or other objects, you might need to code some special handling boiler plate in there too.

Next I needed to be able to compare questions in different revisions, so I could have added another abstract function compareTo:

        compareTo: function(that) {
          if (this.getLifeAtDepot() != that.getLifeAtDepot())
            return false;
          if (this.getLifeAtStore() != that.getLifeAtStore())
            return false;
          /* ... */
          return true;
        }

This is even more tedious, especially as I'm iterating over the same set of properties.  So here's the same code with annotations:

    qx.Class.define("mypackage.ShelfLifeQuestion", {
      extend: mypackage.Question,
      
      properties: {
        lifeAtDepot: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtDepot",
          "@": [ "mypackage.businessField" ]
        },
        lifeAtStore: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtStore",
          "@": [ "mypackage.businessField" ]
        },
        confirmedLifeAtDepot: { /* ... */ },
        confirmedLifeAtStore: { /* ... */ },
        uploadedValidationData: { /* ... */ },
        explain: { /* ... */}
      }
    });

And then in `mypackage.Question` I have:

      members: {
        compareTo: function(that) {
          var self = this;
          var names = this.getListOfBusinessRuleProperties();
          return names.every(function(name) {
            var src = self["get" + qx.lang.String.upname(name)]();
            var dest = that["get" + qx.lang.String.upname(name)]();
            return src == dest;
          });
        }
      }

In these examples, I can force the class hierarchy any way I want, but Annotations are even more interesting when you write a library to handle code that has not been written yet.  This next  example is in Java, not Javascript, but I'm including here to show just how useful they can be:

    @StoreDocument
    public static final class ChemicalParamsQuestion extends Question {
      @Property(anno="mypackage.businessField", arrayType=ChemicalParam.class)
      @StoreProperty(componentType=ChemicalParam.class, value="parameter", accessible=true, create=true)
      private ArrayList<ChemicalParam> parameters;

      public ArrayList<ChemicalParam> getParameters() {
        return parameters;
      }
    }
    
    @StoreDocument
    public static final class ChemicalParam implements Proxied {
      @Property(anno="sparqa.field") @StoreProperty private String test;
      @Property(anno="sparqa.field") @StoreProperty private String unit;
      @Property(anno="sparqa.field") @StoreProperty private String min;
      @Property(anno="sparqa.field") @StoreProperty private String max;
      @Property(anno="sparqa.field") @StoreProperty private String freqOfTesting;
      
      /* all getXxxx() and setXxxx() methods deleted for clarity */
    }

The `@Property` annotation is from my Qooxdoo ServerObjects contrib and handles all of the I/O between the client and the server; it creates class definitions in Javascript on the client I can do things like:

    var param = new mypackage.ChemicalParam().set({ unit: "mg", min: "05" });
    productSpec.getChemicalParamQuestion().getParameters().push(myNewChemical);

and it is automatically replicated on the server.

Also the `@StoreDocument` and `@StoreProperty` - they allow this class to be able to be serialised to a database, with one library that handles the equivalent of SQL CREATE TABLE and INSERT/UPDATE/DELETE (including the one to many relationship with the ChemicalParam).  Note how `@StoreProperty` in ChemicalParamsQuestion can be used to provide specific information about the property

That Java example is complete - there are no other SQL DDL/DML, and no REST APIs to build and test because the libraries can look into the class and take care of it all.  It took some investment to create the libraries but the payoff is huge because I now spend most of my time working on the client rather than DDL/DML/REST donkey work.

I was a sceptic when I first read about them, ie not sure about the big picture either, but having used them for years now, I firmly believe that Annotations are one of those things that appear a little "so-so" at first (especially as they seem to overlap what interfaces are for), but the more you use them the more they can turnout to be a real game changer in some scenarios.

johnspackman:
I've just thought of another potential example - Qooxdoo supports local storage, but obviously local storage only stores JSON data.  You can implement qx.data.marshal.IMarshaler in your classes, but that still requires you to write your own serialisation/deserialisation code - much of which is boilerplate.

There is nothing to stop the existing marshalling code from using reflection to iterate the properties and reading/writing them - except that you almost certainly do *not* want *all* properties serialised in that way.  So a better way would be to be able to say which properties need to be serialised automatically and that's where Annotations would be come in, in many cases removing the need for your own serialisation/deserialisation code altogether.

derrell:
Thanks for the thorough description and examples! This looks like a nice new concept to learn more about. I'd request waiting until after the weekend, to give time for this concept to sink in a bit, before we vote on it. (I haven't looked at the code at all yet, wanting instead to understand the purpose first.)

johnspackman:
no worries, I'm in no rush, especially for fundamental things like this :)

cajus:
Cool! I've been missing them in qx ;-) It would be "optical" nicer to let the generator extract them from the source, so that they don't have to occur as member attributes. No idea if that would be a way to go.

johnspackman:
That would look a lot better - I think that's possible with qxcompiler, because we can add to the babel plugin to transform the more java-esque syntax into the members object

derrell:
I spent time during the weekend coming to grips with annotations. I think there are some clear benefits here, once one learns to take advantage of them. I wouldn't want to see any _requirement_ for using annotations in qooxdoo core, but adding this ability seems like a win.

I don't particularly like the syntactical clutter. Can we consider what/where automatic annotation / reflection capability we might want to add (annotations produced by the generator), that would reduce the amount of extra, visible annotation that must be added by the user? Also to reduce clutter, maybe the user could add a single "annotation generator" function that is called by the generator with each function, property, etc. That function might return, for example, for any function name beginning with "_business", an array of annotations containing the string "myapp.business-logic" and the class myapp.BusinessLogicRequirements. Just thinking out loud here...

johnspackman:
There's no reason why the generator can't be used to add meta data in the form of annotations where it has it available, and that would be a good way to provide any extra meta data (ie rather than having another, parallel meta data mechanism) - but Im not sure how much more the generator can add.  All I can think of at the moment is that the generator could glean bits from the doc comments for each method, although that is still extra syntax, it's just in comments rather than in code.

An annotation generator could be added, and I'd suggest that the best way to do this would actually be by using an annotation :)  For example:

    qx.Class.define("mypackage.ShelfLifeQuestion", {
      extend: mypackage.Question,
    
      properties: {
        lifeAtDepot: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtDepot"
        },
        lifeAtStore: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtStore"
        },
        confirmedLifeAtDepot: { /* ... */ },
        confirmedLifeAtStore: { /* ... */ },
        uploadedValidationData: { /* ... */ },
        explain: { /* ... */}
      },
      
      statics: {
        "@generateAnnotations": [ qx.core.annotations.AnnotationGenerator.getInstance() ],
        generateAnnotations: function(clazz, type, name) {
          if (type == "property") {
            if (name == "lifeAtDepot" || name == "lifeAtStore") {
              return [ "mypackage.businessField" ];
            }
          }
          return null;
        }
      }
    });

This example would be implemented by a change to qx.Class.

It's interesting to consider that an annotation generator function could be implemented by using annotations - in this case, because we would need a static method to define the annotations it would not be possible to use interfaces, so while we could add another top-level method and bake that into the `qx.Class.define` data structure, annotations allows the developer to nominate an ordinary static method.

While an annotation generator would work for some property/method styles, personally I think that one of the benefits of annotations is being able to add the meta data directly on the property or method that needs to be annotated.  In the earlier examples, Im just using a string of "mypackage.businessField" as an annotation, which is perfectly functional but only provides one piece of information - i.e. is it present or not - but in many cases you want to provide at least one or two parameters. 

Let's assume that in my example, the `compareTo()` should work on all fields, but the `copyFrom()` should only copy certain fields because it is a requirement that the user re-enter them so that we are 100% sure there isn't a copy & paste mistake (which is one of the things we worry about in this application, and shelf life is a very big deal in the retail sector).  Rather than a string as annotation, it would be better to have an object which can be interrogated at runtime but which is configured on a per property basis in code, EG:

    qx.Class.define("mypackage.ShelfLifeQuestion", {
      extend: mypackage.Question,
    
      properties: {
        lifeAtDepot: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtDepot",
          // This property must not be copied
          "@": [ new mypackage.annos.QuestionProperty(false) ]
        },
        lifeAtStore: {
          init: 0,
          nullable: false,
          check: "Integer",
          event: "changeLifeAtStore",
          // This property must not be copied
          "@": [ new mypackage.annos.QuestionProperty(false) ]
        },
        confirmedLifeAtDepot: { /* ... */ },
        confirmedLifeAtStore: { /* ... */ },
        uploadedValidationData: { /* ... */ },
        explain: {
          init: "",
          nullable: false,
          check: "String",
          event: "changeExplain",
          // This property can be copied
          "@": [ new mypackage.annos.QuestionProperty(true) ]
        }
      }
    });

Obviously you don't have to use new, a static instance would probably suffice but either way having an actual class instance is a more structured programming style and better for forward compatibility because it allows for more complexity to be stored in a single annotation.

As the complexity of the annotation increases, the benefit of having the annotation right next to (or part of) the property definition increases; the annotation becomes part of the property definition rather than something outside it.

johnspackman:
The annotations at the moment are designed to fit in Javascript's objects, but with a babel plugin in the compiler we could achieve a more natural version, EG:

    @new qx.serverio.SerialiseTable(true, "CAT")
    qx.Class.define("qx.test.Cat", {
       extend: qx.test.Pet,
       
       properties: {
         @qx.test.OtherAnnotation
         hasFur: {
           refine: true
         }
       },
       
       members: {
         @qx.test.OtherAnnotation
         doSomething: function() {
           return false;
         }
       }
     });

@cajus I think this what you were thinking also?  

But now that I've mocked it up Im reconsidering because Im not sure that it's that much clearer plus no editor will understand the syntax, any coder without Qooxdoo experience will wonder whats going on, and either we document two forms or everyone will be forced to use the compiler all the time.

cajus:
@johnspackman yes, that's what I was thinking of. Feels a bit cleaner IMHO.

derrell:
The only thing I would be **dead set against** is having an annotation syntax that does not fit JavaScript syntax. If it messes up my editor, I'd be really upset, so the latter idea, with non-JavaScript syntax, is a total non-starter for me.

I think the annotation generator concept may end up having merit, but it's an add-on to your basic annotations. I'm on board, and I think we should go with the basic set of annotation features as in this PR.

documentation/manual/source/pages/core/annotations.rst
```diff
@@ -0,0 +1,114 @@
...
+
+You do not want the UI code to handle copying the data to the server and back because it happens in so many places, it would be much better is Customer just "knew" about the properties.
+
+To automate this, you could start by just synchronising *all* properties in Customer to the server, but the likelyhood is that there are properties which only exist in the UI, and anyway the naming conventions might mean that property names on the server are always 100% the same as that on the server.
+
+By adding annotations to each property that needs to be synchronised, you can describe which need to be transferred, what their name is at the server, whether they are one way, etc.  Once piece of generic code can then handle synchronising for Customer or any server object that works in this way.  
+
```
oetiker: code to go with the examples would be great

johnspackman:

level420:
@johnspackman the annotations as described: do they work without using generator or qx-compiler?

johnspackman:
yes - I've built and run the testrunner with generate.py and not changed the generator at all

derrell:
Voted thumbs up. Still not sure whether it will be generally useful, but I don't see anything that should break with this.

level420:
Five votes. Merging.
Thank you @johnspackman

petitben:
Hi,

Just my 2cents here even though if it's already merged... 

I'm a bit skeptic about this kind of addition as I've a feeling it's application/JAVA-based backend specific. Are there so many people in the need of this kind of feature ?

Like @derrell I'd hate to add complexity without a clear benefit, and I don't yet see that benefit... I mean we already have Interfaces so why not use it ?

Also as some members mention it would be clearly nicer to use a /* @annonation: */ - like stuff and let the python generator (or your qxcompiler) do the job to avoid introducing these awful "@*" members - not mean to be rude of course :wink: but really I don't like it at all...

What I see however in additions like this is that we should introduce a more generic way to "upgrade" the qx Class system (a bit like mixins or plugins). I haven't honestly given it much thought but that would allow your addition to be integrated for your specific purpose without forcing to patch qx.Class and qx.core.Property ... So people no wanting to include/use annotations could continue ignore it and those in the need of this kind of feature could include it in a simple way...

Voilà! Nothing personal of course... Just wanted to share my opinion about it - maybe I don't see the real benefit?

johnspackman:
No worries, your opinion is welcome :)

Annotations come from Java as a design concept, but they have nothing to do with the actual backend at all; they have been around for a while (in Java) and there are some occasions when they really can add value.  Like any tool, particularly one where the use is not immediately obvious, there is plenty of opportunity for misuse but speaking from personal experience Annotations can be hugely useful when used wisely.

They also differ from Interfaces and can provide solutions when Interfaces simply will not work; eg Interfaces allow you to work with object instances which have not yet been designed, but Annotations allow you to work with classes that have not yet been designed.  This is because Annotations allow you to add meta data to class definitions (or parts of a class definition) in a structured, extensible way.

I completely agree that the "@myanno" style is pretty ugly :) and I would love to not have to do it - but the reason I chose that is specifically so that the code will work without out of the box, IE without requiring qxcompiler to transpile non-javascript syntax.  Qxcompiler could transpile annotations in comments into the ugly syntax at some point in the future (and I am tempted by the thought ;) but the first phase is to get the Annotations system out there

Unfortunately, it was not possible to write Annotations without patching qx.Class and Im not sure that upgrading Mixins would have allowed it either - it certainly would have been a separate project to get to that stage. (PS - I would have implemented it as a contrib if I possibly could, years ago).  But using it is not mandatory so people can ignore it if they want to.

PR #9109, opened 2016-07-21
Fix url in doc

hkollmann:
To fix the links in doc


oetiker:
I think that would be one of these instances where squashing would be nice do you agree @derrell ?

oetiker:
@hkollmann can you please also create a PR for the release_5_0_x

derrell:
Indeed. Very appropriate here.

documentation/manual/source/_theme.indigo/layout.html
```diff
@@ -152,83 +147,56 @@
...
-          </li>
-          <li>
-            <a href="http://qooxdoo.org/download">Download</a>
-          </li>
-          <li>
-            <a href="http://news.qooxdoo.org">Blog</a>
+            <a class="" href="http://qooxdoo.org/demos.html">demos</a>
           </li>
           <li>
-            <a href="http://bugs.qooxdoo.org">Bugs</a>
```
derrell: It looks like we lost Bugs, which should probably point to the github Issues page

hkollmann:
@oetiker: How?

hkollmann:
Tried it - hopefully it's correct

hkollmann:
I like the idea to automate the Building of the doc page. Any idea how? And when? I suggest the manual Change for the first try - then automate it

hkollmann:
Is there any way to shorten the merge history?

johnspackman:
This seems to be the same as https://github.com/qooxdoo/qooxdoo/pull/9110, why do we need two pull requests?

woprandi:
@johnspackman To run tests suite on both I guess

cajus:
@johnspackman they target different branches. As they fix the sphinx generated documentation, the commits fix the qx manual. It is regenerated per branch, so you need two PRs. One for master, one for 5_0_*.

cajus:
@hkollmann the docs are automatically built by travis already. Not the PRs, though.

oetiker:
@hkollmann here is some info on rebasing and squashing your commits ... https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

documentation/manual/source/_theme.indigo/layout.html
```diff
@@ -32,25 +32,20 @@
...
     <nav>
       <ul>
-        <!--
         <li>
-          <a class="" href="http://qooxdoo.org/">home</a>
+          <a class="" href="/demos.html">demos</a>
         </li>
-        -->
         <li>
-          <a href="http://news.qooxdoo.org">blog</a>
```
johnspackman: The Blog still exists, at https://medium.com/qooxdoo

johnspackman:
@cajus ok ISWYM but that's different to how I thought we were going to do it, which was to mark each PR that is to be in 5.0.2 with release_5_0_2 milestone, and then use this to update the branch_5_0_x and create the release_5_0_2 tag.

Creating separate PRs for each previous PR to merge into the branch would work but it seems like double the effort when it comes to reviewing and approving something that's already been reviewed and approved.  

I definitely think that in cases like this where the content is the same (or largely the same) the second PR should not be created until the first PR has been approved, otherwise which one do we comment on?  I commented on the other PR before I saw this one, but it's the same commit!

oetiker:
@johnspackman I suggested to create the second PR in this instance primarily to get the website fixed asap ... I think an approach where the initial PR is for master, gets discussed and when agreement is reached, the author creates an additional PR for the active branch ... OR maybe even branches. Since if we follow semantic versioning, we would have bugfixes going into the current branch and features going into the next minor version and breaking changes into master only.

cajus:
@johnspackman, you're right. I didn't compare the commit IDs. Seem to be completely the same. Would be something to do cherry pick.

documentation/manual/source/_theme.indigo/layout.html
hkollmann: You are right - but the blog is missing in www.qooxdoo.org either. I copied the footer section from there to have both equals.

oetiker: the blog, btw will soon exist on news.qooxdoo.org again ... medium is setting this up for us

hkollmann:
I can add the Blog entry in 2 weeks - holidays since today. Andy no computer allowed in vacancies. Or could someone else add it?

hkollmann:
@oetiker: Thanks for the hint. Next Time:-)

documentation/manual/source/_theme.indigo/layout.html
```diff
@@ -42,13 +45,17 @@
           <a class="active" href="/current">docs</a>
         </li>
         <li>
+         <a class="" href="/current/apiviewer">api</a>
+        </li>
+        <li>
           <a class="" href="http://stackoverflow.com/questions/tagged/qooxdoo">support</a>
         </li>
         <li>
           <a class="" href="https://gitter.im/qooxdoo/qooxdoo">gitter</a>
```
derrell: It's not one of the PR's changed lines, but...

I suspect that few know what gitter is. How about changing the visible text to "chat room" or "interactive discussion" or some such thing?

oetiker: good idea

derrell:
CALL FOR VOTES

Issue #9104, opened 2016-07-14
Provide dialog widgets


yybalam:
As others JS libraries, will be nice and useful if qooxdoo provide an own set of dialog widgets.
Actually there is  a lot a dialog implementations, some as contrib, some inside projects, etc.

As a lab I create on my fork a branch with a possible proposal of how I think can be qx dialogs. At moment I think is some premature make a PR, but also want to do the proposal.

https://github.com/qooxdoo/qooxdoo/compare/master...yybalam:dialoglab

Also I create a little demo app: http://yybalam.net/qxlab/dialoglab/

If you think that my proposal have future I will be glad on take tips, fix the bugs still remain on my lab, and add tests and documentation.

johnspackman:
I agree, I think we've all probably got our own dialog classes and it's an addition that's long overdue.  

I have some suggestions however:

at https://github.com/qooxdoo/qooxdoo/compare/master...yybalam:dialoglab#diff-db36f17ac6773546a7d9d40012eb8b0dR96 there is no need for aspect.icon, because it's duplicating what is available by:

   new qx.ui.dialog.Confirm(...).set({ icon: "icon" });

Rather than `buttons` property being a string of "yes", "yesno", "yesnocancel" etc, why not make it an array - this makes it possible to expect the dialog and add new button types (it would simplify `__composeButtons` also)

`__getButtons` should be private to allow new button types in derived classes

The design is to provide a map of button IDs to labels (or labels and icons) in `aspectOptons` and a separate map of button IDs to callbacks in `handlers`, and a context which is in that `handlers` map.  

IMHO that spreads the detail of each button around in two or three different places and will make the code confusing; it is more "qooxdoo-ish" to have a single event of (eg) "buttonClicked" and provide the button ID (ie "ok"/"cancel"/etc).  It is useful to provide a callback and context to the constructor so that these can be attached automatically.

    new qx.ui.dialog.Confirm("Title", "Message", [ "ok", "cancel" ], function(){}, this);

or with more control over button properties:

    new qx.ui.dialog.Confirm("Title", "Message", [
      { id: "ok", label: "Accept", icon: "icons/myicon.png" },
      { id: "cancel", label: "No Thanks", icon: "icons/myicon.png" },
    ], function(){}, this);

If you want to have individual callbacks per button, then combine them in the button array, eg:

    new qx.ui.dialog.Confirm("Title", "Message", [
[...]

PR #9111, opened 2016-07-23
[RFC] Issue #9104 Provide dialog widgets

yybalam:
This PR try to provide to qooxdoo a default basic dialog widgets. Basically this PR add :
1. A Message dialog for alert messaged.
2. A Confirm dialog with support for attach function to buttons.
3. A BusyIndicator.

A `qx.ui.Dialog` static class provide a short way for launch dialogs.:

```
qx.ui.Dialog.alert("Yay, some dialog!");
qx.ui.Dialog.error("Some was wrong. :(");
```

Also this PR include an `MBusyBlocker` mixin. This mixin extend `MBlocker` for provide an loading gif while the widget is blocked. Also is possible to show a custom text if `allowBusyPopup` is set to `true`.

Right now there is only two test, one for `Message` and another for `Confirm`.
## Whats missing?

Well, some documentation of course. Also I thing a test for  `MBusyBlocker`, but I don found a test for `MBlocker` so I'm not sure how to test it.

Also when I compile using 'Tango' icon theme, the icons are shown, but I can't get it work for Oxygen.

I know this PR actually is not complete (sorry) to be merge, but maybe heps to get the way.

Cheers.


cajus:
It is better to use the qx childControl mechanism for widgets, buttons, etc. and move everything styling related to the appearances/decorators. Additionally the dialogs expose the complete children handling to the outside world (add/remove/...). Not sure if we want this.

Besides the fact that you don't have to manually dispose the widgets you've created (missing in your code), you've the advantage that it is completely styleable by dialog users if you use child-controls. No need for custom methods like _getButtonBar, _getAtom, etc.

As dialogs are "one shot", so it's maybe a good idea to add an *autodispose* flag to the dialog base class.

framework/source/class/qx/ui/dialog/Abstract.js
```diff
@@ -0,0 +1,207 @@
...
+
+    this.add(this._getAtom());
+    this.add(this._getButtonsBar());
+
+    this._initDialog(title, message);
+
+    if (icon != undefined) {
+      this._getAtom().setIcon(icon);
+    }
+  },
```
oetiker: you may want to follow the structure outlined [in the qooxdoo manuel](http://www.qooxdoo.org/current/pages/desktop/ui_develop.html) using ChildControls

framework/source/class/qx/ui/dialog/Abstract.js
```diff
@@ -0,0 +1,207 @@
...
+     */
+    _initDialog : function(title, message) {
+      this.__title = title;
+      this.setMessage(message);
+    },
+
+    /**
+     * @param title {String} Title of dialog.
+     */
+    setTitle : function(title) {
```
johnspackman: `title` and `message` would be better implemented as a property

johnspackman:
looking good @yybalam

oetiker:
you may want to have a look at the color popup for inspiration https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/control/ColorPopup.js

oetiker:
looks much more standard now ... you had some live demo, where is it ?

also for integration we need updates to the manual and ideally also to the demobrowser ... 

and you should look into squashing the many commits

yybalam:
Hello.

Here is a live demo: http://yybalam.net/qxlab/dialoglab/
Here is the code of the demo: https://gitlab.com/yybalam/dialoglab/tree/master

So, I will to write the docs for manual and examples to demobrower, also add support for the rest of themes and the squashing of my lot of commits (at moment I don't kno how to do this last one, need to study :sweat_smile:)

woprandi:
@yybalam you can check how to squash commits here : https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

oetiker:
looking at the demo ... like the optics ... I wonder about the placement of the buttons ... I would have expected them to be right aligned.

framework/source/class/qx/bom/Cookie.js
```diff
@@ -62,7 +62,7 @@ qx.Bootstrap.define("qx.bom.Cookie",
         end = document.cookie.length;
       }
 
-      return decodeURI(document.cookie.substring(len, end));
+      return unescape(document.cookie.substring(len, end));
     },
```
oetiker: why ?

framework/source/class/qx/bom/Cookie.js
```diff
@@ -80,7 +80,7 @@ qx.Bootstrap.define("qx.bom.Cookie",
     set : function(key, value, expires, path, domain, secure)
     {
       // Generate cookie
-      var cookie = [ key, "=", encodeURI(value) ];
+      var cookie = [ key, "=", escape(value) ];
```
oetiker: why ?

framework/source/class/qx/bom/Cookie.js
yybalam: Nope. That is fixed by another PR that is no merged yet, so I think this PR must avoid that change, so this PR don't break anything with actual sate of master. :sweat_smile:

oetiker: ah this is a revert to your own patch series ... you definitely need to squash these bevor we can merge ...

yybalam: Yes. I'll try to do it right in my next try. Squash is absolutely something new for me. :sweat_smile:

zaucker:
Are the block and busy variants styled differently than the other dialogs on purpose in the demo? Title and min/max/close buttons on block/busy (extending qx.ui.window.Window() and not on the others (probably extending Popup)?

yybalam:
Yeah. The idea behind bolck and busy is a mixin (MBusyBlocker) for any widget, so do not have any special style, is just for block a Widget and shown a centred gif or popup with a tiny gif and a label. I just used Windows for convenience in the demo.

zaucker:
Thanks for the explanation, sounds good.

On Wed, 31 Aug 2016, Yeye Balam wrote:

> Yeah.  The idea behind bolck and busy is a mixin (MBusyBlocker) for any
> widget, so do not have any special style, is just for block a Widget and
> shown a centred gif or popup with a tiny gif and a label.  I just used
> Windows for convenience in the demo.

-- 

cboulanger:
Hello @yybalam, still interested in contributing a widget set to qooxdoo for the 6.0 release? This would require that you update the PR to the current master and address the remaining issues raised in the discussion. You could also contribute your widget set as a "contrib", which would require to change the top-level namespace. It would be great to have two different dialog widget sets as contribs, so people have a choice... If you're interested, let me know.

yybalam:
@cboulanger Oh, yeah. I'm still interesed in contributing the widget. But instead of update this PR, can I close this PR and create a new one?

cboulanger:
@yybalam Sure!

level420:
As @yybalam suggested: closing this PR.

cboulanger:
@yybalam please consider creating a contrib based on a child controls. It would be great to have stylable dialogs, I don't think I will have the time to ever work on my own Dialog widget and add styleability. Publishing the contrib will be really easy with the new `qx contrib` CLI commands which I hope to finish soon.

Issue #7175, opened 2013-01-12
Add ISO8601 parsing support to qx.land.normalize.Date (BZ#7131)



qx-bug-importer:
_George Nikolaidis (gnikolaidis) wrote:_

In our projects we are using ISO-8601 dates extensively. We have created a small bootstrap class to enhance the Date class with ISO features.

Since JavaScript 1.8.5 added support in Date.parse() for ISO date parsing, 

https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Date/parse

we propose including code from (for example)

https://github.com/csnover/js-iso8601/blob/master/iso8601.js

into qx.lang.normalize.Date

We kindly ask Qooxdoo developers to weigh the cost of adding more code to the core framework against the benefits of having said functionality available at all times, and decide if the enhancement proposal is worth considering.

_assigned to Richard Sternagel (@rsternagel)_

gnikolaidis:
Please consider reopening

cajus:
Would you mind creating a pull request for this feature? Sounds like you already have some code available.

gnikolaidis:
Sure, coming up soon

cajus:
Cool. Thanks! Doh. Forgot to re-open.

schnepe2:
Hello again,

I've had time to look over this again and I would like share my thoughts:
What about a user configuration switch (for `config.json`) that lets the application developer decide what behavior is preferred. This would allow the original PR poster to keep his application code unchanged.
Something like `qx.date.parse.mode`, which can be set to either `"standard"` or `"chrome"`  defaulting to `"standard"`
The normalizing code ("`qx.lang.normalize.Date`") can than check what needs to be done to fulfill the "users wish" like:
```
qx.Bootstrap.define("qx.lang.normalize.Date", {

  statics : {

    parse_standard : function (dateString) {
      // ...
    },

    parse_chrome : function (dateString) {
      // ...
    }
  },


  defer : function (statics)
  {
    // ...

    //
    // Date.parse
    //

    // Users "wish"
    var polyfill = qx.core.Environment.get("qx.date.parse.mode");
    // Native "situation"
    var hasParse = qx.core.Environment.get("ecmascript.date.parse");
    var isStd = hasParse && qx.core.Environment.get("ecmascript.date.parse.isstandard");

    switch ( hasParse+"-"+isStd )
    {
      case "false-false": break; // take "standard" or "chrome" polyfill
[...]

PR #9112, opened 2016-07-26
Added normalization for JavaScript 1.8.5 Date.parse()

gnikolaidis:
Bug #7175

Thanks to Colin Snover for original code
https://github.com/csnover/js-iso8601

### DIFF
--- a/framework/source/class/qx/lang/normalize/Date.js
+++ b/framework/source/class/qx/lang/normalize/Date.js
@@ -45,5 +45,45 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
     if (!qx.core.Environment.get("ecmascript.date.now")) {
       Date.now = statics.now;
     }
+
+    /**
+     * Date.parse with progressive enhancement for ISO 8601 <https://github.com/csnover/js-iso8601>
+     * © 2011 Colin Snover <http://zetafleet.com>
+     * Released under MIT license.
+     */
+    var origParse = Date.parse, numericKeys = [ 1, 4, 5, 6, 7, 10, 11 ];
+    Date.parse = function(date) {
+      var timestamp, struct, minutesOffset = 0;
+
+      // ES5 §15.9.4.2 states that the string should attempt to be parsed as a Date Time String Format string
+      // before falling back to any implementation-specific date parsing, so that’s what we do, even if native
+      // implementations could be faster
+      //            1 YYYY            2 MM       3 DD         4 HH    5 mm       6 ss      7 msec      8 Z 9 ±    10 tzHH    11 tzmm
+      if ((struct = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(date))) {
+          // avoid NaN timestamps caused by “undefined” values being passed to Date.UTC
+          for (var i = 0, k; (k = numericKeys[i]); ++i) {
+            struct[k] = +struct[k] || 0;
+          }
+
+          // allow undefined days and months
+          struct[2] = (+struct[2] || 1) - 1;
+          struct[3] = +struct[3] || 1;
+
+          if (struct[8] !== 'Z' && struct[9] !== undefined) {
+            minutesOffset = struct[10] * 60 + struct[11];
+
+            if (struct[9] === '+') {
+                minutesOffset = 0 - minutesOffset;
+            }
+          }
+
+          timestamp = Date.UTC(struct[1], struct[2], struct[3], struct[4], struct[5] + minutesOffset, struct[6], struct[7]);
+      }
+      else {
+          timestamp = origParse ? origParse(date) : NaN;
+      }
+
+      return timestamp;
+    };
   }
 });
--- a/framework/source/class/qx/lang/normalize/Date.js
+++ b/framework/source/class/qx/lang/normalize/Date.js
@@ -50,8 +50,9 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
     parse : function(date) {
       // Match input against ISO8601 regular expression
       var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(date);
-      if(!captureGroups)
+      if(!captureGroups) {
         return Date.originalParse(date);
+      }
 
       // Avoid invalid timestamps caused by undefined values being passed to Date.UTC
       [ 1, 4, 5, 6, 7, 10, 11 ].forEach(function(i) {
@@ -64,8 +65,9 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
       var minutesOffset = 0;
       if(captureGroups[8] !== "Z" && captureGroups[9] !== undefined) {
         minutesOffset = captureGroups[10] * 60 + captureGroups[11];
-        if(captureGroups[9] === "+")
+        if(captureGroups[9] === "+") {
           minutesOffset = - minutesOffset;
+        }
       }
       // Return the number of milliseconds since Epoch
       return Date.UTC(captureGroups[1], captureGroups[2], captureGroups[3], captureGroups[4], captureGroups[5] + minutesOffset, captureGroups[6], captureGroups[7]);
--- a/framework/source/class/qx/lang/normalize/Date.js
+++ b/framework/source/class/qx/lang/normalize/Date.js
@@ -13,6 +13,7 @@
 
    Authors:
      * Martin Wittemann (wittemann)
+     * Colin Snover (original Date.parse polyfill)
 
 ************************************************************************ */
 /**
@@ -37,6 +38,37 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
      */
     now : function() {
       return +new Date();
+    },
+
+    /**
+     * Parses a string representation of a date and return number of milliseconds since Epoch or NaN if string is unrecognised
+     *
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse">MDN documentation</a>
+     *
+     * Derived from <https://github.com/csnover/js-iso8601>: ©2011 Colin Snover <http://zetafleet.com>, MIT license
+     */
+    parse : function(date) {
+      // Match input against ISO8601 regular expression
+      var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(date);
+      if(!captureGroups)
+        return Date.originalParse(date);
+
+      // Avoid invalid timestamps caused by undefined values being passed to Date.UTC
+      [ 1, 4, 5, 6, 7, 10, 11 ].forEach(function(i) {
+        captureGroups[i] = +captureGroups[i] || 0;
+      });
+      captureGroups[2] = (+captureGroups[2] || 1) - 1; // Allow undefined months
+      captureGroups[3] = +captureGroups[3] || 1;// Allow undefined days
+
+      // Handle timezone offsets
+      var minutesOffset = 0;
+      if(captureGroups[8] !== "Z" && captureGroups[9] !== undefined) {
+        minutesOffset = captureGroups[10] * 60 + captureGroups[11];
+        if(captureGroups[9] === "+")
+          minutesOffset = - minutesOffset;
+      }
+      // Return the number of milliseconds since Epoch
+      return Date.UTC(captureGroups[1], captureGroups[2], captureGroups[3], captureGroups[4], captureGroups[5] + minutesOffset, captureGroups[6], captureGroups[7]);
     }
   },
 
@@ -46,44 +78,10 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
       Date.now = statics.now;
     }
 
-    /**
-     * Date.parse with progressive enhancement for ISO 8601 <https://github.com/csnover/js-iso8601>
-     * © 2011 Colin Snover <http://zetafleet.com>
-     * Released under MIT license.
-     */
-    var origParse = Date.parse, numericKeys = [ 1, 4, 5, 6, 7, 10, 11 ];
-    Date.parse = function(date) {
-      var timestamp, struct, minutesOffset = 0;
-
-      // ES5 §15.9.4.2 states that the string should attempt to be parsed as a Date Time String Format string
-      // before falling back to any implementation-specific date parsing, so that’s what we do, even if native
-      // implementations could be faster
-      //            1 YYYY            2 MM       3 DD         4 HH    5 mm       6 ss      7 msec      8 Z 9 ±    10 tzHH    11 tzmm
-      if ((struct = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(date))) {
-          // avoid NaN timestamps caused by “undefined” values being passed to Date.UTC
-          for (var i = 0, k; (k = numericKeys[i]); ++i) {
-            struct[k] = +struct[k] || 0;
-          }
-
-          // allow undefined days and months
-          struct[2] = (+struct[2] || 1) - 1;
-          struct[3] = +struct[3] || 1;
-
-          if (struct[8] !== 'Z' && struct[9] !== undefined) {
-            minutesOffset = struct[10] * 60 + struct[11];
-
-            if (struct[9] === '+') {
-                minutesOffset = 0 - minutesOffset;
-            }
-          }
-
-          timestamp = Date.UTC(struct[1], struct[2], struct[3], struct[4], struct[5] + minutesOffset, struct[6], struct[7]);
-      }
-      else {
-          timestamp = origParse ? origParse(date) : NaN;
-      }
-
-      return timestamp;
-    };
+    // Date.parse
+    if (!qx.core.Environment.get("ecmascript.date.parse")) {
+      Date.originalParse = Date.parse || function(date) { return NaN; };
+      Date.parse = statics.parse; 
+    }
   }
 });
--- a/framework/source/class/qx/bom/client/EcmaScript.js
+++ b/framework/source/class/qx/bom/client/EcmaScript.js
@@ -208,6 +208,15 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
 
 
     /**
+     * Checks if 'parse' supports ISO-8601 parsing
+     * @internal
+     * @return {Boolean} <code>true</code>, if the method supports ISO-8601 dates.
+     */
+    getDateParse : function() {
+        return Date.parse("1970-01-01T00:00:00Z") === 0;
+    },
+
+    /**
      * Checks if 'startsWith' is supported on the String object.
      * @internal
      * @return {Boolean} <code>true</code>, if the method is available.
@@ -254,6 +263,7 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
 
     // date polyfill
     qx.core.Environment.add("ecmascript.date.now", statics.getDateNow);
+    qx.core.Environment.add("ecmascript.date.parse", statics.getDateParse);
 
     // error bugfix
     qx.core.Environment.add("ecmascript.error.toString", statics.getErrorToString);
--- a/framework/source/class/qx/core/Environment.js
+++ b/framework/source/class/qx/core/Environment.js
@@ -278,6 +278,10 @@
  *       <td>{@link qx.bom.client.EcmaScript#getDateNow}</td>
  *     </tr>
  *     <tr>
+ *       <td>ecmascript.date.parse<td><i>Boolean</i></td><td><code>true</code></td>
+ *       <td>{@link qx.bom.client.EcmaScript#getDateParse}</td>
+ *     </tr>
+ *     <tr>
  *       <td>ecmascript.error.toString</td><td><i>Boolean</i></td><td><code>true</code></td>
  *       <td>{@link qx.bom.client.EcmaScript#getErrorToString}</td>
  *     </tr>
--- a/framework/source/class/qx/test/lang/normalize/Date.js
+++ b/framework/source/class/qx/test/lang/normalize/Date.js
@@ -26,7 +26,7 @@ qx.Class.define("qx.test.lang.normalize.Date",
   {
     testParse: function()
     {
-      var sixHours       = 6 * 60 * 60 * 1000,
+      var sixHours = 6 * 60 * 60 * 1000;
       var sixHoursThirty = sixHours + 30 * 60 * 1000;
 
       // Date part
--- a/framework/source/class/qx/test/lang/normalize/Date.js
+++ b/framework/source/class/qx/test/lang/normalize/Date.js
@@ -1,0 +1,93 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2007-2008 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+
+************************************************************************ */
+
+/**
+ * @require(qx.lang.normalize.Date)
+ */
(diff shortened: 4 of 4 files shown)


oetiker:
Does ecma specify anything regarding date parsing?

oetiker:
can you please add a test too

gnikolaidis:
Sure, coming up later today - I have to figure out how to write tests for the framework first

cajus:
`framework/source/class/qx/test/lang/normalize/Date.js` would be the file to add. Looks like the original `qx.lang.normalize.Date` is so simple, that there's no test for it...

oetiker:
way to go! like the capability check

gnikolaidis:
And a blunder snuck in seconds before the commit... pushing again :-/

oetiker:

schnepe2:
The simple check `qx.core.Environment.get("ecmascript.date.parse")` is not enough I think.
FF e.g. has a parse method on Date, but it is not delivering the expected results.
A specialized check for "date.parse.CanParseDateOnly" or "date.parse.EcmaConfirm" or something like that might be necessary.

It seems that the Date.parse tests only pass with Chrome...(Windows 10).
FF **has** a parse method, but it is not delivering the expected results.

oetiker:
@schnepe2 the ecmascript.date.parse property is set depending on a test that is part of this PR ... so you are saying that test should be more rigorose ?

schnepe2:
@oetiker no, the  "ecmascript.date.parse" check is O.K. (as it checks for presence of parse)
I recommend a more rigorose test at qx.lang.normalize.Date.defer ...
something like (untested!):
```
    // Date.parse
    if (!qx.core.Environment.get("ecmascript.date.parse")) {
      Date.originalParse = Date.parse || function (dateString) { return NaN; };
      Date.parse = statics.parse;
    }
    // Date.parse not ECMA confirm?
    if (!qx.core.Environment.get("ecmascript.date.parseEcmaConfir0")) {
      // override with our implementation...
      Date.parse = statics.parse;
    }

```

framework/source/class/qx/lang/normalize/Date.js
```diff
@@ -37,6 +38,39 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
...
+    },
+
+    /**
+     * Parses a string representation of a date and return number of milliseconds since Epoch or NaN if string is unrecognised
+     *
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse">MDN documentation</a>
+     *
+     * Derived from <https://github.com/csnover/js-iso8601>: ©2011 Colin Snover <http://zetafleet.com>, MIT license
+     */
+    parse : function(date) {
```
schnepe2: I would recommend to use `dateString` as parameter name, as it clearly tells the type and is also used at MDN "reference".
Thus making the API Doc and first lines of code like:

```
    /**
     * Parses a string representation of a date and return number of
     * milliseconds since Epoch or NaN if string is unrecognised
     *
     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse">MDN documentation</a>
     *
     * Derived from <https://github.com/csnover/js-iso8601>: ©2011 Colin Snover
     * <http://zetafleet.com>, MIT license
     *
     * @param dateString {String} A string representing an RFC2822 or ISO 8601
     *   date (other formats may be used, but results may be unexpected).
     * @return {Number|NaN} A number representing the milliseconds elapsed
     *   since January 1, 1970, 00:00:00 UTC and the date obtained by parsing
     *   the given string representation of a date.
     *   If the argument doesn't represent a valid date, NaN is returned.
     */
    parse : function (dateString)
    {
      // Match input against ISO8601 regular expression
      var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(dateString);
      if (!captureGroups) {
        return Date.originalParse(dateString);
      }
```

oetiker:
@schnepe2 but it also checks whether a particular string can be parsed ... I think the idea is that if it does not conform then it will not be able to parse the string ... why would you want to have an extra feature for the presence of a potentially broken parse function ...

schnepe2:
@oetiker Ah, yes it does check `Date.parse("1970-01-01T00:00:00Z") === 0` too...
Nevertheless the provided tests of this PR for these values fail (on FF, Edge, ...):

- "2001-02-03T04:05"
- "2001-02-03T04:05:06"
- "2001-02-03T04:05:06.007"
- "2001T04:05:06.007"
- "2001-02T04:05:06.007"
- "-010000T04:05"
- "-010000-02T04:05"
- "-010000-02-03T04:05"

So one can argue that the "`ecmascript.date.parse`" test might be extended somewhat...
like...
```
getDateParse : function () {
  return Date.parse("1970-01-01T00:00:00Z") === 0
     &&  Date.parse("-010000-02-03T04:05") == -377733876900000;
}
```
...or whatever 2nd check is best and simplest to check for "compatibility".

My idea was to have two dedicated checks, one for "parse" and one for "parse does confirm our expectations"
```
  // qx.core.Environment.get("ecmascript.date.parse")
  getDateParse : function () {
    return Date.parse("1970-01-01T00:00:00Z") === 0;
  },

  // qx.core.Environment.get("ecmascript.date.parseconfirms")
  getDateParseConfirms : function () {
    return Date.parse("-010000-02-03T04:05") == -377733876900000;
  }
```
Whatever feels better in your opinion is O.K.  :smile:

schnepe2:
BTW: "[...]potentially broken parse function[...]" sounds a little to harsh :grinning: 
Those special cases (see above) are **special** cases, usually a "broken" parse function will work almost always

schnepe2:
FYI: The failing tests might be due to timezone offset problems:

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#ECMAScript_5_ISO-8601_format_support) :

[...] If the string is date and time in ISO 8601 format, it will be treated as local.[...]
So "2001T04:05:06.007" is interpreted as local (for which the test cases must account!)
and "2001T04:05:06.007Z" is clearly UTC

I think the whole issue is "just" non-perfect test cases here...I'll take a look at how to easily fix'em to work in all timezones.

schnepe2:
With these changes to `qx.test.lang.normalize.Date` my FF, Edge & PaleMoon behave well,
but Chrome now fails :grinning:
Chrome seems to treat ISO Strings without timezone specifier as UTC!

So maybe this environment check might be needed
```
  // qx.core.Environment.get("ecmascript.date.parsedefaultstoutc")
  getDateParseDefaultsToUTC : function () {
    return Date.parse("2001-02-03T04:05:06.007" == 981169506007; // != 981173106007;
  }
```


Here's my changes to `qx.test.lang.normalize.Date` :
```
    "test parse()" : function ()
    {
      var sixHours = 6 * 60 * 60 * 1000;
      var sixHoursThirty = sixHours + 30 * 60 * 1000;

      var february = new Date(new Date().getFullYear(), 1, 1);
      var localOffset = february.getTimezoneOffset() * 6e4; // [milliseconds]
      // qx.log.Logger.info("localOffset:" + localOffset);

      // Date part

      this.assertIdentical(Date.parse("1970-01-01"), Date.UTC(1970, 0, 1, 0, 0, 0, 0), "Unix epoch");

      this.assertIdentical(Date.parse("2001"), Date.UTC(2001, 0, 1, 0, 0, 0, 0), "2001");
      this.assertIdentical(Date.parse("2001-02"), Date.UTC(2001, 1, 1, 0, 0, 0, 0), "2001-02");
      this.assertIdentical(Date.parse("2001-02-03"), Date.UTC(2001, 1, 3, 0, 0, 0, 0), "2001-02-03");

      this.assertIdentical(Date.parse("-002001"), Date.UTC(-2001, 0, 1, 0, 0, 0, 0), "-002001");
      this.assertIdentical(Date.parse("-002001-02"), Date.UTC(-2001, 1, 1, 0, 0, 0, 0), "-002001-02");
      this.assertIdentical(Date.parse("-002001-02-03"), Date.UTC(-2001, 1, 3, 0, 0, 0, 0), "-002001-02-03");

      this.assertIdentical(Date.parse("+010000-02"), Date.UTC(10000, 1, 1, 0, 0, 0, 0), "+010000-02");
      this.assertIdentical(Date.parse("+010000-02-03"), Date.UTC(10000, 1, 3, 0, 0, 0, 0), "+010000-02-03");
      this.assertIdentical(Date.parse("-010000-02"), Date.UTC(-10000, 1, 1, 0, 0, 0, 0), "-010000-02");
      this.assertIdentical(Date.parse("-010000-02-03"), Date.UTC(-10000, 1, 3, 0, 0, 0, 0), "-010000-02-03");

      this.assertTrue(isNaN(Date.parse("asdf")), "invalid YYYY (non-digits)");
      this.assertTrue(isNaN(Date.parse("1970-as-df")), "invalid YYYY-MM-DD (non-digits)");

      // Time part

      this.assertIdentical(Date.parse("2001-02-03T04:05"), Date.UTC(2001, 1, 3, 4, 5, 0, 0) + localOffset, "2001-02-03T04:05");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06"), Date.UTC(2001, 1, 3, 4, 5, 6, 0) + localOffset, "2001-02-03T04:05:06");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007"), Date.UTC(2001, 1, 3, 4, 5, 6, 7) + localOffset, "2001-02-03T04:05:06.007");

      this.assertIdentical(Date.parse("2001-02-03T04:05Z"), Date.UTC(2001, 1, 3, 4, 5, 0, 0), "2001-02-03T04:05Z");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06Z"), Date.UTC(2001, 1, 3, 4, 5, 6, 0), "2001-02-03T04:05:06Z");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007Z"), Date.UTC(2001, 1, 3, 4, 5, 6, 7), "2001-02-03T04:05:06.007Z");

      this.assertIdentical(Date.parse("2001-02-03T04:05-00:00"), Date.UTC(2001, 1, 3, 4, 5, 0, 0), "2001-02-03T04:05-00:00");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06-00:00"), Date.UTC(2001, 1, 3, 4, 5, 6, 0), "2001-02-03T04:05:06-00:00");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007-00:00"), Date.UTC(2001, 1, 3, 4, 5, 6, 7), "2001-02-03T04:05:06.007-00:00");

      this.assertIdentical(Date.parse("2001-02-03T04:05+00:00"), Date.UTC(2001, 1, 3, 4, 5, 0, 0), "2001-02-03T04:05+00:00");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06+00:00"), Date.UTC(2001, 1, 3, 4, 5, 6, 0), "2001-02-03T04:05:06+00:00");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007+00:00"), Date.UTC(2001, 1, 3, 4, 5, 6, 7), "2001-02-03T04:05:06.007+00:00");

      this.assertIdentical(Date.parse("2001-02-03T04:05-06:30"), Date.UTC(2001, 1, 3, 4, 5, 0, 0) + sixHoursThirty, "2001-02-03T04:05-06:30");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06-06:30"), Date.UTC(2001, 1, 3, 4, 5, 6, 0) + sixHoursThirty, "2001-02-03T04:05:06-06:30");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007-06:30"), Date.UTC(2001, 1, 3, 4, 5, 6, 7) + sixHoursThirty, "2001-02-03T04:05:06.007-06:30");

      this.assertIdentical(Date.parse("2001-02-03T04:05+06:30"), Date.UTC(2001, 1, 3, 4, 5, 0, 0) - sixHoursThirty, "2001-02-03T04:05+06:30");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06+06:30"), Date.UTC(2001, 1, 3, 4, 5, 6, 0) - sixHoursThirty, "2001-02-03T04:05:06+06:30");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007+06:30"), Date.UTC(2001, 1, 3, 4, 5, 6, 7) - sixHoursThirty, "2001-02-03T04:05:06.007+06:30");

      this.assertIdentical(Date.parse("2001T04:05:06.007"), Date.UTC(2001, 0, 1, 4, 5, 6, 7) + localOffset, "2001T04:05:06.007");
      this.assertIdentical(Date.parse("2001-02T04:05:06.007"), Date.UTC(2001, 1, 1, 4, 5, 6, 7) + localOffset, "2001-02T04:05:06.007");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007"), Date.UTC(2001, 1, 3, 4, 5, 6, 7) + localOffset, "2001-02-03T04:05:06.007");
      this.assertIdentical(Date.parse("2001-02-03T04:05:06.007-06:30"), Date.UTC(2001, 1, 3, 4, 5, 6, 7) + sixHoursThirty, "2001-02-03T04:05:06.007-06:30");

      this.assertIdentical(Date.parse("-010000T04:05"), Date.UTC(-10000, 0, 1, 4, 5, 0, 0) + localOffset, "-010000T04:05");
      this.assertIdentical(Date.parse("-010000-02T04:05"), Date.UTC(-10000, 1, 1, 4, 5, 0, 0) + localOffset, "-010000-02T04:05");
      this.assertIdentical(Date.parse("-010000-02-03T04:05"), Date.UTC(-10000, 1, 3, 4, 5, 0, 0) + localOffset, "-010000-02-03T04:05");

      this.assertTrue(isNaN(Date.parse("1970-01-01T00:00:00,000")), "invalid date-time (comma instead of dot)");
      this.assertTrue(isNaN(Date.parse("1970-01-01T0000")), "invalid date-time (missing colon in time part)");
      this.assertTrue(isNaN(Date.parse("1970-01-01T00:00.000")), "invalid date-time (msec with missing seconds)");
    },

```

schnepe2:
Grrrr I hate it when this happens! I am not sure what client is doin' it right and what is doin' it wrong!
But what numeric values do we have to expect? Here's my current thought. Please correct me if I'm wrong :stuck_out_tongue_winking_eye:

"2001T04:05:06.007" should be treated as local => `978321906007 - localOffset` right?
"2001T04:05:06.007Z" is clearly UTC => `978321906007`
`localOffset` in milliseconds for the given date (DST taken into account)

I still believe that Chrome is wrong 'cause for Chrome
`Date.parse("2001T04:05:06.007Z") == Date.parse("2001T04:05:06.007")`
which should **not** be **true** (until you live in an UTC timezone)

schnepe2:
Hello again,

I've had time to look over this again and I would like share my thoughts:
What about a user configuration switch (for `config.json`) that lets the application developer decide what behavior is preferred. This would allow the original PR poster to keep his application code unchanged.
Something like `qx.date.parse.mode`, which can be set to either `"standard"` or `"chrome"`  defaulting to `"standard"`
The normalizing code ("`qx.lang.normalize.Date`") can than check what needs to be done to fulfill the "users wish" like:
```
qx.Bootstrap.define("qx.lang.normalize.Date", {

  statics : {

    parse_standard : function (dateString) {
      // ...
    },

    parse_chrome : function (dateString) {
      // ...
    }
  },


  defer : function (statics)
  {
    // ...

    //
    // Date.parse
    //

    // Users "wish"
    var polyfill = qx.core.Environment.get("qx.date.parse.mode");
    // Native "situation"
    var hasParse = qx.core.Environment.get("ecmascript.date.parse");
    var isStd = hasParse && qx.core.Environment.get("ecmascript.date.parse.isstandard");

    switch ( hasParse+"-"+isStd )
    {
      case "false-false": break; // take "standard" or "chrome" polyfill
      case "false-true" : break;// should never happen ;)
      case "true-false" : if (polyfill == "chrome") { polyfill = null; } break;
      case "true-true"  : if (polyfill != "chrome") { polyfill = null; } break;
    }
    if (polyfill) // polyfill on need
    {
      Date.originalParse = Date.parse || function (dateString) { return NaN; };
      Date.parse = statics["parse_"+polyfill];
    }
  }
});
```

The Environment checks are:
```
x.Bootstrap.define("qx.bom.client.EcmaScript",
{
  statics :
  {
    /**
     * Checks if 'parse' supports ISO-8601 parsing
     * @internal
     * @return {Boolean} <code>true</code>, if the method supports ISO-8601 dates.
     */
    getDateParse : function () {
      return Date.parse("1970-01-01T00:00:00Z") === 0;
    },


    /**
     * Checks if 'parse' takes ISO-8601 date strings without timezone specifier
     * and treats them as local (as per specification)
     * @internal
     * @return {Boolean} <code>true</code>, if the method supports ISO-8601 dates.
     */
    getDateParseIsStandard : function () {
      return Date.parse("2001-02-03T04:05:06.007") != Date.parse("2001-02-03T04:05:06.007Z");
       // && Date.parse("2001-02-03T00:00:00Z") == Date.parse("2001-02-03"); // treat 'date only' as UTC
    },

  },


  defer : function(statics) {
    // ...

    // date polyfill
    qx.core.Environment.add("ecmascript.date.now", statics.getDateNow);
    qx.core.Environment.add("ecmascript.date.parse", statics.getDateParse);
    qx.core.Environment.add("ecmascript.date.parse.isstandard", statics.getDateParseIsStandard);

    // ...
  }
});
```

What do you guys think of it?
I would prefer to only have the **standard** version, but the code from the PR acts like chrome and seems to be in production use, so why not have both possibilities.
The name for the 2nd -non standard- option "chrome" however might be subject to change :grin:

schnepe2:
BTW: With those changes the tests pass on FF, Chrome, IE, Edge and PaleMoon :smile: 
("standard" mode selected, ...don't know how to write tests for different "compiler options")

gnikolaidis:
Hi everyone, I feel somewhat guilty that this PR is provoking such a big discussion and concern :-/ 

The fact is that all comments are focusing on a real problem; we too have stumbled upon the same difficulties with erroneous implementations of Date.parse from some browser vendors.

I wish to state that our use case depends on the "standard" implementation, that is one where a date without a trailing Z is considered as at client local timezone and not at UTC0.

gnikolaidis:
@schnepe2 one can imagine a case where getDateParseIsStandard() returns true but in fact is not, for countries at UTC+0 (Ireland, Portugal and UK), posing a risk to the behavior of clients during the DST change from winter to summer (I am nitpicking a lot here!).

schnepe2:
> one can imagine a case where getDateParseIsStandard() returns true but in fact is not, for countries at UTC+0 (Ireland, Portugal and UK), posing a risk to the behavior of clients during the DST change from winter to summer (I am nitpicking a lot here!).

No, this is a real issue and should be addressed...however I don't yet know how.
maybe checking for a date that **is** in summer might work ("2001-07-03T..." instead of "2001-02-03T...")?
Do you have any idea for that check?

> I wish to state that our use case depends on the "standard" implementation, that is one where a date without a trailing Z is considered as at client local timezone and not at UTC0.

Fine. So we only have to provide **one** polyfill :+1:
This one seems to do the job (it might be implemented nicer, though):
```javascript
    /**
     * Parses a string representation of a date and return number of
     * milliseconds since Epoch or NaN if string is unrecognized.
     *
     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse">MDN documentation</a>
     *
     * Derived from <https://github.com/csnover/js-iso8601>: ©2011 Colin Snover
     * <http://zetafleet.com>, MIT license
     *
     * @param dateString {String} A string representing an RFC2822 or ISO 8601
     *   date (other formats may be used, but results may be unexpected).
     * @return {Number|NaN} A number representing the milliseconds elapsed
     *   since January 1, 1970, 00:00:00 UTC and the date obtained by parsing
     *   the given string representation of a date.
     *   If the argument doesn't represent a valid date, NaN is returned.
     */
    parse : function (dateString)
    {
      // Match input against ISO8601 regular expression
      var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(dateString);
      if (!captureGroups) {
        return Date.originalParse(dateString);
      }

      // Just a date without time?
      var noTime = [ 4, 5, 6, 7 ].every(function (i) {
        return captureGroups[i] === undefined;
      });

      // Avoid invalid timestamps caused by undefined values being passed to Date.UTC
      [ 1, 4, 5, 6, 7, 10, 11 ].forEach(function (i) {
        captureGroups[i] = +captureGroups[i] || 0;
      });
      captureGroups[2] = (+captureGroups[2] || 1) - 1; // Allow undefined months
      captureGroups[3] =  +captureGroups[3] || 1;      // Allow undefined days

      // No timezone offset given and *not* just a date (without time)
      if (captureGroups[8] !== "Z" && captureGroups[9] === undefined && !noTime)
      { // => Treat as local
        return new Date( captureGroups[1], captureGroups[2], captureGroups[3],
                         captureGroups[4], captureGroups[5], captureGroups[6],
                         captureGroups[7] ).getTime();
      }

      // Handle timezone offsets
      var minutesOffset = 0;
      if (captureGroups[8] !== "Z")
      {
        minutesOffset = captureGroups[10] * 60 + captureGroups[11];
        if (captureGroups[9] === "+") {
          minutesOffset = - minutesOffset;
        }
      }

      // Return the number of milliseconds since Epoch
      return Date.UTC( captureGroups[1], captureGroups[2], captureGroups[3],
                       captureGroups[4], captureGroups[5] + minutesOffset,
                       captureGroups[6], captureGroups[7] );
    },
```

gnikolaidis:
This is our current implementation:

		parse : function(dateString) {
			// Match input against ISO8601 regular expression
			var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(dateString);
			if(!captureGroups) {
				return Date.originalParse(dateString);
			}

			// Avoid invalid timestamps caused by undefined values being passed to Date.UTC
			[ 1, 4, 5, 6, 7, 10, 11 ].forEach(function(i) {
				captureGroups[i] = +captureGroups[i] || 0;
			});
			captureGroups[2] = (+captureGroups[2] || 1) - 1; // Allow undefined months
			captureGroups[3] = +captureGroups[3] || 1; // Allow undefined days

			// Handle timezone offsets
			var minutesOffset;
			if(captureGroups[8] === "Z") {
				minutesOffset = 0;
			}
			else if(captureGroups[9] !== undefined) {
				minutesOffset = captureGroups[10] * 60 + captureGroups[11];
				if(captureGroups[9] === "+")
					minutesOffset = 0 - minutesOffset;
			}
			else {
				minutesOffset = Date.TIMEZONE_OFFSET
			}

			// Return the number of milliseconds since Epoch
			return Date.UTC(captureGroups[1], captureGroups[2], captureGroups[3], captureGroups[4], captureGroups[5] + minutesOffset, captureGroups[6], captureGroups[7]);
		}


with an added line in defer:

	defer: function(statics) {

		Date.TIMEZONE_OFFSET = new Date().getTimezoneOffset();
		...

gnikolaidis:
I have no idea either about the correct way of detecting a good implementation of Date.parse, I will look into it

gnikolaidis:
Just realized that our implementation suffers from the same problem with DST changes (minutesOffset = Date.TIMEZONE_OFFSET will be wrong after the TZ change) !

schnepe2:
Wouldn't the line `Date.TIMEZONE_OFFSET = new Date().getTimezoneOffset();`  just get the timezone offset of the current date (now) ? Or is that intended?

What about this one:
```
    /**
     * Returns the timezone offset of the standard time
     * (without 'daylight savings time').
     *
     * @return {Integer} timezone offset (in minutes)
     */
    getStandardTimezoneOffset : function ()
    {
      var jan = new Date(this.getFullYear(), 0, 1);
      var jul = new Date(this.getFullYear(), 6, 1);
      return Math.max(jan.getTimezoneOffset(), jul.getTimezoneOffset());
    },

```
...which gives `-60` here in Germany independent if it is issued in July ("GMT+02:00") or December ("GMT+01:00").

gnikolaidis:
@schnepe2 you are absolutely right. Please disregard my previous code snippet.

schnepe2:
@gnikolaidis Yeah wrapping ones head around other timezones etc. if fun isn't it :stuck_out_tongue_winking_eye: 

The main problem is: For stings without a timezone specifier one has to get the timezone offset **after** you set the date (so DST / no DST) can be taken into account. Therefore my implementation just passes the values to the Date c'tor (without `Date.UTC(...)` ) in that special case.

```javascript
      // No timezone offset given and *not* just a date (without time)
      if (captureGroups[8] !== "Z" && captureGroups[9] === undefined && !noTime)
      { // => Treat as local
        return new Date( captureGroups[1], captureGroups[2], captureGroups[3],
                         captureGroups[4], captureGroups[5], captureGroups[6],
                         captureGroups[7] ).getTime();
      }
```

oetiker:
how about updating the PR with the latest ideas ?

schnepe2:
...I could make a PR with the 'current' state. But the discussion of **this PR** is lost then I think...
Although I don't want to take any credits for things @gnikolaidis did :grin:
The original poster can however stop me from doing so :smile:

gnikolaidis:
Please @schnepe2, take the credits and make a PR from the current state! Your version is the correct one, and time constraints do not allow me to me make the PR as good as Qooxdoo deserves :-)

PR #9113, opened 2016-07-27
Arrays

johnspackman:
This PR synchronises `qx.lang.Array` and `qx.data.Array` and adds a new `.replace()` method

The intention is to remove clutter that can crop up when dealing with arrays, especially when you are not able to know in advance whether the array is a `qx.data.Array` or a native array; for example:

```
if (arr instanceof qx.data.Array)
  arr = arr.toArray();
if (otherArray instanceof qx.data.Array)
  otherArray = otherArray.toArray();
return qx.lang.Array.equals(arr, otherArray);
```

becomes:

```
return qx.lang.Array.equals(arr, anotherArray);
```

Similarly, `qx.data.Array.append` already supported native arrays as the source array, but other methods like `.equals` did not.

The `.replace` method is added because when editing a qx.data.Array it is often desirable to avoid multiple `change` events caused by adding and removing multiple entries; it's easy to achieve this is by splice-ing in a new array, but it's clunky code that should be kept out of the application.

Also missing unit tests for some methods

### DIFF
--- a/framework/source/class/qx/data/Array.js
+++ b/framework/source/class/qx/data/Array.js
@@ -141,13 +141,14 @@ qx.Class.define("qx.data.Array",
     /**
      * Concatenates the current and the given array into a new one.
      *
-     * @param array {Array} The javaScript array which should be concatenated
+     * @param array {qx.data.Array|Array} The javaScript array which should be concatenated
      *   to the current array.
      *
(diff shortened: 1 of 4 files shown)


level420:
LGTM

cajus:

cajus:
Two business days have passed. No vote against.

PR #9119, opened 2016-08-11
Allow the parent of MMoveable to be scrollable and calculate the scroll in.

jochumdev:
Allow the parent of MMoveable to be scrollable and calculate the scroll in.

This is required when you have for example a Desktop scrollable.

### DIFF
--- a/framework/source/class/qx/ui/core/MMovable.js
+++ b/framework/source/class/qx/ui/core/MMovable.js
@@ -68,6 +68,9 @@ qx.Mixin.define("qx.ui.core.MMovable",
     __parentLeft : null,
     __parentTop : null,
 
+    __parentScrollLeft : 0,
+    __parentScrollTop : 0,
+
     __blockerAdded : false,
     __oldBlockerColor : null,
     __oldBlockerOpacity : 0,
@@ -159,8 +162,8 @@ qx.Mixin.define("qx.ui.core.MMovable",
         viewportLeft : parseInt(viewportLeft, 10),
         viewportTop : parseInt(viewportTop, 10),
 
-        parentLeft : parseInt(viewportLeft - this.__parentLeft, 10),
-        parentTop : parseInt(viewportTop - this.__parentTop, 10)
+        parentLeft : parseInt(viewportLeft - this.__parentLeft, 10) + this.__parentScrollLeft,
+        parentTop : parseInt(viewportTop - this.__parentTop, 10) + this.__parentScrollTop
       };
     },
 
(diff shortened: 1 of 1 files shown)


oetiker:
how does the behavior change vs the existing behavior? your change does not seem to be configurable in any way ...

pcdummy:
The current implementation just doesn't work with scrollable. Need to say that Desktop disables scrolling by default.

level420:
@pcdummy this, I think, implies that you derived from the standard desktop and enabled scrolling, right?

pcdummy:
@level420 correct, yes.

oetiker:
what worries me is that it is a change of behavior in a core component ... which gets applied regardless ... is there a way to figure if the parent intends to be scrollable ... 

I guess at the bottom of my unease is that this is going for the dom not looking at any qooxdoo info ...

level420:
I've tested the PR locally with current IE11, Firefox, Chrome on Windows 7, Edge on Windows 10, Safari on OSX 10.11. I've moved windows around left, top, bottom, right up to one px outside of the desktop and could not find any problem.

I'm occasionally getting requests from customers where they ask if it would be possible to have the desktop scroll able for devices with low height (eg ultrabooks). So optimally I'd like to also have @pcdummy s implementation of the scrollable desktop in.

level420:
... I've only done tests with the PR from @pcdummy and the standard desktop to see if It does any harm to my apps (it does not), but have not tested it with a scrollable desktop yet.

@pcdummy would you mind creating another PR with your scrollable desktop class? That would be great and would allow testing this PR with a real scrolling desktop.

oetiker:
When users want a scrollable desktop, I do this:

```javascript
                var layout = new qx.ui.layout.VBox();
                var desktop = new qx.ui.container.Composite().set({
                        layout: layout
                });
                var scroll = new qx.ui.container.Scroll(desktop);
                var root = this.getRoot();
                root.add(scroll, {
                    width  : '100%',
                    height : '100%'
                }); 
```

level420:
@pcdummy would the solution of @oetiker above also fit your needs? If yes it would maybe obsolete this PR?!?!

level420:
@oetiker does your solution also work for qx.ui.window.Windows? If I have a window that eg higher than the browser window?

pcdummy:
I can't see how the solution from @oetiker should work with Windows, it will work for non Moveable but for Windows?

With your solution @oetiker you don't use qx.ui.window.Desktop and qx.ui.window.Window so you don't have a MMoveable in it, right?

oetiker:
```javascript
var desktop = new  qx.ui.window.Desktop().set({
  width: 400,
  height: 400
});

var scr = new qx.ui.container.Scroll(desktop).set({
    width: 100,
    height: 100
});

var root = this.getRoot();
root.add(scr, {
    top: 10,
    left: 10
});
var win = new qx.ui.window.Window()
desktop.add(win);
win.open();
```

so I would have assumed something like this should work ... but it does not ... will  your patch make this work ? Or does it already work and I just did something wrong in my code ?

pcdummy:
This is the code that i thought doesn't work:

```javascript
var desktop = new  qx.ui.window.Desktop().set({
  width: 1500,
  height: 1500
});
desktop.getContentElement().enableScrolling();


var root = this.getRoot();
root.add(desktop, {
    top: 0,
    left: 0
});
var win = new qx.ui.window.Window()
desktop.add(win);
win.open();
```

Now i have to figure out why it works and why i made this patch.

level420:
@pcdummy is this PR still valid?

pcdummy:
I close it and reopen/create new if needed.

PR #9122, opened 2016-08-12
Let others know when we moved in q.u.c.MMovable.

jochumdev:

### DIFF
--- a/framework/source/class/qx/ui/core/MMovable.js
+++ b/framework/source/class/qx/ui/core/MMovable.js
@@ -26,6 +26,19 @@
  */
 qx.Mixin.define("qx.ui.core.MMovable",
 {
+
+  /*
+  *****************************************************************************
+     EVENTS
+  *****************************************************************************
+  */
+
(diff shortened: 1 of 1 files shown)


cajus:
This will flood the event queue with many events that don't contain much more than the information that the mouse has moved. But not where, etc..

 So if someone wants to use it, he might need to add additional code to find out where. From a performance point of view, I'm not sure if it's good to do that additionally. What about just listening for pointermove in your own code?

pcdummy:
Thanks for your input @cajus, i can't listen on pointermove as this does e.stopPropagation().
Before this i've overwritten onMovePointerMove and then called this.base(arguments, e). not sure whats better.

cajus:
Did you try _caputure=true_ vs _capture=false_ in the listener?

pcdummy:
Thank you @cajus, `capture=true` works for me which makes this PR obsolete.

cajus:
Cool ;-)

PR #9123, opened 2016-08-12
Fix null assertion when setting q.u.w.Desktop WindowManager.

jochumdev:
I'm not sure if this is the right way to fix this.

[Here](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/window/MDesktop.js#L110) we reset the old WindowManager by setting it to null.

This will raise an error in the assertInterface code.

### DIFF
--- a/framework/source/class/qx/ui/window/IWindowManager.js
+++ b/framework/source/class/qx/ui/window/IWindowManager.js
@@ -29,10 +29,12 @@ qx.Interface.define("qx.ui.window.IWindowManager",
     /**
      * Connect the window manager to the window desktop
      *
-     * @param desktop {IDesktop} The connected desktop
+     * @param desktop {IDesktop|null} The connected desktop or null
      */
     setDesktop : function(desktop) {
-      this.assertInterface(desktop, qx.ui.window.IDesktop);
+      if (desktop !== null) {
+        this.assertInterface(desktop, qx.ui.window.IDesktop);
+      }
     },
 
     /**


cajus:
I'm wondering if it is valid to set the desktop to _null_ at all. The DesktopManager code will break if it is set to _null_.

level420:
@pcdummy could you please explain in which situation you use setWindowManager? I suppose you use you own derived window manager class? Or do you simply use qx.ui.window.Desktop instances?

pcdummy:
I've [forked](https://github.com/drawstack/qxjoint/tree/master/source/class/qxjoint/widget/viewport) qx.ui.window.Desktop for QxJoint as need some more features and removed some for my usecase uneeded code.

I had multiple `qx.ui.window.Desktop` where all of them shared a single Manager, this is where i used that.

level420:
@cajus: I think this are the position where it may be necessary to set the desktop to null:

Here setting the desktop to null first before setting it to the new value:
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/window/MDesktop.js#L109-L114

and here where method updateStack is called while this.__desktop is null:
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/window/Manager.js#L43-L44

where the single use of this is to push the manager twice in the widget queue by calling
```javascript
```

level420:
No wait! The use of https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/window/MDesktop.js#L109-L114
is to remove the connection of the desktop from the old manager and initialize the new one.

So I thin @pcdummy 's PR is correct.

johnspackman:
i agree, null is legitimate (even if there are very few operations which are legitimate until a non-null value is set)

framework/source/class/qx/ui/window/IWindowManager.js
```diff
@@ -32,7 +32,9 @@ qx.Interface.define("qx.ui.window.IWindowManager",
      * @param desktop {IDesktop} The connected desktop
      */
     setDesktop : function(desktop) {
-      this.assertInterface(desktop, qx.ui.window.IDesktop);
+      if (desktop) {
```
level420: I think the parameter desktop should only be null or having the interface IDesktop implemented, so it should fail as well for undefined.

I'll change the PR to
```javascript
    if (desktop !== null) {
```

level420:
@pcdummy could you please modify the PR so that the assert is done only if desktop is explicitly !== null ?
Thank you!

framework/source/class/qx/ui/window/IWindowManager.js
```diff
@@ -32,7 +32,9 @@ qx.Interface.define("qx.ui.window.IWindowManager",
      * @param desktop {IDesktop} The connected desktop
```
level420: @pcdummy please also document here that null is explicitly allowed as parameter like
```javascript
      * @param desktop {IDesktop|null} The connected desktop
```

pcdummy:
@level420 and @johnspackman done thx for your both's input.

level420:

level420:
Five votes. Merging.
Thank you @pcdummy

PR #9124, opened 2016-08-12
sync changes from master to current

oetiker:
this bring current (5.0.x) into sync with master including all the changes by @cajus and @hkollmann .


oetiker:
it will be interesting to see if this will do the right thing on travis @cajus ... should I try?

cajus:
If it does locally, I don't see any reason why it will fail on travis. Give it a go ;-)

PR #9132, opened 2016-08-25
Forward property selectable from Atom to its label child control

zaucker:
Setting the selectable property on Atom has no effect so far. It must be "forwarded" to its child control "label".

### DIFF
--- a/framework/source/class/qx/test/ui/basic/Atom.js
+++ b/framework/source/class/qx/test/ui/basic/Atom.js
@@ -1,0 +1,56 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016- Oetiker+Partner AG, Switzerland, http://www.oetiker.ch
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+   * Fritz Zaucker (fritz.zaucker@oetiker.ch)
+
+************************************************************************ */
+
+/**
+ * @asset(qx/test/webfonts/*)
+ */
+
+qx.Class.define("qx.test.ui.basic.Atom",
+{
+  extend : qx.test.ui.LayoutTestCase,
+
+  include : [qx.dev.unit.MRequirements, qx.dev.unit.MMock],
+
+  members :
+  {
+    tearDown : function()
+    {
+      this.base(arguments);
+      this.getSandbox().restore();
+      qx.bom.webfonts.Manager.getInstance().dispose();
+      delete qx.bom.webfonts.Manager.$$instance;
+    },
+
+    testSelectableSet : function() {
+      var a = new qx.ui.basic.Atom();
+      var l = a.getChildControl('label');	
+      a.setSelectable(true);
+      this.assert(l.getContentElement().getAttribute("qxselectable"));
+      l.dispose();
+    },
+
+    testSelectableUnset : function() {
+      var a = new qx.ui.basic.Atom();
+      var l = a.getChildControl('label');	
+      l.setSelectable(false);
+      this.assert(! l.getContentElement().getAttribute("qxselectable"));
+      l.dispose();
+    },
+
+  }
+});
--- a/framework/source/class/qx/test/ui/basic/Atom.js
+++ b/framework/source/class/qx/test/ui/basic/Atom.js
@@ -16,10 +16,6 @@
 
 ************************************************************************ */
 
-/**
- * @asset(qx/test/webfonts/*)
- */
-
 qx.Class.define("qx.test.ui.basic.Atom",
 {
   extend : qx.test.ui.LayoutTestCase,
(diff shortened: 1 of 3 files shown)


framework/source/class/qx/ui/basic/Atom.js
```diff
@@ -121,6 +121,15 @@ qx.Class.define("qx.ui.basic.Atom",
     },
 
 
+    /** this sets the property on the label child control */
+    selectable :
+    {
+      refine : true,
+      init : false
+      apply : "_applySelectable",
```
level420: the comma is needed one line above and has to be deleted here.

framework/source/class/qx/ui/basic/Atom.js
```diff
@@ -293,6 +302,16 @@ qx.Class.define("qx.ui.basic.Atom",
...
+    _applySelectable : function(value, old)
+    {
+      var label = this.getChildControl("label", true);
+      if (label) {
+        label.setSelectable(value);
+      }
+    },
+
+
+    // property apply
```
cajus: Instead of _applySelectable you can bind the value to the label when creating it in _createChildControlImpl. Would save some lines ;-)

level420:
I think you also should take the situation in account where the selectable property is set and the label instance is not created yet. Then you need to set it on label instance creation.

framework/source/class/qx/ui/basic/Atom.js
level420: Yes, but doesn't this imply that an event changeSelectable is avalable?

cajus:
@level420 which is covered if you use a binding in _createChildControlImpl.

level420:
@cajus didn't knew that. Thank you for the hint! So I don't need to declare the change event explicitly?

zaucker:
@cajsu: can you tell me how to bind it in _createChildControlImpl?

cajus:
@zaucker yes, you need a change event in your property definition. Then you can just add a

```
this.bind("selectable", control, "selectable");
```

in the "label" branch of _createChildControlImpl. The binding is removed on dispose automatically.

cajus:
@level420 yes, you need a change event. Sorry if I got that wrong ;-)

level420:
But we also need a
```javascript
control.setSelectable(this.getSelectable());
```
or not?

cajus:
No. That's done by the bind.

level420:
The question is if we want an additional binding with event firing and so on for such a basic widget.

level420:
... being always worried regarding widget instantiation performance

cajus:
@level420 Good question. That was just a proposal for saving some lines of code.

oetiker:
@level420 is binding an expensive process ?

level420:
But wait! Have a look into https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/basic/Atom.js#L334

there is already code in there for forwarding the selectable state!

level420:
And https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/Widget.js#L790 already has a selectable property which should be inherited to Atom.

zaucker:
Hmm, didn't see the code for _applySelectable(). Why is it not working properly? I'll check, I guess it is the missing control.setSelectable(this.getSelectable()) suggested by level420

level420:
@zaucker Yes I think the initialization of the selectable property in ```_createChildControlImpl``` is missing

framework/source/class/qx/test/ui/basic/Atom.js
```diff
@@ -0,0 +1,52 @@
...
+      l.dispose();
+    },
+
+    testSelectableUnset : function() {
+      var a = new qx.ui.basic.Atom();
+      var l = a.getChildControl('label');	
+      l.setSelectable(false);
+      this.assert(! l.getContentElement().getAttribute("qxselectable"));
+      l.dispose();
+    },
```
level420: @zaucker you still have a dangling comma here. This won't compile

zaucker:
Yes, this fixed it. In my app I used new qx.ui.basic.Atom().set({ selectable : true}). Hadn't thought about trying to set it separately. Which I guess is good, as I thus found a bug ... sorry for not realizing what the real problem was (which my inital attempt wouldn't have fixed). BTW, in the mean-time I switched my app to using qx.ui.basic.Label instead of Atom, as I don't need Atom's Icon anyway. Now I only have to get the tests right and then figure out about squashing ...

zaucker:
@level420 Thanks ...

oetiker:
now you may want to update the test to actually test the thing you fixed

framework/source/class/qx/test/ui/basic/Atom.js
```diff
@@ -20,12 +20,17 @@ qx.Class.define("qx.test.ui.basic.Atom",
 {
   extend : qx.test.ui.LayoutTestCase,
 
+<<<<<<< HEAD
   include : [qx.dev.unit.MMock],
+=======
+  include : [qx.dev.unit.MRequirements, qx.dev.unit.MMock],
+>>>>>>> 4db393876743f81ce2fcf21465b8512054d3df0b
```
level420: @zaucker somehow diff/patch fragments slipped into your code

zaucker:
I am closing this PR as I can't get the commit squashing to work. Will create new PR. Thanks for all the input.

PR #9133, opened 2016-08-25
Add test for property selectable on Atom

zaucker:
Forward property selectable from Atom to it's label child control
Add test for setting/unsetting selectable property on Label
Fix typo in test name for Label

### DIFF
--- a/framework/source/class/qx/test/ui/basic/Atom.js
+++ b/framework/source/class/qx/test/ui/basic/Atom.js
@@ -1,0 +1,71 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016- Oetiker+Partner AG, Switzerland, http://www.oetiker.ch
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+   * Fritz Zaucker (fritz.zaucker@oetiker.ch)
+
+************************************************************************ */
+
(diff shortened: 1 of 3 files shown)


level420:
@zaucker the travis tests fail exactly in you newly added tests.

zaucker:
I guess my test must be wrong. Just don't see (yet) why ...

level420:
maybe the dom is not ready yet, and does not reflect the attributes. You maybe have to flush the queues (just guessing)?

zaucker:
Giving it a try ...

level420:
Have a look into this test. I think there is no dom created until you add the label to the root and then flush it:
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/bom/Font.js#L146

zaucker:
@level420 Thanks, another try.

level420:
and it should be ```.getContentElement().getDomElement().getAttribute('qxselectable')```
Tested this here locally in my app

zaucker:
@level420 Thanks again, I guess testing locally would have been more efficient than going through Travis several times ...

level420:

zaucker:
Hmm, not sure if voting for my own PR is really appropriate ... :-)

oetiker:
@zaucker your suspicion is correct you can not vote for your own PR as per our rules in https://github.com/qooxdoo/qooxdoo-wiki/blob/master/Developer-Documentation.md

zaucker:
OK, I just unvoted ... :-)

level420:
Five votes. Merging.
Thank you @zaucker

PR #9140, opened 2016-08-29
Add method getSubcontrolId returning the id used in _createChildControlImpl

level420:
This allows, e.g. while listening to the createChildControl event, retrieval of the used id during child control creation.

### DIFF
--- a/framework/source/class/qx/ui/core/Widget.js
+++ b/framework/source/class/qx/ui/core/Widget.js
@@ -3687,6 +3687,19 @@ qx.Class.define("qx.ui.core.Widget",
     },
 
 
+    /**
+     * Return the ID of the current widget if it is a created as a 
+     * subcontrol of another widget. 
+     * 
+     * See the first parameter id in {@link qx.ui.core.Widget#_createChildControlImpl} 
+     *
+     * @param id {String|null} ID of the current widget or null if it was not created as a subcontrol
+     */
+    getSubcontrolId : function()
+    {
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/core/Widget.js
```diff
@@ -3687,6 +3687,19 @@ qx.Class.define("qx.ui.core.Widget",
     },
 
 
+    /**
+     * Return the ID of the current widget if it is a created as a
```
oetiker: "if it is a created as a"  ?

level420: yes. But the name is called id in e.g. _createChildControlImpl https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/Widget.js#L3629
So I thought it's the right naming

oetiker: OK, then this:

> If the widget has been created as a ChildControl, return the id (name) of the ChildControl.

the problem is that your current sentence does not really work ... it is missing words.

level420: OK then It's my bad english. Will try to do it better.

oetiker:
much better

oetiker:

voger:
How do I vote? LGTM

level420:
@voger Sorry, voting is only possible for the core-team members.

voger:
Oh I didn't know that. Still looks good.

PR #9134, opened 2016-08-26
Add ISO8601 parsing support to qx.lang.normalize.Date #7175

schnepe2:
This is the PR in response to the discussion at https://github.com/qooxdoo/qooxdoo/pull/9112

Note: That this PR does only provide one (standard confirm) `parse`. (see discussion above)

Note: The check `qx.bom.client.EcmaScript#getDateParse` might return `false` when run in a timezone equal to GMT, but then the only thing that happens is that the (already compliant) `Date.parse` method might get replaced by our (also compliant) `Date.parse`. So it's fail-save.
But if someone has a better idea,...bring it on :grinning:

PR #9145, opened 2016-08-30
Added ISO8601 parsing support to qx.lang.normalize.Date #7175

schnepe2:
This is the (2nd) PR in response to the discussion at #9134


pcdummy:
Nice one.

schnepe2:
I would have added a :+1: vote, 'cause initially this PR was from @gnikolaidis,
but I believe this would be considered as a "self-vote" :smile:

gnikolaidis:
On the other hand, I am free vote without breaking the law :-)

level420:
Three votes, more than 2 business days.
Merging.
Thank you @schnepe2

Issue #9118, opened 2016-08-10
Checkbox shouldn't toggle when pressing enter


RemiHeugue:
When you focus a checkbox and press enter, it's toggling between checked and unchecked.

This is inconsistent with nearly all GUI toolkits (and even the browser native checkbox) and it prevents things like submitting a form by pressing enter if a checkbox is focused.

That may be the case for other widgets too (I've seen select boxes doing the same).

PR #9146, opened 2016-08-30
Prevent check box toggling and stopping event when pressing enter

RemiHeugue:
PR for #9118.

The check box doesn't toggle and doesn't stop the event when pressing enter while it's focused anymore.

### DIFF
--- a/framework/source/class/qx/ui/form/CheckBox.js
+++ b/framework/source/class/qx/ui/form/CheckBox.js
@@ -107,6 +107,50 @@ qx.Class.define("qx.ui.form.CheckBox",
       "toolTipText",
       "value",
       "menu"
-    ]
+    ],
+
+    /**
+     * Listener method for "keydown" event.<br/>
+     * Removes "abandoned" and adds "pressed" state
+     * for the keys "Enter" or "Space"
+     *
+     * @param e {Event} Key event
+     */
+    _onKeyDown : function(e)
+    {
+      switch(e.getKeyIdentifier())
+      {
+        case "Space":
+          this.removeState("abandoned");
+          this.addState("pressed");
+
+          e.stopPropagation();
+      }
+    },
+
+
+    /**
+     * Listener method for "keyup" event.<br/>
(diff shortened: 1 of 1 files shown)


oetiker:
makes sense to me ... I wonder ... is this a breaking change ? maybe it would be good to add a note to the top of the class that interaction has been changed from enter to space for qx 6.0

level420:
It is a breaking change and also applies for radio buttons having the focus. 

So the change has to be done in ToggleButton https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/ToggleButton.js#L309 and in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/ToggleButton.js#L334
where the enter key is explicitly used to trigger.

This behavior seems to be there for a long time.

level420:
@RemiHeugue We have to discuss this change, maybe in the conversation of a new PR

RemiHeugue:
The checkbox was already toggling on the space event, I just removed the toggling on the enter key.
I don't know if we should also remove this behaviour from the ToggleButton, there is no native toggle buttons but native buttons fire the click event when pressing enter or space (at least on chrome and firefox).
Native checkboxes only react when pressing space.

level420:
@RemiHeugue what about native radiobuttons? Do they react on ENTER key?
Please note that ToggleButton is the abstract base class for the form checkbox and the toolbar checkbox.
It should be consistent through all buttons with toggle state that's checkbox and radiobutton.

RemiHeugue:
@level420 I did some tests with native radiobuttons:
- Enter does nothing
- You can only focus the selected button of a group with the tab key (or the first one if none is checked)
- Arrow keys change the selected radio button in the group
- Click event is fired on the button that become selected
- Space fire the click event on firefox and does nothing on chrome

level420:
Still, even for the checkbox, I think the right place to solve this is ToggleButton, not the checkbox class itself. And additionally the PR should also cover radiobuttons to have a consistent behaviour.

PR #9147, opened 2016-08-30
Integrate eslint-config-qx with Qooxdoo and lint.

jochumdev:
This is my try to integrate [eslint-config-qx](http://npmjs.com/package/eslint-config-qx) and lint the framework.

First I've run:

``` bash
find . -type f -iname '*.js' | xargs perl -0777 -i.original -pe 's/qx.Class.define\((.*),\s*\n\{/qx.Class.define($1, {/mg' {}
```

to make all qx.Class.define open curly braces on the same line.

Then I've linted: qx.application.*.

I'll lint the whole code if everyone is ok with that style.


pcdummy:
The curly brace change is needed as eslint and pretty sure all other linters don't liked mixed brace-styles.

See: http://eslint.org/docs/rules/brace-style for more on available styles.

zaucker:
I wonder if "everybody" will be able to agree on the same brace-style. My personal favorite is Stroustrup ... and 4 spaces indent (and no tabs), please ... ;-)

pcdummy:
Hehehe

oetiker:
If we reformat the source, cherry picking code back into the stable branch will be much harder as no patches will apply automatically anymore ... 

I would suggest, instead of actually reformatting the code, only preparing the reformat and doing it right before the 6.0 release.

woprandi:
https://google.github.io/styleguide/javascriptguide.xml

level420:
... and please don't change the indention!

johnspackman:
There is a difference between coding style guide compliance and code-safety, and while there is (somewhere) a Qooxdoo coding style guide it has not always been followed; I suspect that trying to enforce it now will lead to a massive patch.  IMHO even if we do this right before the 6.0 release, it will still make comparisons with previous versions difficult for very little benefit and I do not think that it is worth it.

Code safety is another matter of course, especially one which can be integrated into the toolchain - isn't there a linter which is configurable about which things to complain about?

pcdummy:
eslint is it gives you safety-errors and code-style errors, you can configure it to only give safety-errors.

pcdummy:
Please vote for that:

Make a config for eslint-config-qx which is just for "safety-errors" and integrate that into Qx.

If we get enough votes i'll make that happen.

pcdummy:
Will close this PR in favor of an upcoming "safety-errors" PR.

thron7:
Just to continue with my Cassandra work: There is a '**generate.py lint**' already.

And if you care to look at the [available options](http://www.qooxdoo.org/5.1/pages/tool/generator/generator_config_ref.html#lint-check) you will find that they support specifically features that are specific to qooxdoo, and are not easily found in shrink-wrapped linters: Checking for deprecated symbols; checking if arguments to qx.core.Environment.get() and the like are not literals (and therefore the environment dependency cannot be inferred); if you use reference types to initialize class variables (as those are shared between instances); if you use JSDoc keys that are unknown to the Generator. To name just a few.

pcdummy:
On the other hand eslint finds errors not detected by **generate.py lint** (or didn't we use it on the framework?). Like double decleration of variables, no decleration of vars and so on.

It also finds to complex code and usage of private members from other classes, also unused variables '==' comparison to null and other stuff i forgot. :)

We need in eslint: 
* deprecated checks.
* reference type to initialize class variables.
* JSDoc keys
* anything @thron7 will mention :)

thron7:
I'm certain there are useful things eslint can do above generate.py lint. And, without looking at it, there is probably a good way to teach eslint new things that are useful for qooxdoo. I was the one to push for replacing Python with JS, years before it was actually started. What I find difficult, though, is not analyzing requirements carefully when going forward. And since the project has been transferred to the community there is a bit of a tendency to ignore existing things rather than to learn from them.

And no, the framework itself was hardly maintained using automatic tools. It was all carefully hand-crafted, a labor of love :-).

johnspackman:
There are a number of checks that should be added to qxcompiler that only it can sensibly do (same as there are things in the generator that only it could sensibly do) - some of these are already in place but not being reported, and they need to be incorporated pretty soon.  My experience with generate.py is that the linting is good but not as extensive as a dedicated linter (with the exception of the qooxdoo-specific validations) so it is a good design choice to split linting into qooxdoo-specific and eslint)

This isn't ignoring existing things, just a matter of pragmatism getting things going - @thron7 if you could detail any of those extra, non-standard features that would be really helpful

johnspackman:
PS I read the other day on the javascript news mailing list about a minifier produced by the babel team https://github.com/babel/babili that's getting ready for production - as minification is currently a separate step to the compiler I'll be thinking about integrating that, but I also found https://github.com/babel/babel-eslint - this uses babel as the parser so presumably can be integrated into the parse/transpile stream

pcdummy:
I wouldn't integrate "full" linting into QxCompiler as eslint is very slow, see: https://github.com/qooxdoo/qooxdoo/pull/9156

johnspackman:
Wow that's an astonishing time!  That's around 1000 lines per minute for the full lint, I wondered at first whether that was running the linter on the combined output of the build but I can see from the logs that's not the case.  

I am concerned that this will mean that eslint is not suitable in the general case, which is a shame considering it's flexibility in configuration but I can see that speed being an issue

johnspackman:
just commented on https://github.com/qooxdoo/qooxdoo/pull/9156 as this is closed

Issue #9149, opened 2016-08-31
current devel playground throws exception on startup


level420:
When loading http://www.qooxdoo.org/devel/playground/ I'm getting three assert errors in console (Chrome 52, Firefox 48, IE11 on Windows7x64):

```
playground.js:213 001717 Assertion error! : Called assertTrue with 'false'
 Stack trace: 
qx.core.AssertionError.constructor://www.qooxdoo.org/devel/playground/scrip…anonymous(),anonymous(),qx.bom.request.Script.prototype._emit(),anonymous()
```

The visible impact is that the pre defined example list on the left side is empty.

level420:
I've backed out @johnspackman s memory management PR https://github.com/qooxdoo/qooxdoo/pull/8986 as it "felt" to have something to do with hashcodes by using a fresh master clone like this:
```
git clone https://github.com/qooxdoo/qooxdoo.git
cd qooxdoo
git revert 88d974a1c8f0ea7236374ea7c0db9e142ae670f2 -m 1
```
and voila the playground works again as expected.

I think the relevant location where the problem is showing up is https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/html/Element.js#L572-L573

@johnspackman could you please have a look into this? Thank you.

johnspackman:
no problem, ill have a look later on this evening

johnspackman:
Fixed in https://github.com/qooxdoo/qooxdoo/pull/9150

level420:
Thank you @johnspackman

PR #9150, opened 2016-08-31
Fix broken playground after merging automatic memory management

johnspackman:
fixes a bug in playground where it reuses a DomElement for the Inline playground application which causes an exception because that's not allowed; the changes for automatic memory management added assertion checks which caught the problem.


oetiker:

derrell:
If I understand correctly, this is a bug that has existed for a decade without being detected? If so, maybe it is useless code??? I'll vote for it, but it seems very odd that it wasn't causing any ill effects until an assert was added...

level420:
5 Votes. Merging.
Thank you @johnspackman

level420:
Fixes #9149

PR #9152, opened 2016-09-01
Wrong syntax in qx.test.ui.basic.Atom test method declaration

level420:
There is accidentaly a string literal "test" in the method declaration.

### DIFF
--- a/framework/source/class/qx/test/ui/basic/Atom.js
+++ b/framework/source/class/qx/test/ui/basic/Atom.js
@@ -47,7 +47,7 @@ qx.Class.define("qx.test.ui.basic.Atom",
       a.destroy();
     },
 
-    testSelectableSet : function('test') {
+    testSelectableSet : function() {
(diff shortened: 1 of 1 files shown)


level420:
@zaucker could you please verify it this was a typo or something by intention? Thank you!

zaucker:
@level420 This was unintentional. However, the string 'test' should be an argument of the call to qx.ui.basic.Atom('test'). I wonder how to fix this. I don't seem to be able to edit the file directly. Should I create a new pull request with the fix or can you fix your fix?

level420:
@zaucker done (I think)

zaucker:
Thanks. Let's see what travis says ... strange that the old version passed before.

zaucker:
Travis timed out. Is there something to set the timeout higher? I restarted the one run that didn't finish.

level420:

zaucker:
Merged as we have more than 5 thumbsUp.

Issue #9153, opened 2016-09-01
testrunner in source version ommits silently tests for testfiles containing javascript syntax errors


level420:
I found the bug in the qx.ui.basic.Atom test while fighting with testrunner source version and having console open. The syntax error occurred in the Atom.js testfile while the testrunner loaded each single testfile. Surprisingly the result was that the Atom.js file was not loaded at all and no tests where shown in the tree for qx.ui.basic.Atom (The Atom node was missing in the tests tree). I don't know but suspect that in this case the testrunner simply omitted all atom tests.

After the bug was in, there where no more tests executed for qx.ui.basic.Atom.

Not the entire test suite failed, but the all the tests in the file the syntax error was in. So syntax error in Atom.js -> no tests are executed from Atom.js

level420:
Surprisingly the tests ARE there in the build version and do not fail.

Here is the source code of the syntax error relevant test in qx.test.ui.basic.Atom
```javascript
    testSelectableSet : function('test') {
      var a = new qx.ui.basic.Atom();
      a.setSelectable(true);
      this.getRoot().add(a);
      this.flush();
      var l = a.getChildControl('label');	
      this.assertEquals("on", l.getContentElement().getDomElement().getAttribute("qxselectable"));
      a.destroy();
    },
```

level420:
The generator creates from the faulty function definition the following code:

```
testSelectableSet:function(b)
```
which turns the faulty into a correct syntax

level420:
it turns out that travis uses the karma testrunner which seems to also have this flaw.
Maybe @cajus and @peuter could have a look into this? Thank you.

pcdummy:
I don't know how, [here](https://github.com/qooxdoo/qooxdoo/blob/master/.travis/test-framework.sh#L13) we lint the Framework but not the tests, can't we lint tests files too there?

cajus:
I guess this can easily be fixed by changing the order of how karma/lint runs. If we run lint first and let karma only proceed, it would be OK.

pcdummy:
Lint runs first, but does it lint tests?

level420:
@cajus but then travis should have failed in the lint run, but it didn't

cajus:
What do I have to do to get the error locally?

level420:
git clone current master, then generate the source version of the testrunner and watch the console while testrunner is loading. You'll see a syntax error in Atom.js.
[...]

PR #9154, opened 2016-09-01
Fix lint run in framework

cajus:
This commit makes the linter constant aware in function arguments
and enables 'lint-test' to run properly.

refs #9153


level420:
Contained in https://github.com/qooxdoo/qooxdoo/pull/9158 closing.
Thanky you @cajus

PR #9156, opened 2016-09-01
Add eslint in safety mode

jochumdev:
This PR adds [eslint](http://eslint.org/) in [safety mode](https://github.com/drawstack/eslint-config-qx/blob/master/safety.js) to Qooxdoo.

We have [discussed](https://github.com/qooxdoo/qooxdoo/pull/9147#issuecomment-243711908) that.

One problem with eslint and qooxdoo is that Qooxdoo is big and eslint a bit slow with large files, so I added an "eslint-light" which still takes around 15 minutes on my machine to run through.

To run eslint-light with debug infos (debug gives you a status - which file/how long):

``` bash
npm run eslint-light -- --debug .
```

To run eslint with debug infos:

```
npm run eslint -- --debug .
```

I'm using the github Atom editor, theres a nice [linter-eslint](https://atom.io/packages/linter-eslint) Plugin for it.

Here is the complete output of a "eslint-light" run:
[qooxdoo-lint-errors.txt](https://github.com/qooxdoo/qooxdoo/files/450473/qooxdoo-lint-errors.txt)

And statistics to the rule hits: [RAW CSV](https://github.com/qooxdoo/qooxdoo/files/450480/qooxdoo-lint-errors.csv.txt)

**Rule**    **Count**
[comma-dangle](http://eslint.org/docs/rules/comma-dangle)   1
[complexity](http://eslint.org/docs/rules/complexity)   6
[computed-property-spacing](http://eslint.org/docs/rules/computed-property-spacing) 2
[default-case](http://eslint.org/docs/rules/default-case)   34
[eol-last](http://eslint.org/docs/rules/eol-last)   87
[func-call-spacing](http://eslint.org/docs/rules/func-call-spacing) 1
[guard-for-in](http://eslint.org/docs/rules/guard-for-in)   60
[max-depth](http://eslint.org/docs/rules/max-depth) 6
[max-params](http://eslint.org/docs/rules/max-params)   2
[max-statements-per-line](http://eslint.org/docs/rules/max-statements-per-line) 8
[new-cap](http://eslint.org/docs/rules/new-cap) 9
[new-parens](http://eslint.org/docs/rules/new-parens)   30
[no-alert](http://eslint.org/docs/rules/no-alert)   2
[no-caller](http://eslint.org/docs/rules/no-caller) 9
[no-cond-assign](http://eslint.org/docs/rules/no-cond-assign)   1
[no-constant-condition](http://eslint.org/docs/rules/no-constant-condition) 4
[no-debugger](http://eslint.org/docs/rules/no-debugger) 1
[no-else-return](http://eslint.org/docs/rules/no-else-return)   87
[no-empty](http://eslint.org/docs/rules/no-empty)   1
[no-eq-null](http://eslint.org/docs/rules/no-eq-null)   278
[no-eval](http://eslint.org/docs/rules/no-eval) 5
[no-extend-native](http://eslint.org/docs/rules/no-extend-native)   5
[no-fallthrough](http://eslint.org/docs/rules/no-fallthrough)   1
[no-global-assign](http://eslint.org/docs/rules/no-global-assign)   1
[no-implicit-coercion](http://eslint.org/docs/rules/no-implicit-coercion)   69
[no-lonely-if](http://eslint.org/docs/rules/no-lonely-if)   23
[no-mixed-operators](http://eslint.org/docs/rules/no-mixed-operators)   132
[no-negated-condition](http://eslint.org/docs/rules/no-negated-condition)   72
[no-nested-ternary](http://eslint.org/docs/rules/no-nested-ternary) 10
[no-new](http://eslint.org/docs/rules/no-new)   7
[no-prototype-builtins](http://eslint.org/docs/rules/no-prototype-builtins) 8
[no-redeclare](http://eslint.org/docs/rules/no-redeclare)   167
[no-script-url](http://eslint.org/docs/rules/no-script-url) 1
[no-self-assign](http://eslint.org/docs/rules/no-self-assign)   1
[no-sequences](http://eslint.org/docs/rules/no-sequences)   1
[no-undef](http://eslint.org/docs/rules/no-undef)   31
[no-unneeded-ternary](http://eslint.org/docs/rules/no-unneeded-ternary) 12
[no-unreachable](http://eslint.org/docs/rules/no-unreachable)   1
[no-unused-vars](http://eslint.org/docs/rules/no-unused-vars)   503
[no-useless-call](http://eslint.org/docs/rules/no-useless-call) 4
[no-warning-comments](http://eslint.org/docs/rules/no-warning-comments) 3
[one-var](http://eslint.org/docs/rules/one-var) 74
[operator-assignment](http://eslint.org/docs/rules/operator-assignment) 7
[operator-linebreak](http://eslint.org/docs/rules/operator-linebreak)   21
qx-rules/no-illegal-private-usage   43
qx-rules/no-refs-in-members 28
[radix](http://eslint.org/docs/rules/radix) 13
[semi](http://eslint.org/docs/rules/semi)   3
[semi-spacing](http://eslint.org/docs/rules/semi-spacing)   9
[space-unary-ops](http://eslint.org/docs/rules/space-unary-ops) 43
[valid-typeof](http://eslint.org/docs/rules/valid-typeof)   1


pcdummy:
We can disable any of these rules, please let me know which you don't like.

oetiker:
I don't like the bigfile exclusion ... just because a file is large that does not mean it will benefit less from linting :) how about altering the travis run of  eslint to only lint the files that have been changed by the commit triggering the run ... 

I think `git diff --name-only HEAD^` should spit out the files that have been changed in the last commit.

pcdummy:
A full run of eslint with bigfiles takes 5,5 hours on my older test machine (Core 2 Quad 2,7GHz, SSD).

Your idea sounds good to me.

johnspackman:
Wow that's an astonishing time! That's around 1000 lines per minute for the full lint, I wondered at first whether that was running the linter on the combined output of the build but I can see from the logs that's not the case.

I am concerned that this will mean that eslint is not suitable in the general case, which is a shame considering it's flexibility in configuration but I can see that speed being an issue

pcdummy:
I'll give JSHint a try, don't think it will be much faster with big files. The `git diff` Tobi proposed will work with eslint as nobody will push 100 Files (except linters :)).

johnspackman:
What I've read today suggests that jshint is 2-3 times faster (or was at the time the articles were written) but it also looks like eslint is developing fast; JSCS has merged with eslint, and eslint's github repo certainly has much more commit activity than jshint.

I'm just a bit disappointed I guess because I'd hoped to be able to include lint output as part of the constant compile of qxcompiler; one of the goals is IDE integration, for which super-fast feedback is important.

It would be interesting to see what the difference in speed is, but despite my comments this morning it sounds like eslint is the better product

pcdummy:
WOW JSHint is very fast, 33 seconds for a cached run - i don't know where it stores the cache.
The problem i see it gives only 3.424 Errors while eslint gave me 8.353 errors.

I don't see IDE integration with eslint as point against it as eslint is very fast with small files.

thron7:
On a first glance through qooxdoo-eslint-out.txt two things stand out:
eslint skips around 500 of the 1,500 files, not sure why.
The longest lists of findings are in third-party code, like Sizzle and PdfDetect.

thron7:
I'm heading over from #9147 as the discussion seems to continue here.

John asked about a list of non-standard  checks. I'm just going through the [checks the current Generator supports](http://www.qooxdoo.org/5.1/pages/tool/generator/generator_config_ref.html#lint-check) and comment on each of them. (The idea of most options is that you want them on, and you have to actively ignore them.):

-  **allowed-globals** :
    A global list of symbols that are ok, and should not be complained about as unresolved.
-  **ignore-catch-param** :
    The Ecmascript spec says that the variable `e` in a `try {...} catch(e) {...}` creates a new scope that is only accessible within the catch block, but in some implementations `e` would leak into the enclosing scope. This can lead to surprising effects when you have a variable `e` in the enclosing scope anyway. -- Not sure if this is still an issue with modern browsers or Node.
-  **ignore-deprecated-symbols** : 
    Don't complain about `eval`, `alert` and the likes. -- `no-eval` and `no-alert` in eslint.
-  **ignore-environment-nonlit-key** : 
    This is very qooxdoo-specific: Calls to static methods of `qx.core.Environment` trigger the corresponding environment check be included in the dependencies of the using class. E.g. `qx.core.Environment.get("css.pointerevents")` causes qx.bom.client.Css to be added to the dependencies as the corresponding check is implemented there. This only works if the argument, the "key", to the .get call is a literal string. As the Generator generally doesn't do resolution of variable values, `var a="css.pointerevents"; qx.core.Environment.get(a);` wouldn't work. -- An important check.
-  **ignore-multiple-mapkeys** :
    Warn about the same map key appearing multiple times in a map. A popular bug in big class maps, defining e.g. the same method twice and then being surprised why the first definition is never applied :-). -- General JS issue, not qooxdoo-specific.
-  **ignore-multiple-vardecls** :
    Declaring the same variable multiple times in the same scope with `var`. -- Not qooxdoo-specific.
-  **ignore-no-loop-block** :
    Allow if loop blocks are not enclosed by `{...}`, which is only sensible if the block consists of single statement. A popular bug is to later add a second statement at the right level of indentation and being surprised that it is executed independent of the loop construct. -- Not qooxdoo-specific.
-  **ignore-reference-fields** :
     If you use a reference data type (map or array or class instance) to initialize a member in a class map, this value is shared across all instances of the class. -- qooxdoo-specific and important.
-  **ignore-undeclared-privates** :
    Using `this.__foo` in a class method without declaring `__foo` as a member in the class map. -- Prob. qooxdoo-specific but I can't remember what problems it might create.
-  **ignore-undefined-globals** :
     Ignore symbols that belong to the global scope, and are not recognized as known built-in symbols or class names (You usually want to avoid those). -- qooxdoo-specific; only qooxdoo tries to maintain a comprehensive list of all global symbols of an application to check for unresolved ones.
-  **ignore-shadowing-locals** :
    A similar topic here: With a `var qx = {};` in a function you cut yourself off from the global qx, so no framework class can be used while you are in that scope. -- qooxdoo-specific, because qooxdoo knows all globals.
-  **ignore-unused-parameter** :
    Ignore parameters of functions or catch statements that are not used in their respective body. -- Unspecific.
-  **ignore-unused-variables**  :
    The same for scoped variables: Ignore if variables are declared in a scope but not used. -- Unspecific.
-  **warn-unknown-jsdoc-keys** :
    The Generator supports a specific set of [JSDoc-style comment annotations](http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_ref.html#section-reference). An unknown `@<key>` annotation might be deliberate, but can also be a typo which is hard to spot. -- qooxdoo-specific. Mind that `@ignore` has its [own manual page](http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_at_ignore.html). There's more to be said about JSDoc hints further down.
-  **warn-jsdoc-key-syntax** :
    qooxdoo-specific JSDoc keys have certain [syntax rules](http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_ref.html) to them. You usually want a warning if those are violated, as e.g. the Apiviewer generation will not work as expected. -- qooxdoo-specific.

Aside from these general config settings the linter obeys two JSDoc hints, `@lint` and `@ignore`, which allow you to override global options locally for a given scope. The [@lint](http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_ref.html#lint) hint supports a couple of sub-keys that correspond to lint's global options above: **environmentNonLiteralKey, ignoreDeprecated, ignoreNoLoopBlock, ignoreReferenceField, ignoreUnused, ignoreJsdocKey**.

The [@ignore](http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_ref.html#ignore) JSDoc hint concerns unresolved global symbols, corresponds to the `allowed-globals` above, and is shared between the linter and the compiler. E.g. when you want the linter to ignore a symbol, you usually also don't want the compiler to try to resolve it and follow its dependencies. It supports name globs like `qx.bom.*` to ignore entire name spaces, and the special symbols **auto-require** and **auto-use**.

So in total, the current qooxdoo linter is special in supporting dependency handling and unknown globals, part. taylored through JSDoc hints. I think this you would want to also support in the future.

As all these functions are based on **syntax trees** of the code you could easily share self-written checks with standard checks (eslint, ...) as long as both rely on the same tree structure (estree, babel, Mozilla AST, ...). The Generator has certain AST optimizations like "shadow ast's" to make specific checks part. fast.

Sorry for another long-ish post, but I hope this will make it easier to draw the boundary between qooxdoo-specific and standard linting.

thron7:
Re: **minifiers**  -- I would endorse using a third-party minifier like yuicompressor or babili or something, whatever meets expectations when you take it for test drive. While it's not terribly difficult to implement the most common comression techniques, it's nice to leave that to people that do nothing else, and get you some extra milage on top. And as @johnspackman said, it's a function that is easily outsourced.

The best performance-wise is probably still Google's Closure compiler for which also an NPM module exists. But that would draw in Java as an dependency of the tool chain ...

pcdummy:
@thron7 Thanks for your knowledge and input about the generators linter!

I try to address all "special" cases you wrote here:

- **allowed-globals**: can be done like [here](https://github.com/pcdummy/qooxdoo-tweets-tutorial/blob/master/.eslintrc-qx#L3) in eslint.

- **ignore-catch-param**: haven't found docs for that [Except that](http://stackoverflow.com/questions/6100230/javascript-catch-parameter-already-defined) think we can ignore it, it seams to be an IE8 error.

- **ignore-deprecated-symbols**: `no-eval` and `no-alert` help here, both are enabled by default and can be disabled in a custom config.

- **ignore-environment-nonlit-key**: Need to read it again and again :)

- **ignore-multiple-mapkeys**: is implemented in eslint and works already.

- **ignore-multiple-vardecls**: Works fine with eslint.

- **ignore-no-loop-block**: by default enabled in eslint.

- **ignore-reference-fields**: I've stolen a check for that from the next repo.

- **ignore-undeclared-privates**: same as above, have stolen it from the next repo.

- **ignore-undefined-globals**: We don't have a check for that yet (if theres none in QxCompiler) @johnspackman can you implement that?

- **ignore-shadowing-locals**: ~~We don't have a rule for that~~, theres a eslint [no-shadow](http://eslint.org/docs/rules/no-shadow) rule which i just enabled.

- **ignore-unused-parameter**: By default we complain about these.

- **ignore-unused-variables**: Same as above, we complain.

- **JSDoc**: Nothing for that @johnspackman  ?

- **@ignore**: See http://eslint.org/docs/user-guide/configuring#specifying-globals on howto do that with eslint.

Thanks again @thron7

thron7:
I've looked at the patch of this pull request. The list of big files is hard-coded and would need manual update from time to time. I also don't understand `qx/safety` in the eslintConfig, but that's probably just my lack of knowledge.

In general I'm not sure it is of high priority to lint the framework code. It might be interesting to run lint from time to time on it, but I'm even less sure if all findings should be resolved. Some things are there for a purpose, and there is value in hand-crafting framework code.

Much more important than linting framework code is IMO to provide the users of qooxdoo (meaning: coders that develop apps with qooxdoo) with a consistent and powerful lint utility. Achieving that would advance the feature set of the JS-based tool chain, and would bring the project closer to switching away from Python.

pcdummy:
[eslint-config-qx](https://www.npmjs.com/package/eslint-config-qx) is available to everyone who want to use eslint with qx not only for the framework.

The `qx/safety` include is a special case of `eslint` with hopefully all formatting rules disabled.

If you want to see how an application with qxcompiler, grunt-qx and eslint will look have a look at my [qooxdoo-tweets sample](https://github.com/pcdummy/qooxdoo-tweets-tutorial/).

Tonight i'll remove the dependency of downloading qooxdoo as i made a scoped npm package for that.

pcdummy:
Think we can close this, its now included in the "qx" tool.

PR #9158, opened 2016-09-03
Fix lint run in framework and lint source/class/qx/test/*

jochumdev:
This is a follow up of #9154.

I couldn't fix these warnings, please some else fix that:
- Warning: qx.test.ui.basic.Atom (50, 25): Constant as parameter name in function declaration
- Warning: qx.test.Annotation (40, 25): Reference values are shared across all instances: '@methodA'
- Warning: qx.test.Annotation (66, 23): Reference values are shared across all instances: '@methodB'
- Warning: qx.test.Annotation (87, 23): Reference values are shared across all instances: '@methodB'
- Warning: qx.test.Annotation (163, 23): Reference values are shared across all instances: '@methodA'
- Warning: qx.test.bom.Attribute (163, 55): qx.core.Environment.select: second parameter is not a map.
- Warning: qx.test.Mixin (243, 15): Using an undeclared private class feature: '__b'
- Warning: qx.test.Mixin (244, 27): Using an undeclared private class feature: '__b'
- Warning: qx.test.Class (398, 11): Using an undeclared private class feature: '__p'
- Warning: qx.test.Class (406, 20): Using an undeclared private class feature: '__p'
- Warning: qx.test.Class (465, 11): Using an undeclared private class feature: '__p'
- Warning: qx.test.Class (473, 20): Using an undeclared private class feature: '__p'

### DIFF
--- a/framework/source/class/qx/test/Interface.js
+++ b/framework/source/class/qx/test/Interface.js
@@ -71,6 +71,8 @@ qx.Class.define("qx.test.Interface",
         properties : { color : { } }
       });
 
+      new qx.test.i.Audi("audi");
+
       this.assertTrue(qx.Interface.classImplements(qx.test.i.Audi, qx.test.i.ICar));
       qx.Class.undefine("qx.test.i.Audi");
     },
@@ -98,6 +100,7 @@ qx.Class.define("qx.test.Interface",
 
           properties : { color : { } }
         });
+
       this.assertTrue(qx.Interface.classImplements(qx.test.i.Bmw1, qx.test.i.ICar));
       qx.Class.undefine("qx.test.i.Bmw1");
     },
@@ -171,7 +174,7 @@ qx.Class.define("qx.test.Interface",
       if (this.isDebugOn())
       {
         this.assertException(function() {
-          qx.test.i.ICar();
+          new qx.test.i.ICar();
         }, Error);
 
 
--- a/framework/source/class/qx/test/Interface.js
+++ b/framework/source/class/qx/test/Interface.js
@@ -71,8 +71,6 @@ qx.Class.define("qx.test.Interface",
         properties : { color : { } }
       });
 
-      var audi = new qx.test.i.Audi("audi");
-
       this.assertTrue(qx.Interface.classImplements(qx.test.i.Audi, qx.test.i.ICar));
       qx.Class.undefine("qx.test.i.Audi");
     },
@@ -173,7 +171,7 @@ qx.Class.define("qx.test.Interface",
       if (this.isDebugOn())
       {
         this.assertException(function() {
-          var i = new qx.test.i.ICar();
+          qx.test.i.ICar();
         }, Error);
 
 
@@ -365,7 +363,9 @@ qx.Class.define("qx.test.Interface",
             members :
             {
               getValue : function() {},
-              setValue : function(value) {}
+              setValue : function(value) {
+                value = value;
+              }
             }
           });
         });
@@ -377,7 +377,9 @@ qx.Class.define("qx.test.Interface",
         members :
         {
           getValue : function() {},
-          setValue : function(value) {}
+          setValue : function(value) {
+            value = value;
+          }
         }
       });
 
@@ -400,7 +402,9 @@ qx.Class.define("qx.test.Interface",
         members :
         {
           getValue : function() {},
-          setValue : function(value) {}
+          setValue : function(value) {
+            value = value;
+          }
         }
       });
     },
--- a/framework/source/class/qx/test/Mixin.js
+++ b/framework/source/class/qx/test/Mixin.js
@@ -234,15 +234,16 @@ qx.Class.define("qx.test.Mixin",
       {
         members :
         {
+          _b: null,
+
           sayJuhu : function() { return this.base(arguments) + " Kinners";},
 
-          /** @lint ignoreUndeclaredPrivates(__b) */
           foo : function(dontRecurs)
           {
             var s = "";
             if (!dontRecurs) {
-              this.__b = new qx.Patch2();
-              s += "++" + this.__b.foo(true) + "__";
+              this._b = new qx.Patch2();
+              s += "++" + this._b.foo(true) + "__";
             }
 
             s += this.base(arguments);
@@ -275,7 +276,7 @@ qx.Class.define("qx.test.Mixin",
       // the mixin member
       var o = new qx.Patch1();
       this.assertEquals("++bar__foo", o.foo());
-      o.__b.dispose();
+      o._b.dispose();
       o.dispose();
     }
   }
--- a/framework/source/class/qx/test/bom/webfonts/Manager.js
+++ b/framework/source/class/qx/test/bom/webfonts/Manager.js
@@ -28,30 +28,10 @@ qx.Class.define("qx.test.bom.webfonts.Manager", {
 
   members :
   {
-    __fontDefinitions :
-    {
-      finelinerScript :
-      {
-        family : "FinelinerScriptRegular",
-        source: [ qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.woff"),
-                  qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.ttf"),
-                  qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.eot") ]
-      },
-      fontawesome :
-      {
-        family : "FontAwesome",
-        source: [ qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.woff"),
-                  qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.ttf"),
-                  qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.eot") ]
-      },
-      invalid :
-      {
-        family : "MonkeypooBold",
-        source: [ "404.woff",
-                  "404.ttf",
-                  "404.eot" ]
-      }
-    },
+    __fontDefinitions : null,
+    __nodesBefore : null,
+    __sheetsBefore : null,
+    __manager : null,
 
     __findRule : function(familyName)
     {
@@ -89,6 +69,30 @@ qx.Class.define("qx.test.bom.webfonts.Manager", {
       this.__nodesBefore = document.body.childNodes.length;
       this.__sheetsBefore = document.styleSheets.length;
       this.__manager = qx.bom.webfonts.Manager.getInstance();
+
+      this.__fontDefinitions = {
+        finelinerScript :
+        {
+          family : "FinelinerScriptRegular",
+          source: [ qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.woff"),
+                    qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.ttf"),
+                    qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fineliner_script-webfont.eot") ]
+        },
+        fontawesome :
+        {
+          family : "FontAwesome",
+          source: [ qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.woff"),
+                    qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.ttf"),
+                    qx.util.ResourceManager.getInstance().toUri("qx/test/webfonts/fontawesome-webfont.eot") ]
+        },
+        invalid :
+        {
+          family : "MonkeypooBold",
+          source: [ "404.woff",
+                    "404.ttf",
+                    "404.eot" ]
+        }
+      };
     },
 
     tearDown : function()
--- a/framework/source/class/qx/test/core/Validation.js
+++ b/framework/source/class/qx/test/core/Validation.js
@@ -99,14 +99,21 @@ qx.Class.define("qx.test.core.Validation",
 
   members :
   {
+    __model: null,
 
     testNumber : function()
     {
       var model = this.__model;
       // test for some false inputs
-      this.assertException(function() {model.setNumber("test");}, qx.core.ValidationError, null, "A String is no number.");
-      this.assertException(function() {model.setNumber(new Date());}, qx.core.ValidationError, null, "A Date is no number.");
-      this.assertException(function() {model.setNumber(this);}, qx.core.ValidationError, null, "this is no number.");
+      this.assertException(
+        function() {model.setNumber("test");},
+        qx.core.ValidationError, null, "A String is no number.");
+      this.assertException(
+        function() {model.setNumber(new Date());},
+        qx.core.ValidationError, null, "A Date is no number.");
+      this.assertException(
+        function() {model.setNumber(this);},
+        qx.core.ValidationError, null, "this is no number.");
 
       // test an positive integer
       model.setNumber(12);
@@ -125,10 +132,18 @@ qx.Class.define("qx.test.core.Validation",
       var model = this.__model;
 
       // test some wrong addresses
-      this.assertException(function() {model.setEmail("test");}, qx.core.ValidationError, null, "test is a mailadress?");
-      this.assertException(function() {model.setEmail("@affe.de");}, qx.core.ValidationError, null, "@affe.de is never a mailadress!");
-      this.assertException(function() {model.setEmail("hans@@wurst.de");}, qx.core.ValidationError, null, "Are two @ allowed?");
-      this.assertException(function() {model.setEmail("m@a.d");}, qx.core.ValidationError, null, "m@a.d?");
+      this.assertException(
+        function() {model.setEmail("test");},
+        qx.core.ValidationError, null, "test is a mailadress?");
+      this.assertException(
+        function() {model.setEmail("@affe.de");},
+        qx.core.ValidationError, null, "@affe.de is never a mailadress!");
+      this.assertException(
+        function() {model.setEmail("hans@@wurst.de");},
+        qx.core.ValidationError, null, "Are two @ allowed?");
+      this.assertException(
+        function() {model.setEmail("m@a.d");},
+        qx.core.ValidationError, null, "m@a.d?");
 
       // test some working addresses
       model.setEmail("affe@zoo.de");
@@ -145,10 +160,18 @@ qx.Class.define("qx.test.core.Validation",
       var model = this.__model;
 
       // test some wrong inputs
-      this.assertException(function() {model.setString(1);}, qx.core.ValidationError, null, "A number is not a string!");
-      this.assertException(function() {model.setString(this);}, qx.core.ValidationError, null, "This test is not a string!");
-      this.assertException(function() {model.setString(true);}, qx.core.ValidationError, null, "A boolean is not a string!");
-      this.assertException(function() {model.setString(new Date());}, qx.core.ValidationError, null, "A Date-Object is not a string!");
+      this.assertException(
+        function() {model.setString(1);},
+        qx.core.ValidationError, null, "A number is not a string!");
+      this.assertException(
+        function() {model.setString(this);},
+        qx.core.ValidationError, null, "This test is not a string!");
+      this.assertException(
+        function() {model.setString(true);},
+        qx.core.ValidationError, null, "A boolean is not a string!");
+      this.assertException(
+        function() {model.setString(new Date());},
+        qx.core.ValidationError, null, "A Date-Object is not a string!");
 
       // Test some working inputs
       model.setString("affe@zoo.de");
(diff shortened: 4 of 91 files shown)


pcdummy:
Think that Travis error isn't my fault, right?

level420:
Restarting travis run.
Good night.

level420:
@pcdummy the ```Using an undeclared private class feature:``` could be fixed by declaring the member variable within in the ```members``` part. For the first point in your list in qx.test.Mixin this would be adding

```javascript
      qx.Mixin.define("qx.MPatch",
      {
        members :
        {
          // ADD THIS HERE:
          __b : null,

          sayJuhu : function() { return this.base(arguments) + " Kinners";},

          foo : function(dontRecurs)
          {
            var s = "";
            if (!dontRecurs) {
              this.__b = new qx.Patch2();
              s += "++" + this.__b.foo(true) + "__";
            }

            s += this.base(arguments);
            return s;
          }
        }
      });
```

pcdummy:
Same as with the others, easy. Thanks.

level420:
@pcdummy : the very first lint warning in your list is exactly the situation which caused all this. The test was fixed with https://github.com/qooxdoo/qooxdoo/pull/9152

pcdummy:
Will try to rebase this evening.

level420:
@pcdummy : I think ```Warning: qx.test.bom.Attribute (163, 55): qx.core.Environment.select: second parameter is not a map.``` is, because the generator does not know in this situation that ```setUp``` where __maxLengthValues is initialized, is executed before ```testRemoveAttribute```.
So it may help to assign ```this.__maxLengthValues``` to a local variable before using it as a parameter for ```qx.core.Environment.select```.
Or do something like
```javascript
qx.core.Environment.select("engine.name", (this.__maxLengthValues||{}));
```

level420:
@pcdummy forget it what I've wrote in the comment before does not work. Still investigating.

level420:
No I've understood: as there is explicitly declared int the doc header of qx.core.Environment.select that the second parameter has to be a map, we have to write:
```javascript
    testRemoveAttribute : function()
    {
      var Attribute = qx.bom.element.Attribute;

      Attribute.set(this._input, "maxLength", 10);
      Attribute.set(this._input, "maxLength", null);

      var maxLengthValue = qx.core.Environment.select("engine.name", {
        "mshtml"  : this.__maxLengthValues["mshtml"],
        "webkit"  : this.__maxLengthValues["webkit"],
        "default" : this.__maxLengthValues["default"]
      });
...
```
for that line

johnspackman:
Or just disable the warning for that function?  See http://www.qooxdoo.org/5.1/pages/development/api_jsdoc_ref.html#lint but you should be able to do something like:
````
/**
 * @ignore environmentNonLiteralKey()
 */
testRemoveAttribute : function() {
```

pcdummy:
It doesn't work sadly John :/

thron7:
I think `@ignore environmentNonLiteralKey()` really does only apply to the environment **key**, i.e. the first argument like *"engine.name"* in this case. I'm not sure there is an `@ignore` option for the second, map-type argument in *qx.core.Environment.select()*.

level420:
@pcdummy  and we could also close https://github.com/qooxdoo/qooxdoo/pull/9154 ?

pcdummy:
@level420 yes, this pr incorporates that one.

level420:
@pcdummy @oetiker had that idea. Please change the test to:
```javascript
      qx.Mixin.define("qx.MPatch",
      {
        members :
        {
          sayJuhu : function() { return this.base(arguments) + " Kinners";},

          foo : function(dontRecurs)
          {
            var s = "";
            if (!dontRecurs) {
              this['__b'] = new qx.Patch2();
              s += "++" + this['__b'].foo(true) + "__";
            }

            s += this.base(arguments);
            return s;
          }
        }
      });
```
Acessing this['__b'] instead of this.__b is the trick!

pcdummy:
Integrated that one, hopefully Travis likes us now. Thanks a lot @level420

level420:

derrell:
This is a huge change -- 91 files -- and many of the changes are whitespace changes which make it really difficult to see what's really relevant...

framework/source/class/qx/test/Interface.js
```diff
@@ -365,7 +363,7 @@ qx.Class.define("qx.test.Interface",
             members :
             {
               getValue : function() {},
-              setValue : function(value) {}
```
derrell: Really? A setValue function that doesn't take a value? If lint is complaining about value being unused, this is the wrong fix. If the function really needn't do anything, then I'd declare it as:
```
setvalue : function(value) { value = value; }
```
(Then we'll see if lint complains about a "useless statement"...)

framework/source/class/qx/test/Mixin.js
```diff
@@ -240,8 +240,8 @@ qx.Class.define("qx.test.Mixin",
           {
             var s = "";
             if (!dontRecurs) {
-              this.__b = new qx.Patch2();
-              s += "++" + this.__b.foo(true) + "__";
+              this['__b'] = new qx.Patch2();
```
derrell: This looks like it's trying to fix a "__b is not a member" problem because __b is not declared in the members section?

framework/source/class/qx/test/bom/webfonts/Manager.js
```diff
@@ -89,6 +69,30 @@ qx.Class.define("qx.test.bom.webfonts.Manager", {
       this.__nodesBefore = document.body.childNodes.length;
       this.__sheetsBefore = document.styleSheets.length;
       this.__manager = qx.bom.webfonts.Manager.getInstance();
+
+      this.__fontDefinitions = {
+        finelinerScript :
+        {
```
derrell: Is __findRule called only once? If __fontDefinitions needs to be different for different instances, then typically the initialization would be done in the constructor.

framework/source/class/qx/test/Mixin.js
level420: If you declare __b as member variable then the patch mechanism ```qx.Class.include(qx.Patch1, qx.MPatch)``` complains about __b is not allowed to be overridden.

framework/source/class/qx/test/core/Validation.js
```diff
@@ -145,10 +160,18 @@ qx.Class.define("qx.test.core.Validation",
       var model = this.__model;
 
       // test some wrong inputs
-      this.assertException(function() {model.setString(1);}, qx.core.ValidationError, null, "A number is not a string!");
-      this.assertException(function() {model.setString(this);}, qx.core.ValidationError, null, "This test is not a string!");
-      this.assertException(function() {model.setString(true);}, qx.core.ValidationError, null, "A boolean is not a string!");
-      this.assertException(function() {model.setString(new Date());}, qx.core.ValidationError, null, "A Date-Object is not a string!");
+      this.assertException(
```
derrell: Yes! Get rid of those excessively long lines!!! Thank you!

framework/source/class/qx/test/core/Validation.js
```diff
@@ -221,9 +259,12 @@ qx.Class.define("qx.test.core.Validation",
       var model = this.__model;
 
       // test some wrong inputs (allowed are male and female)
-      this.assertException(function() {model.setArray(0.999999999);}, qx.core.ValidationError, null, "A 0.999999999 is not in ['male', 'female']");
-      this.assertException(function() {model.setArray("malle");}, qx.core.ValidationError, null, "'malle' is not in ['male', 'female']");
-      this.assertException(function() {model.setArray("");}, qx.core.ValidationError, null, "A empty string is not in ['male', 'female']");
+      this.assertException(
+        function() {model.setArray(0.999999999);}, qx.core.ValidationError, null, "A 0.999999999 is not in ['male', 'female']");
```
derrell: Sigh... But here those excessively long lines remain. :-(

framework/source/class/qx/test/io/part/MockPackage.js
```diff
@@ -4,13 +4,16 @@ qx.Bootstrap.define("qx.test.io.part.MockPackage",
   {
     this.id = id;
     this.delay = delay || 0;
-    this.error = !!error;
+    this.error = Boolean(error);
```
derrell: Why this change?

pcdummy: eslint. :)

framework/source/class/qx/test/ui/form/AbstractVirtualBox.js
```diff
@@ -21,12 +21,15 @@ qx.Class.define("qx.test.ui.form.AbstractVirtualBox",
...
+    __selectBox : null,
+
     setUp : function()
     {
-      this.__selectBox = new qx.ui.form.VirtualSelectBox;
+      this.__selectBox = new qx.ui.form.VirtualSelectBox();
       this.getRoot().add(this.__selectBox);
 
-      this.__comboBox = new qx.ui.form.VirtualComboBox;
+      this.__comboBox = new qx.ui.form.VirtualComboBox();
```
derrell: Yes. Much more readable.

framework/source/class/qx/test/Interface.js
level420: @pcdummy does the generator linter really complain here? A few lines below there is still an empty function with parameter value : https://github.com/pcdummy/qooxdoo/blob/297d25359f857c4d4396fbbe5a69e7f60d4bbd36/framework/source/class/qx/test/Interface.js#L401

framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
```diff
@@ -286,7 +286,7 @@ qx.Class.define("qx.test.ui.tree.virtual.WidgetProvider",
       var newWidget2Bindungs = widget2.getBindings().length;
       var newModelBindungs = this.getLookupTable().getBindings().length;
       this.assertEquals(oldWidget1Bindungs, newWidget1Bindungs, "Binding on first widget is not removed!");
-      this.assertEquals(oldWidget1Bindungs, newWidget1Bindungs, "Binding on second widget is not removed!");
+      this.assertEquals(oldWidget2Bindungs, newWidget2Bindungs, "Binding on second widget is not removed!");
```
derrell: Shouldn't we, while we're fixing these lines, fix the spelling of Bindings in this function???

framework/source/class/qx/test/util/DateFormat.js
```diff
@@ -22,35 +22,37 @@ qx.Class.define("qx.test.util.DateFormat",
...
-    {'date' : new Date(-20,10,14), 'result' : {'era': {'abbrev': 'BC', 'fullName': 'Before Christ', 'narrow': 'B'}}},
-    {'date' : new Date(2012, 4, 24, 11, 49, 57, 1), 'result' : {}},
-    {'date' : new Date(2012, 4, 24, 11, 49, 57, 12), 'result' : {}},
-    {'date' : new Date(2012, 4, 24, 11, 49, 57, 123), 'result' : {}}
-
-  ],
+    __dates : null,
+
+    setUp : function() {
+      // result contain an object with what should be expected, the result to test the date against
```
derrell: Once again... Confirm this function is called only once. Typically such initialization is done in the constructor.

derrell:
Ok. I've gone through the whole thing and added a number of line comments. I don't think I'm quite ready to vote on this yet. My comments should probably be addressed first.

derrell:
BTW, this was a huge job. Thanks for doing it!

framework/source/class/qx/test/util/DateFormat.js
level420: @derrell but do we have a constructor in the test classes?

level420:
@derrell added some comments to your line notes ;-)

framework/source/class/qx/test/util/DateFormat.js
derrell: Maybe not, and maybe this is the right place for it. I just wanted confirmation that this function is called only once, and that the data given it at different times varies. If it's invariant, then the original code -- granted, a reference type initialized in the members section -- is a fine place to put it to avoid multiple assignments. If this setup function is called only once, this is also a fine place to put it, particularly if there's no constructor available in which to put it.

framework/source/class/qx/test/Mixin.js
derrell: That's a problem. __b is a private variable. Private variables are not accessible even to mixins, IIRC. The __b in MPatch is different than the __b in Patch1.

framework/source/class/qx/test/Interface.js
pcdummy: No ESLint complained (i used ESLint to fix those files and run the generator after).

framework/source/class/qx/test/util/DateFormat.js
pcdummy: I see setUp as constructor for tests, yes.

framework/source/class/qx/test/bom/webfonts/Manager.js
pcdummy: This is part of "setUp" not "__findRule", the diff isn't clear enough.

framework/source/class/qx/test/core/Validation.js
pcdummy: We should start another Lint with formatting after.

framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
pcdummy: New patch with that will come.

framework/source/class/qx/test/core/Validation.js
thron7: From experience I can only recommend to put formatting changes in "whitespace-only" commits. When you mix them with real changes it is very hard to review those commits.

level420:
@derrell please ad a dissenting vote to this PR to avoid someone is accidentally merging it.

derrell:
Down-voted to avoid PR being merged prematurely.

framework/source/class/qx/test/util/DateFormat.js
level420: @derrell I think this change is OK as (as stated in the docs) creating the array with test dates again for each test does not harm and each test has reliable the same initial values.

framework/source/class/qx/test/bom/webfonts/Manager.js
level420: Yes. Seems reasonable to have a fresh initialized testdata array for each test.

framework/source/class/qx/test/Mixin.js
level420: @derrell The fact is if we declare it as a member, then the exception https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/Mixin.js#L258 is thrown. But why? The class qx.Patch1 where qx.MPatch is included, does not have a private member __b itself. And, as discussed with @oetiker private members are valid in mixins.

What if in the case of the test, the intention was not to write clean code, but trick the testsuite somehow to be able to implement that test? In this case it may be valid to have a private variable not be declared member and hide it from the linter via a @lint hint.

We could merger @thron7 's PR https://github.com/qooxdoo/qooxdoo/pull/9164 first, leave the code as it was, asking @pcdummy t revert this changes and to rebase this PR after merging https://github.com/qooxdoo/qooxdoo/pull/9164, which also includes the fix by adding the lint hint.

framework/source/class/qx/test/Interface.js
level420: @pcdummy I think @derrell is right here. As the test is to check if a class fulfills the requirements of the interface it would be valid to have some code in the implementing method. Even such code like
```javascript
value += '';
```

level420:
@derrell please re-evaluate this PR. Thank you.

framework/source/class/qx/test/util/DateFormat.js
derrell: Agree. The change is fine.

derrell:
I think all but one of my questions have been answered. The one pending is setvalue(value). Excluding optional parameters is common practice, but mandatory parameters should always be listed even if unused. If that means we have to "use" them with some trivial, idempotent "change" to shut up the linter, then that's what we should do.

level420:
@pcdummy would you please be so kind and change the PR discussed in https://github.com/qooxdoo/qooxdoo/pull/9158#discussion_r77768439 and above? 
Thank you.

level420:
@pcdummy merget https://github.com/qooxdoo/qooxdoo/pull/9164

level420:
@pcdummy do you have some time to work again on this PR?

After some back and forth and discussion with @derrell it would be good if you can change the private variable __b to a protected one _b and add it to the members of qx.MPatch ( in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/Mixin.js#L233-L280 )
then I'll start a new voting round.

That would be great!

pcdummy:
@level420 all done if travis is will be green.

level420:
@pcdummy Thank you.

But I can't find the change ```qx.test.Mixin.testPatchOverwritten``` where the private member ```__b``` is changed to a protected one ```_b``` and is explicitly declared as a member ( https://github.com/pcdummy/qooxdoo/blob/fix-lint-constants/framework/source/class/qx/test/Mixin.js#L244 ).

There is also missing the change where in https://github.com/pcdummy/qooxdoo/blob/fix-lint-constants/framework/source/class/qx/test/Interface.js#L380 the ```setValue``` method should have a parameter again, maybe with some dummy code in the function body like ``` value = value; ``` to make eslint happy.

At https://github.com/pcdummy/qooxdoo/blob/fix-lint-constants/framework/source/class/qx/test/Interface.js#L403 there is also such an empty ```setValue(value)``` function declaration. Doesn't eslint complain here as well?

And please note: until we resolve issue https://github.com/qooxdoo/qooxdoo/issues/9167 the travis run will never complete successfully.

level420:
@pcdummy would be nice if you have some time to add the last changes so we can merge and close this PR. Thank you.

pcdummy:
This should be finaly ready to Vote and Merge.

framework/README.md
```diff
@@ -0,0 +1,13 @@
+# pcdummy's qooxdoo
```
cajus: So this file may need to be removed, I guess.

framework/package.json
```diff
@@ -1,11 +1,27 @@
 {
```
cajus: This one shouldn't get changes, too for this commit, or am I wrong? ;-)

pcdummy:
Made a badass mistake, rebased on the wrong master branch (mine instead of upstream). Will fix.

framework/source/class/qx/test/dev/unit/Sinon.js
```diff
@@ -235,9 +235,6 @@ qx.Class.define("qx.test.dev.unit.Sinon",
       var func = function() {};
       var obj = {"a": function() {}};
 
-      var spy = this.spy(func);
-      var stub = this.stub(obj, "a");
-      var xhr = this.useFakeXMLHttpRequest();
```
level420: What if this lines test Sinon itself? Of course the variable ```spy``` is not used anymore, but the call ```this.spy(func)``` itself could be the test. If ```this.spy(func)``` fails because of an exception, then the Sinon integration may be faulty.

framework/source/class/qx/test/Interface.js
```diff
@@ -71,8 +71,6 @@ qx.Class.define("qx.test.Interface",
         properties : { color : { } }
       });
 
-      var audi = new qx.test.i.Audi("audi");
```
level420: I don't know if we should delete this line. The instantiation itself could fail with an exception which then would maybe be a test itself.

framework/source/class/qx/test/Interface.js
```diff
@@ -173,7 +171,7 @@ qx.Class.define("qx.test.Interface",
       if (this.isDebugOn())
       {
         this.assertException(function() {
-          var i = new qx.test.i.ICar();
+          qx.test.i.ICar();
```
level420: Is this really equivalent? Does simply writing ```qx.text.i.ICar()``` create a new instance of that class? I'm asking because I really don't know that.

framework/source/class/qx/test/io/rest/Resource.js
```diff
@@ -228,7 +223,7 @@ qx.Class.define("qx.test.io.rest.Resource",
       this.require(["debug"]);
 
       this.assertException(function() {
-        var res = new qx.io.rest.Resource([]);
+        qx.io.rest.Resource([]);
       });
```
level420: Here again: is ```qx.io.rest.Resource([]);``` really equivalent to ```new qx.io.rest.Resource([]);```?
I don't thinks so.

framework/source/class/qx/test/mobile/core/Widget.js
```diff
@@ -136,7 +136,7 @@ qx.Class.define("qx.test.mobile.core.Widget",
       if (qx.core.Environment.get("qx.debug"))
       {
         this.assertException(function() {
-          var widget2 = new qx.ui.mobile.core.Widget().set({id:"affe"});
+          qx.ui.mobile.core.Widget().set({id:"affe"});
         });
```
level420: And again: Is calling ```qx.ui.mobile.core.Widget().set({id:"affe"});``` equivalent to ```new qx.ui.mobile.core.Widget().set({id:"affe"});```?

framework/source/class/qx/test/theme/simple/Appearance.js
```diff
@@ -443,13 +443,6 @@ qx.Class.define("qx.test.theme.simple.Appearance",
       this.assertIdentical(qx.theme.simple.Image.URLS["treevirtual-cross"], style.icon);
     },
 
-    testTreeVirtualCross : function()
```
level420: Why is this test completely removed?

level420 (changes requested):

level420:
@pcdummy it would be nice if you have some time to finish this work. it would be a great pity if all this work will we lost! thank you.

cboulanger:
@pcdummy Another year has passed and we still haven't merged your PR, after all these iterations. Would you be so kind to update your branch?

cboulanger:
No feedback for over a month, closing,

Issue #9161, opened 2016-09-05
qx.lang.Function.create speedup in "debug" build


schnepe2:
I've found on my application, that `qx.lang.Array.toString()` was using up most of the time while creating a data Model for a List (via `qx.data.marshal.Json.createModel()`)...

After a little digging I found that it was issued in the debug assertion check at `qx.lang.Function.create` :
```
      // ...
      return function(event)
      {
        if (qx.core.Environment.get("qx.debug"))
        {
          if (qx.core.Object && options.self && qx.Bootstrap.isObject(options.self) && options.self.isDisposed && qx.Bootstrap.isFunction(options.self.isDisposed))
          {
            qx.core.Assert && qx.core.Assert.assertFalse(
              options.self.isDisposed(),
              "Trying to call a bound function with a disposed object as context: " + options.self.toString() + " :: " + qx.lang.Function.getName(func)
            );
          }
        }
```
The 2nd parameter (`msg`) of `qx.core.Assert.assertFalse` is **always** evaluated, independent whether the assertion triggers or not.

What about changing it to:
```
      // ...
      return function(event)
      {
        if (qx.core.Environment.get("qx.debug"))
        {
          if (qx.core.Object && options.self && qx.Bootstrap.isObject(options.self) && options.self.isDisposed && qx.Bootstrap.isFunction(options.self.isDisposed))
          {
            if (options.self.isDisposed()) {
              qx.core.Assert && qx.core.Assert.fail(
                "Trying to call a bound function with a disposed object as context: " + options.self.toString() + " :: " + qx.lang.Function.getName(func)
              );
            }
          }
        }
```
The assertion is still checked, but now the (expensive!) options.self.toString() is only executed if the assertion really failed!

On my real-life example this changed the creation time of a model from 10164ms to 2318ms!

I would love to see this change in qooxdoo, it speeds up debug code a lot (10 seconds is an awful long time :snail: ; while 2.5 seconds is workable!)

oetiker:
how about creating a PR ? ... maybe this pattern exists in other locations too ?

schnepe2:
@oetiker : I did a quick search for similar pattern, but did not find any more that could have such an impact. Although this was by far not a **full scan of all assert calls**!
I am sorry, but I don't have the time to do a full check of **all** message creations of **all** assert calls.
Most are "harmless" however, I believe.

oetiker:
no problem :) thanks for the PR

level420:
[...]

PR #9162, opened 2016-09-05
changed assertion message generation to be executed only if assertion fails issue #9161

schnepe2:
PR for issue #9161 
- changed assertion message generation to be executed only if assertion fails.


oetiker:

level420:
@oetiker and you vote ;-)

level420:
4 votes
more than two business days
merging
thank you @schnepe2

PR #9164, opened 2016-09-05
adding ignoreUndeclaredPrivates to @lint hint

thron7:
This solves the issue with qx.test.Mixin's __b. See the Mixin class for @lint syntax.
I also fixed the framework's "lint-test" job so that it acutally works :-).

I'll update the manual once the PR is accepted.

### DIFF
--- a/documentation/manual/source/pages/development/api_jsdoc_ref.rst
+++ b/documentation/manual/source/pages/development/api_jsdoc_ref.rst
@@ -763,6 +763,9 @@ Description
              * **ignoreJsdocKey**
 
                JSDoc @ keys which are either unknown (i.e. not documented on this page) or do not comply with the syntax given here.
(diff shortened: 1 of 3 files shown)


level420:
@thron7 Thank you for you PR. As from the gitter chat: I still think this PR is still needed. We should revert the ugly "hack" this.__b vs. this["__b"] which is only there to trick lint.

oetiker:
@level420 and note that as @derrell pointed out, mixins can not access private variables of the class they are being mixed into

level420:
@oetiker yes thats true, but this is not the case here

thron7:
@level420 Done as requested (including squashing commits).

level420:
Thank you thron7.

level420:

level420:
4 Votes. More than 2 business days.
Merging.
Thank you @thron7

Issue #9165, opened 2016-09-07
generator.py compiler does not detect faulty string literals in function definitions


level420:
As discussed in https://github.com/qooxdoo/qooxdoo/issues/9153#issuecomment-244052804 the generator does not detect an error if the function definition contains a string literal like this:

```javascript
  myFunc : function('bla') {
  },
```
which should lead to an error during compilation.

thron7:
I'll try to look into this. (Don't know how to assign it to me here on Github.)

thron7:
(@thron7 Self-mentioning, so I have it in my list of 'mentioned' issues ...)

level420:
fixed via https://github.com/qooxdoo/qooxdoo/pull/9168 and landed master via 02d75a54faa9040f19e06f77a0b0dc92028b5963

PR #9168, opened 2016-09-07
[#9165]: detection of faulty string literals in function definitions for generator.py

thron7:
I don't know what the current test regimen for the tool chain is. I've tested this manually, on the framework test runner and api viewer among others. If you have more/automated tool chain tests, please run them after merging.


level420:
@thron7 we have automatic testing with travis. See https://travis-ci.org/qooxdoo/qooxdoo/builds/158292196

Currently a test related to chrome 54 and flash fails, but the other tests pass. So this should be OK.

level420:

level420:
3 votes and more than 2 business days
merging
thank you @thron

Issue #9167, opened 2016-09-07
travis run for chrome unstable (as for now version 54) fails on qx.test.ui.embed Flash:testEvents


level420:
See https://travis-ci.org/qooxdoo/qooxdoo/jobs/158131683

The log says:
```
Chrome 54.0.2840 (Linux 0.0.0) qx.test.ui.embed Flash:testEvents FAILED
	expected loaded to have been called at least once but was never called: Called fail().
		qx.core.AssertionError.constructor:43:38
		qx.core.Assert.__fail:71:19
		qx.core.Assert.fail:139:12
		qx.type.BaseString.fail:50:22
		qx.dev.unit.Sinon.failAssertion:4754:20
		anonymous()
		anonymous()
		qx.dev.unit.TestResult.run:237:36
		anonymous()
		anonymous()
		qx.event.dispatch.Direct.prototype.dispatchEvent()
		wrappedFunction()
		qx.event.Manager.prototype.dispatchEvent()
		qx.event.Registration.fireEvent()
		qx.core.Object.prototype.fireEvent()
		anonymous()
		qx.event.Timer.prototype._oninterval()
		anonymous()
```

level420:
This may be due to the upcoming changes in flash support for chrome:

https://chrome.googleblog.com/2016/08/flash-and-chrome.html

level420:
I've run the failing test (and all others) with testrunner on Windows 7 with Chrome 54.0.2840.16 beta-m (distrubuted today via the beta channel) where the tests succeeded.

So it seems to be specific to either the travis environment or the linux version of Chrome.

We should maybe wait for the next beta to see if this is a bug in chrome.

Also note that all other flash related tests work.

cajus:
Looks like it's no more reproducible locally. I had it yesterday, but now as I want to debug it, the test is green and I can't get it to fail in chrome 45. Doh.

level420:
@cajus how did you server? both tests with php?

cajus:
yes

cajus:
So it looks like the tests are enabled, but the *PercentLoaded()* method returns a *TypeError*.

level420:
@cajus is it reproducible again?

cajus:
only on travis

cajus:
Ok. It's only on Ubuntu. No idea why I've seen it exactly once on my Fedora installation, though.

[...]

PR #9171, opened 2016-09-09
[#9167] Raise timeout for event test in Flash tests

level420:
Attempt to raise the timeout for the qx.test.ui.embed.Flash event tests in order to fix the failing tests in travis.

Maybe a fix for https://github.com/qooxdoo/qooxdoo/issues/9167


level420:
This does not help. Tests still failing.

Issue #8936, opened 2015-09-14
Several issues when closing Window widget by ESC key (BZ#9208)



qx-bug-importer:
_Joseph Karras (stylkarr) wrote:_

The problem is harder to describe than see in action, so here is a playground application that demonstrates it:

http://tinyurl.com/nsszdt6

1st bug: Click the button to open the window, hover the hover area and press ESC on your keyboard -&gt; an exception is thrown

2nd bug: Click the button to open the window, hover your mouse over the edge of the window to change the cursor to its resize variant, press ESC on your keyboard -&gt; cursor is not reset

level420:
Reopened due to user request.

level420:
@JosKar I'm able to reproduce the second bug, but not the first one. Maybe because I can't understand what is meant by "hover the hover area".
Could you please re evaluate and report back here? Thank you!

JosKar:
Hi @level420 
I just re evaluate the playground code in FF ESR 45.3.0 and Chromium  52.0. I got the same results.
Steps to reproduce the first bug:
1) Click the button click me. A window should appear.
2) Hover your mouse over the label "Hover me and press ESC". A pop up should  appear.
3) With the pop up opened press the ESC key to close the window.

In your browser console you will see the following error:
> qx.ui.core.queue.Layout: Error in the 'Layout' queue:TypeError: Cannot read property 'setStyles' of null TypeError: Cannot read property 'setStyles' of null

level420:
It might be that the 1st issue in the description was fixed by PR https://github.com/qooxdoo/qooxdoo/pull/168

level420:
Fixed in master via https://github.com/qooxdoo/qooxdoo/pull/9173 and https://github.com/qooxdoo/qooxdoo/pull/168

PR #9173, opened 2016-09-11
Reset global cursor after a window closes

joskar:
This is a fix for an old bug see second issue of this [bug](https://github.com/qooxdoo/qooxdoo/issues/8936). Demonstration of the bug in the playground [here](http://tinyurl.com/zekfy25).

### DIFF
--- a/framework/source/class/qx/ui/core/MResizable.js
+++ b/framework/source/class/qx/ui/core/MResizable.js
@@ -618,7 +618,10 @@ qx.Mixin.define("qx.ui.core.MResizable",
 
   destruct : function()
   {
-    this.getApplicationRoot().resetGlobalCursor();
+    if(this.getCursor()) {
+      this.getApplicationRoot().resetGlobalCursor();
+    }
+    
     if (this.__resizeFrame != null && !qx.core.ObjectRegistry.inShutDown)
     {
       this.__resizeFrame.destroy();
--- a/framework/source/class/qx/ui/core/MResizable.js
+++ b/framework/source/class/qx/ui/core/MResizable.js
@@ -618,6 +618,7 @@ qx.Mixin.define("qx.ui.core.MResizable",
 
   destruct : function()
   {
+    this.getApplicationRoot().resetGlobalCursor();
     if (this.__resizeFrame != null && !qx.core.ObjectRegistry.inShutDown)
     {
       this.__resizeFrame.destroy();


level420:
@JosKar Reopened https://github.com/qooxdoo/qooxdoo/issues/8936

zaucker:
@JosKar: what if another window has also changed the cursor to some other cursor? Then your fix would probably not have the desired effect.
I am not saying that the current behaviour is correct. But your fix might break other applications rather unexpectedly.

cajus:
What to do with this one? If I understand the comments correctly, it fixes some bad behaviour, but is not general enough to be accepted just now? Is it generic fixable at all?

johnspackman:
the bug is because the popup has been disposed of, but the layout is trying to position it's contentElement (and there is no element any more because it's been disposed).  I don't know why the proposed patch is fixing this, maybe it's by just changing the timing slightly (ie the lack of exception is a side effect rather than actually solving the problem).

This does not happen on Chrome 54 (dev channel) but does on FF 49.0.2.

I suspect that the correct fix is to resolve why the popup is being disposed before it is removed from the widget layout queue, rather than just change the cursor which as @zaucker says could have unintended side effects.  IMHO it requires more investigation before applying this fix.

level420:
After studying the code in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/MResizable.js a bit, I came to the conclusion that this PR is totally valid, regarding the error with the cursor (as @JosKar wrote).

This is not a fix for the complete issue https://github.com/qooxdoo/qooxdoo/issues/8936 , only for the part with the cursor staying in resize mode when ESC is typed during resize.

Resetting the global cursor via ```this.getApplicationRoot().resetGlobalCursor();``` is valid, as it's done in the same way as it#s done if resizing is released or if the pointer is moved outside the resize frame/area: 
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/MResizable.js#L598-L605.

One addition though: I'd take the code from the sequence above and change that part in the destruct method to
```javascript
    if(this.getCursor()) {
      this.getApplicationRoot().resetGlobalCursor();
    }
```
which only resets the global cursor if we changed it.

level420:
Start voting

oetiker:
ok makes sense now

oetiker (approved):

cajus (approved):

zaucker (approved):

level420 (approved):

johnspackman (approved):

derrell (approved):

level420:
Unbelievable!  A full quorum of 6 approvals!
Merging. Thank you @JosKar

Issue #9174, opened 2016-09-11
combo box scrollbars did not work under edge


hkollmann:
The scrollbars of the combo box did not work correct under edge.
To reproduce you can go to the widget browser http://www.qooxdoo.org/current/widgetbrowser/,
open the combo box and try to use the slider. The combo box is closed when hitting the scroll area

level420:
Verified on Windows 10, Microsoft Edge 39.14915.1000.0

level420:
To be more precise: the list popup closes unexpectedly if you click on the scrollbar when trying to scroll the list content.

level420:
I suspect that the listener which should stop the propagation of the events are either not complete (some events are missing) or simply do not work, which means that they do not really stop the propagation.

See https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/scroll/NativeScrollBar.js#L61-L63

level420:
This is only true for native scrollbars.

level420:
I've created a jsfiddle to check if an event which is fired through the scrollbar is not stoppable via stopPropagation or there is some new event which is not handled via the native scrollbar listeners, but everything seems OK and reacts as it does in IE 11:

https://jsfiddle.net/4x75mg7c/2/

level420:
This may be related to https://github.com/qooxdoo/qooxdoo/issues/8786 which was fixed with https://github.com/qooxdoo/qooxdoo/commit/f025242a702a5ccc98d00aaf50a3fdebee5f2f89

level420:
Did some more tests. In MS Edge when trying to use the scrollbar it seems that the scrollbar itself or some widget behind is receiving the focus, which then leads to the ```blur``` event fired for the scrollbar which is closing the popup of the scrollbar ( https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/AbstractSelectBox.js#L275 )

level420:
One more debug sessions where I tried add some debug code to ```qx.ui.core.EventHandler``` and find out which native events are fired. Additionally I've added native event handlers to  ```qx.ui.core.scroll.NativeScrollBar``` where I registered all available native dom elements of the scrollbar widget itself and the scroll pane element inside by using the follwing helper method: 
```javascript
    _addAllListeners : function(element, name) {
      var events = [];
      for (var property in element) {
          var match = property.match(/^on(.*)/)
          if (match) { 
              events.push(match[1]);
          }
[...]

PR #9178, opened 2016-09-15
Ugly fix for issue #9174 selectbox closes popup unexpectidly when clicking on the list scrollbar when using native scrollbars

level420:
This is more a workaround until the real cause in issue #9174 is fixed.

### DIFF
--- a/framework/source/class/qx/ui/form/AbstractSelectBox.js
+++ b/framework/source/class/qx/ui/form/AbstractSelectBox.js
@@ -123,6 +123,10 @@ qx.Class.define("qx.ui.form.AbstractSelectBox",
 
   members :
   {
+    // inhibit blur flag for MS Edge fix for issue [ISSUE #9174]
+    __inhibitBlur : false,
+
+
     // overridden
     _createChildControlImpl : function(id, hash)
     {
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/form/AbstractSelectBox.js
```diff
@@ -272,11 +282,42 @@ qx.Class.define("qx.ui.form.AbstractSelectBox",
...
     },
 
 
     /**
+     * Handler for the pointerdown event of scrollbar-y child of the list 
+     * MS Edge only fix for [ISSUE #9174] 
+     * 
+     * @param e {qx.event.type.Pointer} The pointerdown event.
+     */
+    __onScrollbarYPointerDown : function(e) {
```
derrell: Since these two pointer functions are explicitly only for edge, maybe add them in the constructor, in Environment-specific code?

```
if(qx.core.Environment.get("browser.name") == "edge") {
  this.__onScrollbarYPointerDown = function(e) { ... };
  this.__onScrollbarYPointerUp = function(e) { ... };
}
```

level420: I could also wrap them with a ```qx.core.Environment.select("engine.name"``

derrell (changes requested):
Marking this as Request Changes because @level420 indicated in #9174 that this is to solve a particular problem for  @hkollmann, but not be merged.

level420:
@derrel this is not only a fix for @hkollmann , but for everyone using qooxdoo with select boxes, combo boxes on MS Edge!

derrell:
@level420 Point taken. You did want to avoid merging this, though, right? I thought that was what I understood from your note to him.

level420:
@derrell I want this PR to really be obsolete and I hope that someone finds a better fix for the issue. So yes I want this PR to be there for everybody running into this issue but not merged yet.

framework/source/class/qx/ui/form/AbstractSelectBox.js
level420: Done!

johnspackman:
LGTM

A general case, event level equivalent would be good but this seems like a very effective fix and a lot easier to evaluate and test than changes to low level event manger code.

level420:
Retracting this PR in favor of https://github.com/qooxdoo/qooxdoo/pull/9180

PR #9181, opened 2016-09-18
add manual entry on using external JS libs

thron7:
This is just a doc PR, but since I answered this question a couple of times (recently [here](http://stackoverflow.com/a/39555502/127465)) I thought it was worth a manual entry.

Caveat: The section about wrapping an external lib in a qooxdoo class is incomplete. I've looked through some of Tobi's contributions, but they are quite elaborate especially when loading is concerned, and they also rely on older qooxdoo versions (at least in parts). Maybe he can come up with an updated, simplified version which only shows the bare minimum you need to do, and  fill in the missing pieces.

### DIFF
--- a/documentation/manual/source/pages/development.rst
+++ b/documentation/manual/source/pages/development.rst
@@ -106,4 +106,5 @@ Miscellaneous
    development/reporting_bugs
    development/aspects_template
    development/internet_explorer_specific_settings
+   development/using_non_qx_libs
 
--- a/documentation/manual/source/pages/development/using_non_qx_libs.rst
+++ b/documentation/manual/source/pages/development/using_non_qx_libs.rst
@@ -1,0 +1,96 @@
+Using non-qooxdoo, third-party libraries
+*****************************************
+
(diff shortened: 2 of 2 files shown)


oetiker:
cool will work through ... this ...

oetiker:
I think we should get @level420 s script loader mixin mentioned too

oetiker:
hi @thron7

now that the dynamic script loader is merged, could you update your documentation patch ?

thron7:
I can do that but would need some input from you guys, as I'm not acquainted with the DynamicScriptLoader. Is there still a usage example in the class comment?

As for this PR, I'm surprised it is not already merged so I could do a new patch :-) . Seriously, is there any reason to not merge incremental improvements? This PR is already sooo old ... 

But if you wish I can try and enhance it. Can I just add new commits, or do I have to re-merge master before that?

oetiker:
ok lets merge it

oetiker (approved):

level420 (approved):

level420:
documentation only PR. 2 approvals.
merging
Thank you @thron7

PR #9183, opened 2016-09-22
Added qx.core.MProperty#isPropertyInitialized

mbgonicus:
If you got a property `foo` which is not initialized and you call `getFoo()`, you'll get an error that the property is not initialized. You must either enclose it in a try-catch block or set an init value for the property. 

This PR introduces a small function to check if a property is initialized.

### DIFF
--- a/framework/source/class/qx/core/MProperty.js
+++ b/framework/source/class/qx/core/MProperty.js
@@ -137,7 +137,9 @@ qx.Mixin.define("qx.core.MProperty",
      */
     isPropertyInitialized : function(prop)
     {
-      qx.core.Assert.assertString(prop);
+      if (qx.core.Environment.get("qx.debug")) {
+        qx.core.Assert.assertString(prop);
+      }
 
       if (!this["get" + qx.Bootstrap.firstUp(prop)]) {
         throw new Error("No such property: " + prop);
--- a/framework/source/class/qx/core/MProperty.js
+++ b/framework/source/class/qx/core/MProperty.js
@@ -125,6 +125,26 @@ qx.Mixin.define("qx.core.MProperty",
 
 
       this[resetter[prop]]();
+    },
+
+    /**
+     * Checks if the property is initialized, i.e. has a defined init value or
+     * has got a value by a setter method.
+     *
+     * @param prop {String} Name of the property
+     * @returns {Boolean} If the property is initialized
+     * @throws {Error} If the property defined does not exist
+     */
+    isPropertyInitialized : function(prop)
+    {
+      qx.core.Assert.assertString(prop);
+
+      if (!this["get" + qx.Bootstrap.firstUp(prop)]) {
+        throw new Error("No such property: " + prop);
+      }
+
+      return this["$$user_" + prop] !== undefined ||
+        this["$$init_" + prop] !== undefined;
     }
   }
 });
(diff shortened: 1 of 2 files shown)


level420 (approved):

level420:
LGTM

cajus (approved):

stahl-riesa:
IMHO: qx.core.Assert.assertString(prop) should not be called.

level420:
@stahl-riesa could you please explain why?

derrell:
It seems more like a coding error to me if one is getting a property value before setting it (somehow -- through init or a setter). Adding code that /shouldn't/ be needed may not be the best thing to do. Thoughts?

derrell (changes requested):
See my comment in "Conversation"

mbgonicus:
@derrell Well, that's one point of view. But in Javascript uninitialized values can always be checked by comparing them to undefined. Qooxdoo breaks with that idiom.

Anyway - with the function, one has the possibility to use uninitialized properties as a valid state. Not allowing that would be a dictation to the user, but I think that is his decision.

All in all its just a short additional function that noone _must_ use.

level420:
@derrell even if this method should/would not be used in common scenarios, it allows us to introspect the initialization state of a property through an official API. Sounds useful to me.

derrell:
I'm not inherently opposed to adding this, but it seems like a rarely used function that just adds bloat and needs to be maintained. We have a large footprint already, so we should carefully consider whether to add new functions (vs. adding classes which are optimized out if not used). Let's just think carefully before adding it. If, after consideration of this, everyone thinks it's still worthwhile, it'll have my vote too.

derrell:
This function is tiny, so it likely makes little difference whether we add it, but lots of snacks do add up to gaining weight...

framework/source/class/qx/core/MProperty.js
```diff
@@ -125,6 +125,26 @@ qx.Mixin.define("qx.core.MProperty",
...
+     * Checks if the property is initialized, i.e. has a defined init value or
+     * has got a value by a setter method.
+     *
+     * @param prop {String} Name of the property
+     * @returns {Boolean} If the property is initialized
+     * @throws {Error} If the property defined does not exist
+     */
+    isPropertyInitialized : function(prop)
+    {
+      qx.core.Assert.assertString(prop);
```
johnspackman: this should be surrounded by a check for qx.debug

johnspackman (changes requested):
the qx.core.Assert call should be qx.debug only

framework/source/class/qx/core/MProperty.js
stahl-riesa: > this should be surrounded by a check for qx.debug

Sorry, thats what I mean!

johnspackman:
while i agree that uninitialised properties is often a coding design issue, this is not always the case, especially when dealing with arbitrary properties eg via reflection.

framework/source/class/qx/core/MProperty.js
mbgonicus: fair enough - added the check

mbgonicus:
qx.debug check is added

johnspackman (changes requested):
Oops, sorry I didnt spot this earlier Markus - please can you put the check for a getter in the debug also?  As it's also a sanity check

mbgonicus:
Sure - moved getter check in qx.debug check

johnspackman (approved):

patrick-fischer-hirschstein (approved):

derrell (approved):
Discussion consensus seems to be to include this even though it will likely be unused by most users. It's a tiny function with most elided in build version. Ok with me.

oetiker (approved):

level420:
5 approvals.
merging.
Thank you @mbgonicus

PR #9188, opened 2016-09-23
add a dynamic script loader mixin for non qx javascript libraries

level420:
allows simple loading of external javascript libraries like jquery where loading sequence is preserved

### DIFF
--- a/framework/source/class/qx/io/MDynamicScriptLoader.js
+++ b/framework/source/class/qx/io/MDynamicScriptLoader.js
@@ -202,6 +202,7 @@ qx.Mixin.define("qx.io.MDynamicScriptLoader",
         //
         else {
           this.constructor.__LOADING[script] = this;
+          qx.io.MDynamicScriptLoader.__GLOBAL_LOADING[script] = this;
           
           var uri = qx.util.ResourceManager.getInstance().toUri(script);
           
--- a/framework/source/class/qx/io/MDynamicScriptLoader.js
+++ b/framework/source/class/qx/io/MDynamicScriptLoader.js
@@ -147,7 +147,7 @@ qx.Mixin.define("qx.io.MDynamicScriptLoader",
       var script = codeArr.shift();
 
       if (script) {
-        // the script was alreay loaded, recurse to the next script
+        // if the script was alreay loaded, recurse to the next script
         //
         if (this.constructor.__LOADED[script] === true) {
           this.__loadScriptArr(codeArr);
@@ -155,13 +155,13 @@ qx.Mixin.define("qx.io.MDynamicScriptLoader",
         // the script is currently loading, initiated by another instance
         // 
         else if (this.constructor.__LOADING[script]){
-          // start over to the next script, when this script is successfully loaded by
+          // start again at the next script, when this script is successfully loaded by
           // the other instance
           this.constructor.__LOADING[script].addListenerOnce('scriptLoadingSuccess', function(){
             this.__loadScriptArr(codeArr);
           }, this);
         }
-        // the loader was not yet startet for this script
+        // the loader was not yet started for this script
         //
         else {
           this.constructor.__LOADING[script] = this;
--- a/framework/source/class/qx/io/MDynamicScriptLoader.js
+++ b/framework/source/class/qx/io/MDynamicScriptLoader.js
@@ -1,0 +1,218 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016 Visionet GmbH, http://www.visionet.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Dietrich Streifert (level420)
+
+************************************************************************ */
+
+
+/**
+ * Dynamically load non qx scripts on first instance init
+ *
+ * Usage example:
+ * 
+ * <code>
+ *  ... assets ...
+ * /**
+ *  * @asset(myapp/jquery/*)
+ *  * @asset(myapp/highcharts/*)
+ *  *
+ *  * @ignore(jQuery.*)
+ *  * @ignore(Highcharts.*)
+ *  ...
+ *
+ *  
+ *  qx.Class.define("myapp.myClass.Chart", {
+ *    extend : myapp.myClass,
+ *
+ *    include : [
+ *      qx.io.MDynamicScriptLoader
+ *    ],
+ * 
+ *    construct : function()
+ *    {
+ *      this.base(arguments);
+ * 
+ *      // other useful contructor code here
+ * 
+ *      // do something usefull if all scripts are loaded
+ *      this.addListener('scriptsReady', this.__onScriptsReady, this);
+ *   
+ *      // in debug mode load the uncompressed unobfuscated scripts
+ *      var src = '';
+ *      var min = '.min';
+ *      if (qx.core.Environment.get("qx.debug")) {
+ *        src = '.src';
+ *        min = '';
+ *      }
+ *
+ *      // initialize the script loading
+ *      this._loadScriptsDynamic([
+ *        "myapp/jquery/jquery"+min+".js",
+ *        "myapp/highcharts/highcharts"+src+".js",
+ *        "myapp/highcharts/highcharts-more"+src+".js",
+ *        "myapp/highcharts/highcharts-modifications.js"
+ *      ]);
+ *    }
+ *  });
+ *      
+ * </code>
+ */
+qx.Mixin.define("qx.io.MDynamicScriptLoader",
+{
+  /*
+  *****************************************************************************
+     EVENTS
+  *****************************************************************************
+  */
+
+  events : {
+    /**
+     * fired when a specific script is loaded successfully
+     */
+    scriptSuccess: 'qx.event.type.Data',
+
+    /**
+     * fired when a specific script fails loading
+     */
+    scriptFailed: 'qx.event.type.Data',
+
+    /**
+     * fired when all given scripts are loaded, the first time all scripts reported
+     * load success
+     */
+    scriptsLoadingComplete: 'qx.event.type.Event',
+    
+    /**
+     * fired when all given scripts are loaded, each time loadScriptsDynamic is called
+     */
+    scriptsReady: 'qx.event.type.Event'
+  },
+
+
+  /*
+  *****************************************************************************
+     MEMBERS
+  *****************************************************************************
+  */
+
+
+  members : {
+  
+    /**
+     * Load scripts dynamically, typically used in the instance constructor
+     * 
+     * @param codeArr {Array} an array with the uri names of the scripts
+     * 
+     */
+    _loadScriptsDynamic: function(codeArr) {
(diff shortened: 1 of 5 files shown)


framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+ *      qx.io.MDynamicScriptLoader
+ *    ],
+ * 
+ *    construct : function()
+ *    {
+ *      this.base(arguments);
+ * 
+ *      // other useful contructor code here
+ * 
+ *      // do something usefull if all scripts are loaded
```
oetiker: I would say 'when' all scripts are loaded ... the 'if' sounds so tentative

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+  *****************************************************************************
+     EVENTS
+  *****************************************************************************
+  */
+
+  events : {
+    /**
+     * fired when a specific script is loaded successfully
+     */
+    scriptSuccess: 'qx.event.type.Data',
```
oetiker: for consistent naming how about `scriptLoadingSuccessful`

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * fired when a specific script fails loading
+     */
+    scriptFailed: 'qx.event.type.Data',
```
oetiker: for consistent naming how about `scriptLoadingFailed`

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+          this.fireEvent('scriptsLoadingComplete');
+        }
+
+        // scriptsReady event is fired for each instance
+        // this.debug("__loadScriptArr: all scripts already loaded, fireing event scriptsReady");
+        this.fireEvent("scriptsReady");
+      }
+    }
+  }
+});
```
oetiker: add ending newline

oetiker (commented):
looks good ... some comments

level420:
@oetiker done!

oetiker (approved):

patrick-fischer-hirschstein (approved):

cajus (commented):
Any reason why the events are all prefixed with "scripts"?  When taking a look at the current qx event naming scheme, it seems like short names are prefered - like "error", "success", etc.

Not sure if that would make some conflicts, though.

oetiker:
the 'problem' is that it is a mixin ... but I wonder ... does it have to be a mixin ?

cajus:
Hehe. Yes. Why is it one?

oetiker (commented):
does this have to be a mixin ?

level420:
It is mainly a mixin, because I use it in widgets which are wrappers for external libraries like highcharts and ckeditor.

It came handy to leave my wrapper classes as they are and use the mixin to add the dynamic script loading capabilities to them.

Remember the usage: you add some ```@asset``` hints to the class you want to be 'dynamificated', assign a listener to the ```scriptsReady``` event and initiate the loading by calling the method ```_loadScriptsDynamic```.

Additionally there sequential loading and state is held in some class variables ( ```__LOADING, __LOADED, __LOADING_COMPLETE```) which is specifically to the class the mixin is mixed in.

level420:
A global dynamic loader class would for sure be a lot more complicated, as you then have to get track of the different script sets to be loaded. Lets say instance a of Class A requests two scripts and instance b of Class B also requests three scripts while one script is common to both instances.

level420:
And another point: why not a mixin?

cajus:
That just came up with the question about event naming.

oetiker:
since the loaded scripts are global by nature, wouldn't a singleton be more suitable ? you would still do the asset thing and all as  you described ... but there would certainly be no name space issues ... and it would work smoothly if two different classes were about to load the same extrenal script

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+     * Recursively called until the array of scripts is consumed
+     * 
+     * @param codeArr {Array} an array with the uri names of the scripts
+     */
+    __loadScriptArr: function(codeArr){
+
+      var script = codeArr.shift();
+
+      if (script) {
+        // the script was alreay loaded, recurse to the next script
```
derrell: missing word: IF the script was already loaded...

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+        //
+        if (this.constructor.__LOADED[script] === true) {
+          this.__loadScriptArr(codeArr);
+        }
+        // the script is currently loading, initiated by another instance
+        // 
+        else if (this.constructor.__LOADING[script]){
+          // start over to the next script, when this script is successfully loaded by
```
derrell: s/start over to/start again at/

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+          // the other instance
+          this.constructor.__LOADING[script].addListenerOnce('scriptLoadingSuccess', function(){
+            this.__loadScriptArr(codeArr);
+          }, this);
+        }
+        // the loader was not yet startet for this script
```
derrell: s/startet/started/

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+      this.__initStateHashes();      
+      this.__loadScriptArr(codeArr);
+    },
+
+    
+    /**
+     * Initialize the state hashes if not yet done
+     */
+    __initStateHashes : function() {
+      if(!this.constructor.__LOADING) {
```
derrell: The /statics/ section, to include __LOADING and __LOADED, is missing.

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -0,0 +1,218 @@
...
+          loader.open("GET", uri);
+          loader.send();
+        }
+      }
+      // no more scripts in codeArr, which means that all the
+      // scripts were loaded.
+      else {
+        // ensure that the scriptsLoaded event is only fired once
+        //
+        if(this.constructor.__LOADING_COMPLETE !== true) {
```
derrell: __LOADING_COMPLETE should also be in the /statics/ section.

derrell (changes requested):
This will be very useful, as a standardized way to load external scripts! See my few comments.

framework/source/class/qx/io/MDynamicScriptLoader.js
level420: @derrell: this does not work in mixins. If i declare a statics section the mixin errors.

framework/source/class/qx/io/MDynamicScriptLoader.js
level420: same as before. the mixin does not allow a static section.

framework/source/class/qx/io/MDynamicScriptLoader.js
level420: done

framework/source/class/qx/io/MDynamicScriptLoader.js
level420: done

framework/source/class/qx/io/MDynamicScriptLoader.js
level420: done

derrell:
I had also missed the fact that this is a mixin. I don't grok how I'd use it as a mixin; it wouldn't match anything I've ever done. In the past, I would have collected those scripts that the entire app needed, and issued the requests from one place. I do, however, much prefer the singleton approach, which seems to have the benefits you explain regarding mixins without having to actually mix anything in. Tracking which scripts are requested by various other classes becomes trivially easy with the existing statics.

level420:
I'm still in favor of the mixin solution. If you like to see how it works I recommend reading the tests I've done.

One of the benefits is that the class where you mixed this in "knows" that all the scripts it needs are loaded, loads the script only if you create an instance of the class and coordinates loading of the scripts if multiple instances are created, as the loading itself is asynchronously.

@adeliz is using the mixin solution in his qxchart contrib and seems to be very happy with this.

Another aspect for me is that I currently have no time and need (I'm using that mixin ;-) ) to try implementing this as a singleton. So this PR will remain as is as a mixin. That will be then another PR.

framework/source/class/qx/io/MDynamicScriptLoader.js
derrell: Right. That's because it's not clear what this.constructor is, in a mixin. If I'm not mistaken, what this is actually doing is modifying the statics of the class that mixed it in, meaning that it's not global between various classes that use this mixin. That would mean that if two different classes use this mixin and include the same script, that script will actually be downloaded twice. That sounds undesirable, if it's what is actually happening.

Here's an example of how I'd do it, which both makes it much more clear that the statics are of the mixin, and ensures that there is only one global copy of the maps for all classes using the mixin. Note that it does not use this.constructor:

```
qx.Mixin.define("Mixer",
{
  statics :
  {
    myData : "hello world"
  },
  
  members :
  {
    showStatics : function()
    {
      this.debug("myData=" + Mixer.myData)
    }
  }
});

qx.Class.define("Clazz",
{
  extend : qx.core.Object,
  include : [ Mixer ],
  members :
  {
    doit : function()
    {
      this.showStatics();
    }
  }
});

var c = new Clazz();
c.doit();
```

level420:
Retracted this PR in favour of @oetiker s upcoming loader implemented as a PR.

level420:
reopen and close to allow the latest change to come in.

framework/source/class/qx/io/MDynamicScriptLoader.js
```diff
@@ -103,6 +103,25 @@ qx.Mixin.define("qx.io.MDynamicScriptLoader",
...
   *****************************************************************************
+     STATICS
+  *****************************************************************************
+  */
+
+  statics : {
+    /**
+     * a global hash with all scripts currently being processed 
+     */
+    __GLOBAL_LOADING : {},
```
derrell: This would solve the "global" problem. Makes sense. I don't see, however, where __GLOBAL_LOADING[script] ever gets set to anything other than null. Have I just missed it someplace?

level420: Sorry. Fixed it. I'm just having to much versions of this all here locally

level420:
Oddly I have to reopen the PR in order to get the sync with my repo.

derrell (commented):
Ok, I think this is technically clean now. I suspect it is somewhat confusing to the reader who didn't have the benefit of this whole discussion, e.g., why are there both __LOADING and __GLOBAL_LOADING variables, and what's the difference between them. You understand it, so I guess since it's no longer a PR, that's all that matters. :-)

Issue #9187, opened 2016-09-23
lint warning in qx.lang.normalize.Array prevents successful travis run  


level420:
It seems that this came in with https://github.com/qooxdoo/qooxdoo/pull/9177/files
```
Running lint...
  - Warning: qx.lang.normalize.Array (389, 36): qx.core.Environment.get: first argument is not a string literal.
Lint warnings are treated as errors! Please fix!
```
Here one of the latest travis runs: https://travis-ci.org/qooxdoo/qooxdoo/jobs/161879413

level420:
@johnspackman could you please have a look? Thank you!

johnspackman:
no problem

level420:
Fixed via https://github.com/qooxdoo/qooxdoo/pull/9192

PR #9192, opened 2016-09-27
Fix lint warnings in qx.lang.normalize.Arrays

johnspackman:
See https://github.com/qooxdoo/qooxdoo/issues/9187

### DIFF
--- a/framework/source/class/qx/lang/normalize/Array.js
+++ b/framework/source/class/qx/lang/normalize/Array.js
@@ -384,6 +384,9 @@ qx.Bootstrap.define("qx.lang.normalize.Array", {
     }
   },
 
+  /**
+   * @lint environmentNonLiteralKey()
+   */
   defer : function(statics) {
     var install = function(key, name) {
       if (!qx.core.Environment.get(key)) {
--- a/framework/source/class/qx/lang/normalize/Array.js
+++ b/framework/source/class/qx/lang/normalize/Array.js
@@ -24,6 +24,7 @@
  * MDN documentation &copy; Mozilla Contributors.
  *
  * @group (Polyfill)
+ * @use(qx.bom.client.EcmaScript)
  */
 qx.Bootstrap.define("qx.lang.normalize.Array", {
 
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/lang/normalize/Array.js
```diff
@@ -384,49 +384,29 @@ qx.Bootstrap.define("qx.lang.normalize.Array", {
...
-
-    if (!qx.core.Environment.get("ecmascript.array.reduce")) {
-      Array.prototype.reduce = statics.reduce;
-    }
-
-    if (!qx.core.Environment.get("ecmascript.array.reduceright")) {
-      Array.prototype.reduceRight = statics.reduceRight;
-    }
+    var install = function(key, name) {
+      if (!qx.core.Environment.get(key)) {
```
level420: Wouldn't it be useful to allow ```qx.core.Environment.get``` to also accept variables as a parameter?
Does anyone know why the restriction to only use string literals is there?

johnspackman: it's because it prevents the generator from automatically adding dependencies based on the key - which reminds me, I need to add an explicit dependency...

level420: Ah! OK. Then...

level420 (approved):

level420:
I'd love to have travis runs succeed again :-)
This PR is the last one which is needed to allow travis runs to succeed again! 
Please review!

derrell (approved):

zaucker (approved):

cajus (approved):

oetiker (approved):

framework/source/class/qx/lang/normalize/Array.js
```diff
@@ -384,49 +384,29 @@ qx.Bootstrap.define("qx.lang.normalize.Array", {
     }
   },
 
+  /**
+   * @lint environmentNonLiteralKey()
```
schnepe2: Maybe nitpicking here, but shouldn't this lint exception be more "focused"? Like:
```
  /*
   * @lint environmentNonLiteralKey(key)
   */
```
[Note the **key**]

PR #9194, opened 2016-09-27
DynamicScriptLoader

oetiker:
DynamicScriptLoader based on MDynamicScriptLoader as propos ed by @level420 in #9188

### DIFF
--- a/framework/source/class/qx/io/DynamicScriptLoader.js
+++ b/framework/source/class/qx/io/DynamicScriptLoader.js
@@ -1,0 +1,247 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2016 Visionet GmbH, http://www.visionet.de
+     2016 OETIKER+PARTNER AG, https://www.oetiker.ch
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Dietrich Streifert (level420)
+     * Tobias Oetiker (oetiker)
+
+************************************************************************ */
+
+
+/**
+ * Dynamically load non qx scripts on first instance init
+ *
+ * Usage example:
+ *
+ * <code>
+ *  ... assets ...
+ * /**
+ *  * @asset(myapp/jquery/*)
+ *  * @asset(myapp/highcharts/*)
+ *  *
+ *  * @ignore(jQuery.*)
+ *  * @ignore(Highcharts.*)
+ *  ...
+ *
+ *
+ *    // in debug mode load the uncompressed unobfuscated scripts
+ *    var src = '';
+ *    var min = '.min';
+ *    if (qx.core.Environment.get("qx.debug")) {
+ *      src = '.src';
+ *      min = '';
+ *    }
+ *    var callback = function(){
+ *        console.log("Got it");
+ *    };
+ *    // initialize the script loading
+ *    var dynLoader = qx.io.DynamicScriptLoader.getInstance();
+ *
+ *
+ *    var readyId = dynLoader.addListenerOnce('ready',function(e){
+ *      console.log("all scripts have been loaded!");
+ *    });
+
+ *    dynLoader.addListener('failed',function(e){
+ *      var data = e.getData();
+ *      dynLoader.removeListenerById(readyId);
+ *      console.log("failed to load "+data.script);
+ *    });
+ *
+ *    var lastScript = dynLoader.load([
+ *        "myapp/jquery/jquery"+min+".js",
+ *        "myapp/highcharts/highcharts"+src+".js",
+ *        "myapp/highcharts/highcharts-more"+src+".js",
+ *        "myapp/highcharts/highcharts-modifications.js"
+ *    ],callback,this);
+ *
+ *    if (lastScript === null){
+ *      var loadId = loader.addListener('loaded',function(e){
+ *        var data = e.getData();
+ *        if (data.script == lastScript){
+ *          console.log('Last Script Loaded');
+ *          dynLoader.removeListenerById(loadId);
+ *        }
+ *      }, this);
+ *    }
+ *    else {
+ *      console.log('Scripts have already been loaded');
+ *    }
+ * </code>
+ */
+qx.Class.define("qx.io.DynamicScriptLoader", {
+  extend: qx.core.Object,
+
+  type: "singleton",
+
+  /**
+   * Initialize the state hashes if not yet done
+   */
+
+  construct: function () {
+    this.base(arguments);
+    this.__ADDED = {};
+    this.__LOADED = {};
+    this.__QUEUE = [];
+    this.__LOADING = false;
+  },
+
+  /*
+  *****************************************************************************
+     EVENTS
+  *****************************************************************************
+  */
+
+  events: {
+    /**
+     * fired when a specific script is loaded successfully
+     */
+    loaded: 'qx.event.type.Data',
+
+    /**
+     * fired when a specific script fails loading
+     */
+    failed: 'qx.event.type.Data',
+
+    /**
+     * fired when all given scripts are loaded, each time loadScriptsDynamic is called
+     */
+    ready: 'qx.event.type.Event'
+  },
+
+
+  /*
+  *****************************************************************************
+     MEMBERS
+  *****************************************************************************
+  */
+
+
+  members: {
+
+    __ADDED: null,
+    __LOADED: null,
+    __QUEUE: null,
+    __LOADING: null,
+
+
+    /**
+     * Load scripts dynamically, typically used in the instance constructor
+     *
+     * @param codeArr {Array} an array with the uri names of the scripts
+     * @param callback {Function?undefined} a callback function to call when all the scripts in the array have been loaded
+     * @param context {Object?undefined} a contex for calling the callback
+     *
+     * @return {String|null} name of the last script to be loaded
+     *     Returns the name of the last script you can expect to be loaded if all scripts are already loaded null will be returned.
+     *
+     */
+    load: function(codeArr, callback, context) {
+      var lastScript = null;
+      codeArr.forEach(function(script) {
+        if (this.__ADDED[script] !== true) {
+          this.__QUEUE.push(script);
+          this.__ADDED[script] = true;
+        }
+        if (.this.__LOADED[script] !== true) {
+          lastScript = script;
+        }
+      }, this);
+      if (callback) {
+        var cb = new qx.util.DeferredCall(callback, context);
+        if (lastScript !== null) {
+          var id = this.addListener('loaded', function(e) {
+            var data = e.getData();
+            // this.debug("Loading " + data.script + " completed");
+            if (data.script == lastScript) {
+              this.removeListenerById(id);
+              cb.schedule();
+            }
+          }, this);
+        }
+        else {
+          cb.schedule();
+        }
+      }
+      this.__loadScripts();
+      return lastScript;
+    },
+
+
+    /**
+     * Chain loading scripts.
+     *
+     * Recursively called until the array of scripts is consumed
+     *
+     * @param codeArr {Array} an array with the uri names of the scripts
+     */
+    __loadScripts: function() {
+
+      if (this.__LOADING === true) {
+        return;
+      }
+
+      var script = this.__QUEUE.shift();
+
+      if (script) {
+        this.__LOADING = true;
+        var uri = qx.util.ResourceManager.getInstance().toUri(script);
+
+        var loader = new qx.bom.request.Script();
+
+        loader.on("load", function(request) {
+          this.__LOADED[script] = true;
+          this.__LOADING = false;
+          this.fireDataEvent('loaded', {
+            script: script,
+            uri: uri,
+            status: request.status
+          });
+          // start the next load event after the current thread has ended
+          var call = new qx.util.DeferredCall(this.__loadScripts, this);
+          call.schedule();
+        }, this);
+
+        loader.on("error", function(request) {
+          this.__LOADING = false;
+          this.fireDataEvent('failed', {
+            script: script,
+            uri: uri,
+            status: request.status
+          });
+        }, this);
+
+        loader.on("timeout", function(request) {
+          this.__LOADING = false;
+          this.fireDataEvent('failed', {
+            script: script,
+            uri: uri,
+            status: request.status
+          });
+        }, this);
+        // this.debug("Loading " + script + " started");
+        loader.open("GET", uri);
+        loader.send();
+      }
+      // no more scripts in codeArr, which means that all the
+      // scripts were loaded.
+      else {
+        // even when we 'fall through' make sure the event gets fired AFTER the current
+        // code is done.
+        var ready = new qx.util.DeferredCall(function(){this.fireEvent("ready")},this);
+        ready.schedule();
+      }
+    }
+  }
+});
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -201,7 +201,9 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
       }
 
       uri = qx.util.ResourceManager.getInstance().toUri(script);
+
       loader = new qx.bom.request.Script();
+
       loader.on("load", function(request) {
         cl.__LOADED[script] = true;
         delete cl.__IN_PROGRESS[script];
@@ -212,19 +214,16 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
         this.__loadScripts();
       },this);
 
-      loader.on("error", function(request) {
-        this.fireDataEvent('failed', {
-          script: script,
-          status: request.status
-        });        
-      }, this);
- 
-      loader.on("timeout", function(request) {
+      var onError = function(request) {
+        delete cl.__IN_PROGRESS[script];
         this.fireDataEvent('failed', {
           script: script,
           status: request.status
         });
-      }, this);
+      };
+
+      loader.on("error", onError,this);
+      loader.on("timeout", onError,this);
 
       // this.debug("Loading " + script + " started");
       loader.open("GET", uri);
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -139,7 +139,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
      */
     start: function() {
       this.start = function(){
-        throw new Error('you can only call start once per instance');
+        throw new Error(this.tr('you can only call start once per instance'));
       }
       this.__loadScripts();
     },
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -138,7 +138,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
      *
      */
     start: function() {
-      if (this.$$disposed) {
+      if (this.isDisposed()) {
         return;
       }
       this.start = function(){
@@ -181,7 +181,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
       if (dynLoader){
 
           id1 = dynLoader.addListener('loaded',function (e) {
-            if (this.$$disposed) {
+            if (this.isDisposed()) {
               return;
             }
             var data = e.getData();
@@ -194,7 +194,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
           },this);
 
           id2 = dynLoader.addListener('failed',function (e) {
-            if (this.$$disposed) {
+            if (this.isDisposed()) {
               return;
             }
             var data = e.getData();
@@ -214,7 +214,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
       loader = new qx.bom.request.Script();
 
       loader.on("load", function(request) {
-        if (this.$$disposed) {
+        if (this.isDisposed()) {
            return;
         }
         cl.__LOADED[script] = true;
@@ -227,7 +227,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
       },this);
 
       var onError = function(request) {
-        if (this.$$disposed) {
+        if (this.isDisposed()) {
            return;
         }
         delete cl.__IN_PROGRESS[script];
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -46,44 +46,29 @@
  *        console.log("Got it");
  *    };
  *    // initialize the script loading
- *    var dynLoader = qx.io.DynamicScriptLoader.getInstance();
+ *    var dynLoader = new qx.util.DynamicScriptLoader();
  *
  *
- *    var readyId = dynLoader.addListenerOnce('ready',function(e){
+ *    dynLoader.addListenerOnce('ready',function(e){
  *      console.log("all scripts have been loaded!");
  *    });
 
  *    dynLoader.addListener('failed',function(e){
  *      var data = e.getData();
- *      dynLoader.removeListenerById(readyId);
  *      console.log("failed to load "+data.script);
  *    });
  *
- *    var lastScript = dynLoader.load([
+ *    dynLoader.load([
  *        "myapp/jquery/jquery"+min+".js",
  *        "myapp/highcharts/highcharts"+src+".js",
  *        "myapp/highcharts/highcharts-more"+src+".js",
  *        "myapp/highcharts/highcharts-modifications.js"
- *    ],callback,this);
- *
- *    if (lastScript === null){
- *      var loadId = loader.addListener('loaded',function(e){
- *        var data = e.getData();
- *        if (data.script == lastScript){
- *          console.log('Last Script Loaded');
- *          dynLoader.removeListenerById(loadId);
- *        }
- *      }, this);
- *    }
- *    else {
- *      console.log('Scripts have already been loaded');
- *    }
+ *    ]);
+ *    
  * </code>
  */
-qx.Class.define("qx.io.DynamicScriptLoader", {
+qx.Class.define("qx.util.DynamicScriptLoader", {
   extend: qx.core.Object,
-
-  type: "singleton",
 
   /**
    * Initialize the state hashes if not yet done
@@ -91,8 +76,6 @@ qx.Class.define("qx.io.DynamicScriptLoader", {
 
   construct: function () {
     this.base(arguments);
-    this.__ADDED = {};
-    this.__LOADED = {};
     this.__QUEUE = [];
     this.__LOADING = false;
   },
@@ -120,6 +103,10 @@ qx.Class.define("qx.io.DynamicScriptLoader", {
     ready: 'qx.event.type.Event'
   },
 
+  statics: {
+    __ADDED: {},
+    __LOADED: {}
+  },
 
   /*
   *****************************************************************************
@@ -130,8 +117,6 @@ qx.Class.define("qx.io.DynamicScriptLoader", {
 
   members: {
 
-    __ADDED: null,
-    __LOADED: null,
     __QUEUE: null,
     __LOADING: null,
 
@@ -147,35 +132,11 @@ qx.Class.define("qx.io.DynamicScriptLoader", {
      *     Returns the name of the last script you can expect to be loaded if all scripts are already loaded null will be returned.
      *
      */
-    load: function(codeArr, callback, context) {
-      var lastScript = null;
+    load: function(codeArr) {
       codeArr.forEach(function(script) {
-        if (this.__ADDED[script] !== true) {
-          this.__QUEUE.push(script);
-          this.__ADDED[script] = true;
-        }
-        if (.this.__LOADED[script] !== true) {
-          lastScript = script;
-        }
-      }, this);
-      if (callback) {
-        var cb = new qx.util.DeferredCall(callback, context);
-        if (lastScript !== null) {
-          var id = this.addListener('loaded', function(e) {
-            var data = e.getData();
-            // this.debug("Loading " + data.script + " completed");
-            if (data.script == lastScript) {
-              this.removeListenerById(id);
-              cb.schedule();
-            }
-          }, this);
-        }
-        else {
-          cb.schedule();
-        }
-      }
+         this.__QUEUE.push(script);
+      },this);
       this.__loadScripts();
-      return lastScript;
     },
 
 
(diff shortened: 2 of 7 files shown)


framework/source/class/qx/io/DynamicScriptLoader.js
```diff
@@ -0,0 +1,245 @@
...
+        loader.open("GET", uri);
+        loader.send();
+      }
+      // no more scripts in codeArr, which means that all the
+      // scripts were loaded.
+      else {
+        // even when we 'fall through' make sure the event gets fired AFTER the current
+        // code is done.
+        var ready = new qx.util.DeferredCall(function(){this.fireEvent("ready")},this);
+        ready.schedule();
```
derrell: And there's no need to defer calling the callback here, either, since we don't have any more scripts to start loading.

framework/source/class/qx/io/DynamicScriptLoader.js
```diff
@@ -0,0 +1,245 @@
...
+            var data = e.getData();
+            // this.debug("Loading " + data.script + " completed");
+            if (data.script == lastScript) {
+              this.removeListenerById(id);
+              cb.schedule();
+            }
+          }, this);
+        }
+        else {
+          cb.schedule();
```
derrell: No more scripts to request, so callback can be called immediately without the need to schedule it.

framework/source/class/qx/io/DynamicScriptLoader.js
```diff
@@ -0,0 +1,245 @@
...
+        }
+      }
+      this.__loadScripts();
```
derrell: __loadScripts can't be a member function. It must be a function within the load() function, so that events are on a per-load() basis, not on a global basis.

derrell (changes requested):

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      var script = this.__QUEUE.shift();
+      if (!script){
+        this.fireEvent("ready")
+        return;
+      }
+
+      if (cl.__LOADED[script]){
+        this.__loadScripts();
```
derrell: fire 'loaded' event here. caller doesn't know nor care whether he is first or 23rd user of this script.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,231 @@
...
+          return;
+      }
+
+      var uri = qx.util.ResourceManager.getInstance().toUri(script);
+      var loader = new qx.bom.request.Script();
+
+      loader.on("load", function(request) {
+        cl.__LOADED[script] = true;
+        delete cl.__ADDED[script];
+        this.fireDataEvent('loaded', {
```
derrell: It looks like 'loaded' may be fired twice? Once here, and once on line 176?

derrell (changes requested):

framework/source/class/qx/io/DynamicScriptLoader.js
```diff
@@ -0,0 +1,245 @@
...
+ *          dynLoader.removeListenerById(loadId);
+ *        }
+ *      }, this);
+ *    }
+ *    else {
+ *      console.log('Scripts have already been loaded');
+ *    }
+ * </code>
+ */
+qx.Class.define("qx.io.DynamicScriptLoader", {
```
oetiker: done

framework/source/class/qx/io/DynamicScriptLoader.js
oetiker: api has changed

framework/source/class/qx/io/DynamicScriptLoader.js
oetiker: api has changed

framework/source/class/qx/io/DynamicScriptLoader.js
```diff
@@ -0,0 +1,245 @@
...
+        loader.on("load", function(request) {
+          this.__LOADED[script] = true;
+          this.__LOADING = false;
+          this.fireDataEvent('loaded', {
+            script: script,
+            uri: uri,
+            status: request.status
+          });
+          // start the next load event after the current thread has ended
+          var call = new qx.util.DeferredCall(this.__loadScripts, this);
```
oetiker: api has changed

framework/source/class/qx/io/DynamicScriptLoader.js
oetiker: api has changed

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: on line 176 you get a loaded event for instances that piggyback on the loading initiated by this instance. here is the original one ... triggering the one in 176

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: ok

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,232 @@
...
+            if (data.script === script){
+              dynLoader.removeListenerById(id2);
+              dynLoader.removeListenerById(id1);
+              this.fireDataEvent('loaded',data);
+              this.__LOADING = false;
+              this.__loadScripts();
+            }
+          },this);
+          var id2 = dynLoader.addListener('failed',function(e){
+            var data = e.getData();
```
derrell: data declared as var twice. This shows up as an error in some editors (like mine) and it would be nice to avoid  that problem when possible. How about declaring dynLoader, id1, id2, data, uri, and loader at the top of the function?

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,232 @@
...
+          status: 'preloaded'
+        });
+        this.__loadScripts();
+        return;
+      }           
+
+      this.__LOADING = true
+
+      var dynLoader = cl.__ADDED[script];
+      if (dynLoader){
```
derrell: Nothing ever gets added to __ADDED. It's only referenced three times in this class, and none of those places assign to __ADDED[script]

derrell (changes requested):

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: oops ... added

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+    failed: 'qx.event.type.Data',
+
+    /**
+     * fired when all given scripts are loaded, each time loadScriptsDynamic is called.
+     */
+    ready: 'qx.event.type.Event'
+  },
+
+  statics: {
+    __ADDED: {},
```
derrell:
```
/**
 * Map of scripts being added at the present time. Key is script name; value is instance of this class which 
 * is loading it. 
 */
```

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+    __LOADED: {}
```
derrell:
```
/**
 * Map of scripts that have fully loaded. Key is script name; value is true
 */
```

framework/source/class/qx/util/DynamicScriptLoader.js
derrell: Might this better be called __IN_PROGRESS instead of __ADDED ?

derrell (changes requested):

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: no, data is declared within a callback function ... 

I have moved the declaration of the other variables to the top of the function ... is that standard practice in qx?

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: done

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: done

framework/source/class/qx/util/DynamicScriptLoader.js
derrell: There is, unfortunately, no standard for that in qooxdoo. Personally, I find it much easier to know what's going on in my function and especially what scope a variable is declared in, if I need only look for it at the top of each function.

oetiker: the good old days of C

derrell (approved):
Looking good. I'm still waiting for @level420 's reply on gitter, regarding whether this implementation meets the needs that his original mixin implementation met, but I think this does.

oetiker:
there is actually a coole effect of this implementation ... you can now implement complex loading scenarios by using multiple loader instances

derrell:
Yeah. Maybe we should have the loader return a Promise? :-)

adeliz:
Hi,
The [callback ](https://github.com/oetiker/qooxdoo/blob/0266b4e61852a50ef94cc976e2459b4f81de0ec1/framework/source/class/qx/util/DynamicScriptLoader.js#L45-L47) function in the example is never used? Should it be removed as well as in the load method's [parameters ](https://github.com/oetiker/qooxdoo/blob/0266b4e61852a50ef94cc976e2459b4f81de0ec1/framework/source/class/qx/util/DynamicScriptLoader.js#L135-L136)?

oetiker:
done

level420:
I think one difference between what I've implemented and @oetiker s version, is that there exists a global queue ```__QUEUE```. I don't think it is really needed if that what is submitted via load in the parameter ```codeArr``` is (recursively) passed to the method ```__loadScripts``` (as it is currently documented).

I'll do a few more comments in the code.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+        this.__loadScripts();
+        return;
+      }           
+
+      this.__LOADING = true
+
+      dynLoader = cl.__IN_PROGRESS[script];
+      if (dynLoader){
+
+          id1 = dynLoader.addListener('loaded',function(e){
```
level420: using ```addListenerOnce``` here would be sufficient, i think. Then it would also be enough to remove then opposite listener. In the loaded listener remove the failed listener and vice versa.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Load scripts dynamically, typically used in the instance constructor
+     *
+     * @param codeArr {Array} an array with the uri names of the scripts
+     *
+     */
+    load: function(codeArr) {
+      codeArr.forEach(function(script) {
+         this.__QUEUE.push(script);
+      },this);
```
level420: I think instead of storing the script array in the ```__QUEUE``` it would be sufficient to call ```__loadScripts``` (recursively) with ```codeArr``` as parameter. But that may be just a matter of taste.

oetiker:
@level420 hmmm interesting idea ... then you could even to parallel loading with a single instance

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+    },
+
+
+    /**
+     * Chain loading scripts.
+     *
+     * Recursively called until the array of scripts is consumed
+     *
+     * @param codeArr {Array} an array with the uri names of the scripts
+     */
```
level420: If the way of storing the scripts in the ```__QUEUE``` member is choosen, then the ```@param``` hint should be removed.

level420 (changes requested):

framework/source/class/qx/util/DynamicScriptLoader.js
level420: In my last changes to my mixin I changed the parameter name from ```codeArr``` to ```scriptArr```. I really don't know where this name came from.

level420: One more comment ;-) But I'll leave the copy from one array to the other. If an array variable is submitted to the ```load``` method, then it is empty after the call.

oetiker: that is why I am copying the array ... but I should simply use the copy util and not do it by hand ... in an earlier version I did other stuff in this loop as well

level420:
And one more idea for the tests: I know I said that it is OK that one test within a test class is depending on the previous test being run. But if the user decides to run the tests multiple times or to run only a single test, then the tests fail.

Maybe a static method ```clearCache``` would help here, which simply deletes ```__LOADED``` would help here to allow to reset the state.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+  extend: qx.core.Object,
+
+  /**
+   * Initialize the state hashes if not yet done
+   */
+
+  construct: function () {
+    this.base(arguments);
+    this.__QUEUE = [];
+    this.__LOADING = false;
```
level420: I think ```__LOADING``` is only used to assign ```false``` or ```true``` to, but never queried.

level420:
Removed wrong comment on ```__LOADING``` not being used.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,241 @@
...
+    __loadScripts: function() {
+      var cl = qx.util.DynamicScriptLoader;
+      var script;
+      var dynLoader;
+      var id1, id2;
+      var uri;
+      var loader;
+
+      if (this.__LOADING === true) {
+        return;
```
level420: What if we reach here while we are loading a script? Then the loading process is simply stopped?
Or can this not happen for some reason?

oetiker: this would happen if one called the load method twice ... BUT I just pushed update ... now there is actually parallel loading with a single instance by calling load repeatedly

oetiker: it may be a bit confusing though since you will then get multiple ready events ... so I am not quite sure if this is a good change

oetiker: maybe I should revert to the queue design ....

oetiker:
back to the __QUEUE ...

framework/source/class/qx/util/DynamicScriptLoader.js
level420: Maybe the ```load``` method should test ```if(this.__LOADING)``` and return false if this is the case, to indicate that the loader is currently not accepting another request. And of course return true to signal loading was accepted and started.

level420: Hmm. I think I still do not understand how the ```this__LOADING``` flag is used and what it is intended for. Tobi can you please explain a bit? thank you!

derrell: __LOADING looks to be preventing multiple calls to load() from starting a new script loading. I'm wondering if load() is even needed? How about if the constructor took the list of scripts to be loaded and deferred the call to __loadScripts? __LOADING would be unnecessary, and an extra, really unnecessary call to load() would be avoided.

level420: @derrell as ```this.__LOADING``` is set to false and true within one load process, I just could not find out if the situation of ```this.__LOADING === true``` can occur while the load process is running.

level420: @derrell regarding omitting the ```load``` call: we need to first assign an event listener and then initiate the loading. So we could maybe have the ```scriptArray``` as a parameter in the constructor and a ```load``` call which does not need any parameters. This way we can not start multiple loads with different script sets, which would avoid some potential problems like parallel initiated loading.

Additionally the loader could hold an "I'm already done" flag, which leads to subsequent calls to ```load``` returning immediately with false.

derrell: The first call to load() causes __loadScripts to be called. When a script name to be loaded is found, that is not already loaded, __LOADING is set to true, and loading for that script (and, subsequently, any following not-previously-loaded ones on the list) is begun. If a new call to load() comes in on this same instance of DynamicScriptLoader, with additional scripts to be loaded, __loadScripts will be called and find __LOADING is already true, so those scripts which were added to the queue will be loaded when the first instance gets around to them. This new load() call can and does return immediately.

There is a bug, however. __LOADING is set to false in the "loaded" and "failed" handlers, whereas it should only be set to false in the "ready" handler (or rather, when there is nothing left on the queue, whether all scripts successfully loaded or not).

Again, though, if we got rid of the load() method and passed the list of scripts to be loaded directly to the constructor, this whole issue (and the __LOADING member) would disappear.

level420: Yes but how do we assign the "ready" event listener if we do not separate instantiation and the call to load?

derrell: That's exactly why the call to __loadScripts must be deferred.

derrell: You'd instantiate the loader and set its ready handler (and any others). Since __loadScripts is deferred, it won't run until after the caller returns to the JavaScript engine.

oetiker: hihi ... yes ... remember my original code? that was exactly why I did that initially ... but you are right, by not using the load call the whole calling it twice problem can be avoided

derrell: Alternatively (and yes, I jest):

```
load: function(scriptArr) {
     scriptArr.forEach(function(script) {
        this.__QUEUE.push(script);
     },this);
     this.__loadScripts();

     // only allow a single call to load()
     this.load = function() { throw new Error("load() may only be called once."); }
   },
```

oetiker: and if load is called repeatedly it just gets ignored ?

derrell: I think you're right. The implicit style isn't commonly used, so would be unexpected. I think maybe just fix __LOADING to prevent __loadScripts() from being called while it's already running is the right way to do it.

derrell: If load() doesn't throw or return an error, it should just add stuff to the queue and then exit (like it appears it's trying to do now)

level420: ... and we may end up having a singleton again:
```javascript
qx.util.DynamicScriptLoaderManager.getInstance().load( ['s1.js', 's2.js' ], this.__onLoaderReady, this );
```
which creates internally on each load call a new instance of ```qx.util.DynamicScriptLoader```

derrell: To fix the current code, just ensure that __LOADING is set to false only when the queue is finally empty; not on each "loaded" and "failed" method.

derrell: The problem with passing onLoaderReady to either the constructor or to load() is that we lose the individual events for each file loaded, which could be useful -- particularly the "failed" events. If we do it this way, onLoaderReady would need to be passed a list of scripts that failed to load, and a list of scripts that succeeded, or some such thing. Sounds like too much record keeping internally, for my taste.

oetiker: lets go for 

* add array on create
* but call load() to initiate loading
* if load is called a second time, it gets ignored

oetiker: or maybe 'start' instead of 'load'

level420: OK then I'd vote for the call to load() to only be possible once and the script set to be passed to the constructor.
Multiple calls to load should lead to a warning and return false, while the first call should return true.
We could also rename load to execute or init  or something that implies the "single call" character of that method.

level420: yes start is good!

level420: We could also delete the load or start method from the instance after the first call ;-)
Burn after using.

derrell: start() is good, and I'd go farther than just returning false if start() has already been called. It's definitely a programmer error if it's called more than once, so I'd throw an error.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,233 @@
...
+ *    
+ * </code>
+ */
+qx.Class.define("qx.util.DynamicScriptLoader", {
+  extend: qx.core.Object,
+
+  /**
+   * Initialize the state hashes if not yet done
+   *
+   * @param scriptArr {Array} an array with the uri names of the scripts
```
derrell: Simple usage will be a single script, so let's allow that:

Array|String

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,233 @@
...
+     * Array of the scripts to be loaded
+     */
+    __QUEUE: null,
+
+    /**
+     * Start loading scripts. This may only be called once!
+     *
+     */
+    start: function() {
+      this.load = function(){
```
level420: shouldn't this be ```this.start = function() {```

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,233 @@
...
+  extend: qx.core.Object,
+
+  /**
+   * Initialize the state hashes if not yet done
+   *
+   * @param scriptArr {Array} an array with the uri names of the scripts
+   */
+
+  construct: function (scriptArr) {
+    this.base(arguments);
```
derrell:
```
this.__QUEUE = (typeof strings == "string"
                ? [ strings]
                : qx.lang.Array.clone(scripts));
```

level420: or even better:

```javascript
this.__QUEUE = (qx.lang.Type.isString(scriptArr)
                ? [ scripts ]
                : qx.lang.Array.clone(scripts));
```

framework/source/class/qx/util/DynamicScriptLoader.js
derrell: Agree. This looks like it should be this.start, not this.load.

derrell (changes requested):
s/this.load =/this.start =/

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,235 @@
...
+        cl.__LOADED[script] = true;
+        delete cl.__IN_PROGRESS[script];
+        this.fireDataEvent('loaded', {
+          script: script,
+          status: request.status
+        });
+        this.__loadScripts();
+      },this);
+
+      loader.on("error", function(request) {
```
level420: I think here is missing
```javascript
delete cl.__IN_PROGRESS[script];
```

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,235 @@
...
+        this.fireDataEvent('failed', {
+          script: script,
+          status: request.status
+        });        
+      }, this);
+ 
+      loader.on("timeout", function(request) {
```
level420: I think here is missing
```javascript
delete cl.__IN_PROGRESS[script];
```

thron7:
Without being into all the nitty-gritty of dynamic loading, let me just point you to `qx.Part`. While geared towards part loading, it solves all the underlying problems: loading of a list of script files, in parallel, then evaluate them in order so dependencies resolve properly. The hook for houskeeping is `qx.$$loader.closureParts`, the wrapping of the scripts into `qx.Part.$$notifyLoad()` is done in the Generator (`CodeGenerator.py`). -- I understand that for arbitrary scripts the wrapping is not feasible as they are not handled by the compiler. But this makes loading them *soo* much easier, so you might want to think of something else to achieve the same effect (e.g. wrapping on the client side, or some `jsonp`-style request support).

framework/source/class/qx/util/DynamicScriptLoader.js
oetiker: no, since all load events trigger here ... I remove it once the script I am waiting for has been loaded.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,234 @@
...
+        delete cl.__IN_PROGRESS[script];
+        this.fireDataEvent('loaded', {
+          script: script,
+          status: request.status
+        });
+        this.__loadScripts();
+      },this);
+
+      var onError = function(request) {
+        delete cl.__IN_PROGRESS[script];
```
level420: maybe add here ```cl.__LOADED[script] = false;``` to avoid trying to load the script next time again?

This would also need to distinguish the two states false and true in the pre-test if loading needs to be started:
```javascript
      if (cl.__LOADED[script] === true){
         this.fireDataEvent('loaded',{
           script: script,
           status: 'preloaded'
         });
         this.__loadScripts();
         return;
       }
       if (cl.__LOADED[script] === false){
         this.fireDataEvent('failed',{
           script: script,
           status: 'failed'
         });
         return;
       }
```

oetiker: yea the whole strategy for dealing with failed loads ... my idea for now is that I want to allow the code to try again ... maybe a script failed to load due to a temporary condition on the server ...

level420: OK.

level420:
Ther testrunner currently reports an error in the tests: see https://travis-ci.org/qooxdoo/qooxdoo/jobs/163846442

It's related to the introduction of ```this.tr``` in the exception code. I think ```this.tr``` should be removed again. It's meant for the developer and not for the end user.

level420 (approved):

cajus (commented):
The DynamicScriptLoader class shoud remove some references on _destruct_, so that pending events received after dispose will not run into errors.

I'm not sure wheter @johnspackman's changes remove the need for assigning __QUEUE & Co with null, too.

derrell (approved):

level420:
@cajus So the edge case is that someone disposes actively a loader during pending requests

cajus:
Yes. That's the edge case I meant.

derrell (changes requested):
See @cajus comment. There are three listeners (id1, id2, loader), at least, that should stop listening if this object is disposed. (It would be very strange for the caller to dispose this instance while loading is in progress, but we should still handle it.)

level420:
Even if I'm able to catch all the problems resulting from a loader being disposed during requests, there is still the ```qx.bom.request.Script``` instance which will fire its events and run the corresponding listeners.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,234 @@
...
+          },this);
+
+          return;
+      }
+
+      uri = qx.util.ResourceManager.getInstance().toUri(script);
+
+      loader = new qx.bom.request.Script();
+
+      loader.on("load", function(request) {
```
derrell: Insert here:
```
if (this.__disposed) {
  return;
}
```

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,234 @@
...
+
+      loader.on("error", onError,this);
+      loader.on("timeout", onError,this);
+
+      // this.debug("Loading " + script + " started");
+      loader.open("GET", uri);
+      cl.__IN_PROGRESS[script] = this;
+      loader.send();
+    }
+  }
```
derrell: Add here:
```
destruct : function()
{
  this.__disposed = true;
}

File: framework/source/class/qx/util/DynamicScriptLoader.js [hunk 25]
```diff
@@ -0,0 +1,234 @@
...
+  */
+
+
+  members: {
+
+    /**
+     * Array of the scripts to be loaded
+     */
+    __QUEUE: null,
+
```
derrell: Add here:
```
/**
 * Set to true if this object is disposed, to prevent calling event handlers
 */
__disposed : false,

File: framework/source/class/qx/util/DynamicScriptLoader.js [hunk 26]
```diff
@@ -0,0 +1,234 @@
...
+          status: 'preloaded'
+        });
+        this.__loadScripts();
+        return;
+      }           
+
+      dynLoader = cl.__IN_PROGRESS[script];
+      if (dynLoader){
+
+          id1 = dynLoader.addListener('loaded',function(e){
```
derrell: Insert here:
```
if (this.__disposed) {
  return;
}

File: framework/source/class/qx/util/DynamicScriptLoader.js [hunk 27]
```diff
@@ -0,0 +1,234 @@
...
+            var data = e.getData();
+            if (data.script === script){
+              dynLoader.removeListenerById(id2);
+              dynLoader.removeListenerById(id1);
+              this.fireDataEvent('loaded',data);
+              this.__loadScripts();
+            }
+          },this);
+
+          id2 = dynLoader.addListener('failed',function(e){
```
derrell: Insert here:
```
if (this.__disposed) {
  return;
}

**Review by derrell — CHANGES_REQUESTED** [review]
See insert/add here messages which use a __disposed flag to specify whether to process listeners after an instance of the dynamic script loader has been disposed.

File: framework/source/class/qx/util/DynamicScriptLoader.js [hunk 28]
```diff
@@ -0,0 +1,234 @@
...
+    __QUEUE: null,
+
+    /**
+     * Start loading scripts. This may only be called once!
+     *
+     */
+    start: function() {
+      this.start = function(){
+        throw new Error('you can only call start once per instance');
+      }
```
derrell: Insert here, to prevent starting a new load after disposing the object:
```
if (this.__disposed) {
  return;
}

File: framework/source/class/qx/util/DynamicScriptLoader.js (hunk 23 above)
Comment (cajus) [inline framework/source/class/qx/util/DynamicScriptLoader.js]: Or just use

```
if (this.$$disposed) {
  return;
}
```

File: framework/source/class/qx/util/DynamicScriptLoader.js (hunk 28 above)
Comment (derrell) [inline framework/source/class/qx/util/DynamicScriptLoader.js]: Could add an error here if we wanted, to, or redefine this.start to throw an error in the destructor function.

File: framework/source/class/qx/util/DynamicScriptLoader.js [hunk 29]
```diff
@@ -177,7 +180,10 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
       dynLoader = cl.__IN_PROGRESS[script];
       if (dynLoader){
 
-          id1 = dynLoader.addListener('loaded',function(e){
+          id1 = dynLoader.addListener('loaded',function (e) {
+            if (this.$$disposed) {
```
derrell: $$disposed is undocumented and should really be considered private. However, there is a public method for accessing whether the object is disposed. Let's change these to:
```
if (this.isDisposed()) {
```

derrell (approved):

cajus (approved):

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,258 @@
...
+
+    /**
+     * Start loading scripts. This may only be called once!
+     *
+     */
+    start: function() {
+      if (this.isDisposed()) {
+        return;
+      }
+      this.start = function(){
```
johnspackman: please can you stored a `this.__started` flag and test for it instead of changing the `start` function; although the started state cannot be turned off, it's unexpected behaviour for a declared method to suddenly point to a non-method.

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -0,0 +1,258 @@
...
+
+
+    /**
+     * Chain loading scripts.
+     *
+     * Recursively called until the array of scripts is consumed
+     *
+     */
+    __loadScripts: function () {
+      var cl = qx.util.DynamicScriptLoader;
```
johnspackman: please can `cl` be renamed to `DynamicScriptLoader` (or at least be capitalised as `DSL`) to remind the reader that it's being treated as a specific, 'constant' static class rather than a variable

derrell: @oetiker is away on vacation. May I suggest that we accept this PR as is, and then make @johnspackman 's recent two modifications. (I agree with them, particularly this one about cl vs DynamicScriptLoader. I should have caught that too.)

level420: Yes, please accept it as is and create a new PR afterwards.

johnspackman (approved):

patrick-fischer-hirschstein (approved):

level420:
Voting label added 6 days ago.
4 approvals.
Merging.
Thank you @oetiker

PR #9200, opened 2016-10-06
Added missing '@' in doc strings

### LINKED ISSUE #9199: qx.Annotation partly lacks parameter description
If one inspects http://www.qooxdoo.org/devel/api/#qx.Annotation, it is marked with a yellow "!" and most of the contained methods are marked with a red "x". Should be updated before a release IMHO.

cajus:
fixes #9199


level420 (approved):

johnspackman (approved):

level420:
Doc only fix. Two approvals. Merging.
Thank you @cajus

Issue #9204, opened 2016-10-06
Filtered table not working after filling the table with addRows


peuter:
Please have a look at the following playground example: http://tinyurl.com/zc6xg34

Initially you can activate/deactivate the filtering by clicking the filter-button. But after cklicking the refresh-button, the filtering is broken (table is empty).

The problem is, that by resetting the tableModel with `setData([])` the initial `__fullArr` gets resetted too, but when you fill the tableModel by using addRows/setRows and not setData `__fullArr` does not get updated.

A possible solution would be to remove the overridden setData method from qx.ui.table.model.Filtered. The described problem is fixed then, but I don't know if it has any unwanted impact.

Also I was wondering what would happen if a row was added to the tableModel (by using addRows) when a filter is active. After deactivating the filter, the tableModel would be replaced with `__fullArr` which does not know about the new row?

peuter:
Forget about the mentioned solution as creates more errors than it solves

oetiker:
I think the best thing to be done about the filtered table model is to mark is as deprecated, and advise people to use a different implementation technique as discussed a few months ago ...

derrell:
I would tend to agree. It was contributed and used (extensively, I think) by the contributor, but not well used otherwise, and has had a number of problems reported over the many years since it was included, with little additional attention paid to it by the contributor nor anyone else that I know of.

peuter:
Can you give me a hint about the different implementation technique you are talking about?

level420:
@woprandi could you help here?

woprandi:
@peuter You should use filter on your data array. I modified your example (line 45) : http://tinyurl.com/hedgrr9

peuter:
@woprandi Thank you!

level420:
@peuter can we close this issue?

peuter:
Sure

carlosfvieira:
Hi @woprandi ,

How exactly this works differently from using the Filtered model? Will you not be resetting initial data also? Leading to the same problem stated on the creation of this issue?

Can you provide an example? The provided link on quote is now broken...

Thanks

>@peuter You should use the Array.filter method on your data array. I modified your example (line 45) : http://tinyurl.com/hedgrr9

PR #9206, opened 2016-10-07
Set qx.ui.table.model.Filtered class as deprecated

woprandi:
For https://github.com/qooxdoo/qooxdoo/issues/9204

### DIFF
--- a/framework/source/class/qx/ui/table/model/Filtered.js
+++ b/framework/source/class/qx/ui/table/model/Filtered.js
@@ -68,6 +68,10 @@ search.addListener("changeValue", function(e) {
 
 this.getRoot().add(search, {top : 500, left : 10});
 </pre>
+ *
+ * @deprecated {6.0} You should use
+ * <a href='http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter' target='_blank'>Array.filter</a>
+ * method to filter the table model
  */
 qx.Class.define("qx.ui.table.model.Filtered",
 {


patrick-fischer-hirschstein (approved):

johnspackman (changes requested):
please can you restore the documentation in the comments?  Although this class is being deprecated, there's no need to loose the documentation or otherwise modify it

woprandi:
@johnspackman you're right

woprandi:
@johnspackman updated

johnspackman (approved):
thank you :)

cajus (approved):

level420 (approved):

framework/source/class/qx/ui/table/model/Filtered.js
```diff
@@ -68,6 +68,10 @@ search.addListener("changeValue", function(e) {
 
 this.getRoot().add(search, {top : 500, left : 10});
 </pre>
+ *
+ * @deprecated {6.0} You should use
+ * <a href='http://developer.mozilla.org/docs/Web/JavaScript/Reference/Objets_globaux/Array/filter' target='_blank'>Array.filter</a>
```
level420: The Link to the mozilla developper network does not seem to work

level420 (changes requested):

framework/source/class/qx/ui/table/model/Filtered.js
woprandi: @level420 it works for me

johnspackman: doesnt work for me either

level420: @woprandi the "Objet_globaux" part int he URL seems to be for the french link. My link gets rewritten to start with https://developer.mozilla.org/de/... (german) and is suspect @johnspackman s link gets rewritten to https://developer.mozilla.org/en/ . So wie need the full englisch path in the URL

johnspackman: I think you meant https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter - it looks like MDN localises the URLs and unless the locale is specified explicitly it's not reliable

woprandi: @level420 Oh sorry, I'll fix that

level420 (approved):

level420:
Documentation only fix. Merging.
Thank you @woprandi

PR #9211, opened 2016-10-13
Variable item height rendering for qx.ui.list.List

level420:
derived from a gist shared by @eggnaube.

Adds property `variableItemHeight` which, when set to true, caculates for each list item an individual height.

### DIFF
--- a/framework/source/class/qx/ui/list/List.js
+++ b/framework/source/class/qx/ui/list/List.js
@@ -260,6 +260,19 @@ qx.Class.define("qx.ui.list.List",
       event: "changeGroups",
       nullable: false,
       deferredInit: true
+    },
+
+
+    /** 
+     * Render list items with variable height, 
+     * calculated from the individual item size. 
+     */
+    variableItemHeight :
+    {
+      check : "Boolean",
+      apply : "_applyVariableItemHeight",
+      nullable : false,
+      init : true
     }
   },
 
@@ -317,6 +330,8 @@ qx.Class.define("qx.ui.list.List",
     __defaultGroupUsed : false,
 
     __defaultGroups : null,
+
+    __deferredLayerUpdate : null,
 
 
     /**
@@ -379,6 +394,7 @@ qx.Class.define("qx.ui.list.List",
     _initLayer : function()
     {
       this._layer = this._provider.createLayer();
+      this._layer.addListener("updated", this._onLayerUpdated, this);
       this.getPane().addLayer(this._layer);
     },
 
@@ -565,6 +581,18 @@ qx.Class.define("qx.ui.list.List",
     },
 
 
+    // property apply
+    _applyVariableItemHeight : function(value, old) {
+      if(value) {
+        this._setRowItemSize();
+      }
+      else {
+        this.getPane().getRowConfig().resetItemSizes();
+        this.getPane().fullUpdate();
+      }
+    },
+
+
     /*
     ---------------------------------------------------------------------------
       EVENT HANDLERS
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/list/List.js
```diff
@@ -565,6 +581,18 @@ qx.Class.define("qx.ui.list.List",
     },
 
 
+    // property apply
+    _applyVariableItemHeight : function(value, old) {
+      if(value) {
```
oetiker: how about adding
```
if ( value == old){
   return;
}
```

framework/source/class/qx/ui/list/List.js
```diff
@@ -597,6 +625,28 @@ qx.Class.define("qx.ui.list.List",
...
+
+      if (this.__deferredLayerUpdate === null) {
+        this.__deferredLayerUpdate = new qx.util.DeferredCall(function () {
+          this._setRowItemSize();
+        }, this);
+      }
+
+      this.__deferredLayerUpdate.schedule();
+    },
+
```
oetiker: if there are multiple layer updates before the event loop runs again, the setrowitemsize would run multiple times in sequence? or can a deferred call only be scheduled once ? additional scheduled calls before it is executed get ignored ?

oetiker (changes requested):

framework/source/class/qx/ui/list/List.js
cajus: Property apply should not be called when value == old. Except you do it manually.

framework/source/class/qx/ui/list/List.js
johnspackman: no, only the once.  `DeferredCall.schedule` puts the call on a queue, but the queue is not serviced until the event loop runs.  If you call `.schedule` several times, then nothing happens if it's already on the queue so it will only get called once when the event loop is freed up

johnspackman (approved):

framework/source/class/qx/ui/list/List.js
level420: Yes! The property system only calls the applyer if the property value changed

oetiker (approved):

cajus (approved):

slachtar:
Great addition.
Does it work for customized list items?

level420:
@slachter in my use case the list item is a widget based on ```new qx.ui.container.Composite()``` with a vbox layout containing three child widgets, an atom, a label and a checkbox, where the label contains html which has a variable height. Works great!

framework/source/class/qx/ui/list/List.js
derrell: @johnspackman Are you sure about that in this case? DeferredCall would only put the identical function reference onto its schedule queue once. This function, however, is anonymous. It could be a non-unique reference on each occurrence, the same (but not identical reference) function to be called multiple times. Unless you're sure, we should verify it.

derrell: Disregard. I missed the fact that it's only creating a single DeferredCall (if deferredLayerUpdate is null). The schedule call itself only actually schedules once.

However, is there a reason not to move the schedule() call into the _if_ block so it only gets called once?

johnspackman: Yep, I'm sure - the `DeferredCall` puts the instance of `DeferredCall` on the queue in `DeferredCallManager`, not the function; that makes it unique.

johnspackman: @derrell oops, sorry crossed in the post there.  The `schedule` needs to stay where it is because it needs to be called each time there is an update

derrell (approved):
LGTM

PR #9214, opened 2016-10-18
Support for cooperative concurrency

cajus:
This pull requests implements the _cooperative concurrency_ (draft) detailed in https://w3c.github.io/requestidlecallback/. Please note that it is currently supported only in Chrome (>=47). Firefox support is on its way. The PR takes care about this fact and emulates the behaviour using traditional _setTimeout()_ calls and a pre-defined time bucket.

So - what is it good for? Think about a huge data model that needs some client side processing when read. Processing everything in a linear manner might lead to browser "page is not responsive" popups. So you'll end up splitting stuff into smaller chunks and process them in _setTimeout()_ calls.

That has the disadvantage that you cannot use the time where the browser is IDLE, because _setTimeout()_ has a granularity variing from browser to browser. That's what the W3C draft is good for.

Example:

```
var process = function(deadline) {
  while (deadline.timeRemaining() > 0 && stuffToDo()) {
    doChunkedStuff();  
  }

  if (stuffToDo()) {
    qx.bom.IdleCallback(process);  
  }
};

qx.bom.IdleCallback(process);
```

So the browser will tell you when there's time to do stuff without influencing the rendering. And - in opposite to _setTimeout()_ - it is called as soon as possible and gives you an estimated time frame to do as much as possible in that.

### DIFF
--- a/framework/source/class/qx/bom/IdleCallback.js
+++ b/framework/source/class/qx/bom/IdleCallback.js
@@ -22,13 +22,13 @@
  * https://www.w3.org/TR/requestidlecallback/
  *
  * Here is a sample usage:
- * <pre class='javascript'>var clb = function(deadline) {
+ * <pre class='javascript'>var cb = function(deadline) {
  *   while (deadline.timeRemaining() > 0) {
  *     // ... do some last tasks
  *   }
  * };
  *
- * qx.bom.IdleCallback.request(clb, this);
+ * qx.bom.IdleCallback.request(cb, this);
  * </pre>
  */
 qx.Bootstrap.define("qx.bom.IdleCallback",
@@ -38,12 +38,16 @@ qx.Bootstrap.define("qx.bom.IdleCallback",
   statics :
   {
     /**
-     * The default time in ms the timeout fallback implementation uses.
+     * The default time in ms the setTimeout() fallback implementation uses. There is no
+     * way to detect if the browser has some spare IDLE time, it defaults to zero - meaning
+     * that the callback will be called as soon as possible.
      */
     TIMEOUT : 0,
 
     /**
-     * The default remaining time in ms the timeout fallback implementation uses.
+     * The default remaining time in ms the timeout fallback implementation uses. Since we
+     * cannot know how much time is available, this is a hard coded time bucket available
+     * for actions done in the callback.
      */
     REMAINING : 250,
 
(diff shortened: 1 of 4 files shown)


framework/source/class/qx/bom/IdleCallback.js
```diff
@@ -0,0 +1,109 @@
...
+
+  statics :
+  {
+    /**
+     * The default time in ms the timeout fallback implementation uses.
+     */
+    TIMEOUT : 0,
+
+    /**
+     * The default remaining time in ms the timeout fallback implementation uses.
```
derrell: I've read through all of the code and I'm still not sure what this is. Can the documentation provide some more detail for REMAINING and TIMEOUT here?

framework/source/class/qx/bom/IdleCallback.js
```diff
@@ -0,0 +1,109 @@
...
+        return callback.call(context, deadline);
+      };
+
+      if (qx.core.Environment.get("client.idle")) {
+        return window.requestIdleCallback(clb, timeout);
+      }
+      else {
+
+        var deadline = {
+          started : +(new Date()),
```
derrell: Or, more clearly (without the type conversion): `(new Date()).getTime()` which is, granted, a tiny bit slower, but makes it much more clear what the result it.

framework/source/class/qx/bom/IdleCallback.js
```diff
@@ -0,0 +1,109 @@
...
+
+          timeRemaining : function() {
+            var now = +new Date();
```
derrell: ditto

framework/source/class/qx/bom/IdleCallback.js
```diff
@@ -0,0 +1,109 @@
...
+     *   object. It contains a <code>timeRemaining()</code> call which returns the
+     *   remaining milliseconds and the <code>didTimeout</code> flag which indicates
+     *   whether the callback was fired due to a timeout. The latter is always false
+     *   in case of the emulation.
+     * @param context {var} The context of the callback.
+     * @param timeout {Number} Timeout in milliseconds.
+     * @return {Number} Handle for that request
+     */
+    request : function(callback, context, timeout) {
+      var clb = function(deadline) {
```
derrell: Since you need a name other than `callback`, the most commonly-seen abbreviation for it is `cb`. I had to spend some time figuring out what `clb` meant. Suggest changing it to `cb`.

derrell (changes requested):
This will be nice to have. I wasn't familiar with requestIdleCallback. Once this is merged, I should look at qx.ui.progressive.Progressive to see if it could take advantage of this.

framework/source/class/qx/bom/IdleCallback.js
cajus: That's just copy'n paste like it currently is in qooxdoo. Taken from *qx.bom.AnimationFrame* ;-) But I can change that, sure.

framework/source/class/qx/bom/IdleCallback.js
cajus: Same here. It was used that way in qooxdoo, so I just made the same use of it.

cajus:
Something seems to break travis, while it runs locally. Needs inspection.

derrell (approved):

level420 (approved):

johnspackman (approved):

zaucker (approved):

level420:
merging
thank you @cajus

PR #9216, opened 2016-10-19
Trap error from RPC callback rather than calling callback with its own error

derrell:
This should fix the problem that @oetiker identified, where the RPC callback function threw an error, which caused that callback function to be called again.

### DIFF
--- a/framework/source/class/qx/io/remote/Rpc.js
+++ b/framework/source/class/qx/io/remote/Rpc.js
@@ -528,7 +528,6 @@ qx.Class.define("qx.io.remote.Rpc",
       var argsArray = [];
       var eventTarget = this;
       var protocol = this.getProtocol();
-      var _this = this;
 
       for (var i=offset+1; i<args.length; ++i)
       {
@@ -575,7 +574,7 @@ qx.Class.define("qx.io.remote.Rpc",
             }
             catch(e)
             {
-              _this.error(
+              eventTarget.error(
                 "rpc handler threw an error:" +
                   " id=" + id +
                   " result=" + JSON.stringify(result) +
--- a/framework/source/class/qx/io/remote/Rpc.js
+++ b/framework/source/class/qx/io/remote/Rpc.js
@@ -568,7 +568,18 @@ qx.Class.define("qx.io.remote.Rpc",
             break;
 
           case 1: // async with handler function
-            handler(result, ex, id);
+            try
+            {
+              handler(result, ex, id);
+            }
+            catch(e)
+            {
+              console.error(
+                "rpc handler threw an error:" +
+                  " id=" + id +
+                  " result=" + JSON.stringify(result) +
+                  " ex=" + JSON.stringify(ex));
+            }
             break;
 
           case 2: // async with event listeners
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -568,7 +568,18 @@ qx.Class.define("qx.io.remote.Rpc",
...
 
           case 1: // async with handler function
-            handler(result, ex, id);
+            try
+            {
+              handler(result, ex, id);
+            }
+            catch(e)
+            {
+              console.error(
```
johnspackman: shouldn't this be this.error instead of console?  Im not sure that console is always available (although that may be an IE8 thing Im remembering), but this.error allows for filtering and other outputs

johnspackman (changes requested):

framework/source/class/qx/io/remote/Rpc.js
oetiker: true

oetiker:
I have not looked at the code ... my original Assumtion was, that there might be a `try` somewhere way up in the call chain ... and that this might actually affect other scenarios too

framework/source/class/qx/io/remote/Rpc.js
derrell: Yes, but 'this' isn't what we'd think it is. I could use .bind(this) on the internal function, but instead, I just assigned the context to a separate variable. Thanks.

derrell:
Dumb. Let me fix. There's already a context variable.

derrell:
There is no other try/catch in the Rpc code, and I don't believe there is any farther up within the qooxdoo code. The app using Rpc may have try/catch...

oetiker:
that would mean that my app is calling its own code ... I guess we need a sample :) ...

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -568,7 +568,18 @@ qx.Class.define("qx.io.remote.Rpc",
...
+                "rpc handler threw an error:" +
+                  " id=" + id +
+                  " result=" + JSON.stringify(result) +
+                  " ex=" + JSON.stringify(ex));
```
level420: I think ```qx.lang.Json.stringify``` should be used instead.

level420 (changes requested):

framework/source/class/qx/io/remote/Rpc.js
derrell: Is there ANY reason to ever use qx.lang.Json.* any longer? Is there any supported browser that still doesn't support the JSON object's methods?

level420: Well it is used in a lot of places within the framework and in this file as well: https://github.com/derrell/qooxdoo/blob/27fea506e107f0564981a67bab7cd6254e5b12ed/framework/source/class/qx/io/remote/Rpc.js#L811

derrell: Ok. So it should be a decision to do a large-scale clean-up of qx.lang.Json, but for now, I'll make that change here and defer the large-scale clean-up for another time.

level420 (approved):

cajus (approved):

johnspackman (approved):

oetiker (approved):

zaucker (approved):

level420:
5 approvals
merging
thank you @derrell

PR #9223, opened 2016-10-30
Enhance Window: add autoDestroy and autoCenter properties

derrell:
If the autoDestroy property is true, automatically remove a window from its root container, remove all listeners and contents, and dispose the window, when it is closed.

If the autoCenter property is true, automatically center the window when it is opened.

There is no change to prior behavior, by default, with this PR. Both autoDestroy and autoCenter default to false.

### DIFF
--- a/framework/source/class/qx/ui/window/Window.js
+++ b/framework/source/class/qx/ui/window/Window.js
@@ -91,9 +91,6 @@ qx.Class.define("qx.ui.window.Window",
     // Focusout listener
     this.addListener("focusout", this._onWindowFocusOut, this);
 
-    // Initialize the centerWhen property to its own empty array
-    this.setCenterWhen([]);
-
     // Automatically add to application root.
     qx.core.Init.getApplication().getRoot().add(this);
 
@@ -386,59 +383,24 @@ qx.Class.define("qx.ui.window.Window",
     ---------------------------------------------------------------------------
     */
 
-    /** 
-     * When should the window be automatically centered
-     * 
-     * This property value is formed from the following strings.
-     *
-     *   "appear" - Center the window when it appears (or reappears)
-     *   "resize" - Center the window when its parent container is resized
-     * 
-     * An empty array removes all automatic centering.
-     * 
-     * Otherwise, the property value should be an array containing one or more
-     * of the above strings.
-     * 
-     * Examples:
-     *   Center upon appear and resize:
-     *     win.setCenterWhen([ "appear", "resize" ]);
-     * 
-     *   Center only upon appear:
-     *     win.setCenterWhen([ "appear" ]);
-     * 
-     *   Remove automatic centering:
-     *     win.setCenterWhen([]);
-     */
-    centerWhen :
+    /** Whether this window should be automatically centered when it appears */
+    centerOnAppear :
     {
-      init : null,              // initialized in constructor
-      nullable : false,
-      apply : "_applyCenterWhen",
-      check : function(value)
-      {
-        var i;
-        var allowable = [ "appear", "resize" ];
-
-        // Value must be an array
-        if (! (value instanceof Array))
-        {
-          return false;
-        }
-
-        // The array must contain only allowable values
-        for (i = 0; i < value.length; i++)
-        {
-          if (allowable.indexOf(value[i]) == -1)
-          {
-            // Found a value that is not in our allowable list
-            return false;
-          }
-        }
-
-        return true;
-      }
+      init  : false,
+      check : "Boolean",
+      apply : "_applyCenterOnAppear"
     },
 
+    /** 
+     * Whether this window should be automatically centered when its container
+     * is resized.
+     */
+    centerOnContainerResize :
+    {
+      init  : false,
+      check : "Boolean",
+      apply : "_applyCenterOnContainerResize"
+    },
 
 
 
@@ -548,7 +510,7 @@ qx.Class.define("qx.ui.window.Window",
       this.base(arguments, parent);
 
       // Re-add a listener for resize, if required
-      if (parent && this.getCenterWhen().indexOf("resize") != -1)
+      if (parent && this.getCenterOnContainerResize())
       {
         this.__centeringResizeId =
           parent.addListener("resize", this.center, this);
@@ -1058,8 +1020,7 @@ qx.Class.define("qx.ui.window.Window",
       }
     },
 
-    // overridden
-    _applyCenterWhen : function(value, old)
+    _applyCenterOnAppear : function(value, old)
     {
       var             parent = this.getLayoutParent();
 
@@ -1069,20 +1030,25 @@ qx.Class.define("qx.ui.window.Window",
         this.__centeringAppearId = null;
       }
 
+      // If we are to center on appear, arrange to do so
+      if (value) {
+        this.__centeringAppearId =
+          this.addListener("appear", this.center, this);
+      }
+    },
+
+    _applyCenterOnContainerResize : function(value, old)
+    {
+      var             parent = this.getLayoutParent();
+
       // Remove prior listener for centering on resize
       if (this.__centeringResizeId !== null) {
         parent.removeListenerById(this.__centeringResizeId);
         this.__centeringResizeId = null;
       }
 
-      // If we are to center on appear, arrange to do so
-      if (value.indexOf("appear") != -1) {
-        this.__centeringAppearId =
-          this.addListener("appear", this.center, this);
-      }
-
       // If we are to center on resize, arrange to do so
-      if (value.indexOf("resize") != -1) {
+      if (value) {
         if (parent) {
           this.__centeringResizeId =
             parent.addListener("resize", this.center, this);
--- a/framework/source/class/qx/ui/window/Window.js
+++ b/framework/source/class/qx/ui/window/Window.js
@@ -91,6 +91,9 @@ qx.Class.define("qx.ui.window.Window",
     // Focusout listener
     this.addListener("focusout", this._onWindowFocusOut, this);
 
+    // Initialize the centerWhen property to its own empty array
+    this.setCenterWhen([]);
+
     // Automatically add to application root.
     qx.core.Init.getApplication().getRoot().add(this);
 
@@ -408,7 +411,7 @@ qx.Class.define("qx.ui.window.Window",
      */
     centerWhen :
     {
-      init : [],
+      init : null,              // initialized in constructor
       nullable : false,
       apply : "_applyCenterWhen",
       check : function(value)
@@ -1058,7 +1061,7 @@ qx.Class.define("qx.ui.window.Window",
     // overridden
     _applyCenterWhen : function(value, old)
     {
-      var             parent;
+      var             parent = this.getLayoutParent();
 
       // Remove prior listener for centering on appear
       if (this.__centeringAppearId !== null) {
@@ -1080,7 +1083,6 @@ qx.Class.define("qx.ui.window.Window",
 
       // If we are to center on resize, arrange to do so
       if (value.indexOf("resize") != -1) {
-        parent = this.getLayoutParent();
         if (parent) {
           this.__centeringResizeId =
             parent.addListener("resize", this.center, this);
@@ -1219,10 +1221,6 @@ qx.Class.define("qx.ui.window.Window",
 
     // Remove ourselves from the focus handler
     qx.ui.core.FocusHandler.getInstance().removeRoot(this);
-
-    // Remove the listener for appear, if there is one
-    id = this.__centeringAppearId;
-    id && this.removeListenerById(id);
 
     // If we haven't been removed from our parent, clean it up too.
     parent = this.getLayoutParent();
--- a/framework/source/class/qx/ui/window/Window.js
+++ b/framework/source/class/qx/ui/window/Window.js
@@ -1219,20 +1219,24 @@ qx.Class.define("qx.ui.window.Window",
   destruct : function()
   {
     var id;
-    var parent = this.getLayoutParent();
+    var parent;
+
+    // Remove ourselves from the focus handler
+    qx.ui.core.FocusHandler.getInstance().removeRoot(this);
 
     // Remove the listener for appear, if there is one
     id = this.__centeringAppearId;
     id && this.removeListenerById(id);
 
-    // Remove the listener for resize, if there is one
-    id = this.__centeringResizeId;
-    id && parent.removeListenerById(id);
+    // If we haven't been removed from our parent, clean it up too.
+    parent = this.getLayoutParent();
+    if (parent) {
+      // Remove the listener for resize, if there is one
+      id = this.__centeringResizeId;
+      id && parent.removeListenerById(id);
 
-    // Remove ourselves from the focus handler
-    qx.ui.core.FocusHandler.getInstance().removeRoot(this);
-
-    // Remove ourself from our parent
-    parent && parent.remove(this);
+      // Remove ourself from our parent
+      parent && parent.remove(this);
+    }
   }
 });
--- a/framework/source/class/qx/ui/window/Window.js
+++ b/framework/source/class/qx/ui/window/Window.js
@@ -100,7 +100,7 @@ qx.Class.define("qx.ui.window.Window",
     // Register as root for the focus handler
     qx.ui.core.FocusHandler.getInstance().addRoot(this);
 
-    // change the resize frames appearance
+    // Change the resize frames appearance
     this._getResizeFrame().setAppearance("window-resize-frame");
   },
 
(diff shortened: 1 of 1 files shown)


zaucker (approved):

framework/source/class/qx/ui/window/Window.js
```diff
@@ -632,17 +679,34 @@ qx.Class.define("qx.ui.window.Window",
...
+      // so now. (Note that we explicitly re-obtain the autoDestroy property
+      // value, allowing the user's close handler to enable/disable it before
+      // here.)
+      if (this.getAutoDestroy()) {
+        this.removeAll();                                    // remove contents
+        qx.event.Registration.removeAllListeners(this);      // remove listeners
+        qx.core.Init.getApplication().getRoot().remove(this);// remove self
+        this.dispose();
+      }
     },
```
cajus: For some reason we did not do it this way and used the *onDisappear* event to handle that. I don't recall the reason, because it's so much time has passed now. Will check when I'm back @ work...

framework/source/class/qx/ui/window/Window.js
```diff
@@ -632,17 +679,34 @@ qx.Class.define("qx.ui.window.Window",
...
      */
     open : function()
     {
       this.show();
       this.setActive(true);
       this.focus();
+
+      // If autoCenter is true, then center the window
+      if (this.getAutoCenter()) {
+        this.center();
```
oetiker: at least in my code I usually center in response to the appear event ... and recenter on resize ... as discussed on gitter

level420: The same with our code. I've created a mixin which adds a ```centered``` property and adds an event listener to both ```resize``` and ```appear``` event.
Additionally I've added an applier to the ```centered``` property which centers the window when set to true while the window is open.

framework/source/class/qx/ui/window/Window.js
level420: I think the removal from the focus handler is missing here:

```javascript
qx.ui.core.FocusHandler.getInstance().removeRoot(this);
```

framework/source/class/qx/ui/window/Window.js
level420: The same with our code. I've created a mixin which adds a ```centered``` property and adds an event listener to both ```resize``` and ```appear``` event.
Additionally I've added an applier to the centered property which centers the window when set to true while the window is open.

framework/source/class/qx/ui/window/Window.js
level420: What about packing the destruction code into one overridable method. This way deriving classes could add their own code or even interrupt auto destruction in some cases.

level420 (changes requested):

framework/source/class/qx/ui/window/Window.js
```diff
@@ -632,17 +679,34 @@ qx.Class.define("qx.ui.window.Window",
...
+
+      // If automatically destroying the window upon close was requested, do
+      // so now. (Note that we explicitly re-obtain the autoDestroy property
+      // value, allowing the user's close handler to enable/disable it before
+      // here.)
+      if (this.getAutoDestroy()) {
+        this.removeAll();                                    // remove contents
+        qx.event.Registration.removeAllListeners(this);      // remove listeners
+        qx.core.Init.getApplication().getRoot().remove(this);// remove self
+        this.dispose();
```
johnspackman: Can this be reduced to just `this.dispose()`?  `this.removeAll()` is done by the `Widget.destruct` already, and `qx.event.Registration.removeAllListeners` is handled by `Object.destruct`.  

I would expect that `getRoot().remove()` does need to be called, but in the current implementation it looks like it's *never* called, even if the Window is manually destroyed -- which would sound like a bug you'd expect to have cropped up by now, so maybe it's already handled elsewhere somehow?

However, if something needs to be done during destruction/disposal, the code could be moved into the destructor or overridden `dispose()` method so it's always implemented

level420: Yes it should be reduced to a ```this.dispose()``` and by adding additional code into the ```destruct``` method. The additional code could be
```javascript
qx.ui.core.FocusHandler.getInstance().removeRoot(this);
qx.core.Init.getApplication().getRoot().remove(this);
```

level420: I've digged through the framework code without finding any call to ```qx.core.Init.getApplication().getRoot().remove``` there, so I think windows are added but never removed from application root.

derrell: The fact that I couldn't find anyplace that removed the window from the root was what started me on this path.

Thanks for all of the comments, guys. I'll update this PR in the coming days.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -993,6 +1018,112 @@ qx.Class.define("qx.ui.window.Window",
...
+            // These are expected values
+            break;
+
+          default :
+            throw new Error("Unexpected centerWhen value: " + value);
+          }
+        });
+
+      return values;
+    },
```
johnspackman: this transform is perfectly valid code, but it seems very complicated for what the end goal is trying to achieve, ie a simple pair of booleans.  How about changing the property to `check: [ "appear", "resize", "both", "never" ]`?  It's a common approach to this problem and a lot easier to maintain

johnspackman (changes requested):

framework/source/class/qx/ui/window/Window.js
derrell: I try to avoid that type of thing. It has no future capability to enhance it without greatly complicating the list of options. Consider if we wanted to have center upon resize of the window as another option. Then we'd have [ "appear", "resizeParent", "resizeWindow", "appear+resizeParent", "appear+resizeWindow", "resizeWindow+resizeParent", "appear+resizeWindow+resizeParent" ] ... and you'd have to list them in the correct order. Whereas, with the current code, if we want to add a third option, it's changing only that simple switch statement.

derrell: (oh, and I forgot "never" in my modified list)

framework/source/class/qx/ui/window/Window.js
```diff
@@ -372,6 +372,78 @@ qx.Class.define("qx.ui.window.Window",
...
+     * An empty string or an empty array removes all automatic centering.
+     * 
+     * Otherwise, the property value should be either an array containing one
+     * or more of the above strings, or a string containing one or more of the
+     * above strings, whitespace-delimited.
+     * 
+     * Examples:
+     *   These two yield identical behavior, centering upon appear and resize:
+     *     win.setCenterWhen([ "appear", "resize" ]);
+     *     win.setCenterWhen("appear resize");
```
oetiker: any particular reason for whipping up a DSL here ? (albite a simple one)

framework/source/class/qx/ui/window/Window.js
johnspackman: I see your point - but is it necessary to convert from strings in various forms?  The centerWhen property has a given type (array) and set of allowed values in that array, I don't think it's appropriate to coerce the values from a subset of other unknown values, especially if that means parsing strings into arrays etc.  IMHO it's the caller's requirement to use the correct type.

As a side issue, I agree with the usefulness of sets as property values, it would be great if there was a way to express this in qx property definitions but perhaps we can talk about that on gitter?

johnspackman (changes requested):

framework/source/class/qx/ui/window/Window.js
derrell: BTW, users who want to extend Window can override the transform and/or apply functions to add their own keys here.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -618,12 +732,15 @@ qx.Class.define("qx.ui.window.Window",
...
+     * Close the current window instance.
+     *
+     * Simply calls the {@link qx.ui.core.Widget#hide} method if the
+     * {@link qx.ui.win.Window#autoDestroy} property is false; otherwise 
+     * removes and destroys the window.
      */
     close : function()
     {
-      if (!this.isVisible()) {
+      if (!this.getAutoDestroy() && !this.isVisible()) {
```
cajus: Why can't I close a window where *autoDestroy* is set to *true*? Not sure what that check is good foor, can you explain?

framework/source/class/qx/ui/window/Window.js
```diff
@@ -929,6 +1055,40 @@ qx.Class.define("qx.ui.window.Window",
...
+    // overridden
+    _applyCenterWhen : function(value, old)
+    {
+      var             parent;
+
+      // Remove prior listener for centering on appear
+      if (this.__centeringAppearId !== null) {
+        this.removeListenerById(this.__centeringAppearId);
+        this.__centeringAppearId = null;
+      }
```
cajus: At least the listener on *this* could be omitted. Just calling *this.removeListener("appear", this.center, this)* is enough, as it doesn't fail when it does not exist. But - well, that's a matter of taste, of course.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -1050,5 +1210,29 @@ qx.Class.define("qx.ui.window.Window",
...
+  {
+    var id;
+    var parent;
+
+    // Remove ourselves from the focus handler
+    qx.ui.core.FocusHandler.getInstance().removeRoot(this);
+
+    // Remove the listener for appear, if there is one
+    id = this.__centeringAppearId;
+    id && this.removeListenerById(id);
```
cajus: That two lines can be omitted since listeners to this will be automatically removed.

cajus (changes requested):
If I ignore the _centerWhen_ property, it looks good for me.

A proposal: if it's important to react on _appear_, _resize_, _appear/resize_, then just add two boolean properties - i.e. _autoCenterOnResize_ and _autoCenterOnAppear_. So you're getting rid of all that hassle with checking, shared Array references, etc.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -372,6 +372,96 @@ qx.Class.define("qx.ui.window.Window",
...
+     *     win.setCenterWhen([ "appear", "resize" ]);
+     * 
+     *   Center only upon appear:
+     *     win.setCenterWhen([ "appear" ]);
+     * 
+     *   Remove automatic centering:
+     *     win.setCenterWhen([]);
+     */
+    centerWhen :
+    {
```
level420: I'm here with @cajus: It would be much simpler and understandably if we had two boolean properties ```centerOnAppear``` and ```centerOnResize```. No arrays, no initialization hassle.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -929,6 +1055,40 @@ qx.Class.define("qx.ui.window.Window",
       }
     },
 
+    // overridden
```
level420: The method is not overridden, it's new to this class, I think.

framework/source/class/qx/ui/window/Window.js
level420: I think all this logic could be omitted including the dynamic removal and re-assignment of the listeners. I'll add the listeners for resize and appear anyway and ask in the listener if the corresponding property is true and center or not. I don't think it's worth all that listener tracking.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -929,6 +1055,40 @@ qx.Class.define("qx.ui.window.Window",
...
+
+      // If we are to center on appear, arrange to do so
+      if (value.indexOf("appear") != -1) {
+        this.__centeringAppearId =
+          this.addListener("appear", this.center, this);
+      }
+
+      // If we are to center on resize, arrange to do so
+      if (value.indexOf("resize") != -1) {
+        parent = this.getLayoutParent();
```
level420: It could theoretically happen that the layout parent changes, I think. We'd then have to hook into the ```setLayoutParent``` method, remove the event listener from the old parent and attach the listener to the new one. This gets simpler if we always attach a resize listener to the parent. All this can be done in ```setLayoutParent```

framework/source/class/qx/ui/window/Window.js
level420: It is not necessary to track the listener id for listeners attached to ```this```

framework/source/class/qx/ui/window/Window.js
```diff
@@ -1050,5 +1210,29 @@ qx.Class.define("qx.ui.window.Window",
...
+    // Remove ourselves from the focus handler
+    qx.ui.core.FocusHandler.getInstance().removeRoot(this);
+
+    // Remove the listener for appear, if there is one
+    id = this.__centeringAppearId;
+    id && this.removeListenerById(id);
+
+    // If we haven't been removed from our parent, clean it up too.
+    parent = this.getLayoutParent();
+    if (parent) {
```
level420: If the logic for attaching/detaching the resize listener to the parent is moved to ```setLayoutParent``` we could simply call here ```this.setLayoutParent(null);``` which removes the parent listener if it's there.

level420 (changes requested):

level420:
It would be nice to have a test for the "center on appear" which I think could be done by some calculations on the position within the parent.

framework/source/class/qx/ui/window/Window.js
derrell: The original code just looked at isVisible because hide() and close() were essentially the same thing. If it was already not visible, through either a hide() call or a prior close() call, then there was nothing to do here. With autoDestroy true, we don't want to return. When close() is called, we definitely want it to close and be destroyed even if hide() had previously been called.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -372,6 +372,96 @@ qx.Class.define("qx.ui.window.Window",
...
+     * 
+     *   Center only upon appear:
+     *     win.setCenterWhen([ "appear" ]);
+     * 
+     *   Remove automatic centering:
+     *     win.setCenterWhen([]);
+     */
+    centerWhen :
+    {
+      init : [],
```
derrell: Yes, I'll add a proper init function. Thanks. With regard to the push() scenario, if they want to try to break the code, it's their own fault. This isn't end-user usage; it's programmer usage.

framework/source/class/qx/ui/window/Window.js
derrell: Sorry, commented on this separately. See other explanation.

framework/source/class/qx/ui/window/Window.js
derrell: apply functions are always overridden. They override the default code. Or at least that's the way it is throughout most qooxdoo code. I don't care. I'll remove that, but I think it's "the standard."

framework/source/class/qx/ui/window/Window.js
derrell: That's just a preference of methodology, right? No technical issue here?

framework/source/class/qx/ui/window/Window.js
derrell: Yes, the layout parent can change. That's why there is code in the overridden setLayoutParent method to handle that.

framework/source/class/qx/ui/window/Window.js
derrell: Yes, I need to track the listener id so that I can remove the listener if the property value changes. (It's much faster to use the listener id than to search through the list of listeners for the one that matches a given function and context.) But yes, I can remove these two lines here because all listeners of `this` are removed in higher-layer code.

framework/source/class/qx/ui/window/Window.js
derrell: That removes a few lines of code, while making the code more difficult to understand.

derrell: Also, I still need to retrieve the parent in order to remove `this` from it.

framework/source/class/qx/ui/window/Window.js
level420: Yes, just how I would have done it. Nothing technical.

framework/source/class/qx/ui/window/Window.js
oetiker: I also think having just two properties would be perfectly fine and easy to understand ...

oetiker (changes requested):

framework/source/class/qx/ui/window/Window.js
level420: Afaik applyers are not created automatically and then overriden. If you declare a property and define there an applyer attribute, then using that property via the setter will throw an exception. See http://tinyurl.com/zsd4oaq

framework/source/class/qx/ui/window/Window.js
derrell: I don't like it, but ok, I'll change it to two properties. It may take me a day or three before I can get to it, though.

framework/source/class/qx/ui/window/Window.js
```diff
@@ -929,6 +1020,43 @@ qx.Class.define("qx.ui.window.Window",
       }
     },
 
+    _applyCenterOnAppear : function(value, old)
+    {
+      var             parent = this.getLayoutParent();
```
oetiker: this is not used anymore ...

level420 (approved):

johnspackman (approved):

oetiker (approved):

zaucker:
@derrell I like your last commit message ... ;-)

cajus (approved):

PR #9224, opened 2016-11-03
Bump CLDR to 30.0.2 and extract 'latn' number formats

cajus:
There was a change in CLDR data that breaks the generator code for some languages: number formats are now optionally tagged by a *numberFormat*. If there are multiple definitions, the generator picks the first of them - which might result in *arabic* instead of *latin* number formats.

I used that commit to also bump the CLDR version (again).


RemiHeugue:
It works :+1: (don't forget to run generate.py distclean)

I've got some warnings during the first run though:
```
  - Processing 3 locales  qooxdoo/tool/pylib/misc/cldr.py:138: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not decimalSeparatorNode:
qooxdoo/tool/pylib/misc/cldr.py:146: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not groupSeparatorNode:
qooxdoo/tool/pylib/misc/cldr.py:152: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not percentFormatNode:
```

cajus:
So we've different python versions. Mine does not note anything - but it's easy to fix.

level420 (approved):

level420:
I trust in @cajus :-)

cajus:
@level420 I've expected you to check every single line of CLDR change ;-)

cajus:
The caveat of that solution is that we only support latin number formats. But changing stuff to support multiple of them isn't quite straight forward. I'd wait for a bug report that want's arabic number formats in frech, though ;-)

johnspackman (approved):

oetiker (approved):

derrell (approved):

zaucker (approved):

level420:
Full quorum. Merging.
Thank you @cajus

PR #9225, opened 2016-11-08
treat async functions as normal functions

valette:
ES7 async functions are almost here, and they're great!

Unfortunately, when adding a listener, only functions are allowed right now, not async functions.
I just added the 'AsyncFunction' type, and modified the isFunction assertion to allow async functions and not only functions.

Do we need an isAsyncFunction assertion?


johnspackman:
Hi Valette, thanks for the PR - please can you amend it to add a unit test for this?

oetiker:
JS seems to distinguish between the two ... so I wonder if we can afford just not to ... after all in regular functions we know that things will happen in order ... and at least in my code I rely on that pretty often ... and idea about the implications of this ?

valette:
I would say that the biggest difference between functions and async functions is that functions can return anything, while async functions return a Promise.

So in the context of adding listeners, there is no difference since the return value of the listener is ignored, insn't it?

valette:
Maybe a solution is to change the isFunction to an IsFunctionOrAsyncFunction assertion in the addListener method?

oetiker:
yes I agree for listener functions that should work fine, but your change is such that qooxdoo everywhere will treat them the same ... especially if other code uses the 'isFunction' test to determine if they are faced with a function then the results may be unexpected ... so I would suggest to rather add `isAsyncFunction` and then modify qooxdoo where appropriate to also accept async functions

valette:
I think it's better now. I will try for a unit test, but I'm quite ignorant at this for now...

level420 (changes requested):
@valette Looks good, but please add a test case. Thank you.

valette:
> @valette Looks good, but please add a test case. Thank you.

Will do!

valette:
> @valette do you need advice for the test ?

I'm just wondering how I can write a test including async functions which will not break with non-ES7 browsers, or is it not a problem?

level420:
I think you first need to extend ```qx.bom.client.EcmaScript``` to detect the availability of the async function in the client by offering a ```ecmascript.function.async``` key via ```qx.core.Environment.get```.
The test could be to eval a dummy async function definition and check if it is of type ```AsyncFunction```. If it succeeds, then it should be true.

Then we should extend ```qx.dev.unit.MRequirements``` by adding a ```hasAsyncFunctions``` method which would allow to skip clients not offering async function in the tests.

And after that write the test for which will be skipped if the async functions are not present, like it's done here: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/ui/embed/Flash.js#L98
In this case the line should be ```this.require(["asyncFunctions"]);```

We get two things here: a possibility to detect the availability of async functions and the possibility to skip tests if they are not available.

I'd implement this first start based on this before adding a test.

level420:
an implementation for ```qx.coreEnvironment.get("ecmascript.function.async")``` could be
```javascript
diff --git a/framework/source/class/qx/bom/client/EcmaScript.js b/framework/source/class/qx/bom/client/EcmaScript.js
old mode 100644
new mode 100755
index 794eb43..b361826
--- a/framework/source/class/qx/bom/client/EcmaScript.js
+++ b/framework/source/class/qx/bom/client/EcmaScript.js
@@ -251,6 +251,17 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
      */
     getStringTrim : function() {
       return typeof String.prototype.trim === "function";
+    },
+    
+    /**
+     * Checks if the client supports async functions
+     * @internal
+     * @return {Boolean} <code>true</code>, if async function support is available.
+     */
+    getAsyncFunction : function() {
+      var f;
+      try { eval("f = async function(){};") } catch(e) { return false; };
+      return qx.Bootstrap.getClass(f) === "AsyncFunction";
     }
   },
 
@@ -287,5 +298,8 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
     qx.core.Environment.add("ecmascript.string.startsWith", statics.getStringStartsWith);
     qx.core.Environment.add("ecmascript.string.endsWith", statics.getStringEndsWith);
     qx.core.Environment.add("ecmascript.string.trim", statics.getStringTrim);
+    
+    // ES7 async function support
+    qx.core.Environment.add("ecmascript.function.async", statics.getAsyncFunction);
   }
 });
```

level420:
an implementation of ```qx.dev.unit.MRequirements.hasAsyncFunctions``` would then be:
```javascript
diff --git a/framework/source/class/qx/dev/unit/MRequirements.js b/framework/source/class/qx/dev/unit/MRequirements.js
old mode 100644
new mode 100755
index cdc7832..12dc718
--- a/framework/source/class/qx/dev/unit/MRequirements.js
+++ b/framework/source/class/qx/dev/unit/MRequirements.js
@@ -264,6 +264,16 @@ qx.Mixin.define("qx.dev.unit.MRequirements", {
     hasNoChromeOnLinux : function()
     {
       return (qx.core.Environment.get("browser.name") === "chrome" && qx.core.Environment.get("os.name") === "linux"?false:true);
+    },
+    
+    /**
+     * Checks if the application is running on a client supporting async functions
+     *
+     * @return {Boolean} <code>true</code> if the client supports async functions
+     */
+    hasAsyncFunctions : function()
+    {
+      return qx.core.Environment.get("ecmascript.function.async");
     }
   }
```

level420:
@valette do you need any additional help with the tests?

valette:
No, I'm just a little bit behind my schedule. I plan to do this in the following month, would it be OK?

level420:
@valette of course :smile:

level420:
The recently released Chrome 55 supports async functions: https://www.chromestatus.com/feature/5643236399906816

valette:
OK I implemented the test in framework/source/class/qx/test/event/Emitter.js
One travis build fails because linting outputs a deprecation warning.

Unfortunately I also did a 'git rebase', which seems to pollute the PR... Is there an easy fix to that?

valette:
Wait, something is missing...

valette:
OK, something working at last! But the test just creates an async function, adds it to an event listener, fires the event, removes the listener, fire the event again:

```
    testAddAsyncFunction : function() {<br>
      this.require(["asyncFunctions"]);
      var f = eval("f = async function(){};");
      this.__ee.addListener("test", f, this);
      this.__ee.emit("test");
      this.__ee.removeListener("test", f, this);
      this.__ee.emit("test");
    }
```

but the spy is not used anymore.

oetiker:
@valette if you squash you PR and push it with --force all your intermediate steps will be gone ... there are a bunch of tutorials around on how todo that

level420:
Wouldn't it be better to create a new PR only containing those changes really belonging to the issue and closing this one?

valette:
I just did that!

level420:
@valette thank you! closing this PR now.

level420:
closed in favour of PR #9266

PR #9226, opened 2016-11-09
use the correct start time for animation sequences

peuter:
The start time for an animation sequence should be based on `performance.timing.navigationStart + performance.now()` and not `Date.now()` (if the performance API is available).

As the performance API says `performance.now()` is constantly increasing (independent from changes to system tiime), while the browser is active. It does *not* increase while the browse is suspended. So e.g. after the browser has been suspended for 10s, `Date.now()` would be 10s ahead `performance.timing.navigationStart + performance.now()` and the animation sequence will start with a 10s delay with the current implementation.

And: I replaced `+(new Date())` with `Date.now()` within this fix, because it's more intuitive.

### DIFF
--- a/framework/source/class/qx/bom/AnimationFrame.js
+++ b/framework/source/class/qx/bom/AnimationFrame.js
@@ -82,7 +82,7 @@ qx.Bootstrap.define("qx.bom.AnimationFrame",
     startSequence : function(duration) {
       this.__canceled = false;
 
-      var start = (window.performance && window.performance.now) ? (performance.now() + qx.bom.AnimationFrame.__start) : Date.now();
+      var start = (window.performance && performance.now) ? (performance.now() + qx.bom.AnimationFrame.__start) : Date.now();
       var cb = function(time) {
         if (this.__canceled) {
           this.id = null;
--- a/framework/source/class/qx/bom/AnimationFrame.js
+++ b/framework/source/class/qx/bom/AnimationFrame.js
@@ -25,7 +25,7 @@
  * API the spec describes.
  *
  * Here is a sample usage:
- * <pre class='javascript'>var start = +(new Date());
+ * <pre class='javascript'>var start = Date.now();
  * var cb = function(time) {
  *   if (time >= start + duration) {
  *     // ... do some last tasks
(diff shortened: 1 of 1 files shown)


cajus (approved):
Looks familiar ;-)

framework/source/class/qx/bom/AnimationFrame.js
```diff
@@ -76,11 +76,13 @@ qx.Bootstrap.define("qx.bom.AnimationFrame",
...
      *
      * @param duration {Number} The duration the sequence should take.
+     *
+     * @ignore(performance.*)
      */
     startSequence : function(duration) {
       this.__canceled = false;
 
-      var start = +(new Date());
+      var start = (window.performance && window.performance.now) ? (performance.now() + qx.bom.AnimationFrame.__start) : Date.now();
```
level420: I'm not shure, but this should be tested against Internet Explorer which tends to fail on such tests like 
```javascript
(window.performance && window.performance.now)
```
vs.
```javascript
("performance" in window ...
```

I think it whould be better to have a ```qx.core.Envorinment.get("performance.timer")``` to check against, which also may be handy in other situations.

level420 (changes requested):

framework/source/class/qx/bom/AnimationFrame.js
level420: Hmm. I see that the original code uses the check window.performance in the defer section, so it should be safe then.

@peuter do you think the  qx.core.Environment.get("performance.timer") is worth it?

cajus: `performance.timer` is intended to be used for performance monitoring, only. I guess the only reason why it is used here (the only location in the complete framework) is that the animation frame callback gets a `DOMHighResTimeStamp` parameter. Not sure if it is usable for something else, though. I'm wondering what for you need a 5µ resolution an a web platform.

level420: @peuter @cajus. I'm convinced, maybe besides of the test which should be (I think)

```javascript
(window.performance && performance.now)
```
to get par with the test in the defer section.

level420 (approved):

level420:
@peuter I don't know if this is relevant anymore, but using Date.now will break compatibility for some ancient browsers. See https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/now

peuter:
@level420 Qooxdoo has its own polyfill for `Date.now()` which is `+(new Date())` by the way.

level420:
Cool! then I'm quiet now ;-)
Looks good.

johnspackman (approved):

johnspackman (approved):

level420:
Removed the release_5_0_2 milestone, as it is no bug fix, more an enhancement.

peuter:
This is a bugfix, no enhancement.

level420:
2 approvals. more than 2 business day. merging
Thank you @peuter

PR #9231, opened 2016-11-17
Transform property

johnspackman:
This PR extends the property definition so that the `transform` method also receives the value currently assigned to the property (or undefined) as well as the new value.

This makes `transform` analogous to the `apply` method, and provides information which is not otherwise available; note that it is not ideal to call the property’s `getXxx` method because this can throw exceptions if the property has not yet been initialised, and adding try/catch handlers complicates property implementation unnecessarily (especially when the property system has the existing value to hand)

This is a non-breaking change.

### DIFF
--- a/documentation/manual/source/pages/core/defining_properties.rst
+++ b/documentation/manual/source/pages/core/defining_properties.rst
@@ -396,7 +396,7 @@ Here we define both a check and transform method for the width property. Though 
 
     members :
     {
-       _transformWidth : function(value)
+       _transformWidth : function(value, oldValue)
        {
           if ( qx.lang.Type.isString(value) )
           {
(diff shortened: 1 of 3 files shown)


level420:
@johnspackman Is the ability to declare a transform method new to this release?

derrell:
I added transforms about 10 years ago.

level420:
Oh no! It was already there. Never used it.

level420 (changes requested):
I think the documentation should also be changed accordingly:

here: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/core/Property.js#L85-L93

and here: https://github.com/qooxdoo/qooxdoo/blob/master/documentation/manual/source/pages/core/defining_properties.rst

johnspackman:
@level420 all done

level420 (approved):

level420:
Thank you @johnspackman !

zaucker (approved):

derrell (approved):

oetiker (approved):

cajus (approved):

zaucker (approved):

level420:
Full quorum. Merging.
Thank you @johnspackman

Issue #214, opened 2016-04-29
Promises for Qooxdoo


johnspackman:
Adding a Promises implementation to Qooxdoo would be beneficial because:
- it enables events to be chained in a structured, application specific order;
- there is pressure to switch to using async I/O when communicating and promises encourage async coding;
- it is a popular and well understood pattern

At the least, the implementation should use a short namespace (eg qx.Promise) and support binding "this" to the entire promise.

Also proposed is this [StackOverflow question](http://stackoverflow.com/questions/35156009/controlled-event-handling-flow-in-qooxdoo) which proposes that by returning Promises from event handlers, asynchronous event handlers can be implemented

Similarly, property apply methods could return a Promise to indicate when the apply has completed (effectively delaying the changeXxxx event for the property until applyXxxx has completed)

As well as that SO question, this would be useful for server I/O - as volumes of data increase, browsers time out the connection _unless_ it is asynchronous (especially Safari); for example, an apply method which performs an asynchronous operation, causes the caller to think the apply has completed it's work (which might include loading other objects from the server) - the caller then tries to use these objects but fails milliseconds before the data arrives.

It is reasonable for an apply method to take an unspecified amount of time, and when it comes to server I/O it is increasingly important that this is allowed to be asynchronous. Technically, the powers that be have decided that synchronous I/O is deprecated, and Safari is already adding a fixed, very short timeout to synchronous I/O operations that cannot be overcome.

derrell:
We are using the Promise package from https://github.com/kriskowal/q in our
server apps. It works well for us, and may provide either a starting point,
or some ideas for a native qooxdoo implementation.

Derrell

dteleguin:
We've been using native promises for a while in our qooxdoo project. There are several moments to pay attention to:

1. Most modern browsers [do support Promises, but some still lack it](http://kangax.github.io/compat-table/es6/), most importantly PhantomJS (see this [issue](https://github.com/ariya/phantomjs/issues/12401)). PhantomJS remains a de facto standard in headless testing, and it's unlikely that it gets Promise support until they migrate to [CEF](https://bitbucket.org/chromiumembedded/cef).
    That said, we'll have to provide a fallback implementation. In our project, we used [babel-polyfill](https://github.com/babel/babel/tree/master/packages/babel-polyfill), and it worked perfectly. Considering that there were plans to integrate babel-polyfill into [QxCompiler](https://github.com/johnspackman/qxcompiler), this might be the way to go.

[...]

PR #9232, opened 2016-11-21
Promises 

johnspackman:
This PR implements promises as originally outlined in https://github.com/qooxdoo/qooxdoo/issues/214 - Promises are non-trivial to implement properly, so this class is implemented thanks to BluebirdJS from http://bluebirdjs.com/ (also under MIT license)

The Qooxdoo classes have not been changed to “promisify” them because that can be handled as a separate PR; however this is not true of properties, which can only be promisified with the participation of the Qooxdoo property system.

The changes to the property system, which are backwards compatible and non-breaking, are:

*Asynchronous Apply Methods*
It is reasonable that any `applyXxx` method could take some time to complete (including triggering a server roundtrip), in which case it is necessary to delay the firing the `changeXxx` event until the apply is complete, otherwise the event is fired before the property value is fully applied.

The return value for apply methods is normally ignored, but with this PR if you return a `qx.Promise` the `setXxx` will wait for the promise to be fulfilled before firing the event.

As setXxx method normally returns the value which has been set, it is not possible to return the promise to the caller - which is a disadvantage because it follows that there could be other processes which depend on the successful, and sequential, completion of the property apply.

*A new setXxxAsync Method*
For properties which are known in advance to have asynchronous apply methods, you can enable a new `setXxxAsync` method by setting the `async: true`  value on the property configuration; the normal property accessors operate in the same way that they always did and remain backward compatible, it’s just that there is `setXxxAsync` method that returns a `qx.Promise`.  Note that it always returns a promise, even if the apply method is not asynchronous and does not return anything - this is so that `setXxxAsync` always operates consistently regardless of how it is implemented.

For example:

    properties {
      myProp: {
        init: 0,
        async: true,
        apply: “_applyMyProp”
      }
    },
    members: {
      doStuff: function() {
        this.setMyPropAsync(123).then(function() {
          this.debug(“Set value of myProp to 123”);
        }, this);
      },
      // … snip ….

*Setting property values from Promises*
When you want to set a property to a value fulfilled by a promise, there is this form:

    var myPromise = someOtherValue.getSomethingAsync();
    myPromise.then(function(value) {
      myObj.setMyProp(value);
    });

The above example works, but the code remains relatively lengthy for the task you are trying to achieve.  The setXxx and setXxxxAsync methods can take values either as ordinary values, or they can be passed a `qx.Promise`, in which case they will wait for the promise to be fulfilled before applying it.  The example above can be re-written as:

    var myPromise = someOtherValue.getSomethingAsync();
    myObj.setMyProp(myPromise);

The one exception to this is where the property value is expected to be a promise; this is achieved by defining the property with a `check: “qx.Promise”`, in which case it is stored as-is without waiting for it to be resolved.


level420:
@johnspackman this PR also seems to contain the changes of PR https://github.com/qooxdoo/qooxdoo/pull/9231 with the transform attribute change.

johnspackman:
@level420 ah, yes you're right.  I started this after proposing that PR, I didn't mean to actually include it though.  Can I leave it in?  It makes for a slightly odd change history but shouldn't cause any conflicts.

level420:
@johnspackman I don't know which conflicts will result, but I've planned to merge the transform PR approx. in 3 hours (grace period for waiting for dissenting votes).
Is the transform PR a pre-requisite for this PR?

johnspackman:
its not a prerequisite, but I'd imagine there would be conflicts if this PR was not based on the transform PR.

level420:
@johnspackman just merged the transform PR

level420:
@johnspackman  as I understand the async setter/getter of the async property ```myprop``` would be named ```setMypropAsync``` and ```getMypropAsync```. This cold lead to a conflicting situation if another property exists which is called ```mypropAsync```. 

This situation should be avoided by either choosing a non-conflicting naming scheme e.g. by using the prefix ```setasync``` (all lowercase), or by actively checking for the conflict during setter and getter creation.

johnspackman:
yes - although there is no `getMypropAsync` at the moment (and IMHO there is a case for it in the future), you're right there could be a conflict with `setMypropAsync`; but the `setMypropAsync` will not be generated unless you manually enable the `async:true` setting in the property configuration.

I did consider using a prefix like `setasyncMyprop`, but decided against it because it will work with lots of existing code that can dynamically get and set values based on the prefix of method names; while binding and other code uses detection will often need to be upgraded to make full use of async features, you can make them write any property (or pseudo-property) asynchronously just by adding the "Async" suffix.

Actually, thinking about it - this is a good reason for adding a `getXxxAsync` method and a `changeXxxAsync` event - so that you can write code like:

```
myObj.bind("mypropAsync", myOtherObj, "value");
```

Other reasons are that the Async suffix seems to be a bit of a convention (eg Bluebird's promisifyAll adds Async suffix when processing node modules), plus I felt that the `setXxxAsync` style was more idiomatic.

If we agree to stick with Async suffix, then there should be conflict detection also.

level420 (changes requested):
I think a check should be there to avoid setter name clashes as described in https://github.com/qooxdoo/qooxdoo/pull/9232#issuecomment-262194488

framework/source/class/qx/core/Property.js
```diff
@@ -903,38 +914,51 @@ qx.Bootstrap.define("qx.core.Property",
...
      *
-     * @param instance {Object} the instance which calls the method
      * @param clazz {Class} the class which originally defined the property
      * @param name {String} name of the property
      * @param variant {String} Method variant.
-     * @param args {arguments} Incoming arguments of wrapper method
      * @return {var} Return value of the generated function
      */
-    installOptimizedSetter : function(instance, clazz, name, variant, args)
+    installOptimizedSetter : function(clazz, name, variant)
```
level420: This also is an API change

level420 (changes requested):

framework/source/class/qx/core/Property.js
```diff
@@ -797,30 +811,27 @@ qx.Bootstrap.define("qx.core.Property",
      * Installs a getter into the class prototype, without executing it
      * Supported variants: get
      *
-     * @param instance {Object} the instance which calls the method
      * @param clazz {Class} the class which originally defined the property
      * @param name {String} name of the property
      * @param variant {String} Method variant.
      */
-    installOptimizedGetter : function(instance, clazz, name, variant)
+    installOptimizedGetter : function(clazz, name, variant)
```
johnspackman: ah - yes, that's true; the problem is that an instance is simply not always available when installing the accessor methods, those installOptimised* methods were introduced in the patch for qxcompiler and were simply following the implementation of teh functions they were split off from.  Arguably they shouldn't be exposed in the first place, but as they never made it to a released version, I propose making them private instead.

level420: Maybe it would be best to preserve the public api as a wrapper to the new private methods?!?!

framework/source/class/qx/core/Property.js
oetiker: isn't there a way to preserve the api ?

framework/source/class/qx/core/Property.js
oetiker: If the function has never been in a release, then I would say it is fine to change the api

framework/source/class/qx/core/Property.js
oetiker: is this also a new function which has never been in a release ?

johnspackman: yes

johnspackman:
FWIW I have some changes (and requested changed) to commit, should be later on today or tomorrow

derrell:
I haven't yet had a time to review this in depth, but I have one concern based on the prior comments. We have a few places in qooxdoo where one can't set all properties using `obj.set( { ... } );` which I find extremely annoying. In those few cases, one must use the property's own setter, e.g., `obj.setSomeProperty(value);`. Let's please be sure that any new setters can be set using the `obj.set( { ... } );` notation.

johnspackman:
is that where the object has implemented its own property, ie it has a setXx and getXx but it is not a "real, qooxdoo" property?

(👍  btw)

johnspackman:
@derrell looking at https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/core/MProperty.js#L39-L76 i think the problem there is that it is not sometimes possible to make a property behave the way you want it to; for example, selection in `qx.ui.form.List` and elsewhere was implemented by hand in order to prevent the underlying `qx.data.Array` being replaced (ie `setSelection` would modify the array instead).  That can be easily implemented now with the new `transform` PR (selection code was a motivation for proposing it).

The other time I can think of (from my own code, I can't think of a qooxdoo example at the moment) is when you need to implement your own storage; while you can override how set works via `apply`, you cannot override how `get` works.  (Interestingly, I looked at qooxdoo next this morning and it allows `get` to be defined).  Adding a `get` would allow properties to be completely overridden, while at the same time keeping the definition that there is a property and being able to enforce the full property richness.  I think that's a different PR though, perhaps to be included in a future PR for https://github.com/qooxdoo/qooxdoo/issues/9233

derrell:
@johnspackman Yes, I think the selection example is one I was thinking of. In any case, though, if your new code can allow the standard set() call to work, i.e., you don't munge the naming such that separate calls are required outside of set(), then we should be good from my point of view.

peuter:
I just started to play around with qx.Promise, but I am running into a SyntaxException by the generator when using `.catch(function(error)`

```
>>> Collecting classes  /<class 'ecmascript.frontend.SyntaxException.SyntaxException'> :
Expected an attribute name (pos (178, 17)).
```

Is there a special way to use this, or does the qooxdoo generator need to learn the new syntax?

cajus:
@johnspackman looks like the generator works after changing *tool/pylib/ecmascript/frontend/tokenizer.py +270* to read

```
elif tok.value in lang.RESERVED and self.out_stream[-1]['detail'] != "DOT":
```

johnspackman:
brill, thanks @cajus 👏  I'll add that in the next commit

peuter:
@cajus 
With that fix I can generate my application, but it seems to break other things. Just tried to generate the playground to do more tests and with that fix I got:

```
>>> Collecting classes  /<type 'exceptions.IndexError'> :
list index out of range
```
after I reverted it I can build the playground.

cajus:
@peuter that's because there may be no previous token. Should be checked, too.

cajus:
```
diff --git a/tool/pylib/ecmascript/frontend/tokenizer.py b/tool/pylib/ecmascript/frontend/tokenizer.py
index 0f3d07b..c483b0f 100755
--- a/tool/pylib/ecmascript/frontend/tokenizer.py
+++ b/tool/pylib/ecmascript/frontend/tokenizer.py
@@ -126,6 +126,13 @@ class Tokenizer(object):
 
     def parseToken(self):
         for tok in self.scanner:
+
+            hasNoPreviousDot = True
+            try:
+                hasNoPreviousDot = self.out_stream[-1]['detail'] != "DOT"
+            except (IndexError):
+                pass
+
             # some inital values (tok isinstanceof Scanner.Token())
             token = {
                 "source" : tok.value,
@@ -267,7 +274,8 @@ class Tokenizer(object):
                         token['detail'] = lang.TOKENS[tok.value]
 
                 # JS keywords
-                elif tok.value in lang.RESERVED:
+                elif tok.value in lang.RESERVED and hasNoPreviousDot:
```

peuter:
The last fix works for me.

oetiker (approved):

cajus (changes requested):
We've tested that with success. Great addition! ;-)

That needs to be in to make it good, though:

https://github.com/qooxdoo/qooxdoo/pull/9232#issuecomment-262903627

cajus:
@johnspackman do you want me to update the generator part in your tree?

johnspackman:
@/all sorry for the delay, it took me longer than expected to get these last changes working right.  It adds `getXxxAsync` so that bind can work asynchronously - this also implies that there is a `qx.event.Registration.fireEventAsync` and `qx.core.Object.bindAsync` which are only necessary if the caller wants to get the promises.  IE ordinary `fireEvent` and `bind` work just fine on async properties:

```
var obj = new abc.MyClass();
obj.bind("somePropAsync", myOtherObj, "valueAsync");
```

Also included is @level420 's request for conflict detection (which works for all type of property methods, not just the new "Async" versions)

thanks @cajus i'd already included them in this delayed push

level420:
@johnspackman seems that basic tests fail with this error:
```
Firefox 50.0.0 (Linux 0.0.0) qx.test Interface:testClassImplements FAILED
	members.get is undefined
		qx.core.Property:11950
		qx.core.Property:5761
		qx.Class:37126
		qx.Class:2890
```

level420 (dismissed):

peuter:
I started to evaluate this PR by using it in a real world application (and they work really well, great work @johnspackman). With the last changes I get warnings about unreturned promises, when setting normal properties (not async ones). I managed to narrow it down to a playground example:

```
qx.Class.define("Test", {
  extend: qx.core.Object,
  
  properties: {
    "base": {
      init : "",
      check: "String",
      nullable: true,
      event: "_changedBase"
    },
  }
});
var test = new Test();

var p = new qx.Promise(function(resolve, reject) {
  qx.event.Timer.once(resolve, this, 1000);
  test.setBase("tes1");
}, this);

p.then(function() {
  test.setBase("test");
})
.catch(function(error) {
  this.error(error);
});
```

When you remove the event from the property declaration or the ` catch` block at the end the warning does not occur any more. @johnspackman I hope these hints help you finding out whats going on there.

peuter:
Just one additional note. The last changes seemm to have a negative impact on performance. I just reverted https://github.com/qooxdoo/qooxdoo/pull/9232/commits/8255138de7ec2e14bafbe6fb80a18e508f14bc45 from my local clone and the app performance increased dramatically.

johnspackman:
@peuter Please can you try putting that commit back and then editing qx/core/Property.js to remove these lines: https://github.com/johnspackman/qooxdoo/blob/8255138de7ec2e14bafbe6fb80a18e508f14bc45/framework/source/class/qx/core/Property.js#L563-L591

and let me know if that solves the performance problem?

peuter:
@johnspackman Just tried: removing these lines has no influence on the performance, still slow.

johnspackman:
ok thanks for trying, I take a proper look shortly.  I take it that the slow down is in general operation, not just at startup?  Do you have any easily reproducable test case I can look at?

peuter:
Yes it is a general slow down, but i cannot provide a testcase. Maybe you should try to fix the warning-problem I reported before first (in my comment you can find a playground example for reproduction) and hopefully that fixes the performance problem too. I tend to believe that somewhere in the property applying process Promises were used where they shouldn't and that might slow down everything. But that is just a fast assumption and might be totally wrong.

derrell:
I've been thinking a lot about Promises. A common usage pattern is to directly chain the functions, avoiding indentation hell as with multi-level event handlers, and also avoiding the confusion (at least for me) of a separate promise variable. It seems that to make the code more readable in that situation, it would be useful to swap the context and the function in the qxPromise constructor, so usage would like like this (modifying peuter's example from a prior comment):

```
(new qx.Promise(
  this, 
  function(resolve, reject) 
  {
    qx.event.Timer.once(resolve, this, 1000);
    test.setBase("tes1");
  }).then(
  function(resolve, reject)
  {
    test.setBase("test");
  }).catch(
  function(error)
  {
    this.error(error);
  });
```

The context can still be optional (`if (typeof context == "function") fn = context;`). This puts all of the functions directly adjacent, in usage, instead of having a `this` between the first and second function. Does this sound like a rational change?

derrell:
In fact, maybe qx.Promise could be a static function that returns a new qx.PromiseObj (or something) to completely eliminate the confusing instantiation here. (I'm comparing this to typical usage in nodejs apps, where one doesn't instantiate anything in order to use promises. They are so cool, I just don't want them to be confusing for people.)

peuter:
Another thing that confuses me and took a while to figure out whats going on is that you cannot use a static  class as context. I assume that the undelying problem is the different ways to use the `then` method either with two functions:
```
promise.then(callbackFunctionForResolvedPromise, callbackFunctionForRefusedPromise)
// or
promise.then(callbackFunctionForResolvedPromise, context).catch(callbackFunctionForRefusedPromise)
```
I guess the static class is interpreted as callback function and is tried to be called when the promise gets rejected.
Of course you can workaround the problem by using `.bind(this)` but it would be better to solve it another way, either by detecting if the second parameter is a "normal" function or a class or by just defining that the second parameter is always the context and if the user has to use the `.catch` way to get rejected Promises. I would prefer the second solution because IMHO it is less confusing if there is only one way of using it.

@derrell I think it would be more confusing to change the parameter order, because it is a common pattern in qooxdoo that the context parameter is th one after the function it should be applied to. At least it felt natural to me to use it the ways its currentply implemented.

peuter:
Just another thing I recently found out: the generic way af accessing statics from whithin a Promise does not work (`this.self(arguments)`), that should at least be documented if there is no other solution for that.

framework/source/class/qx/Promise.js
```diff
@@ -0,0 +1,6547 @@
...
+     * Returns a Promise object that is resolved with the given value. If the value is a thenable (i.e. 
+     * has a then method), the returned promise will "follow" that thenable, adopting its eventual 
+     * state; otherwise the returned promise will be fulfilled with the value. Generally, if you 
+     * don't know if a value is a promise or not, Promise.resolve(value) it instead and work with 
+     * the return value as a promise.
+     * 
+     * @param value {Object}
+     * @return {qx.Promise}
+     */
+    resolve: function(value) {
```
derrell: As discussed, `resolve` and 'reject' should take a context parameter to which the Promise is bound, in addition to the value/reason.

derrell (changes requested):

derrell:
I've been using this and really like it. Looking forward to it being merged!

derrell:
@peuter The problem with `this.self` isn't `this.self`, it's `arguments`. This is really a JavaScript issue, not a Promises issue. Each function has its own `arguments` pseudo-array, so `arguments` in the member function is different than `arguments` in the Promise function.

You can solve this problem by storing the constructor's or method's `arguments` as shown the following playground snippet (assuming you've built a playground that includes promises):

```
qx.Class.define("xxx",
{
  extend : qx.core.Object,
  
  construct : function()
  {
    var     _args = arguments;
    this.debug("Before promise: arguments=", arguments);
    this.debug("DATA=" + this.self(arguments).DATA)
    
    qx.Promise.resolve()
    .bind(this)
    .then(
      function()
      {
        this.debug("arguments=", _args);
        this.debug("DATA=" + this.self(_args).DATA)
      })
  },
  
  statics :
  {
    DATA : "hello world"
  }
});

new xxx();
```

It's probably more clear to use `xxx.DATA` instead of `this.self(_args).DATA` if you don't mind the hard-coded class name.

peuter:
@derrell I already thought that the arguments are the problem here and I use the `xxx.DATA`-way to solve the problem. I just wanted to mention it here, that it should be documented somewhere to prevent others running into that problem.

b.t.w. A working solution to access the statics in a generic way in your example is: `this.constructor.DATA` could that be an alternative?

derrell:
@peuter yes, this.constructor not only works, but is a bit more intuitive than this.self(arguments).

As to documenting the `arguments` thing, we could do so, but we'd also then want to document it for all callback functions throughout qooxdoo, which of course have the same issue. It's not specific to promises. (I have never, once, in over a decade of using qooxdoo, used `this.self(arguments)` for anything. It's completely meaningless to any but the most sophisticated qooxdoo users and I couldn't even remember what it was for when you first mentioned it. (I don't like `this.base(arguments)` but we don't have much choice there, and becomes so second-nature that anyone using qooxdoo knows what it means.)

derrell (approved):
I've been using this. Although I'd like for the context to be added to the two method, I'd also like to get this PR accepted and can wait for that as a separate PR. Approving.

level420 (changes requested):
the general slow down reported by @peuter bothers me.
I think it should be solved before merging.

johnspackman:
I agree with @level420 - sorry its taken so long, in theory I get some holiday next week when this release is done.  The good news is that the main work is stable, @derrell 's using it and I will have 2 sites with it in production by tonight ...

level420:
@johnspackman thank you for the last fixes. could you please also resolve the branch conflicts? thank you!

level420:
@johnspackman as long as the branch conflicts exist, travis won't run.

level420:
@peuter could you please review this PR and check if the performance degradation you reported before is still present? Thank you!

peuter:
@johnspackman Are those performance problems solved? I only see an added performance test, but no other changes, or are they somewhere included in the rebase commit?

level420:
I did some tests with this PR applied with my main application containing a big editor containing about 200 fields, partly having dependencies which cause a lot of event being fired. Here are the performance tests I did, using the normal generator.py, by manually stopping the time needed for the first initialization which means from clicking "open editor" to "the editor is fully visibly and ready to be used":

1. **Without this PR** applied
**source version: 5sec**
build version: 5sec

2. **With this PR** applied
**source version: 102sec**
build version: 5sec

So for the build version there is no difference between the version without and with this PR, but for the source version there is a performance degradation with factor 20!

level420:
And here is the result of my profiling with chrome 56:

Self Time (ms) | % | Total Time (ms) | % | Function | File:line
-------------- | - | --------------- | - | -------- | -----------
185292.6 | 88.95 | 	185496.8 | 89.05 |	(anonymous) | Promise.js:0	
184468.3 | 88.56 | 	184672.5 | 88.66 | 	>CapturedTrace | Promise.js:0	
183104.8 | 87.90 | 	183305.0 | 88.00 | 	>>longStackTracesCaptureStackTrace | Promise.js:0	
1363.5   | 0.65 | 	1367.5 | 0.66 | 	>>longStackTracesCaptureStackTrace | Promise.js:2027	
824.3    | 0.40 | 	824.3 | 0.40 | 	>CapturedTrace | Promise.js:2345	
5590.5   | 2.68 | 	5603.1 | 2.69 | 	(anonymous) | Promise.js:2443	
1561.3   | 0.75 | 	1561.3 | 0.75 | 	(garbage collector)	
1081.1   | 0.52 | 	1081.1 | 0.52 | 	(program)	
745.2    |      | 	745.2 | 	| (idle)

level420:
And the performance degradation Oscar goes to this line:

https://github.com/johnspackman/qooxdoo/blob/promises_squashed2/framework/source/class/qx/Promise.js#L751

```javascript
    __attachBluebird: function(Promise) {
      this.Bluebird = Promise;
      var debug = qx.core.Environment.get("qx.debug");
      Promise.config({
        warnings: debug,
        longStackTraces: debug,
        cancellation: true
      });
    },
```
Setting the ```longStackTraces``` to true in the debug case is the cause of the performance degradation. Setting it to false stops its.

In order to keep the source version debug-able, this should be configurable by a `qx.debug.promiseLongStackTraces` option in `config.json`, defaulting to false.

johnspackman:
@level420 thanks for testing, but I didn't mean to imply that the performance issue is solved - I must have "automatic push" turned on on my git client, sorry for the confusion.  I've added performance tests to as a separate branch and then merged it in so that there is a clear before/after set of measurements - there is a perf_properties branch which will be a separate PR into master later.

I see a similar degradation in performance, in the setXxx implementations, which Im working on now.  The point of yesterday's squash and rebase was to resolve the conflicts so that this branch can be used for testing; the API should be stable at this point.

level420:
@johnspackman I think I found the cause, please read my comments.

peuter:
I can confirm that the performance problem is gone when I disable the longStackTraces. Another observation: With `longStacktraces: true` I get warnings about unreturned promises where no promises are used at all (the warning points to this line in my Application.js `this.getRoot().setBlockerColor("#000000");`)

These warning are gone with `longStacktraces: false`.

cajus:
After the last changes, I've these warnings before the app is getting initialized:

```
Native.js:58 003300 Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.table.columnmenu.MenuItem
Native.js:58 003304 Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.menu.CheckBox
Native.js:58 003305 Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.menu.AbstractButton
Native.js:58 003305 Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.core.Widget
Native.js:58 003306 Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.core.LayoutItem
Native.js:58 003845 Conflicting property method qx.ui.progressive.State.getUserData with qx.ui.progressive.State
Native.js:58 003846 Conflicting property method qx.ui.progressive.State.setUserData with qx.ui.progressive.State
Native.js:58 004569 qx.core.Init: Load runtime: 4568ms
Native.js:58 004633 qx.core.Init: Main runtime: 64ms
Native.js:58 004646 qx.core.Init: Finalize runtime: 11ms
```

johnspackman:
@peuter warnings about unreturned promises are a deliberate feature of Bluebird - I’ve added individual environment settings to allow fine grained control of Bluebird warnings, but a work around is to return anything other than undefined from your functions.  The idea is that the promise contains a value which is successively manipulated by calls to `.then()`, and the manipulations themselves may involve more promises; when you return an undefined value, you’re erasing the value held in the promise and this is most likely not what any future usage of that promise might want.  Of course, in your example you *know* that there is no future usage of the promise so it’s OK, but in other examples that warning could be incredibly helpful in tracking down a bug.  The workaround is to return null (or something) from your `.then()` callback

Also the performance slowdown is somewhat solved; thanks for @level420 for analysing it, the `longStackTraces` has a major impact on performance so this is not at all desirable for production builds; by default it is turned on for `./generate.py source` and off for `./generate.py build`.

Even after turning off `longStackTraces`, implementation of `setXxxx` methods was slower because it was always constructing a `qx.Promise` to pass the data onto `changeXxxxAsync` event handlers and was returning a `qx.Promise`.  This was a breaking API change (which is no longer done) and in this release properties which are *not* marked as `{ async: true }` are able to defer creating a promise unless there is a specific need (eg a `changeXxxAsync` listener).  There is a slight slow down when compared to the current master, of around 25-30% more time taken - but we’re talking 25-45ms per 10,000 calls.

Properties which *are* marked as `{ async: true }` support asynchronous returns from events and other complications, and this does involve new instances of qx.Promise and are therefore slower; `setXxx` methods for async properties take around 2x longer than normal.  There is room to improve this, although its slow going with the code-generation nature of properties (I’m *so* looking forward to when https://github.com/qooxdoo/qooxdoo/issues/9233 is implemented…)

The environment settings are called `qx.promise.warnings` and `qx.promise.longStackTraces`, and both default to the value of `qx.debug`.

Also included is a bugfix for passing class instances as a context as reported by @peuter

@derrell apologies if I ignored your comments way back on Dec 8th - I think we talked about it on Gitter please shout if not.  In summary, I think that keeping the parameter order as it is now is a good thing, and I’ve added context parameters to `qx.Promise.resolve` and `qx.Promise.reject`

@cajus that’s odd about those warnings - I couldn’t reproduce it, but I think I spotted a related bug - please can you try again, and if it still happens, do you derive from `qx.ui.table.columnmenu.MenuItem`?

level420:
@johnspackman thank you for your last commits!

But I think we really need a config.json switch which disables the longStackTrace feature in source version. I've done some tests with my application where some large form initializations went from 5sec to over 100sec!

Please rethink about inteoducing a flag In order to keep the source version usable for use cases like I have, e.g qx.debug.promiseLongStackTraces option in config.json, defaulting to false.

johnspackman:
@level420 fair point, that is rather drastic! Changes applied :)

level420:
Just found out that the config.json settings are there. Thanks alot John!

peuter:
@johnspackman Yes I now about the intend of the warning and how to workaround them. They have been very useful during develepment. But the last warning I reported did not use a Promise (it was a setter for a qx property, not my code). But this warning is gone now with your last fixes. Performance is also fine now.

So I can approve this PR. Thanks John really nice feature ;-)

level420:
@johnspackman It would be nice to have both `qx.promise.warnings` and `qx.promise.longStackTraces` added here with their defaults:
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/core/Environment.js#L880

This way we could control their values by adding them to the applications `config.json` file, e.g.:

```json
    "source" :
    {
      "desc"    : "create source version of current application",
      
      "environment" :
      {
        "qx.aspects" : true,
        "qx.debug" : true,
        "qx.debug.databinding" : false,
        "qx.debug.ui.queue" : true,
        "qx.debug.dispose" : true,
        "qx.promise.warnings" : true,
        "qx.promise.longStackTraces" : true
      },
```

You've done that already with the `qx.automaticMemoryManagement` key.

level420:
Just checked together with @johnspackman: setting `qx.promise.warnings` and `qx.promise.longStackTraces` in config.json work as expected! Great!

level420 (approved):

johnspackman:
It looks like this is nearly ready to be merged, so please can the PR https://github.com/qooxdoo/qooxdoo/pull/9279 be reviewed and merged into master before this one?  It's so that we have a before/after tests for performance in master

mbgonicus:
I didn't really engage myself in promises, but we're using it in our project and I just came to an interesting observation.

Although I create a `qx.data.Array` object in `then`, I get a native javascript array in the final `spread` method:

```javascript
new qx.Promise(function(resolve) {
        resolve();
    })
    .then(function() {
        return qx.Promise.all(["foo", new qx.data.Array(["a", "b", "c"])]);
    })
    .spread(function(str, arr) {
        console.log(str, arr);
        console.log(arr instanceof qx.data.Array);
    });

// output: 
// foo ["a", "b", "c"]
// false
```

Is this intended?

johnspackman:
@mbgonicus I'd say that's a bug, or certainly a very undesirable behaviour - i'll change it

zaucker (approved):
I can't say that I grasp all the glory of this PR. As it was heavily tested in the wild I hope it won't break too many things on other people's applications.

level420:
@johnspackman I've merged PR #9279 as you requested, but now we have conflicts in this PR again.

johnspackman:
@level420 done

level420:
@johnspackman what prevents this PR from being merged, ist the bug reported by @peuter here https://github.com/qooxdoo/qooxdoo/pull/9232#issuecomment-276385114 and the approval from @cajus

johnspackman:
@cajus that last commit should fix your multiple warnings; however there is still one warning, which is correct but is a warning about Qooxdoo:

```
Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.core.Widget
```

`MenuItem` has a boolean `visible` property which is conflicting with `Widget.isVisible()`

peuter:
My issues are already fixed, I guess the one reported by @mbgonicus here: https://github.com/qooxdoo/qooxdoo/pull/9232#issuecomment-276385114 is the only one left.

johnspackman:
ah, thanks - that's the one Im thinking of!

johnspackman:
time flies (again), here's the final fix for @mbgonicus

framework/source/class/qx/Promise.js
```diff
@@ -0,0 +1,6551 @@
...
+    
+    /**
+     * One-time initializer
+     */
+    __initialize: function() {
+      if (this.__initialized) {
+        return;
+      }
+      this.__initialized = true;
+      window.addEventListener("unhandledrejection", this.__onUnhandledRejection.bind(this));
```
derrell: As discussed on gitter, this can't call `window.addEventListener` because this code could be running in a non-browser environment. Rather, it should call `qx.bom.Event.addNativeListener` which detects whether's there's a way to do it.

derrell (changes requested):

framework/source/class/qx/Promise.js
derrell: Furthermore, I wonder if it makes sense to handle nodejs here?
```
// Handle nodejs
if (typeof process == "object" && typeof process.on == "function")
{
  process.on("unhandledRejection", this.__onUnhandledRejection.bind(this));
}
```

Or whether `qx.bom.Event` should instead be modified? (Modifying qx.bom.Event has much broader implications.)

johnspackman: I've made the fix we talked about for native listeners, but not the one for node (yet) - is there a better way to detect whether node is running, eg protect against some code defining `process` object?  At least it should be a global `qx.core.Environment` test

peuter (approved):

framework/source/class/qx/Promise.js
level420: @johnspackman you could add 
```javascript```
if (typeof window === 'undefined')
```
to your check.

johnspackman: @level420 sure, but I think that those kind of checks should be isolated into a separate environment check, so that there's one, definitive way of checking.  I tend to agree with @derrell that we should look at whether `qx.bom.Event.addNativeListener` should be deciding whether to add a `process.on()` when node is detected, but that IMHO that should be part of a separate PR.  If we do agree to go down that route, then the latest commit would be fixed too 😀

oetiker: and since we already have the structure where we maintain such information that makes certainly a lot of sense.

derrell (approved):

cajus (approved):

cajus:
Very promising. Thanks @johnspackman ;-)

PR #9234, opened 2016-11-22
Fix scroll momentum behaviour

peuter:
With current implementation you cannot continue (or change the direction)  of a momentum scrolling, it always stops on pointerdown. This PR only stops the momentum but does not cancel the gesture. It also stops the currently running momentum scroll when a new one is started (to allow direction changes).

As an addition i stumbled open the unused method `isBelowTapMaxDistance` while analyzing the problem, this PR removes it (there is a `_isBelowTapMaxDistance` method anyway).

### DIFF
--- a/framework/source/class/qx/event/handler/GestureCore.js
+++ b/framework/source/class/qx/event/handler/GestureCore.js
@@ -73,6 +73,7 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
     this.__gesture = {};
     this.__lastTap = {};
     this.__stopMomentum = {};
+    this.__momentum = {};
     this._initObserver();
   },
 
@@ -88,6 +89,7 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
     __rollImpulseId : null,
     __stopMomentum : null,
     __initialDistance : null,
+    __momentum : null,
 
     /**
      * Register pointer event listeners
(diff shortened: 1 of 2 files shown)


cajus (approved):

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -753,22 +765,6 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
...
-     * @return {Boolean} true if distance is below TAP_MAX_DISTANCE.
-     */
-    isBelowTapMaxDistance: function(event) {
-      var deltaCoordinates = this._calcDelta(event);
-      var clazz = qx.event.handler.GestureCore;
-
-      return (Math.abs(deltaCoordinates.x) <= clazz.TAP_MAX_DISTANCE &&
-              Math.abs(deltaCoordinates.y) <= clazz.TAP_MAX_DISTANCE);
-    },
-
```
oetiker: this is a public method so even if it is not used internally you would have to deprecate it prior to removal

oetiker (changes requested):

framework/source/class/qx/event/handler/GestureCore.js
hkollmann: You are right. But the implemention can not work. TAP_MAX_DISTANCE returns an object  )-:

hkollmann: You are right. But the method will not work because TAP_MAX_DISTANCE is an object.

peuter: And `this._calcDelta` does not exist.

oetiker: ah ... ok ... so no one could have used it anyway ... ok

oetiker (approved):

johnspackman (approved):

level420 (approved):

level420:
merging
thank you @peuter

PR #9236, opened 2016-11-28
fix logger to not use the $$hash value when this object has not bee…

peuter:
…n registered in the Registry.

This fixes the problem, that the logging object was not shown in the log string since the new automatic memory management feature has been introduced.

### DIFF
--- a/framework/source/class/qx/test/log/Logger.js
+++ b/framework/source/class/qx/test/log/Logger.js
@@ -89,18 +89,28 @@ qx.Class.define("qx.test.log.Logger",
       qx.log.Logger.clear();
       qx.log.Logger.register(appender);
 
+      qx.Class.define("test.DisposableObject", {
+        extend: qx.core.Object,
+        implement: qx.core.IDisposable
+      });
+
       var obj = new qx.core.Object();
-      qx.core.ObjectRegistry.register(obj);
+      var dispObj = new test.DisposableObject();
       qx.log.Logger.debug(qx.core.Object, "m1");
       qx.log.Logger.debug(obj, "m2");
       qx.log.Logger.debug(qxWeb(), "m3");
+      qx.log.Logger.debug(dispObj, "m4");
+
 
       var events = appender.getAllLogEvents();
       this.assertEquals(qx.core.Object, events[0].clazz);
-      this.assertEquals(obj.toHashCode(), events[1].object);
+      this.assertEquals(qx.core.Object, events[1].clazz);
       this.assertEquals(qxWeb, events[2].clazz);
+      this.assertEquals(dispObj.toHashCode(), events[3].object);
 
       qx.log.Logger.unregister(appender);
+      dispObj.dispose();
+      qx.Class.undefine("test.DisposableObject");
     },
 
 
--- a/framework/source/class/qx/log/Logger.js
+++ b/framework/source/class/qx/log/Logger.js
@@ -582,7 +582,8 @@ qx.Bootstrap.define("qx.log.Logger",
         // to introduce an unwanted load-time dependency
         if (object.$$hash !== undefined) {
           entry.object = object.$$hash;
-        } else if (object.$$type) {
+        }
+        if (object.$$type) {
           entry.clazz = object;
         } else if (object.constructor) {
           entry.clazz = object.constructor;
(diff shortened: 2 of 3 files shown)


level420 (changes requested):
It seems that this fix let's a test fail in travis:

```
Chrome 55.0.2883 (Linux 0.0.0) qx.test.log Logger:testContextObject FAILED
	Expected '6676-1' but found ''undefined''!
		qx.core.AssertionError.constructor:43:38
		qx.core.Assert.__fail:71:19
		qx.core.Assert.assertEquals:177:33
		qx.type.BaseString.assertEquals:87:22
		qx.test.log.Logger.testContextObject:100:12
		qx.dev.unit.TestFunction:118:23
		qx.dev.unit.TestResult.run:237:36
		qx.dev.unit.TestFunction.run:115:18
```

Wouldn't it be better to fix this where the $$hash gets resolved?

... or if not, the test ```testContextObject```has to  be modified to use a class which implements ```qx.core.IDisposable```

peuter:
It is better to not use the $$hash, when is cannot be resolved anyway. You cannot fix this where $$hash gets resolved because the needed ' clazz' information is missing there, because the Logger either stores the $$hash or the `clazz` but not both.

I will take a look at the test, bit I think the test needs to be fixed too.

level420:
@peuter OK. Understood. Then a fixed test would be good. Thank you!

framework/source/class/qx/test/log/Logger.js
```diff
@@ -89,16 +89,24 @@ qx.Class.define("qx.test.log.Logger",
       qx.log.Logger.clear();
       qx.log.Logger.register(appender);
 
+      qx.Class.define("test.DisposableObject", {
```
level420: could you please undefine the class via ```qx.Class.undefine``` to prevent pollution with test classes? Thank you.

level420 (changes requested):

level420 (approved):

cajus (approved):

derrell:
I'm not convinced this is an appropriate change. We have a means, even with automaticMemoryManagement turned on, to get a hash code, via toHashCode(). Should we not simply call that function to obtain the hash code, for better logger debugging, rather than excluding the hash code altogether?

peuter:
The problem is, that the hashcode is used to retrieve the Object here: https://github.com/qooxdoo/qooxdoo/blob/0bcb969/framework/source/class/qx/log/appender/Util.js#L175

And that does not work when automaticMemoryManagement is turned on, leading to a missing classname in the log message.

derrell:
The hash code is very useful in debugging, particularly when the same message is printed numerous times for different objects. How about calling qx.core.ObjectRegistry.register() in this patch so that the normal fromHashCode is available at line Util.js#L175?

peuter:
Would't you loose the benefits from the memorymanagement, when you register all your logged objects?
I would prefer to add the hash value to the clazz structure if available.

derrell:
For those few objects, used for debugging rather than production, I don't see a problem with losing the benefits of the memory management change. It wouldn't be for all objects; only those used in Logger calls. It appears, however, that all objects have a hash code, assigned either via register() or via toHashCode() in qx.core.Object's constructor. The Logger output, therefore, can use entry.clazz.toHashCode() to retrieve that hash code in the output if memory management is in effect.

johnspackman:
The automatic memory management just means that the object is not logged in a central list, so the hash code is important because it could be used as an index in other lookups.  IMHO the logger should store the hashcode always, and then leave it up to the appender to try and resolve it; this would allow the application to add custom appenders to resolve custom hashcode lookups.

I would definitely not call `.register` in the Logger because it means behaviour of the objects would change depending on the debug level in use - for example, my apps download a set of filters for Logger on a per-user basis so that users with issues can be configured to have their log output increased and uploaded to the server.  IE the logger is used in production; it wouldnt be good for the changed configuration of logging to change the memory management.

Eg teh code should be:

```
        if (object.$$hash !== undefined) {
          entry.object = object.$$hash;
        }
        if (object.$$type) {
          entry.clazz = object;
        } else if (object.constructor) {
          entry.clazz = object.constructor;
        }

```

peuter:
I agree with @johnspackman to not use `register`in the Logger. I will change my PR according to his suggestion and let the appender handle all cases to provide the hash code information even if memory management is enabled.

derrell:
Sounds fine to me. As long as the appender displays a hash code for each object, I'm happy.

derrell (approved):

level420:
3 approvals. more than 2 business days since labeled voting.
merging.
thank you @peuter

PR #9237, opened 2016-11-28
Add support for building release-sdk.zip files on travis

cajus:
This PR generates a *qooxdoo-x.y.z-sdk.zip* - if building for a git tag. *x.y.z* is _not_ taken from a potential version encoded in the *tag*. Instead it uses the contents of */version.txt* for that, so it needs to be adjusted accordingly.

I've tested that one on my github qooxdoo clone - maybe the encrypted API-Key needs to be replaced after a merge. Not 100% sure in which context it is stored on travis side. We'll see, I guess.


cajus:
Note that I had to disable the PDF generation, because there's no LaTeX available in the travis container based builds. So the documentation in the release SDK is HTML only.

oetiker:
get texlive on travis:

before_install:
  - curl -L https://github.com/raphink/travis-texlive/releases/download/2015-07-14_05/texlive.tar.xz  | tar xJC ~

cajus:
It would be good if travis would whitelist some more packages. I had to add some more hacks in the script, just because packages are not available.

The proposal above is a big hack, too. There's a texlive directory containing stuff in our home directory after we've extracted the archive. And then? No paths, no environment variables, etc.? It's unlikely that this works out of the box.

I'm not familiar with "what's needed to make latex happily use files from within a single directory". If you've more information to make it work in the first shot, please tell me.

cajus:
Going to build a new PR for that one.

cajus:
Continued in #9240

PR #9238, opened 2016-11-29
Avoid division by zero for broken fonts

cajus:
There are some fonts around that have broken glyphs (i.e. an aspect ration of zero). This mini PR skips broken glyphs instead of throwing exceptions around.

### DIFF
--- a/tool/pylib/generator/output/Script.py
+++ b/tool/pylib/generator/output/Script.py
@@ -128,7 +128,10 @@ class Script(object):
                 for glyphname, code in res.mapping.iteritems():
                   fdsc = "@%s/%s" % (res.alias, glyphname)
                   if not fdsc in result:
-                      result[fdsc] = [result[resid][1], round(result[resid][2] / code[1]), code[0]]
+                      try:
+                          result[fdsc] = [result[resid][1], round(result[resid][2] / code[1]), code[0]]
+                      except:
+                          pass
                 del result[resid][4]
 
         # ExtMap returns nested maps


level420 (approved):

tool/pylib/generator/output/Script.py
```diff
@@ -128,7 +128,10 @@ def createResourceStruct(resources, formatAsTree=False, updateOnlyExistingSprite
                 for glyphname, code in res.mapping.iteritems():
                   fdsc = "@%s/%s" % (res.alias, glyphname)
                   if not fdsc in result:
-                      result[fdsc] = [result[resid][1], round(result[resid][2] / code[1]), code[0]]
+                      try:
+                          result[fdsc] = [result[resid][1], round(result[resid][2] / code[1]), code[0]]
+                      except:
+                          pass
```
oetiker: with this you are blocking all possible errors ... that might make it hard to notice something new and unexpected is going amiss

oetiker (changes requested):

tool/pylib/generator/output/Script.py
oetiker: There is no possibility that another problem could manifest on this line, rendering valid glyphs seemingly invalid ? Any it might take a long time do discover the reason ?

cajus: If we're getting at this point, one should start to add meaningful outputs for every potential "non inclusion" and telling why. Additionally it should output a warning for each non named glyph. Maybe this is what you wanted to say ;-)

Just before getting off topic:

From my current experience with the generator, it mostly fails with two categories: non-informative-error-messages that don't help you to find configuration problems (esp. when working with parts) or non-informative-stacktraces that need deeper inspection.

If we think we need the generator as it is for the next couple of years, I'd address some of these problems - and make it love python 2/3.

oetiker: ok

oetiker (approved):

derrell (approved):

johnspackman (approved):

level420:
4 approvals, more than 2 business days since labeled voting
merging
thank you @cajus

PR #9239, opened 2016-11-30
add resetSource method to qx.html.Label as it is used by qx.ui.basic.Image when using font icons.

peuter:
This PR adds the missing method which is used here:
https://github.com/qooxdoo/qooxdoo/blob/0bcb969/framework/source/class/qx/ui/basic/Image.js#L877

### DIFF
--- a/framework/source/class/qx/html/Label.js
+++ b/framework/source/class/qx/html/Label.js
@@ -131,6 +131,15 @@ qx.Class.define("qx.html.Label",
      */
     getValue : function() {
       return this._getProperty("value");
+    },
+
+    /**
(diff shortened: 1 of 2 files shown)


cajus (changes requested):
Can you add a test for this scenario?

cajus (approved):
Ah. I've been too slow ;-)

johnspackman (approved):

level420 (approved):

derrell (approved):

oetiker (approved):

cajus:
Thanks @peuter.

PR #9240, opened 2016-11-30
Add support for building release-sdk.zip files on travis

cajus:
This PR generates a qooxdoo-x.y.z-sdk.zip - if building for a git tag. x.y.z is not taken from a potential version encoded in the tag. Instead it uses the contents of /version.txt for that, so it needs to be adjusted accordingly.

I've tested that one on my github qooxdoo clone - maybe the encrypted API-Key needs to be replaced after a merge. Not 100% sure in which context it is stored on travis side. We'll see, I guess.


cajus:
Should now generate the PDF based documentation, too.

level420:
Cool! I'm not good in all this travis/github things: Does this mean that if someone creates a new tag that we have an automatic build release package available here: https://github.com/qooxdoo/qooxdoo/releases ?

cajus:
Yes. That's the plan. It's a bit outdated (no PDF, some minor bugs), but you can see a test release (5.2.6) here: https://github.com/cajus/qooxdoo/releases

cajus:
Note that the version comes from "version.txt", not from the tag. So the zip "seems" to have a wrong naming.

level420 (approved):
The result seems to be good! Not fully walked through all travis and other scripts. though.

derrell:
Instead of a "wrong naming" of the zip file, how about naming it with the commitish (short version of the commit) unless it has a tag; and with the tag if it has one. To find the tag:

git describe --exact-match <commit-id>

oetiker:
what has always bugged me abou the qx release naming is that the initial release was without a third 'digit'  I would love to see us follow http://semver.org/ ... so the name for the new release would be 5.1.0 and not 5.1

cajus:
The current way is to use version.txt, because other parts of qx build on version.txt, too. Besides the naming of the branches (RELEASE_x_y_z, etc.) it has the advantage, that you can see what version-zip you have in your download folder. Adding a git hash makes no sense to me for file naming.

derrell:
Then how about the value from version.txt followed by the git hash? I'm trying to distinguish between a various intermediate versions and the actual release of that version, in the file naming.

derrell:
(And only add the git hash if it's not the actual tagged release)

cajus:
I personally think it's good to have the target version in the filename. I don't see any advantage of cryptic hashes there for end users. Looks more than a disadvantage to me. You can see the hash on the github release page, anyway.

But - well, discuss it, and if you all want hashes there, it's easy to change ;-)

derrell:
Yes, I agree. Maybe you didn't understand what I said in my second message. Put the target version in the file name. Just *append* the hash if it's it's not the actual release, so we know both what the target version is AND how to get back to that exact preparatory version. (We can get back to the release version via its tag, so we don't need the hash.)

Alternatively, a date/time stamp could be appended to the target release for preparatory versions, if you think that's easier. I prefer the hash, from which I can pull the actual code or get the date with git-log, but a timestamp would be ok with me too.

cajus:
@derrell the release archive is only build if you tag a version in git. From this point of view, I guess I still don't understand what you're pointing at. There's just one release. No need for dates/tags/etc.

If we build a release archive for every push, then that makes sense. Not sure where to place such a "master" build to make github happy, though.

derrell:
Ah! It was I who misunderstood. I thought we were creating the zip for each version. I think it would be nice to have the latter, but maybe not necessary. I agree, for the tagged, release versions, there is no need for any additional information on the file name. Sorry 'bout that.

cajus:
So we want that for master only, I guess?

oetiker:
the zip could be packed as 'master.zip' but when unpacked unpack to master-hash

cajus:
The problem would be that "master" is no release at all. Is there a mechanism for that? Or do we just want to place it on the website as "latest" and link it?

level420:
Shouldn't we encourage those who want to live on master to use git pull? 
Why isn't https://github.com/qooxdoo/qooxdoo/archive/master.zip enough?

derrell:
I'd like to see "release candidates" that people can download and play with as an SDK rather than via git. For people keeping up with master, I agree, just using git (or .../archive/master.zip) should be fine. We should probably have release candidates and minor releases more frequently, making this new feature more useful, but that's a separate topic of conversation.

cajus:
@level420 I completely agree with you here. There's no real need for it. The difference between master.zip and the release sdk is:

 * pre build documentation as HTML and PDF
 * cleaned source - no tools/admin, documentation sources, git/travis files, etc.
 * unixified source files
 * pre build components (server/website/desktop)
 * pre build showcase
 * pre build framework API
 * pre build website API

If you use *git*, that's no deal at all.

cajus:
@derrell release candidates would have tags from my understanding. So - besides of the discussion if a new release tag should be named v5.1.0 vs RELEASE_5_1 - they would be "in" already.

level420:
@derrell you can just use  https://github.com/qooxdoo/qooxdoo/archive/master.zip without using git.
And if we want to have a release candidate we simply would create a tag named release_5_1_0_rc1 or similar.

oetiker:
@cajus yes a 'master/latest/bleeding'  release on the website for download would be great ... as you point out this is more than just master.zip from git as it contains all the prebuilt stuff ... and it is also a good thing to make sure we can build this whenever a PR is added an extra test so to say

cajus:
Ok. As you like. I don't see why we're doing this, but I've modified the script to place a *master.zip* on the web root. We've to check whether that works after merging.

level420:
BTW: the task of publishing the website and documentation where the fresh released version will be available as ```current``` is done separately, right?

cajus:
Building the release SDK and deploying the site are different tasks/scripts. Deploy will deploy to

  * devel if we're on master
  * x.y.z if we're on a tag
  * current if we're on a tag, and that is the newest one

Deploy will look up the master.zip if applicable and install it additionally.

This is how it should work. If it does in real life needs to be prooven. I've tested that partly, but it's a bit time consuming to test all these cases on a seperate repository/github user.

cajus:
Do we want to give that one a try? So there would be more time before a release to fix things...

johnspackman (approved):

oetiker (approved):

zaucker (approved):
A bit long for a real review, but I guess there is not much damage that couldn't be repaired if it doesn't work out ...

derrell (approved):
I don't really follow everything this is doing. I'll approve because it doesn't appear that it will break anything that won't be obviously broken.

level420:
Full quorum. merging.
Let's see what happens.
Thank you @cajus

Issue #9135, opened 2016-08-26
localize() vs translate()


cajus:
Just found a regression that is caused by #9002. We're now executing _translate()_ on the label value if the method exists. In case of _dynlocale_ it returns a new _LocalizedString_ instance. This string fails to translate a second time because it does its lookups in the wrong message catalogue.

Example:

```
> qx.locale.Date.getMonthName("wide", new Date().getMonth()).toString()
< "August"
> qx.locale.Date.getMonthName("wide", new Date().getMonth()).translate().toString()
< "cldr_month_format_wide_8"
```

I tend to mark a _LocalizedString_ object whether it was born by _localize()_ in order to be able to call the appropriate method in the _Label_ in this case.

Any opinions on that?

PR #9242, opened 2016-12-01
Fix for issue #9135

cajus:
This PR adds a flag to distinguish between the two possible message catalogues, in order to make the translation of CLDR provided objects possible.

See #9135

### DIFF
--- a/framework/source/class/qx/locale/LocalizedString.js
+++ b/framework/source/class/qx/locale/LocalizedString.js
@@ -28,12 +28,14 @@ qx.Class.define("qx.locale.LocalizedString",
    * @param translation {String} The translated message
    * @param messageId {String} The messageId to translate
    * @param args {Array} list of arguments passed used as values for format strings
+   * @param localized {Boolean} True if the string uses localize instead of translate
    */
-  construct : function(translation, messageId, args)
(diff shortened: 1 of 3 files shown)


level420:
@cajus could you please explain a bit the scenario where this is needed and how it is used? Thank you.

cajus:
https://github.com/qooxdoo/qooxdoo/issues/9135 explains the problem.

cajus:
If that's not enough: every time you want to use CLDR translated date (i.e. a weekday or a monthname), and write it out in a label (or the console), you're getting the cldr_foobar string instead of "Sonntag".

That is because qx holds translations in this.__translation, but the CLDR data lives in this.__locales.

translate() only looks in this.__translation.

cajus:
As a result, I'm no storing the information what the source catalogue was - remembering it in translate().

cajus:
@level420 Ok? ;-)

level420 (approved):
Understood!

johnspackman (approved):

derrell (approved):

zaucker (approved):

level420:
4 approvals. more than 2 business days.
merging.
thank you @cajus

PR #9243, opened 2016-12-01
stop momentum scrolling when the window object is gone

peuter:
when you close e.g. a dialog while its content is scrollign with a momentum, the roll event gets fired on an non existing element, which leads to an error.

This fix stops the momentum scrolling when the internal `__window` object is gone.

### DIFF
--- a/framework/source/class/qx/event/handler/GestureCore.js
+++ b/framework/source/class/qx/event/handler/GestureCore.js
@@ -422,7 +422,7 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
         this.stopMomentum(this.__momentum[domEvent.pointerId]);
       }
       // do nothing if we don't need to scroll
-      if ((Math.abs(deltaY) < 1 && Math.abs(deltaX) < 1) || this.__stopMomentum[oldTimeoutId] || !this.__window) {
+      if ((Math.abs(deltaY) < 1 && Math.abs(deltaX) < 1) || this.__stopMomentum[oldTimeoutId] || !this.getWindow()) {
         delete this.__stopMomentum[oldTimeoutId];
         delete this.__momentum[domEvent.pointerId];
         return;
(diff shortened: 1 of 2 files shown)


level420 (approved):

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -422,7 +422,7 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
         this.stopMomentum(this.__momentum[domEvent.pointerId]);
       }
       // do nothing if we don't need to scroll
-      if ((Math.abs(deltaY) < 1 && Math.abs(deltaX) < 1) || this.__stopMomentum[oldTimeoutId]) {
+      if ((Math.abs(deltaY) < 1 && Math.abs(deltaX) < 1) || this.__stopMomentum[oldTimeoutId] || !this.__window) {
         delete this.__stopMomentum[oldTimeoutId];
```
level420: Hmmm... couldn't find where ```this.__window``` is declared or initialized.

level420 (changes requested):

framework/source/class/qx/event/handler/GestureCore.js
peuter: It is declared in `qx.event.handler.Gesture` which is a subclass of `qx.event.handler.GestureCore`, but as I am writing this I wonder if it is possible to access a private member from a subclass?

If not I guess `__window` needs to be changed to `_window`.

derrell: It is not possible, in the `build`  version, to access privates from other classes.

level420: ... or you need a getter 
```javascript
  getWindow : function() {
    return this.__window;
  },
```

level420: See also the failing travis job
```
Running lint...
  - Warning: qx.event.handler.GestureCore (425, 99): Using an undeclared private class feature: '__window'
Lint warnings are treated as errors! Please fix!
```

level420 (approved):

cajus (approved):

zaucker (approved):

cajus:
Merging. Two businessdays and no dissenting votes. Thank you @peuter.

PR #9246, opened 2016-12-05
Issue #9245: Fix spinner focus handling

### LINKED ISSUE #9245: Spinner Focus problem
Hi all,
I think, there is a problem with Spinner object when we use the method '.focus()'.
The object seems to receive the focus but we can't use the keyboard in order to change the value.
(ok with tab key)

Example:
```
// Create a button
var button1   = new qx.ui.form.Button("First Button", "icon/22/apps/internet-web-browser.png");
var mySpinner = new qx.ui.form.Spinner(0, 0, 99);

var doc = this.getRoot();

doc.add(button1,   {left : 100, top  : 50});
doc.add(mySpinner, {right: 100, top  : 50});

// Add an event listener
button1.addListener("execute", function(e) {mySpinner.focus();});
```

cajus:
Do the focus handling more or less analogue like it's done in the DateField.

Fixes #9245

### DIFF
--- a/framework/source/class/qx/ui/form/Spinner.js
+++ b/framework/source/class/qx/ui/form/Spinner.js
@@ -100,7 +100,7 @@ qx.Class.define("qx.ui.form.Spinner",
     }
 
     // CREATE CONTROLS
-    this._createChildControl("textfield");
+    var textField = this._createChildControl("textfield");
     this._createChildControl("upbutton");
     this._createChildControl("downbutton");
 
@@ -118,6 +118,17 @@ qx.Class.define("qx.ui.form.Spinner",
     } else {
       this.initValue();
     }
+
+    // forward the focusin and focusout events to the textfield. The textfield
+    // is not focusable so the events need to be forwarded manually.
+    this.addListener("focusin", function(e) {
+      textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
+      textField.selectAllText();
+    }, this);
+
+    this.addListener("focusout", function(e) {
+      textField.fireNonBubblingEvent("focusout", qx.event.type.Focus);
+    }, this);
   },
 
 
--- a/framework/source/class/qx/ui/form/Spinner.js
+++ b/framework/source/class/qx/ui/form/Spinner.js
@@ -123,7 +123,6 @@ qx.Class.define("qx.ui.form.Spinner",
     // is not focusable so the events need to be forwarded manually.
     this.addListener("focusin", function(e) {
       textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
-      textField.setTextSelection(0,0);
     }, this);
 
     this.addListener("focusout", function(e) {
@@ -831,6 +830,13 @@ qx.Class.define("qx.ui.form.Spinner",
      */
     gotoValue : function(value) {
       return this.setValue(Math.min(this.getMaximum(), Math.max(this.getMinimum(), value)));
+    },
+
+    // overridden
+    focus : function()
+    {
+      this.base(arguments);
+      this.getChildControl("textfield").getFocusElement().focus();
     }
   },
 
--- a/framework/source/class/qx/ui/form/Spinner.js
+++ b/framework/source/class/qx/ui/form/Spinner.js
@@ -123,7 +123,7 @@ qx.Class.define("qx.ui.form.Spinner",
     // is not focusable so the events need to be forwarded manually.
     this.addListener("focusin", function(e) {
       textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
-      textField.selectAllText();
+      textField.setTextSelection(0,0);
     }, this);
 
     this.addListener("focusout", function(e) {
--- a/framework/source/class/qx/ui/form/DateField.js
+++ b/framework/source/class/qx/ui/form/DateField.js
@@ -77,7 +77,6 @@ qx.Class.define("qx.ui.form.DateField",
     // is not focusable so the events need to be forwarded manually.
     this.addListener("focusin", function(e) {
       textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
-      textField.setTextSelection(0,0);
     }, this);
 
     this.addListener("focusout", function(e) {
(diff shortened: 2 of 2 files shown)


patrick-fischer-hirschstein (approved):

framework/source/class/qx/ui/form/Spinner.js
```diff
@@ -118,6 +118,17 @@ qx.Class.define("qx.ui.form.Spinner",
     } else {
       this.initValue();
     }
+
+    // forward the focusin and focusout events to the textfield. The textfield
+    // is not focusable so the events need to be forwarded manually.
+    this.addListener("focusin", function(e) {
+      textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
+      textField.selectAllText();
```
level420: I think selecting all text in the textfield is inconsistent with other widgets like combobox and datefield where only a focus caused by a tab selects all text.

See http://tinyurl.com/zfxldfl

level420 (changes requested):

framework/source/class/qx/ui/form/Spinner.js
cajus: Yeah. Will change it back.

Do you have an idea why it is this way?  I don't see why tab selection differs from click selection.

cajus:
@level420 do you've an idea why it is that way? The tab/click focus difference feels inconsistent to me.

framework/source/class/qx/ui/form/Spinner.js
cajus: I personally would prefer a "cursor after text" in all textfield related widgets.

level420:
@cajus I think this is simply due to different people with different os and ui background implemented different behavior.

I think it would be good to have this consistent.

level420:
What I wouldn't expect to happen, is that if I focus the textfield via a pointer click all text gets selected. I would expect the cursor to be placed at the mouse click position.

framework/source/class/qx/ui/form/Spinner.js
johnspackman: i think tab vs click differs because it's designed around larger text input areas - EG with a click, the implication is that you could be positioning the cursor at the same time as adding focus, and if you had a long text field or multi-lone text area that could be important; when using tab, the idea is that you are tabbing through a form filling in each field, so select-all is useful so that you can review and quickly replace the content.

cajus:
So we've to define what "constistent" is, actually. I've commited the "cursor before text" like in DateField now. If we'd go for "cursor after text", both - the Spinner and the DateField - need a change.

level420:
focus on tab -> select all
focus on click -> place cursor at click position
programmatic focus -> I'm not sure, but maybe also select all

johnspackman:
programmatic focus should not select-all, otherwise you can't switch focus away and then back again eg to a popup

level420:
combobox does not set the selection at all: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/ComboBox.js#L57-L59

This way the cursor positioning is done (I guess) the OS or browser way, which at least on windows is positioning the cursor at the end of the text.

So for me the fix would be to not do any cursor positioning, selection at all:
```javascript
    this.addListener("focusin", function(e) {
      textField.fireNonBubblingEvent("focusin", qx.event.type.Focus);
    }, this);
```

level420:
this should also be fixed in datefield.

cajus:
If you leave out the text selection in the spinner, the cursor will not jump in on programatic focus. I guess that the combo box will suffer from the same problem.

cajus:
They're overriding `focus()` in the combo box to make it work. That is also inconsistent.

cajus:
So either, we're going to override `focus()` in Spinner/DateField and remove the text selection from the handler, or keep the text selection and do not override `focus()` ;-)

cajus:
Can push that for DateField and Spinner. So all widgets I looked at, place the cursor *after* the text on programatic focus.

cajus:
If I recall correctly, that was my first try and it went red in travis - but lets wait for it...

level420:
So here is a playground sample showing the different textfield based widgets with the behavious before this PR:

http://tinyurl.com/hl462ba

cajus:
And this one after:

http://tinyurl.com/z68999c

Not that it shows a different behaviour on Chrome/Firefox. Firefox puts the cursor *before* and Chrome *after* the text ;-)

level420:
@cajus Firefox puts the cursor on windows _after_ the text. So it's OS dependent with firefox. IE puts the cursor after the text.

cajus:
So - whatever - it's quite not consistent over the OSses, but it's now consistent in itself.

level420:
hmmm not on safari OSX. while all other widgets put the cursor after the text, the textarea places it _before_ the text.

cajus:
... which is maybe out of focus for this issue and may be just a bug in Safari. It would not be the first I've spotted there.

cajus:
Or we start to implement  forced "after text" for all browsers in all text widgets.

level420:
After testing on OSX I'd be in favor of forcing a consistent "force cursor after text" on focus for all platforms.

But we should discuss this.

derrell:
Note that any change here will be a backward-compatibility-breaking change so must go into a major release. Either that, or we should provide a function that fixes these things and have the next major version call that function by default.

level420:
@derrell at least the fix for the spinner is a bug fix imho. I don't think it breaks backward compatibility if we fix it.

The same applies for me with datefield, as it reacted inconsistent for some platforms/browsers by not doing the default, but forcing it o set the cursor at the beginning of the text.

So I would then say ot merge the PR as it is now.

level420 (approved):

level420:
The forced unification for all platforms should be a separate PR then.

derrell:
@cajus If we are sure that everyone would consider it a bug fix then sure, merge as is. If not, it should wait for a major release. I'm good with whichever decision the team makes on this.

cajus:
It's a bug fix. Everything else (the fact that it behaves differently on certain browsers/os'es) is stuff that is/was broken all the time.

johnspackman (approved):

derrell (approved):

level420:
3 approvals, more than 2 business days
merging
thank you @cajus

PR #9248, opened 2016-12-07
Make qx.util.ResourceManager.toDataUri more robust

level420:
avoids exception if the resid is not loaded yet and passes it to toUri

### DIFF
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -232,7 +232,7 @@ qx.Class.define("qx.util.ResourceManager",
     toDataUri : function (resid)
     {
       var resentry = this.constructor.__registry[resid];
-      var combined = this.constructor.__registry[resentry[4]];
+      var combined = resentry?this.constructor.__registry[resentry[4]]:null;
       var uri;
       if (combined) {
         var resstruct = combined[4][resid];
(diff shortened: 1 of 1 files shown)


derrell (approved):

framework/source/class/qx/util/ResourceManager.js
```diff
@@ -232,7 +232,7 @@ qx.Class.define("qx.util.ResourceManager",
     toDataUri : function (resid)
     {
       var resentry = this.constructor.__registry[resid];
-      var combined = this.constructor.__registry[resentry[4]];
+      var combined = resentry?this.constructor.__registry[resentry[4]]:null;
```
cajus: Could you add some spaces to make it more readable?

      var combined = resentry ? this.constructor.__registry[resentry[4]] : null;

Would be the way it seems to be used in 99% of all occurences in the framework ;-)

oetiker (approved):

framework/source/class/qx/util/ResourceManager.js
level420: done!

cajus (approved):

johnspackman (approved):

zaucker (approved):

cajus:
Full quorum. Merging. Thanks!

PR #9251, opened 2016-12-09
Transform property

johnspackman:
This fixes a bug introduced in https://github.com/qooxdoo/qooxdoo/pull/9231 where the comparison with the existing value was broken


johnspackman:
closed because of conflicts

PR #9252, opened 2016-12-09
fix bug caused by #9231

johnspackman:
This fixes a bug introduced in #9231 where the comparison with the existing value was broken


cajus (approved):

level420:
@cajus did you test the fix?

cajus:
@level420 that's the line I reverted locally - so, yes.

level420 (approved):

level420:
2 approvals. more than 2 business days.
merging.
thank you @johnspackman

PR #9255, opened 2016-12-13
avoid exceptions in qx.ui.window.Manager if the desktop was reset

level420:
A bit more description:

I'm trying to replace the window manager of a standalone desktop application ```qx.application.Standalone``` with the following code in the ```main``` method:
```javascript
    main : function() {
      this.base(arguments);

      var desktop = this.getRoot().getWindowManager().getDesktop();
      var windowManager = new qx.ui.window.MyManager();
      windowManager.setDesktop(desktop);
      this.getRoot().setWindowManager(windowManager);
```
which is essentially successfull, but the widget queue throws an exception as it seems that with ```this.getRoot().setWindowManager(windowManager);``` the private ```__desktop``` member is set to null while the old window manager is still queued in the widget queue.

As a result an exception is thrown when the widget queue tries to run ```syncWidget```.

This PR _solves_ this by leaving the ```syncWidget``` method if the desktop is not set.

### DIFF
--- a/framework/source/class/qx/ui/window/Manager.js
+++ b/framework/source/class/qx/ui/window/Manager.js
@@ -41,7 +41,16 @@ qx.Class.define("qx.ui.window.Manager",
     setDesktop : function(desktop)
     {
       this.__desktop = desktop;
-      this.updateStack();
+      
+      if(desktop) {
+        this.updateStack();
+      }
+      else {
+         // the window manager should be removed
+         // from the widget queue if the desktop
+         // was set to null
+         qx.ui.core.queue.Widget.remove(this);
+      }
     },
 
 
@@ -86,12 +95,6 @@ qx.Class.define("qx.ui.window.Manager",
      */
     syncWidget : function()
     {
-      // it may occur that the manager was added to the widget queue
-      // and the desktop was removed meanwhile
-      if(!this.__desktop) {
-        return;
-      }
-      
       this.__desktop.forceUnblock();
 
       var windows = this.__desktop.getWindows();
--- a/framework/source/class/qx/ui/window/Manager.js
+++ b/framework/source/class/qx/ui/window/Manager.js
@@ -86,6 +86,12 @@ qx.Class.define("qx.ui.window.Manager",
      */
     syncWidget : function()
     {
+      // it may occur that the manager was added to the widget queue
+      // and the desktop was removed meanwhile
+      if(!this.__desktop) {
+        return;
+      }
+      
       this.__desktop.forceUnblock();
 
       var windows = this.__desktop.getWindows();


johnspackman (approved):

cajus (approved):

derrell:
Is it not a problem that the old window manager is still queued? Will that old window manager be somewhere/somehow removed from the queue?

level420:
@derrell I've investigated a lot on how I could de-queue the old window manager, without success. It is very hard to uncouple the application from the window manager which has a desktop assigned which holds a reference to the window manager. I've tried ```qx.ui.core.queue.Widget.remove(oldWindowManager)``` right before I assigned the new one, but without any effect.

I'd be glad to get another solution to this, but could not find any.

derrell:
Can you provide a simple playground example that demonstrates the problem? I'm uncomfortable leaving that old window manager in the queue and essentially just putting a band-aid on it. Maybe I or someone else can figure out how to get it properly removed? That would be the correct solution to this, right?

derrell (changes requested):
See comments re: getting the old one removed.

level420:
@derrel you're right. It seems that I've not tried all combinations juggling with the old and new window manager and tested the queue removal at the wrong position.

Here is a working playground sample: http://tinyurl.com/z8pj5gc

derrell:
So is that playground sample showing that this PR is unnecessary?

derrell:
Or maybe, the appropriate PR would do that removal automatically when a new one is set?

level420:
What remains is the question if the window manager ```qx.ui.window.Manager``` could/should handle correctly setting it's internal ```__desktop``` to null, by using calling ```setDesktop(null)```.

The interface definition says so: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/window/IWindowManager.js#L32

If yes, then calling ```qx.ui.window.setDesktop``` should be changed like this:

```javascript
    setDesktop : function(desktop)
    {
      this.__desktop = desktop;
      if(desktop) {
        this.updateStack();
      }
      else {
        qx.ui.core.queue.Widget.remove(this);
      }
    },
```

level420:
Yes this PR is obsolete, and the code change to allow to set the desktop of the window manager to null solves the issue without having to call ```qx.ui.core.queue.Widget.remove(oldWindowManager)``` manually.

derrell:
Without looking in detail at what proposed setDesktop() method is doing internally, it appears correct from that top-level view.

level420:
@derrel thank you for your help. closing this one and will create another PR.

level420:
On the other side: what happens if in the situation of the window manager having __desktop set to null is re queued in the widget queue or someone is calling ```qx.ui.window.Manager.updateStack()``` which also re queues the window manager?

Could we avoid in any situation that ```syncWidget``` is called?

I'm re opening this PR as I think it is still valid, but incomplete.

level420:
I've decided to revert the changes in ```syncWidget``` and handle this within the ```setDesktop``` method.

level420:
Closing this PR in favour of #9265

PR #9261, opened 2017-01-01
seperate error messages during image processing

hkollmann:
during Image processing there are two checks: One for the Image file and one calculating the Image size.
To have better error Messages i seperated this two checks.

### DIFF
--- a/tool/pylib/generator/resource/Image.py
+++ b/tool/pylib/generator/resource/Image.py
@@ -333,7 +333,7 @@ class SvgFile(Image):
         except (struct.error, IOError, KeyError):
             pass
         return (width, height)
-22
+
 
 # http://www.obrador.com/essentialjpeg/HeaderInfo.htm
 class JpegFile(Image):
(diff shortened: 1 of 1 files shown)


tool/pylib/generator/resource/Image.py
```diff
@@ -331,7 +333,7 @@ def size(self):
         except (struct.error, IOError, KeyError):
             pass
         return (width, height)
-
+22
```
level420: I'm not really famialiar with python. What does this line mean?

hkollmann: Sorry, only a typo....

johnspackman (commented):
LGTM but I dont know python and it appears that travis is not triggering automatic tests.  @hkollmann have you run unit tests for this change?  It would be great to have the unit tests extended to check for this new error condition

level420:
@johnspackman I don't think we currently have unit tests which test generator.py itself. The testrunner is javascript only.

hkollmann:
That's the problem. The Generator is'nt tested. Otherwise i had run the Tests :-)

johnspackman:
aha, sorry I'd seen some toolchain tests (which test the output of the generator) and assumed that there were more...

level420 (approved):

johnspackman (approved):

derrell (approved):

cajus (approved):

level420:
4 approvals, more than 2 business days
merging
thank you @hkollmann

PR #9262, opened 2017-01-10
Demobrowser generator fix

cajus:
The API of `path.basename` changed in a way that the input of _undefined_ doesn't return _undefined_ anymore. Instead it raises an exception. If the inspected path just contains one */*, it fails here:

```path.basename(fileNameParts[1], '.html')```


cajus:
Broken PR

PR #9263, opened 2017-01-10
Fix demo browser generation for node >= 6

cajus:
The API of `path.basename` changed in a way that the input of _undefined_ doesn't return _undefined_ anymore. Instead it raises an exception. If the inspected path just contains one */*, it fails here:

```path.basename(fileNameParts[1], '.html')```

### DIFF
--- a/application/demobrowser/tool/lib/DataGenerator.js
+++ b/application/demobrowser/tool/lib/DataGenerator.js
@@ -492,7 +492,7 @@
 
       return {
         category: fileNameParts[0],
-        name: path.basename(fileNameParts[1], '.html')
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

cajus:
Looks like this is incomplete. It fails later on. Oh my node...

level420:
@cajus Is this PR ready for voting?

cajus:
Looks so ;-)

level420 (approved):

johnspackman (approved):

derrell (approved):

oetiker (approved):

level420:
more than 2 days since labeled voting. 4 approvals.
merging.
thank you @cajus

PR #9266, opened 2017-01-13
Allow to add async function as listeners

valette:
This is the clean version of the PR, with a lot of help from @oetiker and @level420 

This PR replaces #9225


valette:
Note that in this PR, the test is very simple and does not cover as much as the test for 'normal' functions, as the spy() method used for testing listeners cannot be used anymore.

level420 (approved):

derrell (approved):
I am not yet too familiar with async functions, but I don't see anything in here that looks like it would cause problems.

oetiker (approved):

level420:
more than 2 business days. 3 approvals.
merging.
thank you @valette

valette:
Well, thank _you_ _all_ for maintaining Qooxdoo!

PR #9267, opened 2017-01-14
Fix copyright in footer

hkollmann:
In the current documentation the footer shows the old copyright notice of 1&1. This should be also merged into the 5.0.2 version!

### DIFF
--- a/documentation/manual/source/_theme.indigo/layout.html
+++ b/documentation/manual/source/_theme.indigo/layout.html
@@ -227,6 +227,7 @@
           <img src="http://resources.qooxdoo.org/images/feed.png">
         </a>
       </div>
+<!--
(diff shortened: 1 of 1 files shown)


oetiker (approved):

level420 (approved):

level420:
2 approvals, more than 2 business days, documentation fix only.
merging.
thank you @hkollmann

hkollmann:
Is this enough to get it into the current website?

cajus:
if you mean "5.0.2" by current: it is not possible, because the 5_0_x branch neither supports automatic deploying, nor is it possible to re-build an existing release. It is already released.

So: needs to be done manually. I'll not be able to do that before the weekend.

Issue #9268, opened 2017-01-16
"click" event no longer fired for a middle mouse button click on chrome 55


woprandi:
Hi,

On this example : http://tinyurl.com/hpe2ugf on chrome (55), the "click" event is no longer fired with a middle or right click (only left). I think it's only since chrome 55 but I didn't test with a previous version. I tried with "tap" event but the behavior is the same. I noticed the recent 5.0.2 release but no changes
On Firefox, no issue.

level420:
@woprandi It seems that Firefox is the only browser which fires a right mouse button click (in windows and osx). Tested with Internet Explorer 11, current Edge, Opera 42, Firefox 50 on windows and Safari 10, Firefox 50 on osx.

Could not test middle mouse clicks, as my mouse has only a left and right button.

The documentation for the click event says that it should react on left and middle mouse clicks:
http://www.qooxdoo.org/current/api/#qx.ui.core.Widget~click!event

The right mouse button should only react on the contextmenu event:
http://www.qooxdoo.org/current/api/#qx.ui.core.Widget~contextmenu!event

So the issue title should be changed maybe to something not specific to chrome or version 55 of chrome.

Are you shure that it worked on chrome 54?

level420:
@woprandi as a workaround you could use the `mousedown` event.
See http://tinyurl.com/jln6ks2

http://www.qooxdoo.org/current/api/#qx.event.type.Mouse

woprandi:
@level420 This code worked for a long time (for middle click). Chrome 55 release seems to be a good culprit. I don't know any other reason but I didn't check since it's not trivial to get a old chrome package

woprandi:
@level420 I can confirm it WORKS with Chrome **53** (could not get a 54 release) and NOT on 55

level420:
@woprandi on which os are you?

woprandi:
@level420 Linux

stahl-riesa:
I have to confirm.  Same issue on Windows 7-64 with Chrome 55 (Opera). Middle and right button does not fire an click event.
My user agent string: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36 OPR/42.0.2393.94

level420:
As of the documentation click should never be fired on right mouse click. On richt mouse click it should fire the contextmenu event.

[...]

PR #9270, opened 2017-01-19
Issue #9268: enable auxclick mouse event support and artificial middle mouse button click

level420:
As of v55 chrome support auxclick events, which are fired if the non primary button (the left button) is clicked. As a consequence the click event is only fired if the left mouse button is clicked.

This PR introduces auxclick events in qooxdoo and solves the problem for the missing middle mouse button click event by firing an artificial click event for the middle mouse button.

See issue #9268 and https://github.com/qooxdoo/qooxdoo/issues/9268#issuecomment-273779337 for further reading.


framework/source/class/qx/event/handler/Mouse.js
```diff
@@ -375,6 +377,17 @@ qx.Class.define("qx.event.handler.Mouse",
...
 
+      /*
+       * In order to normalize middle button click events we
+       * need to fire an artificial click event if the client
+       * fires auxclick events for non primary buttons instead.
+       * 
+       * See https://github.com/qooxdoo/qooxdoo/issues/9268
+       */
+      if (type == "auxclick" && domEvent.button == 1) {
+        this.__fireEvent(domEvent, "click", target);
```
derrell: Can we make this configurable? I never knew that middle mouse button would fire "click" and I consider it undesirable. If that's the way it worked before, it should continue to, but It would be really nice to have a property can can have it generate its own "auxclick" event *instead* of the "click" event. (It probably doesn't hurt for it to always generate the "auxclick" event; it's only "click" that I think should allow for only left-click.

level420:
Playground sample http://tinyurl.com/gsvr99v with a test.

framework/source/class/qx/event/handler/Mouse.js
level420: The api documentation says that a click event will be fired on a left and middle mouse button click. See http://www.qooxdoo.org/devel/api/#qx.ui.core.Widget~click!event
The PR does exactly that, it fires an artificial middle-mouse-button-click if a middle-mouse-button-auxclick ist detected.

level420: btw. I never knew that I'm able perform a middle mouse button click with my logitech wheel, by pushing on the wheel ;-)

framework/source/class/qx/ui/core/Widget.js
```diff
@@ -147,6 +147,10 @@ qx.Class.define("qx.ui.core.Widget",
         {@link qx.event.type.Mouse#getButton} for more details.*/
     click : "qx.event.type.Mouse",
 
+    /** Widget is clicked using the non primary button button.
```
derrell: button button

framework/source/class/qx/event/handler/Mouse.js
level420: ... and of course you're able to listen for the auxclick event the usual way ```widget.addListener("auxclick", ...```

derrell: Yup, I understand that the PR is making it do what it is documented to do, and that I'm asking that a new feature be added "while we're in there mucking with it." :-)

level420: Hmm. what is the feature you are asking for. Maybe I did not understand what you mean.

derrell: A new property, maybe "middleButtonProducesClick" that determines whether your simulated "click" event is generated when the native "auxclick" occurs. The new property would default to true, to do exactly what you have there right now, but if it is set to false, then an "auxclick" would not generate a simulated "click".

level420: But then we consequently need to kill the middle mouse button clicks of non auxclick browsers, like firefox and safari if ```middleButtonProducesClick```.

level420: The current PR gives you the situation back we had before chrome 55 AND adds support for auxclick events, a situation we both were not aware of. I think the configurable behaviour is to much.

derrell: Yeah, didn't think about having to prevent it in the other browsers. Good enough. PR looks good.

derrell: BTW, my Chrome is 54, and also exhibits this new behavior of not passing through the click on middle button. So it began earlier than Chrome 55.

level420: Then may I ask for an approval 😁

derrell: From our policy on PRs:
```text
The commit team (members wth write-access to the repository) will generally wait for at least one full business day (Mon - Fri) before beginning voting to accept the PR, to provide time for people to comment.
```

derrell: I'll be glad to approve it. It just seemed premature, with inadequate time for people who happened to not be around on this particular day to have time to comment.

level420: @derrell I'm respecting our PR policy. After you wrote that the PR looks good, I thought that as a result you would approve this PR (in terms of the github PR workflow) which in my understanding is the statement of one member that the PR is ready to be merged. And of course this PR will not be merged until minimum another member approves it and the two business days grace period is over, which will be next Monday evening. Additionally I would never merge my own PR (don't know if this rule is part of our policy).

level420: Ah OK. Now I understand. We have to wait one business day before labeling a PR for voting.
Retracting the voting label now.

level420:
We should discuss if we want to align the click and auxclick event behavior between clients. 

An artificial auxclick event should be fired if a non auxclick client fires middle and contextmenu mouse button click events (currently firefox, safari, internet explorer).

An artificial middle mouse click event should fired for auxclick enabled browsers (chrome and opera, which is already implemented in this PR).

Firefox fires in addition to the contextmenu event a click event for right mouse button clicks, which then should be suppressed.

This PR introduces ```qx.core.Environment.get("event.auxclick")``` which detects auxclick support in clients and can be used to detect if firing artificial auxclick events is necessary.

qx already does such an event  alignment betwen browsers which support or do not support pointer events.

framework/source/class/qx/event/handler/Mouse.js
derrell: Yeah, it just seems to make sense give time for people to comment, before asking for votes and possibly merging before some have even had a chance to review. Thanks. It's *now* been long enough, though. I'll begin voting (and approve!).

derrell (approved):

level420:
Please note issue #9721 which is about aligning click and auxclick behavior between clients/browsers

tiandian:
> Playground sample http://tinyurl.com/gsvr99v with a test.

when i click the left button and right button it works,
but when i click middle button, nothing happens,
my chrome version is 55.0.2883.87 m, win7 x64.

level420:
@tiandian please note that the playground sample runs on master and shows the current behavior where this PR is not merged yet. If you like to test this PR you have to merge it to your local copy of qooxdoo and run it in your local playground.

cajus (approved):

johnspackman (approved):

cajus (approved):

cajus:
Two businessdays passed. Lets merge. Thanks @level420!

PR #9275, opened 2017-01-24
Fix problem with initially attached Commands

### LINKED ISSUE #5974: setCommand() breaks themed button icon (BZ#5906)
_Mustafa Sak (@msak) wrote:_

On buttons with themed icon property, setting of command set icon property to null, so no icon is see able anymore.

_assigned to Martin Wittemann (@wittemann)_

cajus:
This PR fixes three bugs that are *Command* related:

 * The appearance selector got a wrong value if a *Command* is attached to a widget during its construction inside the *_childControlImpl()* method. This is fixed by a temporary $$needsRefresh marker.
 * *MExecutable* bind the *enabled* property and keep widgets from beeing disabled when they need to. "enabled" is skipped now. As a result, all Widgets using *MExecutable* this way ignore the disabled state and were even clickable.
 * *MExecutable* doesn't care about the enabled state of it's parent widget. As a result it was possible to i.e. trigger a disabled Button with a keyboard shortcut.

This playground example showcases the problem: http://tinyurl.com/hyfojkn Both Buttons should look the same and should be (really) disabled. Not just optically. You can see that the right Button does not get the desired styles and even reacts on clicks/Ctrl+Alt+i when it's disabled.

One problem though: qx has an *enabled* property on the *Command* to be able to enable/disable assigned widgets by disabling the related *Command*. We're loosing this feature with this PR - although I don't think that it's a big loss: it's a designt flaw and mixes scopes.

Closes #5974


framework/source/class/qx/ui/core/MExecutable.js
```diff
@@ -154,7 +154,7 @@ qx.Mixin.define("qx.ui.core.MExecutable",
         // remove the old binding
         if (old != null && !old.isDisposed() && ids[property] != null)
         {
-          old.removeBinding(ids[property]);
+          this.removeBinding(ids[property]);
```
level420: Is now this used instead of old because you changed the binding direction here https://github.com/qooxdoo/qooxdoo/pull/9275/files#diff-3ef8954de08e91faabd2c8f4c7474e0cL180 ?

level420: What if `old.isDisposed()` is true? Is binding removal not needed?

cajus: The original binds from the Command to the Button, i.e. I'm asking myself what the reason for this is, actually. That breaks the enabled state for example. So I changed the direction in the commit. The Command still works and the button still works.

I changed the binding and the unbinding direction. this->value and this->old.

cajus: For the disposal, there might be a problem anyway, because the Button will not be notified when the Command is disposed. You're right in the case of checking. One should change the check above and remove the binding in case of an disposed *old*.

But that doesn't help if you dispose the Command and don't tell it to the Button.

level420: Hmm. travis is failing currently.

cajus: Yes. Because the tests check the opposite direction. The should be failed *if* this is the way to resolve the problem.

While the changes fix the issues I've here, I'm not sure if it's the right way. It looks like the original qx code is either broken since 1.6 for this (legal) usecase, or I'm not seeing the point.

level420: Is the binding direction change needed to solve your problem?

cajus: The $$refreshNeeded solves problem 1. I don't understand why the Command is the master for all these properties. That can't be correct, or am I wrong?

level420:
After reading a bit in the source code of `qx.ui.command.Command` I get the impression that the use case is to have an Instance of `qx.ui.command.Command` which is added to multiple buttons. See:

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/command/Command.js#L20-L23

Let's take as an example a command which has a shortcut for paste, e.g. `CTRL-v`. I think the intention was to have all buttons which are attached to this command be configurable via e.g. `command.setEnabled(false)`. 

This way all buttons which are bound to this command get disabled by disabling the command itself. In our example it could be that the paste command is used in a menu button and in a context menu button.

Besides of this explanation, I'm not shure how to solve this. Maybe by manually adding a bind backward from the button to the command.

level420:
I think the PR should not change bind direction, as it would break exactly the functionality described in https://github.com/qooxdoo/qooxdoo/pull/9275#issuecomment-275117335

framework/source/class/qx/ui/core/MExecutable.js
```diff
@@ -82,7 +82,6 @@ qx.Mixin.define("qx.ui.core.MExecutable",
      */
     _bindableProperties :
     [
-      "enabled",
```
level420: But what if a set of buttons which have the same command instance assigned rely on the enable property being forwarded to all those buttons?

cajus: That's disabled. It's a bad idea to do so anyway because of various scopes. You can disable a Command without any problems. But this will not disable the Button.

The way it is designed, does not work: either the widget is ruling the enabled state, or the command. If the command is ruling, it breaks the ordinary disabled state and leaves the button usable, even it *looks* disabled.

level420: Or did I misunderstood the change?

cajus: In other words: the PR makes disabling a Widget and a Command different things. Disabling the Command disabled just the Command. Disabling the Buttons, disables the button and takes care it cannot be executed by a Command when beeing in this state.

That's at least not as broken as it is now.

cajus: Btw - what's the usecase of assigning one Command to different Buttons at a time? Doesn't make sense IMHO.

level420: I already described here https://github.com/qooxdoo/qooxdoo/pull/9275#issuecomment-275117335 what the use case would be: If you have multiple buttons in different positions regular edit menu AND context menu both having a button for eg 'CTRL-V'

cajus: For me that doesn't make much sense, because both would be executed then. Sound like a bit of chaos. Additionally, they're executable even if they're disabled (i.e. by disabling the outer container).

It is not fixable without loosing the enable flag of the Command - or introducing huge changes containing an effective "computedEnabled" property in all Widget instances.

level420: If the "execute" event is fired multiple times or only once depends if you attach the listener to the buttons or the command. The idea is to attach it to the command and add that command to multiple buttons. I've modified your playground example a bit to show this: http://tinyurl.com/gna4fkv .

cajus: Nevertheless, it is not fixable. The Command overrides the enabled state - which is bad, because it shouldn't have the lead there. It should only treat widgets as enabled, if they are supposed to be enabled.

Even if you attach the "execute" to the Command, it doesn't honour the states. So I'd just drop the "Command can enable or disable Button *states*" behaviour. I personally don't expect a Command to disable Widgets. I'd expect a disabled Command not to react on anything. That's the case.

cajus: The problem is that you've *states* on the one side and *enabled* flags on the other side. This shouldn't be mixed up by secondary sources.

level420: Well it doesn't sound to bad if I have a command which is attached to multiple buttons, is able to set all attached buttons to "disabled" via one command.setEnabled(false). 

I've no example for multiple buttons, but the demo browser example http://www.qooxdoo.org/current/demobrowser/#widget~Menu.html does exactly that, it attaches a command CTRL+V to the paste button and disables the command which also disables the paste button. See the source here:
https://github.com/qooxdoo/qooxdoo/blob/master/application/demobrowser/source/class/demobrowser/demo/widget/Menu.js#L133

Similar code would definitely break!

cajus: Well. It *is* broken currently. Command disabling *must not* control the enabled state of buttons that inherited their state from their parent chain.

I personally expect a disabled button to *not* react on anything. Neither clicks nor keyboard shortcuts. But this is the case in the moment.

cajus: Btw.: I know that this would be an incompatible change that must be documented. I've written it in the PR description ;-)

level420: Hmm well then. I'll stay neutral to this PR. If there are others approving this PR I'll go with it. But I'm not convinced.

cajus: Should definitively not be approved until others shared their thoughts - since it changes established behaviour.

Another solution would be to document the implications that using Command has on the UI consistency and shed light on the posibility to trigger disabled features.

derrell: I'm reading the arguments on both sides of this. I have mixed reactions:

- I agree that it should not be possible, with a Command, to affect a disabled widget in any way (with one exception). A Command for "Paste" must not paste into a text field if that text field is disabled. The exception is that setting up a command to enable or disable all of the widgets to which it is attached seems perfectly reasonable.
- It must remain possible for a single command to be attached to multiple buttons. That's how menus and other buttons that execute the same feature work in general, and is a long-time feature of qooxdoo. Breaking that would likely break many, many programs. It's documented and does exactly what it's supposed to.

So I'd propose that the bug that a disabled widget can be affected (other than its enabled state) by a Command be fixed. I'd further propose (and feel very strongly) that nothing be done that would prevent a single Command from being attached to multiple widgets.

derrell: However... disabling a command should not disable all of the widgets to which it is attached. That, I think, was not the intention of Command.

oetiker: I guess the problem depends on who is the 'master' I would assume that the 'widget' is the master ... so if a widget is disabled then there should be no way of activating it ... if a command for activating the widget is disabled, then that should only affect the command and not the widget ... (and no I did not read the whole thread, only my reaction to reading derrels last comment)

framework/source/class/qx/ui/core/MExecutable.js
```diff
@@ -167,6 +168,7 @@ qx.Mixin.define("qx.ui.core.MExecutable",
             // check also for themed values [BUG #5906]
             if (selfPropertyValue == null) {
               // update the appearance to make sure every themed property is up to date
+              this.$$resyncNeeded = true;
```
johnspackman: I don't agree with this - magic member variables that are injected into other classes to invoke special undocumented features should not be allowed.  If delayed call to `updateAppearance()` is necessary for this mixin then there's an argument that it could be necessary elsewhere - after all it is not a requirement that `MExecutable` is used in something that consumes a `Command`, it's simply a very helpful mixin.  

Ideally the variable in `qx.ui.core.Widget` should be private or protected and operated via an API.

johnspackman (changes requested):

framework/source/class/qx/ui/core/MExecutable.js
johnspackman: Isnt the principal of `MExecutable` that it's added to a `Widget` that needs to pull in changes to `enabled` and other properties from the `Command`?  In which case, it's acceptable for the Widget to bind it's enabled flag or any other property to the Widget it's included into - it's just a design decision.

The problem with enabled is a more general one, IMHO, where different parts of the code want to set enabled and have the values merged in some meaningful way.  I have this problem when defining a UI that enables and disables other parts of the UI based on state (eg a `CheckBox` is off, so therefore a `GroupBox` of fields is not editable and has `enabled` set to false); but then there are also user permissions, and if the current user is not allowed to edit any of the fields everything is disabled but the CheckBox is on so the code wants to enable the fields in the GroupBox....

It sounds like @cajus' problem is the same thing, MExecutable's signalling or enabled is conflicting with the enabled setting of other code; but removing it altogether is not the right answer either because there are good reasons for reflecting Command properties in Widget, which is half the reason MExecutable exists in the first place.

level420: Maybe we could make the bindable properties of a command configurable? They are defined in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/MExecutable.js#L83-L92

level420: BTW: `qx.ui.form.CheckBox` overrides the `_bindableProperties` member in order to avoid forwarding of the icon property from the command to the widget. See https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/CheckBox.js#L98-L109

derrell (changes requested):
With this change, are these conditions met?

- the bug, if we agree to call it that, that a disabled widget can be affected (other than its enabled state) by a Command is fixed. 
- a single Command can be attached to multiple widgets.
- it is still possible to create a command that alters the enabled state of all of its attached buttons

If not, I believe we have backward incompatibility issues to address, aside from deciding if the changes are appropriate.

As additional explanation to my prior comments, I do NOT think it's appropriate for the command.setEnabled() to change the enabled state of the command's attached buttons -- command.setEnabled() should only affect the Command itself -- but it should be possible for the Command's listener to change the state of its attached buttons.

johnspackman:
AIUI points 2 & 3 are "yes" (ie unchanged), but the answer to point 1 is that the widget is still updated, it's just that the enabled property is excluded from the list of properties which are changed.

The `Command` isn't pushing changes into the button, the button is pulling them in by choice, because it chose to include `MExecutable`, and because the user application chose to call `.setCommand(...)`.  

IMHO that's the point of at least 50% of the code in `MExecutable`, to copy property values from the `Command` to the `Widget` - the rest of the code is about copying the "execute" event from the button, so you could say that copying `Command` to a button is the *entire* purpose of `MExecutable`.

The properties which are copied are almost all about the UI - `label`, `icon`, `tooltipText`, and `menu`, plus `value` - tie the behaviour of the `Command` to the buttons pretty closely anyway, and `enabled` is just another one of those UI properties.

`MExecutable` already provides a mechanism for the list of properties being copied to be altered on a per-button basis because `_bindableProperties` can be modified to suit, or the user application can avoid it altogether by simply not calling `button.setCommand(...)` and replace it with it's own bindings.

derrell:
I think we may be talking around each other. Am I correct that you are proposing that it is ok, for example, for a Command to execute a button's listener even though the button is disabled?

If I understand the purpose of this PR, it is to "fix" that: to make it impossible for a Command do a button's action if that button is disabled.

I'm just trying to understand the arguments here...

johnspackman:
no, not at all - that should be fixed by checking the enabled flag before firing the event, it's just that making a change to not copy the enabled property is a breaking change, and Im suggesting that it's valid to copy the enabled property just like all the other properties like `label` etc.

level420:
Just to sum up my proposal discussed on gitter.im/qooxdoo/qooxdoo:

I'd vote for an alternative solution, where the command could modify the bindable properties list during instantiation as an optional, second parameter for the constructor. This way you could individually decide to exclude the enabled property from the properties to be bound. If not present, the default defined in qx.ui.core.MExecutable should be used, which is defined here:

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/MExecutable.js#L83-L92

Calling execute of a disabled widget should be prevented, by adding
```javascript
  if(this.getEnabled()===false) {
    return;
  }
```
here: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/MExecutable.js#L99

I'm not sure if it solves all aspects of his PR, but it would solve the need to not have the enabled property of the widget being bound to the command
This would be:
```javascript
var command = new qx.ui.command.Command("Ctrl+Alt+v",  [
      // "enabled",
      "label",
      "icon",
      "toolTipText",
      "value",
      "menu"
    ]
);
```
where the enabled property is not included in the bind-able properties, or you could completely omit binding via
```javascript
var command = new qx.ui.command.Command("Ctrl+Alt+v", []);
```

framework/source/class/qx/ui/core/MExecutable.js
cajus: I tried various ways. That was the only way to achieve it without changing too much. As it is internal like the other $$ variables, I don't see a real problem there.

cajus:
@level420 I don't think that it is intuitive or obvious why one has to list bindable properties. It's just a try to keep the broken stuff going. I'd just remove the "enabled" from the list of bindings as proposed and document it in the changes. I suspect that too many people are using Command recycling.

framework/source/class/qx/ui/core/MExecutable.js
johnspackman: My objection is that $$resyncNeeded is effectively a private state variable for `Widget` and that by accessing it from outside the `Widget` class you're breaking a fundamental "blackboxing" principal of OO.  

While `MExecutable` is not _really_ suited to be used except in a class derived from `Widget` (although it could be), it is perfectly acceptable for a `Widget` to implement it's own `command` property should it choose to do so - for example, if the developer decided that they disagreed with how properties are bound from the command, they might decide to avoid `MExecutable` altogether.  This would be reasonable because that's how the code is packaged at the moment.  However, now that there is a `$$resyncNeeded` property, the developer would be *forced* to use an internal, private state variable to complete the functionality.

Looking at `Widget` in isolation, `$$resyncNeeded` is largely undocumented, and it's increasingly confusing as to what's going on when there is also `.syncAppearance()` (which is called immediately after `MExecutable` sets `$$resyncNeeded=true`) and also `.updateAppearance()`.

I guess my point is that not having deeply looked at this piece of code before, it's apparent that there are some significant and sometime subtle things going on in `Widget` and IMHO that makes it especially important that behaviours of the `Widget` are very clearly defined and make sense all on their own.

level420:
It could also be implemented as a "do NOT bind those properties" array. So having
```javascript
var command = new qx.ui.command.Command("Ctrl+Alt+v",  [ "enabled" ]);
```
would then exclude the enabled property from being bound.

oetiker:
@cajus is the description at the start of the PR in sync which what it does now?

cajus:
@oetiker yes. "enable" is removed from the bindings and a re-trigger flag is added. Maybe it is possible to use a private class member as @johnspackman suggested, but then we've to add some additional flags to the *updateAppearance()* calls.

level420:
@cajus just a short question: when "enable" is removed from the bindings: is then the $$needRefresh appearance refresh still needed?

cajus:
@level420 yes, these are two different issues.

level420:
@cajus what if there are no bindings between the command and the widget. Would then the issue still be there?

cajus:
@level420 if we completely skip all bindings, it works, too.

cajus:
Ping. From my feeling I answered all questions regarding this PR. Any ideas whether we just close it or if I can do anything to increase the acceptance?

johnspackman:
Re-reading all of this I think that the issues raised revolve around how `Widget.enabled` works (including how I'd *like* it to work 😉 ) and are beyond the scope of this PR; fundamentally, the issues listed at the top of the PR are significant issues and should be resolved.  

This PR solves the issues so I'm withdrawing my objection and approving.

johnspackman (approved):

oetiker (approved):
I agree with @johnspackman ship it

level420:
@cajus: As I remember it would stop disabling widgets if the command is disabled, as the `enabled` property isn't bound anymore, right? This is breaking backwards compatibility.
I'm still not fully convinced by this PR and made some proposals on keeping the backwards compatibility which are not really good as I have to admit.
I'll stay neutral for this PR. Let's wait what @derrel says.

derrell:
I'm not sure that my questions in my Feb 3 comment, repeated here, have been answered, so I'm still not sure of my opinion on the PR:

With this change, are these conditions met?

- the bug, if we agree to call it that, that a disabled widget can be affected (other than its enabled state) by a Command is fixed.
- a single Command can be attached to multiple widgets.
- it is still possible to create a command that alters the enabled state of all of its attached buttons

If not, I believe we have backward incompatibility issues to address, aside from deciding if the changes are appropriate.


As additional explanation to my prior comments, I do NOT think it's appropriate for the `command.setEnabled()` to change the enabled state of the command's attached buttons -- `command.setEnabled()` should only affect the Command itself -- but it should be possible for the Command's listener to change the state of its attached buttons.

cajus:
@derrell the complete Command thingie is a big misconception, because you have multiple sources for the *enabled* flag. I don't think that it's possible to change it without breaking the previous behaviour. I even don't think that your 3rd point is valid at all. Or at least I can't see the usecase. There's nothing changed for your 2nd point. I don't understand the first point, though.

Everything else is in the updated issue description. It fixes three bugs, and maybe we should just fix the appearance queue bug. And document the the other annoying "feature" (no way to really disable buttons with commands attached. I'll keep it in the list of patches that I've to apply for our projects. No problem with that. Or at least better than using to much resources in writing about broken button behaviour ;-)

cajus:
Or what about this approach: lets define how we think that the behaviour would be correct. I'll adapt the PR accordingly. Based on the Playground example:

 * Widgets with a *Command* tend to not respect their appearance. That is a wrong behaviour, because we want these widgets look like we want to. Guess that's something we could agree on.
 * Widgets with a *Command* can be disabled programmatically, look disabled but are clickable and reachable for keyboard shortcuts. This is also something which sounds wrong and we could agree on.

The first can be fixed by a patch that doesn't change any behaviour. The second - well. Maybe we can achieve this behaviour:

 * Widgets with Commands that are disabled (from within the widget) are disabled for whatever interaction and look disabled
 * Widgets with Commands that are disabled (from within the Command) are disabled for whatever interaction and look disabled

That would involve some more modification in some places, but maybe @level420 and @derrell are more happy with it? If not, what is *your* expected behaviour in this case?

derrell:
I am liking this approach, as I think I will understand it better.

Firstly, attaching a command to a widget should have absolutely no, zero, zilch effect on its appearance. A command adds keyboard control and potentially an action to a control, but has nothing to do with what it looks like. If the fact that there's a command attached to the widget is what causes the widget to look different, that's a bug that needs fixing. That one's easy.

If a widget is disabled, there is nothing that the mouse or keyboard should be able to do to cause the widget's actions. An attached Command, acting solely as an attached command, has no special privileges in this regard. Also easy.

That said, a Command can have an "execute" listener. Disabling the widget to which a Command is attached does not (and should not) disable the Command itself, so adding a listener on a Command's "execute" event, that happens to have code that enables a disabled widget (whether or not it's a widget to which the Command is attached) is perfectly acceptable.

And to close the loop, the Command itself can also be disabled, in which case its actions on a widget to which it is attached stop happening, as do firing of its events. But disabling the Command does not by itself disable the widget; the two are independent.

That's my feeling about how this stuff should work. Thoughts?

level420:
OK. I'm nearly convinced to accept @cajus ' PR and am following @derrels last comment. What I'm worried about is if we are breaking something or loosing functionality we had before, as `qx.ui.core.MExecutable` is used in 9 framework classes (see https://github.com/qooxdoo/qooxdoo/search?utf8=%E2%9C%93&q=qx.ui.core.MExecutable ), amongst them `qx.ui.form.CheckBox`, `qx.ui.form.RadioButton` and `qx.ui.form.Button`. Ommitting binding the `enabled` properties in could have effects here which are not desireable.
@cajus Are there any negative effects you found regarding those classes or better widgets?

cajus:
I guess the PR does what @derrell suggests. I've basically removed the binding of "enabled". So *Command* and *Button* go their own way for "enabled". If the buttons `execute` is used, it checks whether it's enabled or not. If the commands `execute` is used, it does not check. So you've to disable the command, too.

cajus:
But you're right - this is definitely a different behaviour like it was before.

cboulanger:
@derrell Can you revisit this issue and make a decision? Thank you.

cboulanger (approved):

derrell (approved):

hkollmann (approved):

rad-pat:
Is there some way this can be made backwards-compatible? Buttons are now enabled throughout our application because this behaviour had been relied upon. I need a quick fix rather than having to trawl all modules fixing up the buttons/menus - I'm happy to do that later. 
Perhaps conditionally defining the _bindableproperties such as: `    _bindableProperties : qx.core.Environment.select("qx.legacycommandbehaviour", {
      "true": [
        "enabled",
        "label",
        "icon",
        "toolTipText",
        "value",
        "menu"
        ],
      "false": [
        "label",
        "icon",
        "toolTipText",
        "value",
        "menu"
      ]}),`

johnspackman:
@rad-pat sounds reasonable to me; please create a PR

derrell:
@rad-pat Just so we fully understand what's happening in your case, are you saying that in your code, you depend on `command.setEnabled(false);` to also disable the widget to which the command is attached? If that's not the issue, what behavior were you depending on that became unavailable with this change?

rad-pat:
Yes @derrell, sorry if I wasn't clear. It's exactly as you stated.

PR #9279, opened 2017-01-26
Perf properties

johnspackman:
Adds a baseline performance test for property setters


component/testrunner/testrunner.json
```diff
@@ -161,7 +161,7 @@
         },
         "uris" :
         {
-          "add-nocache-param" : true
+          "add-nocache-param" : false
```
derrell: This feels wrong. I already have difficulty sometimes getting changes to tests to refresh. Without the nocache parameter, things could only get worse, no?

derrell: Or more generally, why are you making this change?

johnspackman: You're right this should probably be a separate pull request and discussion; it snuck in here because I constantly find myself disabling it if I want to debug code in the unit or performance tests because (in Chrome at least) it means that the breakpoints are not saved from one run to another and it's difficult to find the code to set the breakpoint without running.

Personally I can't think of a time when the nocache param has ever been useful - AFAICR it was introduced because of Chrome's aggressive caching, and maybe Chrome's behaviour has changed since then?

I'm happy to back out this change and make a separate PR, although if no-one objects I'd just as happily leave it in :)

derrell: Well, as I said, it's hard enough to get tests to refresh properly even with that option enabled. I'm not eager to make this particular (really unrelated to the PR) change, although I have no problems with the remainder of the PR.

level420: I'll think the caching change should be another PR then. It makes It difficult to approve. If this change is backed out I'll vote for this PR, bringing the promise PR forward.

level420 (changes requested):

derrell (approved):

component/testrunner/testrunner.json
johnspackman: no problemo ... :)

level420 (approved):

cajus (approved):

zaucker (approved):
Please add a newline at the end of the last line before merging ...

level420:
merging
thank you @johnspackman

PR #9282, opened 2017-01-31
fix assertions used in qx.ui.basic.Image font icon code in app built with qx.debug set to false

level420:
assertions should be guarded with ```qx.core.Environment.get("qx.debug")```.

Building with ```qx.debug``` set to false, assertion functions are not available. Calling ```this.assertXxxx``` then leads to an exception because the function is missing.

See https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/core/Object.js#L42

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -918,7 +918,9 @@ qx.Class.define("qx.ui.basic.Image",
         }
         else {
           var charCode = parseInt(qx.theme.manager.Font.getInstance().resolve(source.match(/@([^/]+)\/(.*)$/)[2]), 16);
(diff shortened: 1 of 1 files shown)


cajus (approved):

zaucker (approved):
Perhaps you could add a comment on the PR why asserts should be protected. I think I can guess, but it might be worthwhile to state explicitely.

oetiker (approved):

level420:
@zaucker See https://github.com/qooxdoo/qooxdoo/pull/9282#issue-204291534

johnspackman (approved):

johnspackman:
merging, thanks @level420

PR #9284, opened 2017-02-02
Adds get/setTimeout to qx.io.remote.transport.XmlHttp

johnspackman:
It is sometimes useful to change the timeout when using XML; this can only be done by directly manipulating the XMLHttpRequest before the request is opened, and this API allows the timeouts to be specified.

This method is deprecated from the start because synchronous I/O itself is deprecated in the W3C spec https://xhr.spec.whatwg.org/ and timeouts are indicative of synchronous I/O and/or other server issues.  However, the timeouts API is still supported by many browsers and this API is useful for code which has not made the transition to asynchronous I/O - which is typically a non-trivial task.

By nature this is platform specific, and therefore it is only added to the `XmlHttp` transport; it is arguably only likely to be used in specific applications or low level libraries (such as ServerObjects contrib 😊  ). 

I have also found this useful in resolving ERR_NETWORK_IO_SUSPENDED in Chrome, which appears to be related to sleep mode.

### DIFF
--- a/framework/source/class/qx/io/remote/transport/XmlHttp.js
+++ b/framework/source/class/qx/io/remote/transport/XmlHttp.js
@@ -105,6 +105,10 @@ qx.Class.define("qx.io.remote.transport.XmlHttp",
     
     /**
      * Sets the timeout for requests
+     * @deprecated {6.0} This method is deprecated from the start because synchronous I/O itself is deprecated
+     *  in the W3C spec {@link https://xhr.spec.whatwg.org/} and timeouts are indicative of synchronous I/O and/or
+     *  other server issues.  However, this API is still supported by many browsers and this API is useful
+     *  for code which has not made the transition to asynchronous I/O   
      */
     setTimeout: function(timeout) {
       this.__timeout = timeout;
--- a/framework/source/class/qx/io/remote/transport/XmlHttp.js
+++ b/framework/source/class/qx/io/remote/transport/XmlHttp.js
@@ -96,6 +96,26 @@ qx.Class.define("qx.io.remote.transport.XmlHttp",
      */
     isSupported : function() {
       return !!this.createRequestObject();
+    },
+    
+    
+    /** The timeout for Xhr requests */
+    __timeout: 0,
+    
+    
+    /**
+     * Sets the timeout for requests
+     */
+    setTimeout: function(timeout) {
+      this.__timeout = timeout;
+    },
+    
+    
+    /**
+     * Returns the timeout for requests
+     */
+    getTimeout: function() {
+      return this.__timeout;
     }
   },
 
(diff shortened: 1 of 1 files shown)


derrell:
I see no technical problems with this PR. I do think, though, that we should be highly discouraging any use of the synchronous interface. With some browsers, it blocks the whole browser, and is one of those capabilities that is generally considered "Really Bad (tm)" to use, these days. (I'm very surprised, actually, that ServerObjects would use it. What is the use case in modern code? I take it there is some a use case that you are not able to accomplish with async operations?)

johnspackman:
With ServerObjects, it's up to the caller to decide whether to make the call synchronous or asynchronous - and given the reasonably quick turn around of I/O given modern bandwidth it's not been too much of a problem to do synchronous work in this way, especially if you have a closed audience like a corporate intranet app.

I completely agree that we should not promote synchronous I/O - Safari/Webkit has started ignoring timeouts on synchronous I/O and Chrome is now issuing deprecated warnings; and as you say, a blocked browser is not attractive.

One of the things I found was that as soon as you start implementing some asynchronous functions, there are times when sequencing becomes very important.  This leads to very hard to reproduce problems where 99% of the time everything is fine, but that other 1% the asynchronous functions change sequencing and the code breaks.  The solution is of course to look at things like promises 😉  but the reality is that if you have a large code base this reworking is quite a task.

This change is something that I wrote over a year ago as I started to deal with the fallout of moving from synchronous, and it helps during the transition (so long as you're not on Safari); so it's a sticking plaster, not the cure.  I've marked it as deprecated so that it's clear it's not the right solution long term

derrell:
Sigh. Yes, promises seems the right solution. I'll approve this when voting starts, and hopefully the deprecated indication will prevent people from using it. I hope you're transitioning ServerObjects to promises. :-)

johnspackman:
Already done :)

framework/source/class/qx/io/remote/transport/XmlHttp.js
```diff
@@ -308,6 +332,12 @@ qx.Class.define("qx.io.remote.transport.XmlHttp",
         return;
       }
 
+      // Apply timeout
+      var timeout = qx.io.remote.transport.XmlHttp.getTimeout();
```
hkollmann: should this not be "this.getTimeout()"?

johnspackman: No, because it's static and being called from a member; AFAICR there is 'self' but personally I find that obscure and not distinct so I prefer the longer form

hkollmann: Sorry, did not see the static..

level420 (approved):

patrick-fischer-hirschstein (approved):

oetiker (approved):

cajus (approved):

derrell (approved):

level420:
merging.
thank you @johnspackman

PR #9286, opened 2017-02-08
Fixed not RFC 7230 conform Reason-Phrase handeling/Improved Compatibi…

RaHehl:
Fixed not RFC 7230 conform Reason-Phrase handeling/Improved Compatibility with Tomcat >=8.5 https://bz.apache.org/bugzilla/show_bug.cgi?id=60183 https://github.com/apache/tomcat/blob/fc6f2ead8463d06ae73954508927e4f2224c8e77/java/org/apache/coyote/http11/Http11OutputBuffer.java#L364

### DIFF
--- a/framework/source/class/qx/bom/request/Xhr.js
+++ b/framework/source/class/qx/bom/request/Xhr.js
@@ -1062,7 +1062,7 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
       if (this._getProtocol() === "file:") {
         error = !this.responseText;
       } else {
-        error = !this.statusText && this.status !== 204;
+        error = this.status === 0;
       }
 
       return error;


framework/source/class/qx/bom/request/Xhr.js
```diff
@@ -1062,7 +1062,7 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
       if (this._getProtocol() === "file:") {
         error = !this.responseText;
       } else {
-        error = !this.statusText && this.status !== 204;
+        error = this.status === 0;
```
level420: Does this change really improve compatibility? Or does it only improve conformance to RFC7230, but maybe leads to failing clients which do not conform to RFC7230. Conformance does not allways mean compatibility. Compatibility is something we may also assure for non conforming clients.

But in the end I fear that I'm not familiar enough with Xhr, so others have to give their opinion on this PR.

johnspackman: I think that this is correct; `status === 0` is because the communication failed - eg TCP timeout rather than the server reporting "500 blah".  Unless a non-conforming client is masking a TCP failure as something else I'd imagine it would be fine.  The code previously was inferring failure from no `statusText`, except where it was `status == 204`because that routinely sends no `statusText`.

rhehl: @level420 the reason why i committed this was that we really had the problem that our application stopped working with Tomcat 8.5 because of the missing (but RFC Conform Missing) statusText

level420: @rhehl thank you for your explanation. As this sounds reasonable and with @johnspackman  s comment I'll go with it.

level420 (approved):

oetiker (approved):

johnspackman (approved):

derrell (approved):

level420:
merging.
thank you @rhehl

PR #9287, opened 2017-02-08
Fixed API-Doc warning

schnepe2:
`generate api` spits out a warning, therefore I...
- added missing return value API-Doc for `qx.bom.client.Event#getAuxclickEvent`
- minor clean-up (whitespaces). No functional change.
- removed individual `@internal` tags, 'cause the whole class is marked `@internal`
...and...
- added missing environment checks documentation for "event.auxclick" / qx.bom.client.Event#getAuxclickEvent at `qx.core.Environment`

### DIFF
--- a/framework/source/class/qx/bom/client/Event.js
+++ b/framework/source/class/qx/bom/client/Event.js
@@ -32,9 +32,10 @@ qx.Bootstrap.define("qx.bom.client.Event",
     /**
      * Checks if touch events are supported.
      *
+     * @internal
      * @return {Boolean} <code>true</code> if touch events are supported.
      */
-    getTouch : function () {
+    getTouch : function() {
       return ("ontouchstart" in window);
     },
 
@@ -42,9 +43,10 @@ qx.Bootstrap.define("qx.bom.client.Event",
     /**
      * Checks if MSPointer events are available.
      *
+     * @internal
      * @return {Boolean} <code>true</code> if pointer events are supported.
      */
-    getMsPointer : function ()
+    getMsPointer : function()
     {
       // Fixes issue #9182: new unified pointer input model since Chrome 55
       // see https://github.com/qooxdoo/qooxdoo/issues/9182
@@ -65,9 +67,11 @@ qx.Bootstrap.define("qx.bom.client.Event",
     /**
      * Checks if the proprietary <code>help</code> event is available.
      *
+     * @internal
      * @return {Boolean} <code>true</code> if the "help" event is supported.
      */
-    getHelp : function () {
+    getHelp : function()
+    {
       return ("onhelp" in document);
     },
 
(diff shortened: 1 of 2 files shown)


schnepe2:
...by the way: Shouldn't Travis (CI) detect api warnings? ...treat them as errors?

framework/source/class/qx/bom/client/Event.js
```diff
@@ -32,21 +32,19 @@ qx.Bootstrap.define("qx.bom.client.Event",
     /**
      * Checks if touch events are supported.
      *
-     * @internal
```
level420: @schnepe2 I think this is still true: this methods should only be used internally. Or does this generate a warning in the API viewer?

framework/source/class/qx/bom/client/Event.js
```diff
@@ -32,21 +32,19 @@ qx.Bootstrap.define("qx.bom.client.Event",
...
      * @return {Boolean} <code>true</code> if touch events are supported.
      */
-    getTouch : function() {
+    getTouch : function () {
```
level420: I don't think this is usual code styling in qooxdoo. A short investigation shows that there is not space between the keyword function an the brackets in the rest of the code.

framework/source/class/qx/bom/client/Event.js
```diff
@@ -108,45 +104,50 @@ qx.Bootstrap.define("qx.bom.client.Event",
...
+    getCustomEvent : function ()
+    {
       if (!window.CustomEvent) {
         return false;
       }
       try {
         new window.CustomEvent("foo");
         return true;
-      } catch(ex) {
+      } catch (ex) {
```
level420: this is also not a common coding style in qooxdoo.

level420:
@schnepe2 thank you for this PR. There are a lot of formatting changes in this PR which do not belong to the problem you have solved with it. I think the only part needed to solve this one is: https://github.com/schnepe2/qooxdoo/blob/3066c55313875ed2fcfec90b11daae7eb3843a0a/framework/source/class/qx/core/Environment.js#L359-L362

Please split your PR in two PRs, one which only adds the missing doc part for `qx.bom.client.Event#getAuxclickEvent` and another one with your proposals for the changes to the coding style, so we can discuss this independently.

Regarding the travis runs: yes it may be good to have travis failing on api viewer generation. If you like to help us implementing this: you're very welcome!

schnepe2:
@level420 O.K. I will revert the white-space changes
Although I find this critics a little bit wrong here;
The file was committed with...
- missing api doc (warning!) 
- some methods marked `@internal` some not
- trailing white spaces
- inconsistent coding style (braces for example) all over the place
...and I got criticized because of **this**?

schnepe2:
> Regarding the travis runs: yes it may be good to have travis failing on api viewer generation. If you like to help us implementing this: you're very welcome!
Sorry, but setting up Travis is nothing I can help with

level420 (approved):

derrell (approved):

level420:
documentation only fix. 2 approvals.
merging.
thank you @schnepe2

PR #9288, opened 2017-02-11
add some test cases for qx.ui.form.TextField

hkollmann:
I added a test class for  qx.ui.form.TextField together with some tests.


framework/source/class/qx/test/ui/form/TextField.js
```diff
@@ -0,0 +1,58 @@
...
+  members:
+  {
+    setUp: function() {
+      this.__field = new qx.ui.form.TextField();
+      this.getRoot().add(this.__field);
+    },
+
+    tearDown: function() {
+      this.__field.destroy();
+      this.base(arguments);
```
level420: I think it would be good to àlso have
```javascript
this.__field = null;
```
here.

hkollmann: I can do this. 
In the other test the variable isnt't set to null. So i did not do this.

level420 (approved):

patrick-fischer-hirschstein (approved):

framework/source/class/qx/test/ui/form/TextField.js
```diff
@@ -0,0 +1,59 @@
...
+      this.assertEquals(4, l);
+      this.__field.resetMaxLength();
+      var l = this.__field.getMaxLength();
+      this.assertEquals(Infinity, l);
+    },
+	
+
+    __field: null
+  }
+});
```
johnspackman: I think that we always ask for a newline at the end of a file?  I don't know why but I've been asked a couple of times, so maybe it's a lint thing (which does not necessarily apply to unit tests)

johnspackman (approved):

framework/source/class/qx/test/ui/form/TextField.js
derrell: Yes, some editors automatically add it, always. Others don't do so automatically. To avoid ending up with diffs that are solely the addition of a newline, it's nice to just always have it there in each commit. It's also defined by the POSIX standard. See the checked answer in this stackoverflow message: http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

derrell (changes requested):

oetiker (approved):

derrell (approved):

level420:
merging.
thank you @hkollmann

PR #9292, opened 2017-03-02
replace context parameter

peuter:
Something I just stumbled upon: I think this is some kind of copy & paste error which causes the native event listener to be added in the capturing phase. If this is a wanted behaviour, we should change the last parameter from `this` to `true`. Otherwise remove it as suggest in this PR.

I don't really know how much impact this change would have on existing applications, so this change needs to be discussed first.

### DIFF
--- a/framework/source/class/qx/event/handler/Gesture.js
+++ b/framework/source/class/qx/event/handler/Gesture.js
@@ -129,7 +129,7 @@ qx.Class.define("qx.event.handler.Gesture",
       // list to wheel events
       var data = qx.bom.client.Event.getMouseWheel(this.__window);
       this.__fireRollWrapped = qx.lang.Function.listener(this._fireRoll, this);
-      qx.bom.Event.addNativeListener(data.target, data.type, this.__fireRollWrapped, this);
+      qx.bom.Event.addNativeListener(data.target, data.type, this.__fireRollWrapped);
     },
 
     /**
--- a/framework/source/class/qx/event/handler/Gesture.js
+++ b/framework/source/class/qx/event/handler/Gesture.js
@@ -129,7 +129,7 @@ qx.Class.define("qx.event.handler.Gesture",
       // list to wheel events
       var data = qx.bom.client.Event.getMouseWheel(this.__window);
       this.__fireRollWrapped = qx.lang.Function.listener(this._fireRoll, this);
-      qx.bom.Event.addNativeListener(data.target, data.type, this.__fireRollWrapped);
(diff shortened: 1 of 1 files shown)


level420:
As `this` is defined in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/Gesture.js#L132 and https://github.com/qooxdoo/qooxdoo/blob/68a9fd0/framework/source/class/qx/bom/Event.js#L87 ``!!useCapture`` is used, I assume the correct fix would be to change the fix to:
```javascript
      qx.bom.Event.addNativeListener(data.target, data.type, this.__fireRollWrapped, true);
```

peuter:
Yes as I said, using `true` would fix this without changing the behaviour, but I think this behaviour has been introduced by accident and not on purpose.

level420:
It's in there from start: https://github.com/qooxdoo/qooxdoo/commit/56a4840c23c312ad57670f361447c16dacd2242b#diff-89346d402646fd82486d8a77f9598e28

level420:
We could introduce tighter checking on parameter 4 of `qx.bom.Event.addNativeListener` like this:
```javascript
if(useCapture !== undefined && useCapture !== false, && useCapture !== true) {
  // e.g. throw an exception
}
```
this way we could detect such situations.

level420 (changes requested):
I currently don't have the use case for gestures and not the time to investigate what the possible negative effect of interpreting `this` as `true` for useCapture may be.
Therefore I think the fix for now should be to replace `this` by `true`.
Additionally It simply seemed to work a long time as this was introduced nearly 3 years ago.

johnspackman:
I agree it looks like a typo, but if we don't have any tests/experience that show that a falsey-value works fine, and we do not have a case where a truthy-value is causing a problem, then it's difficult to justify making it something other than `true`.

level420:
@peuter Just "highjacked" your PR and replaced `this` by `true`.

level420 (approved):

johnspackman (approved):

peuter:
Didn't know that is even possible. Just imagine that I did it myself and we do not hurt your contribution rules ;-)

derrell (approved):

peuter:
Maybe we should add a comment to the changed line, that the event is added to the capturing phase for historical reasons and a link to this ticket?

oetiker (approved):

cajus (approved):

johnspackman (approved):

level420 (approved):

level420:
5 approvals. 
merging.
Thank you @peuter

Issue #9294, opened 2017-03-03
Add possibility to disable promises in build


cajus:
For applications that need to look at the memory they consume or the data that needs to be transferrerd, it would be nice to have some *qx.promises* flag to disable their inclusion. If there's a way to detect the use of pomises (i.e. directly or thry async properties), an automatic detection would be even nicer (maybe in post *generator* times).

level420:
Fixed via https://github.com/qooxdoo/qooxdoo/pull/9295

PR #9295, opened 2017-03-03
Allows promises to be disabled and eliminated from the build

johnspackman:
Please see issue https://github.com/qooxdoo/qooxdoo/issues/9294, this patch allows an environment key called `qx.promise` to turn off the `qx.Promise` from the main body of the Qooxdoo framework.  

This means that properties and events will no longer support promises, and will mean that provided you do not use `qx.Promise` yourself, it will be eliminated from the `build` target.  The `source` target will still include `qx.Promise` (use the `exclude` key of config.json to force exclusion completely)

Note that this does not force `qx.Promise` to be excluded from the build, it merely makes sure that Qooxdoo framework itself does not require it to be included.  

If `qx.promise == false` and your code (or contribs etc) reference `qx.Promise`, the class will still be included and will work, but will not be integrated into the properties or events; a warning will be output in this case.

### DIFF
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -776,6 +776,9 @@ qx.Class.define("qx.Promise", {
       }
       this.__initialized = true;
       qx.bom.Event.addNativeListener(window, "unhandledrejection", this.__onUnhandledRejection.bind(this));
+      if (!qx.core.Environment.get("qx.promise")) {
+        qx.log.Logger.error(this, "Promises are installed and initialised but disabled from properties because qx.promise==false; this may cause unexpected behaviour");
+      }
     },
     
     /**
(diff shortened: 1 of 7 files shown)


level420 (approved):
I've done some tests with one of my apps, setting `"qx.promise" : false,` in `config.json`.
Size of build version before: 3688 kByte, after: 3593 kByte 
As I'm serving a statically gzipped version of my app, the gain reduces from 95 kByte, to compressed 29 kByte over the wire.
So far it seems that everything is working as expected with no negative impact on my app.
Approving.

oetiker (approved):
I have not actually run the code, but from looking through it, it seems reasonable

level420 (approved):

level420:
@cajus would be nice if you have some time to review this PR, as it was created by your request.

level420:
2 approvals, 14 days since labeled VOTING
merging.
Thank you @johnspackman

Issue #9303, opened 2017-03-21
Improve mouse wheel normalization code to adjust scrolling sensitivity across browsers


vvandens:
This issue has been exposed on [Stackoverflow](http://stackoverflow.com/questions/42872306/adjusting-qooxdoo-scrolling-speed-sensitivity) with the premise of a solution by @level420 :

I'm wondering if it is possible to adjust the scrolling sensitivity in qooxdoo applications using qx.ui.container.Scroll or any internally scrollable components, e.g. qx.ui.table.Table.

What I'm experiencing on my MacBook Pro running Sierra 10.12.3 is a blazing fast scrolling speed using the trackpad, and it only appears to be the case in qooxdoo-based applications which makes them very difficult to use. Tried different browsers with comparable results (Chrome, Safari, ...).

As an example, open the [API browser](http://www.qooxdoo.org/5.0.2/api/index.html) and scroll the browsing API tree on the left.

level420:
@vvandens thank you for creating this issue.

The following thread in the apple forum gives some additional information https://discussions.apple.com/thread/7679256?start=0&tstart=0

level420:
@vvandens could you please verify if the following hint solves this?
https://discussions.apple.com/message/30700725#30700725

Just to be sure it's not the problem described in that thread.
Thank you.

vvandens:
@level420 Unfortunately, it doesn't solve anything. I'll take some time to play with the code and see how it goes.

vvandens:
I've taken a quick look to the code. My (little) understanding of the problem is that the way `WheelEvent`s are handled cannot be directly applied to the MacBook Pro trackpad events. Don't know however if the same problem arise on other hardware/os.

After some debugging, I found that the constant `qx.event.handler.GestureCore.ROLL_FACTOR`, that is hardcoded to `18`, is used to determine the number of pixels to actually scroll depending on the number of mouse wheel ticks (i.e. 18 pixels / tick ?). However, the mbp trackpad fires a lot of such `WheelEvent`s; much more than a mouse wheel would actually do which explains the very high perceived sensitivity.

[...]

PR #9304, opened 2017-03-22
#9303 : Improve mouse wheel normalization code to adjust scrolling sensitivity across browsers

vvandens:
This is a fix for #9303
The detection is dynamic and inspired from this [SO solution](http://stackoverflow.com/questions/10744645/detect-touchpad-vs-mouse-in-javascript/31191250#31191250).
The difference here is that the detection is dynamic, which means that the user can use both a trackpad and a mouse alternatively in the same application. The scrolling sensitivity will adjust accordingly.


johnspackman (approved):

johnspackman:
I think that this is a good solution; my only question is around the resetting of the detection - for example, it might be better to have a longer timeout in `TOUCHPAD_WHEEL_EVENTS_TIMEOUT`, but that's easy for a user to change and I think some experience will be necessary to fine tune these parameters.  

Thanks @vvandens 👍

vvandens:
Thx @johnspackman . BTW, I don't know about the back-porting policy but the change can be safely cherry-picked to the 5_0_x branch

level420:
I'll apply this PR locally and see if it has any negative impact on my apps.

level420:
I've done some fast tests on chrome 58 windows 7. I'm using a logitech mouse with a wheel which rotates stepwise in ticks (hope I've described this correct).

From the first impression it seems that I need to rotate the wheel a few steps before scrolling starts, so for the first ~ 0.5sec to 1sec or ~ 8 rotation ticks nothing happens. Then scrolling starts to work normally. After pausing wheel scrolling a while, the scroll start delay is again there.

It seems to be a combination on how fast I'm initaly turning the wheel and how man ticks im turning it.

If I didn't know about your PR I'd think my mouse has a problem.

cajus (changes requested):
That changes scrolling drastically - at least on linux/chrome. The first two mouse wheel clicks nothing happens at all. After that I've micro steps so that I've to rotate the mouse wheel hundreds of times to scroll 5cm. If we change something here, it should not change the normal behaviour.

cajus:
Same behaviour with the touch pad. Normally it scrolls as fast as I move the fingers. Now I've to swipe ten to twenty times to scroll a little bit.

level420 (changes requested):
Maybe this PR should be specific to OSX sierra.

See https://github.com/qooxdoo/qooxdoo/pull/9304#issuecomment-288684677

cajus:
Another observation: scroll speed differs if you switch between mouse wheel and the touch pad. If the touch pad is used, the mouse scrolling gets very very very slow.

cajus:
@level420 just testet on my mac with the latest and greatest Safari/OS X. It doesn't seem to be Mac specific, because the scrolling is very slow there, too. I've to swipe over the touchpad over and over again to make it move a little bit.

cajus:
On windows 10 / edge it is scrolling lightning fast/with big jumps. Only if you move the fingers very slowly, I'm getting normal scrolling behaviour.

Windows 10 / chrome and firefox don't make it possible to scroll slow. Its always lightning fast, so that it's not easy to scroll list content to a special location.

level420:
@cajus your last test on windows 10 and a trackpad: is this with or without the PR?

vvandens:
OK. So we have to review the detection code and adjust the various constants. The initial delay might be the one that is required to count the wheel events and decide if it's a trackpad or a mouse. But it shouldn't exceed 100ms which is the TOUCHPAD_WHEEL_EVENTS_PERIOD constant. If less than TOUCHPAD_WHEEL_EVENTS_THRESHOLD (10) wheel events occur during this period, then it's a mouse wheel, if more, it's a trackpad.
When the detection is done, the scrolling factor is applied accordingly to the delta, i.e. WHEEL_ROLL_FACTOR (18 as before the PR) or TOUCHPAD_ROLL_FACTOR (2).
Then, when no wheel event occurs during the TOUCHPAD_WHEEL_EVENTS_TIMEOUT (500ms), the detection code is re-armed and the next wheel event will re-trigger it (100ms delay again).

Some ideas that we have to dig :
- Do we want the user to be able to switch between trackpad and mouse during the same application session ? If not, the detection code should occur only once and the roll factor might be kept for the whole application life. If yes, then we have to find a way to reduce to its minimum the initial scroll latency (and understand why @level420 experienced a 0.5-1s delay while it should not exceed 100ms).
- I seems that @cajus experienced a failing detection re-run since the roll factor applied was not correctly reset to the device used after a switch.
- Maybe it's also mouse-dependant ? high-res mouse wheels ?
- Regarding the TOUCHPAD_ROLL_FACTOR, it may need some adjustment. A value of 2 is maybe too slow.

level420:
Still if trackpads on other platforms don't have this issues we should maybe omit completely trackpad detection on those platforms.

It would be good to have some tests with trackpad scrolling WITHOUT this PR applied.

cajus:
@level420 w10 with PR-patched qx of course.

oetiker (commented):
sorry to come late to the party ... but looking at the PR description it is not entirely clear what we are trying to solve with this patch ... is touchpad interaction BAD with qooxdoo? worse than with other browser based apps ? or is it bad in general and this patch tries to make it better for qooxdoo ? is the problem limited to certain browsers oses ?

level420:
@oetiker please read https://github.com/qooxdoo/qooxdoo/issues/9303 and http://stackoverflow.com/questions/42872306/adjusting-qooxdoo-scrolling-speed-sensitivity

level420:
@cajus is there any problem with trackpad scrolling WITHOUT this PR?

cajus:
No, I never noticed some odd behaviour on either platforms.
> 
>   
>   
> 
> 
> 
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c5
> 5493e4bb","name":"GitHub"},"entity":{"external_key":"github/qooxdoo/q
> ooxdoo","title":"qooxdoo/qooxdoo","subtitle":"GitHub
> repository","main_image_url":"https://cloud.githubusercontent.com/ass
> ets/143418/17495839/a5054eac-5d88-11e6-95fc-
> 7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent
> .com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-
> b52498112777.png","action":{"name":"Open in
> GitHub","url":"https://github.com/qooxdoo/qooxdoo"}},"updates":{"snip
> pets":[{"icon":"PERSON","message":"@level420 in #9304: @cajus is
> there any problem with trackpad scrolling WITHOUT this
> PR?"}],"action":{"name":"View Pull
> Request","url":"https://github.com/qooxdoo/qooxdoo/pull/9304#issuecom
> ment-288744540"}}}

vvandens:
@oetiker If you have a macbook pro under sierra (10.12.x), you can try the current [widget bowser](http://www.qooxdoo.org/current/widgetbrowser/) either lists, tables, trees and you will see that scolling with the trackpad (2 fingers) is blazing fast making it very difficult to use. Even the outermost scrollpane is. Scrolling with a real mouse wheel is not. That's what I was trying to actually fix...

cajus:
@vvandens I've a current Mac Book Pro and I can't find anything that behaves bad. Strange.

vvandens:
But as I mentioned it in the bug, I don't know if it's a Mac only problem, a Sierra only problem, or any other.

vvandens:
@cajus that's strange... On Sierra also ?

cajus:
If Sierra is the latest release, then yes.

vvandens:
[tree_scrolling.mov.zip](https://github.com/qooxdoo/qooxdoo/files/865293/tree_scrolling.mov.zip)

Here is a sample video of the result of my 2 fingers on the mbp trackpad. Its only 1-2 millimeters move.

vvandens:
Then it might be tied to the trackpad configuration.

cajus:
Will check if I can see if I’ve any special settings when I’m back to the Mac...

vvandens:
Another piece of information: when browsing the current API browser, only the scrolling of the API tree is blazing fast. The main pane scrolling is fine. Haven't dug into the code, but I believe they are constructed differently (maybe with a native scroll for the main pane ?)

oetiker:
@vvandens that sounds promising, so maybe the problem is that the implementation of qooxdoo is too naive and we should use a different approach, using native interaction somehow ... sorry, just speaking theoretically ...

PR #9306, opened 2017-03-25
Issue #9303: Improve mouse wheel normalization code to adjust scrolling sensitivity for touchpads

### LINKED ISSUE #9303: Improve mouse wheel normalization code to adjust scrolling sensitivity across browsers

vvandens:
- Touch pad detection limited to Mac OS after Sierra (>= 10.12). But still don't know if other combinations should be included.

- Events occurring during detection are now buffered and scheduled for fire after the detection period ends if no other wheel events come. This will prevent single wheel events to be swallowed and physical wheels to seem broken.

- If a touch pad is detected, the delta normalization code is ignored and a factor of 1 is applied to the delta of the DOM event. This lets the native delta computation happen. My understanding is that the `qx.util.Wheel.FACTOR` should not be used for the Macbook trackpad (and for others ?). It is computed (and reduced) everytime the `qx.util.Wheel.MAXSCROLL` increases. And it is *one way only*. If a user swipes fast, then the `MAXSCROLL` increases a lot and the `FACTOR` decreases a lot which makes the scrolling sensitivity of the trackpad lowered a lot for the whole session. There is no way to come back. I think this computation is certainly useful for fixed-delta physical wheels, but cannot be applied when the delta varies.

- Constants have been reviewed. The TOUCHPAD_WHEEL_EVENTS_TIMEOUT has been set to 5s in order to trigger a new detection process.

- A new qx.core.Environment entry (qx.debug.touchpad.detection) has been  added to print touchpad debug messages. It is set to false by default.

Since the code is now limited to Mac OS Sierra and above, it wont happen obviously on other platforms. However, I still don't know if this should not be more largely applied (I think so actually). In order to remove the platform limitation for a test, just patch the new method `qx.event.handler.GestureCore._isTouchPadDetectionEnabled` to return true.


framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -682,27 +718,125 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
...
+    },
+
+    /**
+     * Is touchpad detection enabled ? Default implementation activates it only for Mac OS after Sierra (>= 10.12).
+     * @return {boolean} true if touchpad detection should occur.
+     * @internal
+     */
+    _isTouchPadDetectionEnabled: function () {
+      return qx.core.Environment.get("os.name") == "osx" && qx.core.Environment.get("os.version") >= "10.12";
+    },
```
level420: I think this should be a comparison against a number `10.12`, like it's done in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/webfonts/Manager.js#L216-L247

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -59,7 +59,31 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
      * @type {Integer} Factor which is used for adapting the delta of the mouse wheel
      * event to the roll events,
      */
-    ROLL_FACTOR : 18
+    WHEEL_ROLL_FACTOR: 18,
```
level420: Theoretically, this is an API change. `qx.event.handler.GestureCore.ROLL_FACTOR` vs. `qx.event.handler.GestureCore.WHEEL_ROLL_FACTOR` although it is very unlikely that someone uses this outside of the framework.

level420:
This PR "feels" much better than the previous. No negative impact on windows and a slight, but still a noticable delay on first mouse wheel turn on safari under osx 10.12.

cajus:
Will try it this evening on various systems!

framework/source/class/qx/event/handler/GestureCore.js
vvandens: @level420 Is there something I should do about this ?

framework/source/class/qx/event/handler/GestureCore.js
vvandens: @level420 Good. Didn't know about the `os.version` format. Are we sure that it is always parseable as a number ? e.g. there is never `x.y.z` versions reported ?

level420: I just searched in the framework and only found comparisons against numbers. This should work for OSX as we explicitly map the version: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/client/OperatingSystem.js#L128-L155

framework/source/class/qx/event/handler/GestureCore.js
level420: I'm not sure. This is just a comment so other reviewers could discuss this. To preserver compatibility it would be best to leave `qx.event.handler.GestureCore.ROLL_FACTOR` as is and use it and only add the new `qx.event.handler.GestureCore.TOUCHPAD_ROLL_FACTOR`.

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -682,27 +718,125 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
...
+              this.__endTouchPadDetection();
+            }
+          }
+          if (this.__touchPadDetectionPerformed) {
+            if (this.__rollEvents.length === 0) {
+              this.__rollEvents.push(domEvent);
+            }
+            // Detection is done. We can now decide the roll factor to apply to the delta.
+            // Default to a real mouse wheel event as opposed to a touchpad one.
+            var event = this.__performAdaptativeRollScrolling(target);
```
level420: Travis CI and the linter of the build process complains about `var event` not being used. This should be corrected as else travis will never succeed. See: https://travis-ci.org/qooxdoo/qooxdoo/jobs/214923160 (at the end of the log).

level420 (changes requested):

framework/source/class/qx/event/handler/GestureCore.js
derrell: I agree. Although the new name is more descriptive, which I almost always prefer, in this case, backward-compatibility should trump that.

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -91,6 +121,12 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
     __initialDistance : null,
     __momentum : null,
 
+    __rollEvents : null,
+    __rollEventsCountStart : null,
+    __rollEventsCount : null,
+    __touchPadDetectionPerformed : null,
+    __lastRollEventTime: null,
```
derrell: rollEventsCountStart, rollEventsCount, touchPadDetectionPerformed, and last RollEventTime are all intended to hold non-reference types. They should not be initialized to null. Rather, they should be initialized to 0 for the numeric ones, and false for the boolean one... and then there's no need (although it doesn't hurt) to initialize them in the constructor.

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -682,27 +718,125 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
...
+     * Ends touch pad detection process.
+     */
+    __endTouchPadDetection: function () {
+      if (this.__rollEvents.length > qx.event.handler.GestureCore.TOUCHPAD_WHEEL_EVENTS_THRESHOLD) {
+        qx.util.Wheel.IS_TOUCHPAD = true;
+      } else {
+        qx.util.Wheel.IS_TOUCHPAD = false;
+      }
+      if (qx.core.Environment.get("qx.debug.touchpad.detection")) {
+        qx.log.Logger.debug(this, "IS_TOUCHPAD : " + qx.util.Wheel.IS_TOUCHPAD);
```
derrell: Is there a reason for using the long-form `qx.log.Logger.debug(this, ...` rather than the simpler, `this.debug(...` ?

framework/source/class/qx/event/handler/GestureCore.js
```diff
@@ -682,27 +718,125 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
...
-        domEvent.delta = {
-          x: qx.util.Wheel.getDelta(domEvent, "x") * qx.event.handler.GestureCore.ROLL_FACTOR,
-          y: qx.util.Wheel.getDelta(domEvent, "y") * qx.event.handler.GestureCore.ROLL_FACTOR
-        };
-        domEvent.delta.axis = Math.abs(domEvent.delta.x / domEvent.delta.y) < 1 ? "y" : "x";
-        domEvent.pointerType = "wheel";
+        if (this._isTouchPadDetectionEnabled()) {
+          var now = new Date().getTime();
+          if (this.__lastRollEventTime > 0 && now - this.__lastRollEventTime
+            > qx.event.handler.GestureCore.TOUCHPAD_WHEEL_EVENTS_TIMEOUT) {
```
derrell: Possible readability improvement:
```
var now = new Date().getTime();
var timeout = qx.event.handler.GestureCore.TOUCHPAD_WHEEL_EVENTS_TIMEOUT;
if (this.__lastRollEventTime > 0 && now - this.__lastRollEventTime > timeout) {

Comment (derrell) [inline framework/source/class/qx/event/handler/GestureCore.js]: By the way, I prefer all variable declarations at the top of the functions in JavaScript since having them at the beginning of a block implies that there are block-level variables (declared with var) which JavavScript does not support. There are plenty of instances in qooxdoo of the style being used here, so this is not a mandated change, but I think it makes the code more clear.

File: framework/source/class/qx/util/Wheel.js [hunk 7]
```diff
@@ -39,6 +39,11 @@ qx.Bootstrap.define("qx.util.Wheel", {
      */
     FACTOR: 1,
 
+    /**
+     * Is the Wheel actually a touchpad ?
+     * @internal
```
derrell: What does @internal mean? Is that meaningful to the generator or to readers?

framework/source/class/qx/event/handler/GestureCore.js
derrell: (This is done in a few other places too.)

derrell (changes requested):
I don't see any problems here other than the few small things I've identified, but I'm concerned about this PR. It is a low-level change with the potential to have far-reaching effects. How will we test and ensure that this doesn't break anything that has been working?

vvandens:
@level420 @derrell Thanks for the reviews. I will wait for others and compile all the requested changes in a new commit and resubmit a PR.

@derrell As for the non-regression tests, the change is limited to osx Sierra+. So there is little to none risk that it breaks anything on other platforms. But it might be useful to extend its scope to other platform / os / version if required and this would require more extensive testing since you cannot know about the hardware (either the touchpad or the mouse).

framework/source/class/qx/event/handler/GestureCore.js
oetiker: Yes, please don't break compatibility without strong reason

oetiker (commented):
I am still wondering what the source of this all is ... and if it is not rather a bug of some sort in the way the browsers interpret what they get from OSX ... since you mentioned that native scrolling works fine ...

vvandens:
@oetiker You are right, I don't know either... Initially, my assumption was that it was limited to MacOS Sierra that introduced a lot of scrolling problems (see [this discussion](https://discussions.apple.com/thread/7679256?start=0&tstart=0) for instance).
However, since the native MacOS events are handled by browsers and delivered to the javascript engine, it should not break anything once the browser "knows" how to handle the multiplication of events of Sierra.
So now, my preferred "understanding" is that the qooxdoo normalization code should not be applied to the trackpad (at least the MacBook one) as I tried to explain it in the initial description of this PR. The normalization code computes a factor, that cannot do anything else than decrease, based on the events delta it receives during the application lifetime. So if you start to swipe gently on the trackpad, the factor is high (around 2) and the scroll speed is also. If then, you make a big swipe, then the factor lowers to something around 0.7 and the scroll speed becomes too slow. Then the factor *never* increases again. That's why, when detecting a detecting a trackpad, the PR code disables any normalization code and the event delta is handled "as is". When detecting a mouse wheel, the normalization code is applied as before.

framework/source/class/qx/event/handler/GestureCore.js
vvandens: Well, the GestureCore class extends `Object` and not `qx.core.Object`, so there is no shortcut method to the qx logging system available. Same for `qx.util.Wheel` that is a static class and does not extend anything. Any alternative @derrell ?

framework/source/class/qx/util/Wheel.js
vvandens: I don't know, but I thought this was done in order to omit the inclusion in docs and tell the reader that it it is not a reliable public API.
I really just applied the same thing than for the other variables in the class.

framework/source/class/qx/event/handler/GestureCore.js
derrell: Ah. In that case, what you're doing is fine. In some cases, I might recommend changing the superclass, but not here in this very low-level code where any small change could have wide-ranging effects.

framework/source/class/qx/util/Wheel.js
derrell: Sigh. Leave it as it is, but it's wrong for all of them. If variables are not supposed to be touched from outside of a class, they should be private, i.e., with two leading underscores, and then they are "@internal" by default. I don't think it's appropriate to change all of them at this point, but that would really be the proper solution.

framework/source/class/qx/event/handler/GestureCore.js
derrell: I hadn't noticed that. No, the only alternative is changing the superclass to qx.core.Object, but this is too low-level of a class to be considering that at this point. The possibility of wide-ranging bugs is too high. Leave it as it is.

derrell (changes requested):

vvandens:
I've just added a commit that includes all the requested changes by @derrell and @level420

derrell (approved):
I remain paranoid about this, but I can't see any problems, so I'll I'll approve and hope it wasn't a mistake...

level420 (commented):
Does not touch other platforms and feels OK for safari on OSX 10.12.

Don't have a mac with touchpad. I'd like to wait with my approval until @cajus feedback on his mac book.

cajus:
Finally. Sorry for the long delay. @vvandens current master with my Macbook (10.12.4, late 2016). And it works fine on all tested browsers.

cajus (approved):

vvandens:
@cajus No worry, we are all very busy. Great to hear.

level420 (approved):

level420:
3 approvals, more than 2 business days
merging
thank you @vvandens

vvandens:
Great! Thank you all for your reviews.
Any chance for this PR to be back-ported to 5.0.x branch since there is no API change ? If not, no worry, I will patch locally.

PR #9307, opened 2017-03-30
bug fix for complex regex

hkollmann:
if you set the filter to an complex regex like `^(\+|-)?\d*$` validating never stops.
This commit fixes this issue and adds some test cases.

I seperated the the code for validating in `_onHtmlInput` to an own function `_validateInput`. This function could be tested easily.

### DIFF
--- a/framework/source/class/qx/ui/form/AbstractField.js
+++ b/framework/source/class/qx/ui/form/AbstractField.js
@@ -560,20 +560,9 @@ qx.Class.define("qx.ui.form.AbstractField",
       if (this.__oldInputValue && this.__oldInputValue === value) {
         fireEvents = false;
       }
-
-      // check for the filter
-      if (this.getFilter() != null)
+      if (fireEvents)
       {
-        var filteredValue = "";
-        var index = value.search(this.getFilter());
-        var processedValue = value;
-        while(index >= 0)
-        {
-          filteredValue = filteredValue + (processedValue.charAt(index));
-          processedValue = processedValue.substring(index + 1, processedValue.length);
-          index = processedValue.search(this.getFilter());
-        }
-
+        var filteredValue = this._validateInput(value);
         if (filteredValue != value)
         {
           fireEvents = false;
@@ -581,7 +570,6 @@ qx.Class.define("qx.ui.form.AbstractField",
           this.getContentElement().setValue(value);
         }
       }
-
       // fire the events, if necessary
       if (fireEvents)
       {
@@ -942,8 +930,31 @@ qx.Class.define("qx.ui.form.AbstractField",
         qx.ui.form.AbstractField.__addPlaceholderRules();
       }
     },
-
-
+    /**
+     * validates the the input value
+     * 
+     * @param {type} value: the value to check
+     * @returns the checked value
+     */
+    _validateInput : function(value) {
+      // if no filter is set return just the value
+      var filteredValue = value;
+      // check for the filter
+      var filter = this.getFilter();
+      if (filter != null)
+      {
+        filteredValue = "";
+        var index = value.search(filter);
+        var processedValue = value;
+        while((index >= 0) && (processedValue.length > 0))
+        {
+          filteredValue = filteredValue + (processedValue.charAt(index));
+          processedValue = processedValue.substring(index + 1, processedValue.length);
+          index = processedValue.search(filter);
+        }
+      }
+      return filteredValue;
+    },
     /*
     ---------------------------------------------------------------------------
       PROPERTY APPLY ROUTINES
--- a/framework/source/class/qx/ui/form/AbstractField.js
+++ b/framework/source/class/qx/ui/form/AbstractField.js
@@ -231,8 +231,8 @@ qx.Class.define("qx.ui.form.AbstractField",
      * RegExp responsible for filtering the value of the textfield. the RegExp
      * gives the range of valid values.
      * Note: The regexp specified is applied to each character in turn, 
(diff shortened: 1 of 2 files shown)


level420:
@hkollmann it seems that the travis job is currently unable to load/start chrome unstable

level420:
@hkollmann could you please open an issue regarding the failing chrome unstable travis job? thank you.

derrell:
I'm not convinced this is a valid issue. The `filter` property is supposed to be a regexp that has the valid characters for the field:

```
    /**
     * RegExp responsible for filtering the value of the textfield. the RegExp
     * gives the range of valid values.
     * The following example only allows digits in the textfield.
     * <pre class='javascript'>field.setFilter(/[0-9]/);</pre>
     */
```

so any filter regexp that contains `^` and/or `$` is providing "bad" input to the filter property. I believe that as designed, the filter regexp must evaluate to a single character. It must not have any relative positions.

In fact, the code could really be simplified to something like this entirely untested code:
```
var filter = this.getFilter();
filter = new RegExp("^(" + filter.toString() + ")+$");
if (! value.match(filter))
{
    this.getContentElement().setValue(this.__oldInputValue);
}
```

So the question here, is do we want to extend the functionality, and is it entirely backward compatible?

oetiker:
I would go even further and not even anchor the filter ... let the users choose how they want to shoot themselves ...

hkollmann:
The most Intention for this fix is that the check functions returns even if the filter is not correct.
Before you get an endless Loop.
I agree with @derrell  that the Intention was a list of allowed characters. But it should be return in all cases.

Yes it's backward compatible - that's the reason why i added some unit testing.

derrell:
Point taken, that it should never go into an infinite loop. Ok, let's make sure it can work with a regexp that doesn't match only a single character. I don't understand how it works. Does it work if the regexp is `/xy/` (i.e., the only valid complete input is the string `xy`) and the first character, `x` is entered in the field? If it does, then I need an explanation of your new validation function because I don't see it...

hkollmann:
Validation function is not new. It's the old one with a little fix to go out the loop.
new is that it's an seperate function copied from the old _onHtmlInput function so that the validate could be unit tested.

derrell:
Yes, I understand. I'm just trying to ascertain if this fix, which is intended to change the behavior (filter regexp matches a single character -> filter regexp matches more than one character) works in all cases, or only with some regexps. If it works in all cases, I'm all for this. If only certain regexps will work with it, then I think we need to consider it more. If the regexp is `/^xy$/` then what happens when the user enters `x`? `y`? `xy` (if it's possible)?

I don't disapprove of this concept at all; I'm just trying to ensure that it doesn't fix the infinite loop problem, allowing regexps which aren't intended, but then not work with other regexps. In that case, it might be better to just document the fact that the filter regexp **must** match only a single character and attempting to do otherwise is at their own peril.

hkollmann:
what's about writing a test case for the expressions you have problems with? One benefit of the patch is that this is possible now.
Any suggestions?

framework/source/class/qx/ui/form/AbstractField.js
```diff
@@ -560,28 +560,16 @@ qx.Class.define("qx.ui.form.AbstractField",
       if (this.__oldInputValue && this.__oldInputValue === value) {
         fireEvents = false;
       }
-
-      // check for the filter
-      if (this.getFilter() != null)
+      if (fireEvents)
       {
```
level420: @hkollmann could you please explain why the if condition changed from `if(this.getFilter() != null)` to `if(fireEvents)`?

hkollmann: you are right. Should be `if (fireEvents && (this.getFilter() != null)`

level420 (approved):

hkollmann:
Could we get a decision about this? Thank's!

oetiker (changes requested):
yes ... the fix seems fine ... although the whole thing works a lot different from what I thought It would work ... maybe the description should be changed too:

> RegExp responsible for filtering the value of the textfield. the RegExp gives the range of valid values. The following example only allows digits in the textfield.

sounds not like what is happening here ... I get the impressen the person who did the first implementation did not entirely grasp how regexps work ... 

maybe change the documentation to 

>RegExp responsible for filtering the value of the textfield. Note, that the regexp specified is applied to each character in turn, NOT to the entire string. So only regular expressions matching a single character make sense in the context.

hkollmann:
>> I get the impressen the person who did the first implementation did not entirely grasp how regexps work ...

That was my Impression too,

oetiker (approved):

framework/source/class/qx/ui/form/AbstractField.js
```diff
@@ -230,6 +230,9 @@ qx.Class.define("qx.ui.form.AbstractField",
     /**
      * RegExp responsible for filtering the value of the textfield. the RegExp
      * gives the range of valid values.
+     * Note: The regexp specified is applied to each character in turn, 
+	 *       NOT to the entire string. So only regular expressions matching a 
+	 *       single character make sense in the context.
```
oetiker: hmmm it seems something is out of shape with the indentation here ...

oetiker (changes requested):

level420:
@oetiker fixed indention

oetiker (approved):

derrell (approved):
With the addition of the Note in the documentation, I'm find with this. Thanks.

level420:
3 approvals, more than 2 busines days
merging
thank you @hkollmann

Issue #9300, opened 2017-03-17
"Use of Mutation Events is deprecated. Use MutationObserver instead." warning in Firefox


firewave:
I am not really sure which widget this actually triggers this, but it somehow originates from inside of the qooxdoo framework  and not the application code (possibly qx.ui.embed.Iframe related?).
I see at least two mutation events in the Firefox debugger - one "DOMContentLoaded" and "DOMNodeInserted" (this seems to trigger the warning) both on something in qooxdoo-5.0.2-sdk/framework/source/class/qx/lang/Function.js. I looked a bit on the web and it appears this warnings has been around for several years now and it appears that Chrome also gave this warning at some point. I am not seeing it in Chrome though.

I am using qooxdoo 5.0.2 and Firefox 52 64-bit on Windows 10.

level420:
See https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

It could be that DOM-Events will be deprecated in the future.

schnepe2:
Can someone make a test case for this issue?
It seems that the work-around that has been added (for old [Bug 2756](https://github.com/qooxdoo/qooxdoo/commit/0975a69aadfaca68ed2ae6be2e978085666030a8) rsp. issue #2887)
should also be applied for current Chrome[1].
At least when the scenario is like the one described in #2887

[1] Version 57.0.2987.133 (64-bit)

schnepe2:
By the way: The original issue was, that "relocating" the iframe into another container did not re-render a changed source-URL:
```
container1.add(iframe, {flex: 1});
window.setTimeout(function() {
  iframe.setSource("http://www.ajaxian.com");
  container2.add(iframe, {flex: 1});                    // <= this one !!
}, 5000);
```
...but isn't that something that shouldn't be done?

schnepe2:
I've integrated tests into the PR ( #9314 ) as @oetiker suggested.

cboulanger:
@level420 @schnepe2 @firewave Has this issue been fixed in the meantime by @level420's commit?

hkollmann:
No answer - closed

PR #9314, opened 2017-04-18
Update Iframe.js to use MutationObserver (#9300)

schnepe2:
My proposal for #9300 using MutationObserver approach.
This **seems** to work, although the original poster should check if it's working for him/her

### DIFF
--- a/framework/source/class/qx/ui/embed/Iframe.js
+++ b/framework/source/class/qx/ui/embed/Iframe.js
@@ -378,7 +378,13 @@ qx.Class.define("qx.ui.embed.Iframe",
         iframeSource = iframeSource.substring(0, iframeSource.length-1);
       }
 
-      if (iframeSource != this.getSource()) {
+      if (iframeSource != this.getSource())
+      {
+        if ( qx.core.Environment.get("browser.name") != "edge" &&
+             qx.core.Environment.get("browser.name") != "ie" )
+        {
+          qx.bom.Iframe.getWindow(iframeDomElement).stop();
+        }
         iframeDomElement.src = this.getSource();
       }
     },
--- a/framework/source/class/qx/ui/embed/Iframe.js
+++ b/framework/source/class/qx/ui/embed/Iframe.js
@@ -51,6 +51,7 @@ qx.Class.define("qx.ui.embed.Iframe",
   */
 
   /**
+   * @ignore(MutationObserver)
    * @param source {String} URL which should initially set.
    */
   construct : function(source)
--- a/framework/source/class/qx/ui/embed/Iframe.js
+++ b/framework/source/class/qx/ui/embed/Iframe.js
@@ -69,12 +69,25 @@ qx.Class.define("qx.ui.embed.Iframe",
 
     if ((qx.core.Environment.get("engine.name") == "gecko"))
     {
-      this.addListenerOnce("appear", function(e)
+      this.addListenerOnce("appear", function ()
       {
-        var element = this.getContentElement().getDomElement();
-        qx.bom.Event.addNativeListener(element, "DOMNodeInserted", this._onDOMNodeInserted);
-      });
-      this._onDOMNodeInserted = qx.lang.Function.listener(this._syncSourceAfterDOMMove, this);
+        // Mutation record check callback
+        var isDOMNodeInserted = function (mutationRecord) {
+          return mutationRecord.addedNodes;
+        };
+
+        var observer = new MutationObserver(function (mutationRecords)
+        {
+          if (mutationRecords.some(isDOMNodeInserted)) {
+            this._syncSourceAfterDOMMove();
+          }
+        }.bind(this));
+
+        // Observe parent element
+        var element = this.getLayoutParent().getContentElement().getDomElement();
+        observer.observe(element, { childList: true });
+
+      }, this);
     }
   },
 
--- a/framework/source/class/qx/ui/embed/Iframe.js
+++ b/framework/source/class/qx/ui/embed/Iframe.js
@@ -69,25 +69,59 @@ qx.Class.define("qx.ui.embed.Iframe",
 
     if ((qx.core.Environment.get("engine.name") == "gecko"))
     {
-      this.addListenerOnce("appear", function ()
+      if ( qx.core.Environment.get("ecmascript.mutationobserver") )
       {
-        // Mutation record check callback
-        var isDOMNodeInserted = function (mutationRecord) {
-          return mutationRecord.addedNodes;
-        };
-
-        var observer = new MutationObserver(function (mutationRecords)
+        this.addListenerOnce("appear", function ()
         {
-          if (mutationRecords.some(isDOMNodeInserted)) {
-            this._syncSourceAfterDOMMove();
-          }
-        }.bind(this));
+          var element = this.getContentElement().getDomElement();
 
-        // Observe parent element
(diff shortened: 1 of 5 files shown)


framework/source/class/qx/ui/embed/Iframe.js
```diff
@@ -69,12 +69,25 @@ qx.Class.define("qx.ui.embed.Iframe",
...
-        var element = this.getContentElement().getDomElement();
-        qx.bom.Event.addNativeListener(element, "DOMNodeInserted", this._onDOMNodeInserted);
-      });
-      this._onDOMNodeInserted = qx.lang.Function.listener(this._syncSourceAfterDOMMove, this);
+        // Mutation record check callback
+        var isDOMNodeInserted = function (mutationRecord) {
+          return mutationRecord.addedNodes;
+        };
+
+        var observer = new MutationObserver(function (mutationRecords)
```
oetiker: does the MutationObserver object exist on all gecko engine versions ?

schnepe2: ...at least on those who complain about the Mutation Events ;)
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#Browser_compatibility) it's available for Firefox since Version 14.
Should we really add a environment-switch for that?

schnepe2: @oetiker : Independent of my question above, I think I will add a `qx.bom.client.EcmaScript#getMutationObserver` function.
Is that the right place for that information?

level420:
@schnepe2 current chrome is reported as "webkit" and also has `MutationObserver` implemented.

schnepe2:
@level420 yeah I've added the according environment check method and used it
4cd5ababb99092db2279653e8964660214979564 should be something to look at (sorry for the many single commits).

schnepe2:
As asked by me in the issue discussion (#9300):
Can someone make a test case for this issue / work-around?

My test application is basically just this:
```
qx.Class.define("dummy.Application",
{

  members :
  {
    // overridden
    main : function ()
    {
      this.base(arguments);

      var parent = new qx.ui.container.Composite(new qx.ui.layout.VBox());
      var container1 = new qx.ui.container.Composite(new qx.ui.layout.VBox());
      parent.add(container1, {flex:1});
      var container2 = new qx.ui.container.Composite(new qx.ui.layout.VBox());
      parent.add(container2, {flex:1});

      var iframe = new qx.ui.embed.Iframe();

      var doc = this.getRoot();
      doc.add(parent, {edge: 10});

      container1.add(iframe, {flex: 1});
      iframe.setSource("http://www.c64.com");

      window.setTimeout(function() {
        iframe.setSource("http://www.w3c.org");
        container2.add(iframe, {flex: 1});
      }, 5000);
    }
  }
});
```
But it can only be run and observed manually :-1: 
Writing a proper unit test for this would be needed, but that's something I have to assign to the professionals :grin:

framework/source/class/qx/ui/embed/Iframe.js
```diff
@@ -335,7 +380,11 @@ qx.Class.define("qx.ui.embed.Iframe",
 
       if (iframeSource != this.getSource())
       {
-        qx.bom.Iframe.getWindow(iframeDomElement).stop();
+        if ( qx.core.Environment.get("browser.name") != "edge" &&
+             qx.core.Environment.get("browser.name") != "ie" )
+        {
+          qx.bom.Iframe.getWindow(iframeDomElement).stop();
+        }
```
level420: @schnepe2 why is this specific to ie and edge and was not before?
Ist this something specific to the mutation even?

framework/source/class/qx/ui/embed/Iframe.js
```diff
@@ -67,13 +68,57 @@ qx.Class.define("qx.ui.embed.Iframe",
...
+
+        // Observe parent element
+        var parent = this.getLayoutParent().getContentElement().getDomElement();
+        observer.observe(parent, { childList: true });
+
+      }, this);
+    }
+    else // !qx.core.Environment.get("ecmascript.mutationobserver")
+    {
+      this.addListenerOnce("appear", function ()
```
level420: @schnepe2 before your PR the mutation event was attached to all gecko browsers. Now it is attached to every browser NOT having the mutation observer. 
Before the PR the listener was not attached to IE, Safari and Chrome which both report to be webkit.

I think adding this listener should be guarded with
```javascript
if ((qx.core.Environment.get("engine.name") == "gecko"))
```
Or am I wrong here?

level420 (changes requested):

framework/source/class/qx/ui/embed/Iframe.js
schnepe2: Yes and No... the stop() method is not available in "Internet Explorer".
To **remove** the **Mutation Event** is all this issue is about!

framework/source/class/qx/ui/embed/Iframe.js
schnepe2: Please ask the writer of the original work-around why this had been done only for gecko.
Possibly this work-around was only needed for gecko back then, but now it must be applied to all but Internet Explorer (both edge and IE).

level420: OK. Then you think that the else path should be applied to all browser NOT having MutationObserver?

framework/source/class/qx/ui/embed/Iframe.js
level420: Ist this stop() needed for all browsers except ie/edge? Event if MutationObserver ist not present?

framework/source/class/qx/ui/embed/Iframe.js
schnepe2: I don't **think** that the else path should be applied to all browser not having MutationObserver, I have **tested** it (at least with some Browsers) AND applied additional Tests.

_Opinion-ON_
From my point of view this complete work-around should be removed
and anyone who thinks he/she must implement an application were it is needed to **move** an iframe from one area of the DOM into another should handle this "issue" in his/her application.
The original work-around supplier seemed to be important enough so that it was introduced into the framework.
_Opinion-OFF_

level420: OK convinced on this point.

framework/source/class/qx/ui/embed/Iframe.js
level420: And convinced on this point.

level420 (approved):

framework/source/class/qx/ui/embed/Iframe.js
schnepe2: > Ist this stop() needed for all browsers except ie/edge?

To be honest, I don't think so.
But as I have no information about the initial need for this whole work-around I decided to keep it (as much as possible).

In my test(s) it makes no difference, but these test(s) only loaded small HTML pages ...
... I can imagine however, that if the first ("_pre-DOM-move_") URL loads lots of data it can make sense to stop loading that, when a "_DOM-move-and-URL-change_" was detected.
It might still be superfluous (as the URL is changed and the Client might stop loading the old URL anyway), but that's something I can neither confirm nor deny.

oetiker (approved):
looks good

level420:
2 approvals, more than 2 business days
merging
thank you @schnepe2

PR #9316, opened 2017-04-18
Test-case for issue #9300

schnepe2:
Test case approach for issue #9300 Sync-Source-After-DOM-Move Handling


schnepe2:
Note:

1. This test **should fail** on IE and Chrome (and other Webkits ?!) until PR #9314 is merged.
2. This test **does not** check for "Use of Mutation Events is deprecated. Use MutationObserver instead." warning in Firefox (as I don't know how)

oetiker:
hmm can't you add that directly to 9300 ?

schnepe2:
@oetiker sure, but one PR is the fix the other one is a test, so I thought you would like them separate.
I'll close this PR and add the changes to #9314, 'cause merging branch A into branch B is much to weird (with git) for me.

Issue #9313, opened 2017-04-11
Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.core.Widget


schnepe2:
With the current (master) build, using a table the following warning is logged:
`Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible with qx.ui.core.Widget`

The auto generated isVisible() method of the `qx.ui.table.columnmenu.MenuItem`s `visible` property triggers this.

So either the changes of commit 51a8c2261eef5feab112e6ba4e5d07e702ddc4d5 need to be changed,
or the name of `MenuItem`s property could be changed,
or the implementation of issue #1911 might have to be changed.

Options (1) or (3) are both not my area of expertise, so I can't provide a PR for this.

The playground surely does not show this 'cause it's not a debug-build.

level420:
@schnepe2 I think it is not the fault of PR https://github.com/qooxdoo/qooxdoo/pull/9232 here, because this PR only reveals that we have a conflict here. The solution would be to change the implementation of `qx.ui.table.columnmenu.MenuItem`

https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/table/columnmenu/MenuItem.js#L32-L38

schnepe2:
I agree, unfortunately we have to change the Interface as well.
Easy to change that property name to "`shown`". But that name feels kind of edgy. Anyone here with a nicer name?

schnepe2:
I still don't really like the property's name "`show`", but unless someone comes with a better idea / name:
Take it or leave it.

oetiker:
how about calling it `columnVisible`

schnepe2:
@oetiker Done! Still not perfect, but much better than `shown` :+1:

level420:
Fixed via #9317

PR #9317, opened 2017-04-19
Fix "Conflicting property method qx.ui.table.columnmenu.MenuItem.isVisible ..."

schnepe2:
Minimal impact proposal for issue #9313


framework/source/class/qx/ui/table/columnmenu/MenuItem.js
```diff
@@ -29,12 +29,12 @@ qx.Class.define("qx.ui.table.columnmenu.MenuItem",
...
-    visible :
+    columnVisible :
     {
       check : "Boolean",
       init  : true,
-      apply : "_applyVisible",
-      event : "changeVisible"
+      apply : "_applyColumnVisible",
+      event : "changeColumnVisible"
     }
```
level420: This is a breaking API change. We should consider this. carfully.

schnepe2: Yes (this API change bit me too, see e948834).
But I can't see how I can stop the property system from creating that "isVisible" method.
My first thought was: "implement `setVisible`, `getVisible` methods and `"changeVisible"` event and remove the property", but the Interface `qx.ui.table.IColumnMenuItem` likes to have a property named `visible`!
Or is it O.K. to "manually" create the methods and event to fulfill the interface's needs?

schnepe2: Fixing this _without changing the API_ is difficult:
Either `qx.ui.core.Widget#isVisible()` must be renamed (API change),
...or `qx.ui.table.columnmenu.MenuItem#visible` must be renamed (API change)

What might be a solution is the following:
Changing the Interface in a way that keeps any implementing class happy:
```
qx.Interface.define("qx.ui.table.IColumnMenuItem",
{
  events :
  {
    /**
     * Dispatched when a column changes visibility state. The event data is a
     * boolean indicating whether the table column associated with this menu
     * item is now visible.
     */
    changeVisible : "qx.event.type.Data"
  },

  members :
  {
    // Get/Set whether the table column associated with this menu item is visible
    setVisible : function (value) {},
    getVisible : function () {}
  }
});
```
With this, any implementing class that used a property named `visible`, still fulfills the interface's requirements because `setVisible` method, `getVisible` method and `changeVisible` event are auto-created[1]!

What do you think?

[1] Such an application class will raise the warning "Conflicting property method what.ever.Clazz.isVisible with qx.ui.core.Widget", but that's (a) correct and (b) fixable by the application-developer

schnepe2:
Sorry for the last 2 (or so) commits.
I will **never** understand git (or github)! Why on earth other changes (done at master) are incorporated in the inter-branch merge on my fork is beyond me...sorry
I'll try again...

framework/source/class/qx/ui/table/columnmenu/MenuItem.js
```diff
@@ -60,24 +56,57 @@ qx.Class.define("qx.ui.table.columnmenu.MenuItem",
...
       {
-        this.setValue(value);
+        this.__columnVisible = value;
+
+        // avoid recursion if called from listener on "changeValue" property
+        if (! this.bInListener) {
+          this.setValue(value);
+        }
+
+        this.fireDataEvent("changeVisible", value, this.__columnVisible);
```
level420: As you already assigned the new value to ```this.__columnVisible``` in line 93, the data events old value (third param in ```fireDataEvent```) is allways equal to the new value.
You should call
```javascript
this.__columnVisible = value;
```
after fireing the event.

framework/source/class/qx/ui/table/IColumnMenuItem.js
```diff
@@ -37,5 +29,27 @@ qx.Interface.define("qx.ui.table.IColumnMenuItem",
...
+
+
+    /**
+     * Get whether the table column associated with this menu item is visible
+     *
+     * @return {Boolean} whether the table column associated with this menu
+     *   item is visible
+     */
+    getVisible : function () {}
   }
```
level420: To completely mimic a boolean property, we also need ```isVisible``` and ```toggleVisible``` methods.

framework/source/class/qx/ui/table/IColumnMenuItem.js
```diff
@@ -21,14 +21,6 @@
  */
```
level420: It may be valuable to leave a comment here why we do not use a boolean property visible, but a set of setter and getter.

level420 (changes requested):

framework/source/class/qx/ui/table/columnmenu/MenuItem.js
schnepe2: Ooops, sure (no listener seems to care about these `old` values :grin: ). Will fix that

framework/source/class/qx/ui/table/IColumnMenuItem.js
schnepe2: Yes, but then we override `isVisible` and `toggleVisible`,too! The layout system seems to call `isVisible` on Widgets it likes to render, what causes strange effects (as far as I recognize).

"_mimic full boolean property_" is not something that is really needed for the Table classes to work.
Therefore the interface only demands to implement the `getVisible` and `setVisible` methods.

It might not be "_mimic full boolean property_", but this is the closest thing I could think of in terms of "**no API change**"
...I would still vote for a change in the API (Interface demanding `getColumnVisible` and `setColumnVisible` & Event), but it seems that API-change is ruled out :disappointed:

framework/source/class/qx/ui/table/IColumnMenuItem.js
schnepe2: Sure! 100% agree. Once it's settled how this is implemented I will do that

framework/source/class/qx/ui/table/columnmenu/MenuItem.js
schnepe2: BTW: setting the value after firing the event is also wrong (a call to `getVisible` inside a listener would give wrong results). Still this needs to be fixed :blush:

framework/source/class/qx/ui/table/IColumnMenuItem.js
schnepe2: My opinion on this would still be, to completely remove the Interface, let Table directly use the `value` property (`getValue`, `setValue`, `"changeValue"` Event).
Yes it's an API-change, but a "save" one that can not be unnoticed for any application developer.

**Pro:** fixes this issue, less code
**Con:** Any application class, that implemented the interface will get an error (Interface is gone) and might have to use `value` then.

The "Conflicting method ..." warning for any implemented application class using a `visible` property will be thrown independent of this (which is correct!)

schnepe2:
**Just for the record** (to let everybody see what it looks like):
What I would prefer is the following (API change!)
Rename `visible` => `checked` 
```javascript
qx.Interface.define("qx.ui.table.IColumnMenuItem",
{
  properties :
  {
    /**
     * Whether the table column associated with this menu item is visible
     * Should be of type {Boolean}!
     */
    checked : { }
  },

  events :
  {
    /**
     * Dispatched when a column changes visibility state. The event data is a
     * boolean indicating whether the table column associated with this menu
     * item is now visible.
     */
    changeChecked : "qx.event.type.Data"
  }
});
```

```javascript
qx.Class.define("qx.ui.table.columnmenu.MenuItem",
{
  extend     : qx.ui.menu.CheckBox,
  implement  : qx.ui.table.IColumnMenuItem,



  /*
  *****************************************************************************
     CONSTRUCTOR
  *****************************************************************************
  */

  /**
   * Create a new instance of an item for insertion into the table column
   * visibility menu.
   *
   * @param text {String}
   *   Text for the menu item, most typically the name of the column in the
   *   table.
   */
  construct : function(text)
  {
    this.base(arguments, text);

    // Two way binding this.checked <--> this.value
    this.bind("value", this, "checked");
    this.bind("checked", this, "value");
  },



  /*
  *****************************************************************************
     PROPERTIES
  *****************************************************************************
  */

  properties :
  {
    checked :
    {
      check : "Boolean",
      init  : true,
      event : "changeChecked"
    }
  }
});
```


Code that's calling this would look like this:
 (much more clear what is done than the old `setVisible()`/`"changeVisible"` I think)
```javascript
// ...
          var menuButton =
            new qx.ui.table.columnmenu.MenuItem(options.text);
          menuButton.setChecked(options.bVisible);
// ...
      var data = evt.getData();
      if (this.__columnMenuButtons != null && data.col != null &&
          data.visible != null) {
        this.__columnMenuButtons[data.col].setChecked(data.visible);
      }
// ...
        menuButton.addListener(
          "changeChecked",
          this._createColumnVisibilityCheckBoxHandler(col), this);
// ...
```

These change(s) would inform any possible application-class developer about a changed API. So it should not get missed[1].

Just another 2cents from me :grin: 

[1] O.K. If the application class has implemented a `checked` property itself it might get unnoticed (what are the odds?)

schnepe2:
...or even better (proposal):

```javascript
qx.Interface.define("qx.ui.table.IColumnMenuItem",
{
  properties :
  {
    /**
     * Whether the table column associated with this menu item is visible
     * Should be of type {Boolean}!
     */
    value : { }
  },

  events :
  {
    /**
     * Dispatched when a column changes visibility state. The event data is a
     * boolean indicating whether the table column associated with this menu
     * item is now visible.
     */
    changeValue : "qx.event.type.Data"
  }
});
```

```javascript
qx.Class.define("qx.ui.table.columnmenu.MenuItem",
{
  extend     : qx.ui.menu.CheckBox,
  implement  : qx.ui.table.IColumnMenuItem
});
```

Less code => less errors :grin: 
Pro arguments equal the ones from my last proposal.

framework/source/class/qx/ui/table/IColumnMenuItem.js
level420: OK agree with this.

johnspackman:
AIUI the only way to fix this is to rename the property altogether - the "visible" state of a widget refers to the widget itself (ie the `MenuItem` in this case) and not a third-party object (ie the table column) so there is no possibility of synchronising `MenuItem.visible` with the underlying `Widget.visibility` and there would always be a conflict of meaning anyway.

I would favour just renaming the property to `columnVisible` as @oetiker suggested in https://github.com/qooxdoo/qooxdoo/issues/9313#issuecomment-295219814 - I don't like `checked` in this case because that's the visual representation rather than meaning, and `value` is about how its implemented.

If we were to simply rename the property it would not be necessary to roll our own property implementation, which would be a big plus for maintainability.

The downside is that there would be a breaking API change; how about if we marked `MenuItem.visible` as `@deprecated`, wouldn't the generator's migrate job provide compile time warnings about using it?  All we would then have to do is to bind `MenuItem.visible` to `MenuItem.columnVisible` (and vice-versa) until 6.1 (when `MenuItem.visible` is removed) and backwards compatibility is achieved.

But, if the generator cannot determine that the object being used is an instance of `MenuItem`, the warnings will not happen.

schnepe2:
@johnspackman I am fine with the `columnVisible` approach. :+1: 

Marking `MenuItem.visible` as `@deprecated`, however would still create **the warning this issue was all about**. :-1: 

In my view this API-_change_ is not an API-_break_, because no one was able to fulfill the Interface without getting that warning. It currently is a kind of "impossible Interface".

Changing the Interface to...
```javascript
qx.Interface.define("qx.ui.table.IColumnMenuItem",
{
  properties :
  {
    /**
     * Whether the table column associated with this menu item is visible
     * Should be of type {Boolean}!
     */
    columnVisible : { }
  },

  events :
  {
    /**
     * Dispatched when a column changes visibility state. The event data is a
     * boolean indicating whether the table column associated with this menu
     * item is now visible.
     */
    changeColumnVisible : "qx.event.type.Data"
  }
});
```
...would remove the warning for all use-cases where the application did not derive from `qx.ui.table.columnmenu.MenuItem`
and give an "Interface does not fit" error for those who did.

I think the number of applications that implement `qx.ui.table.columnmenu.MenuItem`-derived classes is not that big (just guessing).
There was a time where "migration" scripts handled this :grin:

johnspackman:
Sure, the warning will go away, but the warning is only being introduced in 6.0 so there will be no warning for the user to see if we do not deprecate the "visible" property before eliminating the conflict warning.  With your solution, if someone implements the interface with the old property name then there will be a very early runtime error when the class is defined (and possibly compile time too, although not sure about that) so it will be very visible that there is something to deal with - which is perfect.

However, where there is code like:

```
var myColumn = this.getSomeColumn();
myColumn.setVisible(false);
```

That will fail only when it's executed and the migration scripts will be unable to pick it up because there is no inferred type information for `myColumn`.  My concern here is that I agree with you that the code which hides columns is not very common, and therefore could easily slip past testing and get into a release.  If there is an easy way to highlight the problem early we should do so.

schnepe2:
Are you sure that such a situation...
```javascript
var myColumn = this.getSomeColumn();
myColumn.setVisible(false);
```
is not detectable?
There is no `setVisible()` method for no class anywhere! :grin:
...only `isVisible()` is implemented for `qx.ui.core.Widget` and its derivatives...
So it might be detectable, but don't ask me how.

level420 (approved):

johnspackman (changes requested):
That's great, except that IMHO there should be a visible property which is marked `@deprecated`

level420:
@schnepe2 @johnspackman I'm not sure how we could continue with this PR. Any ideas?

schnepe2:
I am fine with the latest `'columnVisible' property` approach.

@johnspackman threw in the `@deprecated` idea, but I don't know where he would like to have it.
At the Interface **and** the Implementation? And even with it (the `visible` property) being flagged `@deprecated`, I think that will still raise the original warning this whole PR is all about.

oetiker:
the thing about the deprecated warning is that we try not to yank stuff from qx but rather give users fair warning eg one major release ... so the new functionality would be in place but it would still be accessible under the original name ... OR does accessing it under the original name not even work ?

johnspackman:
At the moment, this is a breaking change and IMHO that should be avoided if reasonably possible; I like the work that @schnepe2 has done to call it `columnVisible`, but as I said above I think that there should be a backwards-compatible, deprecated `visible` property shim which works in the short term (ie one version cycle).  

Yes, this will cause the warning to continue to appear and it is frustrating that removing the warning was the original motivation for this PR, but I think that if I was caught out by this change I would much rather have had an extra warning to ignore than have spent the time figuring out why a hidden column was suddenly visible.  Maybe I'm being over-sensitive, but I imagine it as this is something that could be missed until post-release of the application.

We could also improve the warning text, or special case it to handle MenuItem in the mean time.

schnepe2:
The idea about flagging `visible` property `@deprecated` is clear, but that will still raise the original warning this whole PR is all about.
In that case this issue would have to stay "open" as it is **not fixed** until the `visible` property is removed!

In a perfect world[1] this could be handled by a migration script[2] and an added note in the release notes.

[1] once there was a time ... :grin:
[2] see my https://github.com/qooxdoo/qooxdoo/pull/9317#issuecomment-299133978 too

schnepe2:
@johnspackman anyone that uses the `visible` property already knows that he/she is doing something weird because of the warning. So he/she **is** informed already.
And if he/she thought it might be possible to hide a column (as you imagined), he/she would already noticed that that is not the case.

oetiker:
so the current implementation does not even work ? IN that case I am happy with renaming it ... no deprecation required from my point of view.

johnspackman:
My understanding is that it is not possible for `myColumn.setVisible(false)` to be detected by the generator as invalid.  If you can show that it can, then problem solved.

If `false` was passed to the widget's `setVisible` it would throw an exception because Widget takes a string not boolean; depending on the use case, this might make it easy to detect in development, or crash the app when a user is using it.

I am conflicted about this because obviously we **do** make breaking changes, and sometimes they cannot be detected by the generator's migration scripts; but IMHO we should try hard to avoid them wherever possible, and this is only about eradicating a **warning**, not fixing some bug.

> anyone that uses the visible property already knows that he/she is doing something weird because of the warning. So he/she is informed already.

This is NOT true - that warning is only in the master and is not in a release.  The conflict here is because if this PR is applied, the warning is being introduced in 6.0 and also removed in 6.0.

@oetiker AIUI the current implementation does work just fine, it's just that it hides a method in `Widget` and that causes a warning to appear.  So this change will cause code that was working fine in 5.x to work differently (or crash) in 6.0

schnepe2:
> If false was passed to the widget's setVisible it would throw an exception because Widget takes a string not boolean; depending on the use case, this might make it easy to detect in development, or crash the app when a user is using it.

I don't think so! You might have confused this with the property `visibility`.

Summary:
`qx.ui.core.Widget` has a property `visibility` (string) and a method 'isVisible' (returning boolean)

johnspackman:
> I don't think so! You might have confused this with the property visibility.

Oops, you're right!

schnepe2:
So here we are. Right where we've started :smile: :

At some time we've introduced a warning about conflicting method names (which is right!)
but having that conflict in the frameworks `MenuItem` since 5.0 (which wasn't right in the first place).

I don't see why this should not be fixed! A mayor version update always has the potential of changes that are not automatically detected/fixed, but here it does not seem to be the case:

Any user that updates from 5.0 to 6.0 would not be informed about the change of `MenuItem`s `columnVisible` property,...
...**but** if however a user had implemented a `visible` property it would be noticed by the generators warning!
The current fix only avoid this warning to be thrown from framework code. Any "user-space" misuse is still detected.
...or am I wrong here?

johnspackman:
I've written an app to test these theories - the answers are that `.isVisible()` did work just fine in 5.x and the generator will not detect that `.isVisible()` is not valid.

However - I think that all of this is irrelevant because the MenuItems are re-created every time the button to produce the menu is clicked (in `Table._initColumnMenu` [here](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/table/Table.js#L2091-L2139); the `Table` adds listeners to the MenuItem it creates and uses this to make API calls to `Table.getTableColumnModel().setColumnVisible()`.

IMHO this makes it difficult for application code to use those MenuItems and it seems reasonable to think that if you wanted to show and hide columns you would use the columnmodel API and not rely on the MenuItem happening to exist at the moment you wanted to show/hide it.  It's possible for code to trap the `"columnVisibilityMenuCreateEnd"` event and manipulate MenuItems, but that seems rather odd given that any user interaction outside of the menu would collapse the menu, in which case why would the developer be accessing MenuItem for visibility?  

The only remaining case would be if someone writes a class which implements `qx.ui.table.IColumnMenuItem` but does not implement the columnVisible property - the generator will not pick this up, but the runtime of `qx.Class` throws an error when the class is defined so it is  unmissable.

You could almost consider `MenuItem.visible/columnVisible` a private implementation of Table ... either way, I'm removing my request that there is a deprecated `visible` property

johnspackman (approved):

schnepe2:
> You could almost consider MenuItem.visible/columnVisible a private implementation of Table ...

Yes, right! The whole idea of this interface is something I don't see any real benefit in.

The "private nature" of columnVisible would be most obvious in the implementation proposal at https://github.com/qooxdoo/qooxdoo/pull/9317#issuecomment-296173499, ...but let's not re-open that again :grin:
I am currently fine with the interface -which cannot be removed anyway now it has been "in to the wild"

johnspackman:
Yes, it's weird!  It's just one of those things that creep in I guess, but the interface could be removed completely without affecting any other code in a material way (and would arguably be an improvement)

schnepe2:
> the interface could be removed completely without affecting any other code in a material way

I would make that change topic of another issue ("follow up issue")

oetiker (approved):
yes!

derrell (approved):
Apparently this whole discussion took place while all qooxdoo notifications were going into my spam bucket. Sorry 'bout that. All of the points that I might have commented on have been discussed ad nauseatum and I'm on board with the discussion.

level420:
OK then. merging with 4 approvals 
thank you @schnepe2

PR #9321, opened 2017-04-28
Fix Textarea autosize calculation in special case

rkg-mm:
In some cases the textarea's autosize method was called before the textarea had proper bounds applied. In this case a clone of the textarea was created without a width set and the width was also not updated later. This lead to a wrong calculation of textarea height, because the clone used for calculation had a very small width, while the real textarea could be much bigger, which produced a lot of whitespace the longer the text got.

This fix delays the creation of the clone until the bounds are available. Fixed the issue in our project.

### DIFF
--- a/framework/source/class/qx/ui/form/TextArea.js
+++ b/framework/source/class/qx/ui/form/TextArea.js
@@ -8,8 +8,7 @@
      2004-2008 1&1 Internet AG, Germany, http://www.1und1.de
 
    License:
-     LGPL: http://www.gnu.org/licenses/lgpl.html
-     EPL: http://www.eclipse.org/org/documents/epl-v10.php
+     MIT: https://opensource.org/licenses/MIT
      See the LICENSE file in the project's top-level directory for details.
 
    Authors:
@@ -152,7 +151,8 @@ qx.Class.define("qx.ui.form.TextArea",
         e.stop();
       }
     },
-
+	
+	
     /**
      * When the element resizes we throw away the clone and trigger autosize again, otherwise the clone would have
      * another width and the autosize calculation would be faulty.
@@ -179,10 +179,8 @@ qx.Class.define("qx.ui.form.TextArea",
     */
     __autoSize: function() {
       if (this.isAutoSize()) {
-        // make sure to only create the clone when the bounds of this element are calculated. Otherwise clone will have
-        // no width and the autoSize calculation is wrong due to much smaller clone (creates a lot of whitespace at the
-        // end of the textbox)
         var clone = this.__getAreaClone();
+
         if (clone && this.getBounds()) {
 
           // Remember original area height
--- a/framework/source/class/qx/ui/form/TextArea.js
+++ b/framework/source/class/qx/ui/form/TextArea.js
@@ -8,7 +8,8 @@
      2004-2008 1&1 Internet AG, Germany, http://www.1und1.de
 
    License:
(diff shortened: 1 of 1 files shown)


oetiker:
makes total sense to me ... especially since the original author seems to have somewhat handled the problem but not completely :)

oetiker (approved):

level420 (dismissed):

johnspackman (changes requested):
This fixes the problem where there is no bounds initially, but what happens if the `TextArea` is resized?  I.E. won't that mean that the `__areaClone` is again the wrong width and trigger a similar problem?

If that's true, an easy fix would be to dispose `__areaClone` when resizing

rstmm:
Yes you are right, if the textbox resizes the same problem occurs. Also, the change seems to introduce a "Call stack size exceeded" due to (not yet) known reason in rare cases. If you have an idea how to fix this in general, any improvement is welcome.

johnspackman:
Call stack exceeded is very frustrating because it cannot be trapped, but later versions of Chrome do make the stack trace more useful if you add this:

```
Error.stackTraceLimit = 200;
```

in your code (before the exception happens); that way you should at least get more debug output.  Alternatively, I think the latest Canary has some improvements (but not tried that myself yet).

If the exception was caused by your change, I'd suggest adding a mutex at the start of `__autoSize`, ie 
```
    __autoSize: function() {
      if (this.__inAutoSize)
        return;
      this.__inAutoSize = true;
      try {
        // ... snip ... rest of code 
      } finally {
         this.__inAutoSize = false;
      }
    }
```

rstmm:
Please check again, I have reverted my initial fix and added another one, based on the suggestion of @johnspackman, which also covers resizing of the element, e.g. through browser window resizing.

oetiker:
And now, the case where the cloning happens too early is not handled anymore ?

rstmm:
This case is not a problem anymore with the fix, because when the textarea gets finally its size it will recreate the clone, so in the rare cases where this happened it gets fixed automatically as soon as the size gets set to the element.

oetiker:
just to be sure ... you are aware of the fact, that after `new qx.ui.form.TextArea()` this thing does NOT manifest in the DOM yet ... neither does it in general until all the current code has run its course ... so depending on what you do right after creating the TextArea odd things can happen if the code assumes that anything exists in the DOM yet ... 

Reading your first patch, I thought that you were taking care of a corner case that has todo with this ... There is already some code in the object, search for 'appear' ... but that 'get bounds' call might be run out of turn ... 

anyway if your problem was only connected with resizing, then ... lets fix that :)

rstmm:
Hm yes, seems like I was handling a different issue here :)

level420:
@rstmm ist this PR still valid?

rstmm:
Yes it is, solves the issue of wrong size taken for line break calculations in case 
- the textarea size changes (browser resize/manual change)
- the textarea had no size when clone was created

rstmm:
Sorry, I meant height calculation of autosize feature.

johnspackman (approved):

cboulanger (approved):

cajus (approved):

johnspackman (approved):

level420:
4 approvals. more than 2 business days.
merging.

level420:
thank you @rstmm

PR #9327, opened 2017-05-10
workaround for wrong event targets on touch devices

peuter:
Drag&Drop does not work on touch-enabled devices. To check that try to drag an element from one list to another in the qooxdoo demo browser DragDrop example (use a real tablet or the Chrome-Tablet-Simulation mode).

The problem is that the native event always return the element where the dragging started as target and not the one thats really under the cursor position.

This PR adds a workaround to get the real drop target by using `elementFromPoint`

### DIFF
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -615,13 +615,16 @@ qx.Class.define("qx.event.handler.DragDrop",
 
       // find current hovered droppable
       var el = e.getTarget();
+      if (this.__startConfig.target === el) {
+        el = document.elementFromPoint(e.getDocumentLeft(), e.getDocumentTop());
+      }
       var cursor = this.getCursor();
       if (!cursor) {
         cursor = qx.ui.core.DragDropCursor.getInstance();
       }
       var cursorEl = cursor.getContentElement().getDomElement();
 
-      if (el !== cursorEl) {
+      if (el !== cursorEl && !cursorEl.contains(el)) {
         var droppable = this.__findDroppable(el);
 
         // new drop target detected
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -616,7 +616,7 @@ qx.Class.define("qx.event.handler.DragDrop",
       // find current hovered droppable
       var el = e.getTarget();
       if (this.__startConfig.target === el) {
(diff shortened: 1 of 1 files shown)


oetiker:
hmmm looking at the description of the `elementFromPoint` method, I wonder why it is not being used in call cases ... browser support according to 'https://caniuse.com/#search=elementFromPoint' seems excellent ... so maybe switch alltogether ?

peuter:
Well you loose some events because sometimes the DragDrop-curser is returned from elementFromPoint and those cases are ignored obviously. I don't know if this is a real problem, but relying on the native event target might be better if they are correct.

oetiker:
are we sure it does not work on any 'touch' scenario ... I mean if the normal implementation is better, then it might be appropriate to check for the problem condition (origin == target or something) than just decide based on the type of pointer

peuter:
Ok, as the initial target is already stored a changed to behaviour to use `elementFromPoint` when current target == initial target. Then downside is, that the workaround is also used when you are really reagging over the initila target, but that should do no harm.

Also the check that we do not get the DrogDropCursor element needed to be augmented, cause you might get a descendant of the cursor.

oetiker:
looks good to me

cajus (approved):

oetiker (approved):

johnspackman (approved):

framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -615,13 +615,16 @@ qx.Class.define("qx.event.handler.DragDrop",
 
       // find current hovered droppable
       var el = e.getTarget();
+      if (this.__startConfig.target === el) {
+        el = document.elementFromPoint(e.getDocumentLeft(), e.getDocumentTop());
+      }
```
level420: We already use `elementFromPoint` in the framework here:
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/PointerCore.js#L238

The code there uses things like `domEvent.view.document.elementFromPoint` and uses the thouch coordinates of the event to determine the element position.

I don't know why in `qx.event.handler.PointerCore` `domEvent.view.document.elementFromPoint` is used instead of `document.elementFromPoint`, but I think it may be worth investigating. Could it be that in some situation we have `document !== domEvent.view.document`?

peuter: That could happen in iframes a assume. Good finding I'll change my PR accordingly.

level420 (approved):

level420:
4 approvals. more than 2 business days
merging
thank you @peuter

PR #9332, opened 2017-05-16
take a look underneath the cursor element to fix the dragover event f…

peuter:
…iring

This is follow up to #9327 : The problem is that, depending how fast you drag in the direction where the drag-cursor is positioned it is possible the you do not receive the dragover event.

This fix checks if the cursor is returned as drop target and if so, it hides the cursor to see whats underneath it for a short period of time. As this only happens while your dragging the cursor hiding is not visible to the human eye.

### DIFF
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -616,6 +616,7 @@ qx.Class.define("qx.event.handler.DragDrop",
       // find current hovered droppable
       var el = e.getTarget();
       if (this.__startConfig.target === el) {
+        // on touch devices the native events return wrong elements as target (its always the element where the dragging started)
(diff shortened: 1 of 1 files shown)


oetiker:
neat!

oetiker (approved):

level420 (approved):

cajus (approved):

cboulanger:
That the "Travis CI build failed" doesn't mean anything in this case, does it?

cajus:
@cboulanger travis has some kind of hickup from time to time. I re-triggered the firefox build.

cboulanger (approved):

johnspackman (approved):

level420:
4 approvals, more than 2 business days since voting.
merging
thank you @peuter

PR #9336, opened 2017-05-18
Removes the need for DateFormat instances to be manually disposed

johnspackman:
As a simple utility class, it is natural to instinctively expect `DateFormat` to not require manual disposal any more, this PR changes DateFormat to support that.

The reason why it was not automatically disposable before was because it had a binding to `qx.locale.Manager.locale` that was used only to cache the current locale; there is no need for caching so removing the binding was trivial, but this highlighted a potential bug or issue for backward compatibility: while the user is allowed to change the `DateFormat.locale`, `DateFormat` would behave differently depending on the initial locale given to the constructor.

The system locale is only used if no locale is provided to the constructor (see [DateFormat.js](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/util/format/DateFormat.js#L97-L98) ), but this binding is *not* removed if the locale is explicitly set later; so if you set the locale explicitly in the constructor, it will always be that locale - but if you set it explicitly later on it could be silently changed by the browser (see [_applyLocale](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/util/format/DateFormat.js#L434-L437) ).  There is no event handler for changes to `DateFormat.locale` so it cannot be detected by the user.

My guess is that the intention (and reasonable expectation for users) was that an explicitly set locale is *set* and always overrides the system; if it is not set, or is subsequently un-set, then it reverts to the system default.

### DIFF
--- a/framework/source/class/qx/util/format/DateFormat.js
+++ b/framework/source/class/qx/util/format/DateFormat.js
@@ -72,7 +72,7 @@
 qx.Class.define("qx.util.format.DateFormat",
 {
   extend : qx.core.Object,
-  implement : [ qx.util.format.IFormat, qx.core.IDisposable ],
+  implement : [ qx.util.format.IFormat ],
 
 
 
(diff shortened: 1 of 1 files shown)


level420 (approved):

cboulanger (approved):

level420:
2 approvals. more than 2 business days
merging
thank you @johnspackman

PR #9337, opened 2017-05-18
Fixes a bug where an exception is raised if there is no contentElement

johnspackman:
The exception is a side effect of changing the iframe's source


oetiker (approved):

cboulanger (approved):

level420 (approved):

level420:
3 approvals. more than 2 business days.
merging
thank you @johnspackman

Issue #9340, opened 2017-05-22
bug in qx.ui.table.cellrenderer.Number


dvoral:
If the cell is empty (null) and is being edited, then NaN displays. 

From Derrell Lipman:
if (cellInfo.value || cellInfo.value == 0) {
    should be
if (cellInfo.value || cellInfo.value === 0) {
Note the triple '=' so that it catches the number 0 but not something that's falsy like null.

We should also change the else condition where it says cellInfo.value == 0.

johnspackman:
Hi @dvoral, thanks for the bug report - could you try implementing a Pull Request with this change?  That's the fastest way to get a fix integrated into the core

derrell:
Sigh. There is no good solution to this problem.

On the one hand, we can change the `format` method of `qx.util.format.NumberFormat` so that an empty string is not NaN, but that is not backward compatible with apps that expect an empty string to return NaN.

We could also change qx.ui.table.cellrenderer.Number so that it does what its name implies and ensures that every field contains a number. This would mean that if the data is null or an empty string, we put a 0 in the cell. That way, the cell editor (which was the basis of the OP's problem for which this issue was created) would always get a valid initial value, and not display NaN. That is also not backbackward compatible, because programs may well expect the ability to have empty fields in a Number column.

I'm open to ideas.

Derrell

zaucker:
A number column in a table used for data entry MUST allow empty cells.

How about a separat cellrenderer class? That would not break anything.

Cheers,
Fritz

level420:
@dvoral did you find a solution for this bug? Please share you're solution/thoughts. Thank you.

dvoral:
Yes I did. I wrote on the forum what I did. On June 4 I wrote:
"I managed to fix the NaN while editing problem. I wrote an override of the
setValidationFunction of the qx.ui.table.celleditor.TextField."

Nobody asked until now what I did.
My solution:
var intEditor = new qx.ui.table.celleditor.TextField();
intEditor.setValidationFunction (
               function( value ) {
                if (value == "") {
                value = null;
                }
                if (value != null) {
                value = parseFloat(value);
                }
                return value;
               });


Hope this is helpful.

level420:
@dvoral would you please be so kind to create a pull request for `qx.ui.table.celleditor.TextField`. that would be great! thank you!

dvoral:
Can you please send me a link where I do that?
And instructions if it's not clear for a first-time user. :)

[...]

PR #9348, opened 2017-06-14
Issue 9430: bug in qx.ui.table.cellrenderer.Number editor

dvoral:
Correct problem of NaN displaying when value is null string

See issue #9340

### DIFF
--- a/framework/source/class/qx/ui/table/celleditor/TextField.js
+++ b/framework/source/class/qx/ui/table/celleditor/TextField.js
@@ -38,6 +38,11 @@ qx.Class.define("qx.ui.table.celleditor.TextField",
         value = validationFunc( value, cellEditor.originalValue );
       }
 
+      // Correct problem of NaN displaying when value is null string.
+      if (value == "") {
+          value = null;
+      }
+      
       if (typeof cellEditor.originalValue == "number") {
         if (value != null) {
           value = parseFloat(value);
--- a/framework/source/class/qx/ui/table/celleditor/TextField.js
+++ b/framework/source/class/qx/ui/table/celleditor/TextField.js
@@ -37,14 +37,11 @@ qx.Class.define("qx.ui.table.celleditor.TextField",
       if (validationFunc ) {
         value = validationFunc( value, cellEditor.originalValue );
       }
-
-      // Correct problem of NaN displaying when value is null string.
-      if (value == "") {
-          value = null;
-      }
-      
+     
       if (typeof cellEditor.originalValue == "number") {
-        if (value != null) {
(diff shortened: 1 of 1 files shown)


derrell (approved):
This is a breaking change (formerly presented NaN; now presents nothing), but I really do think it was a bug and I can't imagine that anyone actually wanted that behavior. We should be sure to document this change in the release notes for 6.0 but I think this is a good fix.

level420 (approved):

framework/source/class/qx/ui/table/celleditor/TextField.js
```diff
@@ -38,6 +38,11 @@ qx.Class.define("qx.ui.table.celleditor.TextField",
         value = validationFunc( value, cellEditor.originalValue );
       }
 
+      // Correct problem of NaN displaying when value is null string.
+      if (value == "") {
+          value = null;
+      }
+
```
oetiker: "" is not null ... so since the NaN problem only occurs if the value is a number, how about moving this inside the 

```javascript
if (typeof cellEditor.originalValue == "number") {
```

block

framework/source/class/qx/ui/table/celleditor/TextField.js
```diff
@@ -38,6 +38,11 @@ qx.Class.define("qx.ui.table.celleditor.TextField",
         value = validationFunc( value, cellEditor.originalValue );
       }
 
+      // Correct problem of NaN displaying when value is null string.
+      if (value == "") {
+          value = null;
+      }
+      
       if (typeof cellEditor.originalValue == "number") {
         if (value != null) {
```
level420: wouldn't it be less invasive if the check goes in this line? Like:
```javascript
         if (value != null && value != '') {
```

oetiker (changes requested):

framework/source/class/qx/ui/table/celleditor/TextField.js
level420: @oetiker yes!

level420 (changes requested):

level420:
@dvoral if you want to change this PR see how to do it: https://github.com/qooxdoo/qooxdoo/issues/9340#issuecomment-308415941

oetiker (approved):

johnspackman (approved):

level420 (approved):

cboulanger (approved):

patrick-fischer-hirschstein (approved):

level420:
a lot time has passed. 4 approvals. merging.
thank you @dvoral

Issue #9353, opened 2017-07-05
Qooxdoo images are not loaded in IE when app is served in https and base URL is set


ronkie:
We have a qooxdoo app that is served by Tomcat via https and with base URL set in index.html, so all the resources are served from a different location than index.html location.

There is a Checkbox widget on a page that appears not checked when it actually is. We discovered that it was because chekbox image was not loaded (error 404).
Other images we have in the app are loaded correctly.

There is an issue:
https://github.com/qooxdoo/qooxdoo/issues/2564

Code that was added for it adds a prefix for the url to make it absolute, but it does not take into account "base" url, so app tries to load images from the same location as index.html resides.

ronkie:
![image not found ie](https://user-images.githubusercontent.com/6853101/27863986-f1c51610-618c-11e7-883b-c1ab3031fa21.png)

level420:
Fixed via https://github.com/qooxdoo/qooxdoo/pull/9354

derrell:
Unfortunately, this change breaks existing code. The class `qx.util.ResourceManager` is used in non-GUI code, where `document` does not exist. We need to find a different way of solving the original problem, or ensure that this code still works in non-GUI environments.

derrell:
I've tracked this down further. The problem is that the patch puts a dependency in `qx.util.ResourceManager` on `qx.bom.Selector`  which brings in the Sizzle library that tries to do a bunch of stuff at initialization time that assumes a GUI environment. There can not be a dependency on `qx.bom.Selector` in `qx.util.ResourceManager`.

level420:
OK then the faulty line is https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/util/ResourceManager.js#L288

derrell:
Correct, but the whole patch appears to depend on that...

ronkie:
Hi guys,

I've used qxWeb at first, but that was also not good?

https://github.com/ronkie/qooxdoo/commit/167eb14e769ee70b6fd5995e151a3dd9fcab67fc

Maybe we could simply use

document.querySelectorAll("base")

But before checking for the document, if it is there....

level420:
@ronkie could you please verify if the solution in PR https://github.com/qooxdoo/qooxdoo/pull/9415 works for you? Thank you!

PR #9354, opened 2017-07-05
#9353 Qooxdoo images are not loaded in IE when app is served in https…

ronkie:
… and base URL is set

Use base url if it is present

### DIFF
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -285,7 +285,7 @@ qx.Class.define("qx.util.ResourceManager",
 
           var href;
           //first check if there is base url set
-          var baseElements = qxWeb("base");
+          var baseElements = qx.bom.Selector.query("base", document);
           if(baseElements.length > 0)
           {
             href = baseElements[0].href;
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -71,7 +71,7 @@ qx.Class.define("qx.util.ResourceManager",
       if(!registry) {
         return null;
       }
-  
+
       var ids = [];
       for (var id in registry) {
         if (registry.hasOwnProperty(id)) {
@@ -81,7 +81,7 @@ qx.Class.define("qx.util.ResourceManager",
           ids.push(id);
         }
       }
-  
+
       return ids;
     },
 
@@ -283,16 +283,31 @@ qx.Class.define("qx.util.ResourceManager",
             continue;
           }
 
+          var href;
+          //first check if there is base url set
+          var baseElements = qxWeb("base");
+          if(baseElements.length > 0)
+          {
+            href = baseElements[0].href;
+          }
+
           // It is valid to to begin a URL with "//" so this case has to
           // be considered. If the to resolved URL begins with "//" the
           // manager prefixes it with "https:" to avoid any problems for IE
-          if (resourceUri.match(/^\/\//) != null) {
+          if (resourceUri.match(/^\/\//) != null)
+          {
             statics.__urlPrefix[lib] = window.location.protocol;
           }
           // If the resourceUri begins with a single slash, include the current
           // hostname
-          else if (resourceUri.match(/^\//) != null) {
-            statics.__urlPrefix[lib] = window.location.protocol + "//" + window.location.host;
+          else if (resourceUri.match(/^\//) != null)
+          {
+            if(href)
+            {
+              statics.__urlPrefix[lib] = href;
+            } else {
+              statics.__urlPrefix[lib] = window.location.protocol + "//" + window.location.host;
+            }
           }
           // If the resolved URL begins with "./" the final URL has to be
           // put together using the document.URL property.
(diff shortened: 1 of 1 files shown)


ronkie:
The idea is to use "base" url when it is present; if not, then do things as before.

framework/source/class/qx/util/ResourceManager.js
```diff
@@ -283,16 +283,31 @@ qx.Class.define("qx.util.ResourceManager",
             continue;
           }
 
+          var href;
+          //first check if there is base url set
+          var baseElements = qxWeb("base");
```
oetiker: could you use `qx.bom.Selector.query('base',Document)` I would rather not add a `qxWeb` dependency at this point

oetiker (changes requested):

framework/source/class/qx/util/ResourceManager.js
ronkie: Sure, I changed the code and tested it

patrick-fischer-hirschstein (approved):

oetiker (changes requested):
looking through the diff, it seems that you somehow created quote a lot of formatting changes to the code ... some of them violating formatting rules ...  ( opening curly brackets at the start of the line).

ronkie:
I've searched for any code style guide but could not find anything. Url http://www.qooxdoo.org/community/contribution returns page not found.
Fixed some formatting issues that I could find according to general formatting

johnspackman:
the travis lint job would normally have picked this up - is it not running any more?

cboulanger:
@ronkie Good point, we definitively need a easily accessible style guide. It's in the works...

level420 (approved):

cajus:
@johnspackman the lint job just lints. It does not inspect the coding style.

cajus (changes requested):
As @oetiker noted, the existing braces, feeds, etc. should be left in place. It's harder to find the real changes inside a diff that mostly changes the coding style.

framework/source/class/qx/util/ResourceManager.js
```diff
@@ -291,8 +298,16 @@ qx.Class.define("qx.util.ResourceManager",
           }
           // If the resourceUri begins with a single slash, include the current
           // hostname
-          else if (resourceUri.match(/^\//) != null) {
-            statics.__urlPrefix[lib] = window.location.protocol + "//" + window.location.host;
+          else if (resourceUri.match(/^\//) != null)
```
cajus: To match with the existing code style, the braces should not start on a new line.

level420:
@cajus I've tried to revert (most) of the formatting done, except two spaces within an empty line. 
Hope this is enough.

The existing code  style within the file is inconsistent anyway.

cboulanger (approved):

derrell (commented):
I don't grok 'base' so I'm not sure what this is doing. I'll stay out of voting on this one.

framework/source/class/qx/util/ResourceManager.js
level420: OK. And how should it be with the function bodies and the braces? There are inconsistent code styles in there. I'm not sure?!?!?!?

cajus: For now I'm just looking at the existing code. It seems to differ when using curly braces for function bodies, objects, control structures. There's no style guide I can find in the source, though. It is just mentioned for the tech docs. It would be nice to have something specific, because running the *fix* generator job might just create hundreds of changed lines.

derrell: Please don't get into the political argument of where the braces go. qooxdoo is inconsistent, and we have differing views on the *one true way*. When changing an existing file, just try to remain as consistent as possible with the style already used in that file.

cajus: @derrell: that's just what I wanted to say.

johnspackman (approved):

framework/source/class/qx/util/ResourceManager.js
level420: @cajus @derrell OK bak to the current PR. I've just tried to fix all those unnecessary formatings besides the two spaces in empty lines. 
Is the PR now acceptable from this point of view?

derrell: I have no further comments on formatting as long as you don't try to "fix" it (unless, of course, you fix it to be the way *I* like it.) :-)

As previously stated, I don't really understand what this PR is doing with 'base', so I'm not voting on the technical merits of this one.

level420:
@derrell there is a good description on why the PR was done and how it works in the issue: https://github.com/qooxdoo/qooxdoo/issues/9353#issue-240628725

derrell (approved):
Ok, learned about "base". Understand. This all looks good now, technically.

Venting here, and we can discuss elsewhere... The other formatting issue in this file, that bugs the heck out of me, is variables declared willy-nilly throughout the file. That means that one never knows where to look for a variable's declaration. In (our supported version of) JavaScript, variables are function-level (or global) only. There are no block-level variables. I'd sure prefer (and yes, this is another religious issue) that all variables be declared at the top of their respective function so that their declarations are easily located.

We now return you to your normal programming.

oetiker (approved):

cajus (approved):

cajus:
>5 approvals, travis passed - merging. Thanks for your contribution ;-)

ronkie:
Thank you guys too :)

Issue #9351, opened 2017-07-03
Fixed dimensions (width and height) of AbstractScrollArea


mever:
Hi,

When changing a custom widget to support scrolling one can choose to inherit from {qx.ui.core.scroll.AbstractScrollArea} and place the children on the 'pane' instead. This adds scroll capabilities without too much hassle.

There is a catch, however. AbstractScrollArea hard codes the preferred width and height. Making the described change not enough for a drop-in replacement. To work around this one can set the width and height to null in the constructor of the custom widget. Although this solves the issue many people will stumble with this non-intuitive behavior.

This issue is meant to help people and may inspire a change in the defaults. Although the latter may break backwards compatibility (as some implementations may rely on the fixed width and height). It makes Qooxdoo even more of a pleasure to work with.

Cheers,
Martijn

level420:
@mever thank you for bringing this up. 
Note that you also could override (via refine) the `height` and `width` property with `null`.

level420:
@mever if you like to contribute, you could add some documentation to the `qx.ui.core.scroll.AbstractScrollArea` class or you could create an example within the demo browser ( http://www.qooxdoo.org/current/demobrowser/# ) which will for sure help others searching for a similar solution as you did.

mever:
How about removing the refined properties from the AbstractScrollArea and putting it as minor BC break for 6.0? I find it hard to imaging users rely on the exact dimensions of 100x100 in their application. They probably set their own dimensions. A minor price to pay for more intuitive components don't you think?

level420:
Sure! Please create a PR for this. Thank you!

mever:
I use Intellij Webstorm which understands JSDoc (http://usejsdoc.org/index.html). The 'jsdoc' branch introduces some changes that make my IDE help to understand the types used throughout the code. The problem is that the generator isn't able to parse these. It says:
```
      - Warning: qx.test.ui.core.AbstractScrollArea (30): Unable to parse JSDoc entry: @type qx.test.ui.core.AbstractScrollArea.fixture.CustomWidget
      - Warning: qx.test.ui.core.AbstractScrollArea (33): Unable to parse JSDoc entry: @type qx.ui.container.Composite
      - Warning: qx.test.ui.core.AbstractScrollArea (43): Unable to parse JSDoc entry: @param {qx.ui.core.LayoutItem} widget
[...]

PR #9355, opened 2017-07-05
Fixed dimensions (width and height) of AbstractScrollArea (issue 9351)

mever:
As requested in: #9351

### DIFF
--- a/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
+++ b/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
@@ -78,7 +78,7 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
     // since the scroll container disregards the min size of the scrollbars
     // we have to set the min size of the scroll area to ensure that the
     // scrollbars always have an usable size.
-    var size = qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH;
+    var size = qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH * 2 + 14;
     this.set({minHeight: size, minWidth: size});
 
     // Roll listener for scrolling
@@ -259,7 +259,7 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
     /**
      * Returns the dimensions of the pane.
      *
-     * @return {{width: number, height: number}|null} The pane dimension in pixel. Contains
+     * @return {Map|null} The pane dimension in pixel. Contains
      *    the keys <code>width</code> and <code>height</code>.
      */
     getPaneSize : function() {
--- a/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
+++ b/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
@@ -112,22 +112,6 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
     },
 
 
-    // overridden
-    width :
-    {
-      refine : true,
-      init : 100
-    },
-
-
-    // overridden
-    height :
-    {
-      refine : true,
-      init : 200
-    },
-
-
     /**
      * The policy, when the horizontal scrollbar should be shown.
      * <ul>
@@ -273,9 +257,10 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
     */
 
     /**
-     * Returns the boundaries of the pane.
+     * Returns the dimensions of the pane.
      *
-     * @return {Map} The pane boundaries.
+     * @return {{width: number, height: number}|null} The pane dimension in pixel. Contains
+     *    the keys <code>width</code> and <code>height</code>.
      */
     getPaneSize : function() {
       return this.getChildControl("pane").getInnerSize();
--- a/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
+++ b/framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
@@ -75,6 +75,12 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
       this._setLayout(grid);
     }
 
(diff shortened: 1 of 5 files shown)


mever:
Seems I need to fix some problems first.

level420:
@mever could you please review your PR, as it seems that it makes some other tests fail. See https://travis-ci.org/qooxdoo/qooxdoo/jobs/250567447

Those are:
* qx.test.ui.core DragDropScrolling:testIsScrollbarVisible FAILED
* qx.test.ui.core DragDropScrolling:testIsScrollbarExceedingMaxPos FAILED
* qx.test.ui.core DragDropScrolling:testScrollBy FAILED

Maybe those tests rely on the pre defined minimum scroll area dimensions.

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
```diff
@@ -267,9 +257,10 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
     */
 
     /**
-     * Returns the boundaries of the pane.
+     * Returns the dimensions of the pane.
      *
-     * @return {Map} The pane boundaries.
+     * @return {{width: number, height: number}|null} The pane dimension in pixel. Contains
```
derrell: This is a useful depiction when read in the code, but what's in the curly brackets is supposed to be a type. Does this render properly in the API viewer? What would typically be there is:
```
@return {Map|null} The pane dimensions in pixels. The Map contains two members: width and height, the respective number of pixels for that dimension.
```

I like what you've shown, but I don't think we should put it in one place but not globally. It would be a nice global change.

I'm open to accepting this change as is, if others feel it's appropriate.

derrell (changes requested):

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
johnspackman: i'll go with this unless it breaks the API viewer or jsdoc linting - I've often wished for a decent way to describe map members

johnspackman (approved):
approving, subject to consensus on the jsdoc

level420 (changes requested):
@mever please have a look into the failing travis tests. Thank you.

cboulanger (approved):

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
mever: This works well with my code editor (PhpStorm and by extension WebStorm from Jetbrains) which has Qooxdoo support build in. I think it uses JSDoc (http://usejsdoc.org/) to interpret the doc-blocks.

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
```diff
@@ -75,6 +75,12 @@ qx.Class.define("qx.ui.core.scroll.AbstractScrollArea",
       this._setLayout(grid);
     }
 
+    // since the scroll container disregards the min size of the scrollbars
+    // we have to set the min size of the scroll area to ensure that the
+    // scrollbars always have an usable size.
+    var size = qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH;
+    this.set({minHeight: size, minWidth: size});
+
```
level420: Following the logic in the comment above, the use of this is to initialize scroll area with minimal dimensions. It seems that we need place for both a horizontal and a vertical scrollbar. This is minimum `2 * qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH`.

But I doubt that this is enough as the text states that we should have a USABLE size which shows the up/down or left/right button of the scrollbar and maybe a little room for the scroll knob itself.

With having `qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH` set to 14, the former minWidth of 50 is not so bad (14 *2 + 22 pixel to scroll).

And a second point is that this change introduces `minHeight` and `minWidth` for all themes deriving from `qx.theme.simple.Appearance` which did not define `minHeight` and `minWidht` through the `scrollarea` appearance.

level420 (changes requested):

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
level420: @mever did you build the api viewer with your PR? I not: please do so and see if the api viewer shows this as expected, containing your changes. It it is OK leave it in. If not, please revert the change. Thank you.

level420:
I've done some investigations why some tests in `qx.test.ui.core.DragDropScrolling` fail. It is because the list element has no default height set. If we define a height for the list element like this:
```javascript
    _createDummyList : function()
    {
      var list = new qx.ui.form.List();
      list.setHeight(50); // <--- SET HEIGHT HERE
      for (var i=0; i<20; i++) {
        list.add(new qx.ui.form.ListItem("Item " + i));
      }
      return list;
    },
```
Tests are successful again.
For some reason it seems that `qx.ui.core.scroll.AbstractScrollArea` needs a height set in order to calculate visibility of scrollbars.

@mever could you please have a look into the failing tests? You can see them in the travis runs.

level420:
I've run this PR with one of my apps without any problems, but I don't use any `qx.ui.form.List` widgets there.

mever:
I've added some commits that pass the DragDropScrolling tests locally.

level420 (approved):
We previously had a default width and height in the abstract scroll area of 100/200 which is gone and have it now in th list. Seems OK!

framework/source/class/qx/ui/core/scroll/AbstractScrollArea.js
mever: I just build with the API viewer. The latest changes also show the width and height nicely in the documentation page of the form.List. No problems in the API viewer anymore.

derrell (approved):

level420:
4 approvals, more than 2 business days, merging, thank you @mever

PR #9364, opened 2017-08-05
remove all grunt stuff

hkollmann:


oetiker:
@hkollmann your patch breaks the testing system.

level420 (approved):

level420:
Tested by creating a new app and generating build, source, test-source and test without problems. Additionally build the framework test in source and build version.

oetiker (approved):

cboulanger (approved):

derrell:
The grunt stuff was added prior to the last release. Mightn't there be people depending on it? I don't object to removing it, but I suspect it should be only in a major release, and should await qxcompiler being included as an alternative to generate.py.

johnspackman:
We're working on making qxcompiler the default in 6.0, I'm just readying server builds and then API generation and we're pretty much there for alpha testing ... (famous last words :) )

hkollmann:
@derrell: The grunt stuff is only removed during creation of new projects. Old Projects are not touched.

derrell (approved):

level420:
4 approvals, more than 2 business days, merged.
Thank you @hkollmann

PR #9365, opened 2017-08-07
allow qx.test.io.remote.transport.XmlHttp:testSetHeader to be run under php-fpm

level420:
polyfill for getallheaders from https://github.com/ralouphie/getallheaders
published under the MIT license
needed if the test is run under php-fpm
See https://bugs.php.net/bug.php?id=62596

### DIFF
--- a/framework/source/resource/qx/test/xmlhttp/echo_header.php
+++ b/framework/source/resource/qx/test/xmlhttp/echo_header.php
@@ -1,4 +1,56 @@
 <?php
+/**
+ * polyfill for getallheaders from https://github.com/ralouphie/getallheaders
+ * published under the MIT license
+ *
+ * needed if the test is run under php-fpm
+ *
+ * See https://bugs.php.net/bug.php?id=62596
+ */
+if (!function_exists('getallheaders')) {
+    
+    /**
+     * Get all HTTP header key/values as an associative array for the current request.
+     *
+     * @return string[string] The HTTP header key/value pairs.
+     */
+    function getallheaders()
+    {
+        $headers = array();
+        
+        $copy_server = array(
+            'CONTENT_TYPE'   => 'Content-Type',
+            'CONTENT_LENGTH' => 'Content-Length',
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

cboulanger (approved):

derrell (approved):

framework/source/resource/qx/test/xmlhttp/echo_header.php
```diff
@@ -1,5 +1,57 @@
...
+        return $headers;
+    }
+}
+
 $a = getallheaders();
 if (isset($a['COOKIE'])) unset($a['COOKIE']);
 echo json_encode($a);
-?>
\ No newline at end of file
+?>
```
oetiker: in php it is good practice to NOT have the closing ?> since any spaces after the tag would be output to the browser causing confusion.

oetiker (changes requested):

framework/source/resource/qx/test/xmlhttp/echo_header.php
level420: @oetiker but that is not my code. This is how it originally was. As a good practice I've only changed what needed to fix the problem.

Should I remove it anyway?

level420: besides the CR at the end.

oetiker: I would ... without this the chance of it breaking is smaller since even when someone accidently adds a newline at the end (like you) it will still work

oetiker (approved):

framework/source/resource/qx/test/xmlhttp/echo_header.php
level420: OK. done.

level420:
@oetiker could you please merge? Thank you!

oetiker:
thanks!

Issue #9366, opened 2017-08-09
Improved interface for form fields


mever:
Dear developers,

The form components, great as they are, leave some room for improvement (as do all great engineering efforts). Therefore I want to pitch an idea for improving the form package. Goal is to leave existing code backwards compatible and add functionality that make dealing with the form easier.
Conceptually we have two categories of classes, the form and its fields. Functions of the form; like resetting, validation and data binding now rely on the specifics of those fields. When I browse the code I see a lot of if statements for specific uses that make the code fragile. I want to exercise the dependency inversion principle and the interface segregation principle from SOLID to make this better.

I think a new interface in the form package called IField with the following methods should do the trick:
```
+ setInputValue(value anytype|null) void
+ resetInputValue() void
+ getInputValue() anytype|promise|null

// setInputValue updates the view and model value
// resetInputValue updates the view with model value
// getInputValue updates the model with view value and return the model value

// The view value is what is displayed, the model value is what's in the memory of the field.

// getInputValue may return a promise in order to allow fields to be asynchronous (e.g. fetch server data)
```
_(input value is chosen because value is taken and BC is a priority)_

With this new interface resetting and data binding is possible in a field generic way. Validation may require an additional interface (e.g. IFieldValidation) because each field may choose to present feedback about the (invalid) value specifically. My current position is to leave it compatible with IForm.

This proposal makes it easier to create more fields that can be used in the form.

Any thoughts?

derrell:
This looks really good in principle! Have you done code changes to see that it really can be done in a backwards-compatible fashion? I would be very eager to see a PR that implements this.

level420:
@mever could you please give a practical example where your proposal helps? Thank you!

BTW.: it took some time until I found out that BC does not mean "before christ" ;-)

mever:
@level420 I created a separate form manager which implements a resetFields method. This method calls the form.reset() method on the Qooxdoo component and some custom logic to make it work with our custom fields. I like to get rid of resetFields and let Qooxdoo handle it.

johnspackman:
So is this a unification of the `qx.ui.form.IForm`/`IStringForm`/`INumberForm` fields?  IE `qx.ui.form.IForm` sounds like `IFieldValidation`, and `qx.ui.form.IStringForm`/etc would inherit from `qx.ui.form.IField`?

mever:
@johnspackman yes.

[...]

PR #9367, opened 2017-08-09
[New] IField interface

mever:
As requested for issue #9366.


mever:
I found that some field implementations already conform to this interface so I decided to change the name from inputValue to value instead. The setValue returns an error for incompatible types because an exception is too severe and crashes the application when uncaught.
In next commits I will implement the interface for textfield and virtual selectbox. After that I will modify the resetter and validator to support the new interface and add/modify integration tests to prove it.

mever:
So what's in this latest commit. First I changed the form.Resetter class to use the new interface (reducing it by 93 lines). Run the tests, implemented the interface in various fields and repeated this procedure until all tests passed in the ui package. I also added an extra test case for the virtual selectbox.

mever:
Is it possible to run the Travis tests again? The one failing is probably a mishap.

cajus:
Done

mever:
Okay, so the reason why this is failing is because Travis gets a merge from this branch and master: https://github.com/qooxdoo/qooxdoo/commit/4cdbb78366421d941b5a261b97e4ba50c2dc95e2. The master branch can not be build because it contains a problem. See: https://travis-ci.org/qooxdoo/qooxdoo/builds/262999042

johnspackman (commented):
Shouldn't `qx.ui.form.IStringForm`/`IBooleanForm`/etc inherit from `qx.ui.form.IField`?  The individual widgets would not need to be modified to implement `IField` because it would be implicit

mever:
@johnspackman good point. Can sub-interfaces refine methods (i.e. more specific)? Some languages do not allow it. Let's try it anyway though.

level420:
@mever regarding the failing travis runs: this should be fixed now with https://github.com/qooxdoo/qooxdoo/pull/9368

cboulanger (approved):
Very good, thank you! (Maybe adding a section to the manual would be very useful...)

level420 (commented):
In a first test I've applied the PR to my local master clone and tested my app with it and it fails sadly on form handling.

I'm using a derived `qx.ui.form.Form` with a lot of additions and some form widgets which don't seem to be happy with this PR.

I have for example a widget I've called CheckButtonBox which holds several checkboxes where the single checkboxes are addressable by value. 

setValue([ "opt1", "opt3"]) leads to the checkboxes associated to those values being checked, alll others unchecked.

It may be that I've not specified the right interfaces, so I have to check first if they comply to the new IField resp. I<whatever> interface and lead to the right handling in my derived form class.

But I'll be unavailable for the next two weeks, so I'm not approving this PR but let others decide. 

I'm trusting in the other core members making the right decisions in approving or not approving this PR.

derrell (changes requested):
@level420's comment implies that this PR isn't BC for derived classes of the prior Form field implementation. That may be a problem. I'd like to wait on approving this PR until @level420 returns and can further test with his derived classes, so we more fully understand the implications of this PR.

mever:
@derrell Seems like a sensible idea.
To be clear. The form resetter is _not_ fully BC for fields that do not implement the interface.  This change is therefore not totally BC as it changes the interface of the resetter. I have changed the fields in this project to conform to this interface. For fields defined outside of this project, implications are unsure.

derrell:
@mever Are you able to provide instructions for converting existing derived classes to the new interface? With 6.0, we don't need to be strictly BC, but we must at least either provide the migration script for any non-BC changes, or give clear instructions for people upgrading, on how to change their classes to conform. I think this PR is worthwhile and look forward to approving it; I just want to make sure we don't impose (too much) pain on our users when the release is done. 

Thanks!

level420:
@mever do you have some hints I could start searching for incompatibilities?

framework/source/class/qx/ui/form/Slider.js
```diff
@@ -59,6 +59,7 @@ qx.Class.define("qx.ui.form.Slider",
   extend : qx.ui.core.Widget,
   implement : [
     qx.ui.form.IForm,
+    qx.ui.form.IField,
     qx.ui.form.INumberForm,
```
level420: Just curious: why is @qx.ui.fomr.IField@ needed here? As @qx.ui.form.INumberForm@ derives from @qx.ui.form.IField@ adding @qx.ui.form.INumberForm@ should be enough. Or do I misunderstand something here?

mever: @level420  no, your right. It is redundant and can be removed.

framework/source/class/qx/ui/form/RadioButtonGroup.js
```diff
@@ -59,6 +60,7 @@ qx.Class.define("qx.ui.form.RadioButtonGroup",
 
     // attach the listener
     this.__radioGroup.addListener("changeSelection", function(e) {
+      this.fireDataEvent("changeValue", e.getData(), e.getOldData());
       this.fireDataEvent("changeSelection", e.getData(), e.getOldData());
```
level420: I think one of the incompatibilities come from here. I'm deriving from `qx.ui.form.RadioButtonGroup` and add and implement the `qx.ui.form.IStringForm` interface thus adding a `value` property of type string which fires its own change event.

The idea behind my implementation was to have radio buttons which each have their own string "value". Let's say two buttons labeled "male" and "female". Depending on the users choice the string value of the radio group is "female" or "male" or null if nothing was selected.

As I read the code now this means that the value of the `qx.ui.form.RadioButtonGroup` is an instance of `qx.ui.form.RadioButton` which for sure collides with my current implementation.

framework/source/class/qx/ui/form/Slider.js
mever: Give me a second to fix this.

framework/source/class/qx/ui/form/RadioButtonGroup.js
level420: @mever A fix for this would be to not use an inline function as an event listener, but to change it to a protected override-able member function `_onChangeSelection` .

This way it should be possible to change it to my needs.

level420 (changes requested):

mever:
On backwards compatibility:

First of, classes inhering the standard fields do not need to change as they should already be liskov substitutable. This also applies to classes directly implementing the standard interfaces: `IBooleanForm`, `IColorForm`, `IDateForm`, `INumberForm`, `IArrayForm`, `IModelForm` and `IStringForm`. The IField interface is actually the same interface as listed but with weaker parameter and return types.

The only real BC breaker here is the resetter. Because fields that are using a selection are abstracted away (see the three removed private methods in the resetter class). The `qx.test.ui.form.FormManager#testAll` test helped me a lot here in tracking down which fields didn't conform.

A way of finding potential problems is to look at implementations of `qx.data.controller.ISelection` and `qx.ui.core.ISingleSelection`. These implementations should also conform to this new interface. In order to pass the tests earlier mentioned I had to change the `MSingleSelectionHandling` and `MMultiSelectionHandling` mixins. Fields using those now automatically implement the interface when included. Fields implementing both selection interfaces but not the mixin are not compatible as they should provide their own implementation.

level420:
@mever there are, of course, let's name it implicit or potential BC breakers, if a class signature changes by e.g. adding new public or protected members or properties to an existing class, as this may collide with already existing code deriving from the unchanged class.

As I wrote for the `qx.ui.form.RadioButtonGroup` this situation becomes more manageable if those added members are override-able.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -399,6 +433,17 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
+
+
+    /**
+     * @param event {qx.event.type.Data} {@link qx.data.Array} change event.
+     * @private
+     */
+    __updateValue : function(event) {
+      var d = event.getData();
+      var old = (d.removed.length ? d.removed[0] : null);
+      this.fireDataEvent("changeValue", d.added[0], old);
```
level420: As it may come to the situation that some derived class wants to change the value type it would be good to also change this from a private to a protected member.

mever:
@level420 I agreed.

This new commit should allow you to overwrite the `_onChangeSelection` method.

Is it possible to notify the user when collisions happen in classes inhering one of: `qx.ui.form.RadioButtonGroup` or `qx.ui.form.VirtualSelectBox`?

level420:
@mever regarding detection of collisions: not that I'm aware of. Without knowing the history of the class both adding new methods to the derived class or, after the change in the superclass, overriding the methods are valid scenarios.

There is just one point where I had been able to immediately detect the collision: this is if you had chosen to add a property `value` to `qx.ui.form.RadioButtonGroup`, as I've also added a `value` property to the derived class. In this situation the generator would have reported an error.

level420 (approved):
@mever regarding your last commits: this indeed solved all my compatibility problems as all I had to do was to change the `_onChangeSelection` method in my class (it was already there) derived from `qx.ui.form.RadioButtonGroup` and avoid adding the event listener twice as this now happens in the base class.

I'm approving this PR now, but I hope that the other reviewers first test the code in their applications as well and approve then (if appropriate).

level420:
I'm still a bit unhappy with the situation, as what I now have to do (after this PR) is to change the value type from `qx.ui.form.IRadioItem` to `String` of my class derived from `qx.ui.form.RadioButtonGroup`.

This implies a change of the return value of the `setValue` method, which should not be.

Additionally I'm not sure what happens if a form containing `qx.ui.form.RadioButtonGroup` now runs through the cycle of creating a `qx.data.controller.form`, creating a model from this controller and act wit a `qx.data.marshal.Json` on it. Does it at some point try to serialize a `qx.ui.form.RadioButton?

mever:
@level420 I agree. It is bad because it breaks liskov substitution. This next commit **a0725b7e** should fix that.

I have to admit that my focus was more on my own domain. Before releasing this I want to do a few more cycles and make sure BC is intact. I will also look into the form controller.

level420:
@mever the latest push https://github.com/qooxdoo/qooxdoo/pull/9367/commits/ebeb9faa1bf9aa26f5ff72a40b904737bebf8eb9 to this PR seems unrelated

level420:
Additionally the improved resetter error handling in https://github.com/qooxdoo/qooxdoo/pull/9367/commits/06ff20b7ceefb6d5c6c83bca6c5daae8f8618b4cthe leads a few tests to fail. See https://travis-ci.org/qooxdoo/qooxdoo/jobs/265599013

mever:
@level420 yeah my IDE was complaining about a missing argument (capture) in `removeListener` in one of the commits (this one: **be8a1a07**). Fixed it here. Should I revert and place it in a different PR?

level420:
@mever yes please let this PR be what it is named.

mever:
@level420 so you like a force push without the commit or a revert?

level420:
@mever I think the best would be to revert https://github.com/qooxdoo/qooxdoo/commit/ebeb9faa1bf9aa26f5ff72a40b904737bebf8eb9

level420:
re-requesting approvals.

cboulanger (approved):

oetiker (approved):
I like it. Please squash when merging ...

johnspackman (approved):
reading this makes me realise how overdue it is that selection widgets support `value` as a property.  Well done @mever this is really good idea.

derrell (changes requested):
I like this, but I don't think my question of 19 days ago has yet been fully answered. When we release 6.0, we'll need to provide a migration script that modifies user programs to conform to any non-BC changes we've made, or provide explicit instructions on what a user must do to change their code to conform to our changes. Are you able to provide either a simple migration script or instructions for manual modification that can recognize or show the user how to recognize when and where changes are required, and fix them or tell the user what to do to fix them?

Sorry I'm being picky on this, but it's really important that the user not have unexplained errors after following all of our migration instructions.

level420:
@mever could you please review https://github.com/qooxdoo/qooxdoo/pull/9367#issuecomment-322843623 . Thank you!

mever:
Sorry for the delay, I had a lot on my plate.
@derrell it is good you are picky, not accepting just anything makes a great framework.
This commit is squashed as requested by @oetiker. The previous commits are on the https://github.com/mever/qooxdoo/commits/form.bak branch.

In 60af803 and 140d9fc I've systematically checked {qx.data.controller.ISelection} and {qx.ui.core.ISingleSelection}, respectively. In order to verify all previous widgets that could be used in combination with the resetter. I've also checked the {qx.data.controller.Form} as suggested by @level420 .

**Migration instructions are:**
The biggest change is on the {qx.ui.form.Resetter} component. All form widgets that worked previously from the standard library should still work. Check if your class does not conflict with the {qx.ui.form.IField} interface when inheriting from one of these widgets in your application. This interface essentially specifies a value property with a weak type.
When you specify custom form widgets which use the resetter (e.g.: indirectly through the form), then implementing {qx.ui.core.ISingleSelection} or {qx.data.controller.ISelection} is not enough. You need to implement the {qx.ui.form.IField} interface. When your widget(s) includes {qx.ui.core.MSingleSelectionHandling} or {qx.ui.core.MMultiSelectionHandling} it already implements the interface. Just make sure your class also specifies the interface in its definition.

These objects implement the new {qx.ui.form.IField} interface:

**Widgets**: {qx.ui.container.Stack} and {qx.ui.tabview.TabView}. 
**Form fields**: {qx.ui.tree.Tree}, {qx.ui.form.List}, {qx.ui.form.RadioButtonGroup}, {qx.ui.form.RadioGroup}, {qx.ui.form.SelectBox} and {qx.ui.form.VirtualSelectBox}.
_Any classes inheriting these widgets or form fields should work without change when not conflicting with the {qx.ui.form.IField} interface._

**Mixins**: {qx.data.controller.MSelection}, {qx.ui.core.MSingleSelectionHandling}, {qx.ui.core.MMultiSelectionHandling} and {qx.ui.virtual.selection.MModel}.
_Any classes including these mixins should work when not conflicting with the {qx.ui.form.IField} interface. When you need the implementing classes to work with the {qx.ui.form.Resetter} component they should explicitly implement the {qx.ui.form.IField} interface._

All the **form field interfaces** in {qx.ui.form.*}: {IArrayForm}, {IBooleanForm}, {IColorForm}, {IModelForm}, {INumberForm} and {IStringForm}.
_Any class explicitly (indirectly) implementing these interfaces should work without change._

framework/source/class/qx/ui/form/RadioButtonGroup.js
```diff
@@ -287,12 +289,55 @@ qx.Class.define("qx.ui.form.RadioButtonGroup",
...
+
+    /**
+     * Select given value.
+     *
+     * @param item {null|var} Item to set as selected value.
+     * @return {null|Error} The status of this operation.
+     */
+    setValue : function(item) {
+      if (item && 'object' === typeof item && item instanceof qx.ui.form.IRadioItem) {
+        return this.__radioGroup.setValue(item);
```
derrell: Hmmm... RadioGroup does not have a `value` property prior to this PR, that I can find, and this PR doesn't add one. How does this work? I think it should say, `this.__radioGroup.setSelection( [ item ] );` shouldn't it? (And similarly elsewhere in here.)

Additionally, my recent investigation shows that in the item itself, it should be possible to use properties other than the item's `value` property as the one that the RadioGroup apples to. This needn't, I believe, affect your PR greatly, as long as it doesn't *depend* on the item having a `value` property. See #9392 for what I'm talking about.

mever: Hi @derrell,

This PR includes the ```IField``` interface [here](https://github.com/qooxdoo/qooxdoo/pull/9367/files#diff-a5b5b141c7f0ed3db090c3995b0e9cb0R39). The implementation of this interface was added [here](https://github.com/qooxdoo/qooxdoo/pull/9367/files#diff-7084803c1921aa284f73e74ee39f1c21R68)(1) (```MSingleSelectionHandling``` mixin). ```RadioGroup``` already included this mixin ([see master](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/form/RadioGroup.js#L43)). That's why it doesn't show up in the diff. While, indeed, the ```RadioGroup``` is not implementing the ```IField``` methods + event directly. It does so by including the ```MSingleSelectionHandling```.

 As for #9392, as seen in the ```MSingleSelectionHandling```, the implementation of the interface relies on the ```SingleSelectionManager``` (see: 1). Therefore, it should not affect this PR.

I hope this clears it up a bit.

derrell (approved):

level420:
4 approvals, more than 2 business days since voting.
merging. thank you @mever

PR #9373, opened 2017-08-16
Drop down bugfix and consistency check

mever:
In order to solve the following error:
```
qx.ui.core.queue.Layout: Error in the 'Layout' queue:minHeight is larger than maxHeight!: Called assert with 'false' minHeight is larger than maxHeight!: Called assert with 'false'
```
I had to improve a dimension consistency check **844ae28**. By trowing the exception earlier (when setting the dimensions) it produces a more useful trace. Also the model (layout item in this case) should keep its own state consistent.

I also made a change in the abstract virtual box. Previously, you could press the down or up arrow key to open the list when it was closed. This change prevents the list from opening on key down or up when no visible items are present.

### DIFF
--- a/framework/source/class/qx/test/ui/core/SizeHint.js
+++ b/framework/source/class/qx/test/ui/core/SizeHint.js
@@ -179,12 +179,11 @@ qx.Class.define("qx.test.ui.core.SizeHint",
     testMinLargerThanMax : function()
     {
       this.setStretching(true, true);
-      this.setSize(200, 100, 150);
       if (this.isDebugOn())
       {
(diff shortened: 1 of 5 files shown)


mever:
I am closing this PR and opening two separate ones, one for the dimensions check and the other for the drop-down improvement.

mever:
Split in:

name | PR
--------|--------
dimensions check | https://github.com/qooxdoo/qooxdoo/pull/9377
change in drop down | https://github.com/qooxdoo/qooxdoo/pull/9378

PR #9375, opened 2017-08-17
Fixed menu and child menu items leak.

ghost:


derrell:
In master, there was a change pertaining to the Object Registry. I believe that most objects are now not registered there, unless explicitly requested (or unless some interface is included). @johnspackman added this change. Let's get his comments on whether the current PR is still necessary.

johnspackman:
That fix does not apply to Widgets, because all widgets register themselves with the ObjectRegistry (although since then another change has meant that even this could possibly be eliminated also), but for now this is the correct way to clean up.

johnspackman (approved):

mever:
You can also solve it like this:
```
var menu = new qx.ui.menu.Menu();
var button = new qx.ui.form.MenuButton();
button.setMenu(menu);
qx.util.DisposeUtil.disposeTriggeredBy(menu, button);
```
I think when you try to dispose the menu together with the button it may create a problem for users that use them separately. A good rule of thump is to dispose objects only when they are instantiated internally. Otherwise it should be clear by context (e.g.: factory method or other creational patterns) or by documentation. We could also add a comment to the property stating the disposal of the menu.

iceteagroup:
@mever, I agree. I was experiencing memory leaks with the  qx.ui.menubar.Button in qx.ui.menubar.MenuBar and since it extends qx.ui.form.MenuButton I noticed that it didn't destroy the connected menu to I fixed the base class. Maybe the fix should be in qx.ui.menubar.Button instead since the menu is an integral part of the qx.ui.menubar.Button. The same code is already present in qx.ui.menu.AbstractButton.

oetiker (approved):

cajus (changes requested):
It's not the task of the MenuButton to destroy something that has been provided from the outside. Maybe it's going to be re-used. The only thing that could be done in the destructor would be to reset the menu property.

iceteagroup:
@cajus Yes, I agree. As I wrote in response to @mever maybe the fix should be in qx.ui.menubar.Button instead since the menu is an integral part of the qx.ui.menubar.Button. The same code is already present in qx.ui.menu.AbstractButton.

The qx.ui.menubar.Button doesn't destroy the menu. The qx.ui.menu.AbstractButton does. In both cases the menu is provided externally.

johnspackman:
I'm changing my review on this - adding dispose now will break user code that handles disposing the menu item, i.e. there is nothing wrong with the way that users are handling it at present and as @mever and @cajus have said the "normal" strategy is that calling code cleans up - and it is arguable that `qx.ui.menu.AbstractButton` is wrong for disposing what it is given.

The clean solution to this is to implement auto dispose for widgets - I think that this *may* be possible, and opened an issue for it https://github.com/qooxdoo/qooxdoo/issues/9387

patrick-fischer-hirschstein (approved):

level420:
@iceteagroup @johnspackman should this PR be closed?

johnspackman:
yes, I think so

level420:
closing

PR #9377, opened 2017-08-18
Dimensions check

mever:
This PR actually fixes a bug in the virtual drop down list. But in order to discover the problem I had to change some core functionality. That's why I chose this PR title.

Without this change in core functionality I got the following error (before the fix in commit **7bc0d54**):
```
qx.ui.core.queue.Layout: Error in the 'Layout' queue:minHeight is larger than maxHeight!: Called assert with 'false' minHeight is larger than maxHeight!: Called assert with 'false'
```
Although the message is very clear it produced a trace that didn't make sense to me. Because the error was thrown in a part of the software that didn't execute in the same thread of execution as the code causing the problem. It was impossible to know (from looking at the code alone) where the root cause was.

I hope this change will help others to find problems faster and with less frustration.


oetiker (approved):
makes sense to me

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -439,19 +439,24 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
...
-        list.setMaxHeight(availableHeigth);
-      } else if (maxListHeight < availableHeigth) {
-        list.setMaxHeight(maxListHeight);
+      if (maxHeight > availableHeigth) {
+        maxHeight = availableHeigth;
       }
+
+      var minHeight = list.getMinHeight();
+      if (null !== minHeight && minHeight > maxHeight) {
+        maxHeight = minHeight;
```
derrell: Has this been tested at the extremes of container boundaries? Should there be tests of whether the increased max height will still fit in the container?

derrell (changes requested):

level420:
@mever could you please have a look at @derrell s question? Thank you!

level420:
@mever @derrell Could you please take a look again at this PR? Thank you!

derrell:
I would like to approve this PR if the answer to my prior question is that it has been tested at the extremes without ill effect.

mever:
Yesterday I explained in #9378 how this PR came to be. Because an unclear exception was thrown I had to improve that first.

To make the error less unclear I moved the min < max assertion from the place it was read to the place it was written. After this change the error trace led me to this method. Let me try to reproduce the original error. When I do I will provide a test for it.

@derrell When I've written the test I come back to you.

level420:
Thank you @mever !

milandamen:
I have this same problem with the VirtualDropDownList. I have managed to recreate this in a test environment.

To recreate:

1. Check out the master branch (this bug does not happen in 5.0.2 and so will not happen in the playground)
2. Create a fresh application and add the following code to it:
    ```
    var selectBox = new qx.ui.form.VirtualSelectBox();
    selectBox.setLabelPath('b');

    var win = new qx.ui.window.Window();
    win.setLayout(new qx.ui.layout.VBox());

    win.add(selectBox);
    win.open();
    qx.ui.core.queue.Manager.flush();

    var item1 = qx.data.marshal.Json.createModel({a:1, b: 'asd'});
    selectBox.setModel(new qx.data.Array([item1]));
    selectBox.open();
    selectBox.close();
    win.close();
    selectBox.getModel().removeAll();
    win.open();
    ```
3. Run generate and run the application. The following error should appear in the console:
    ```
    Assertion error! minHeight is larger than maxHeight!: Called assert with 'false'

    qx.ui.core.queue.Layout: Error in the 'Layout' queue:minHeight is larger than maxHeight!: Called assert with 'false' minHeight is larger than maxHeight!: Called assert with 'false'
    ```

This happens because the List is still inside the visibility queue and so the size hint will be computated even though it is not *really* visible. Because of this, `var itemsHeight = list.getPane().getRowConfig().getTotalSize();` will return `0` which will be set as the new maxHeight. But because there is no check to see if `0` is not less than minHeight, it will continue on without any warnings or errors. The error only gets thrown when Qooxdoo tries to draw the widget (setMaxHeight() adds it to the layout queue).

This problem has been really difficult to debug and would have been a lot easier to find with @mever's dimensions check.

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: I would argue that this is a non-issue because having maxHeight never be smaller than minHeight would be more important than whether or not it will fit in a container. I also don't think containers will be 42 pixels or less small.

johnspackman: I don't understand where the 42 pixels come from, but the argument that `minHeight` should never be greater than `maxHeight` seems reasonable to me.  

WRT whether `minHeight` should also be checked against the container (ie against `availableHeight`), this is only an issue if one of `maxHeight`/`minHeight` is an invalid value so it seems a bit of a moot point, except to say that adjusting `maxHeight` up at least makes sure that it is visible and the developer can look further (or maybe the user can resize the viewport).

johnspackman:
@milandamen's explanation of *how* the `maxHeight` came to be set below `minHeight` is very useful, and IMHO it suggests that `__adjustHeight` should be fixed because specifying 0 as a maxHeight is clearly not a sensible value, especially as it will trigger an assertion failure elsewhere in the same code.

This PR has been around for nearly 3 months now, presumably @mever has had this under test all during that time and as @milandamen would have benefited from this in production.

Either @mever can you confirm that this has been tested with smaller viewport sizes, and/or @derrell would you consider reconsidering your objection?

For me personally, I think the change is acceptable to approve at this stage.

johnspackman (approved):

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: Excuse me for not explaining where I got the 42 from.

When the `qx.ui.list.List` gets constructed, its base constructor `qx.ui.virtual.core.Scroller` gets called which in turn calls the base constructor of `qx.ui.core.scroll.AbstractScrollArea`. In this constructor a `size` gets calculated as follows:

```
var size = qx.ui.core.scroll.AbstractScrollArea.DEFAULT_SCROLLBAR_WIDTH * 2 + 14;
```

which will result in the number `42`.

This value will then be set as `minHeight` and `minWidth` of the List (LayoutItem).

derrell (approved):

level420 (approved):

level420:
4 approvals, more than 2 business days, merging
thank you @mever

cajus:
I've massive problems with this patch in our applications and I had to revert it locally.

PR #9378, opened 2017-08-18
[Changed] do not show drop down popup when no items are visible

mever:
Previously, you could press the down or up arrow key to open the list when it was closed. This change prevents the list from opening on key down or up when no visible items are present.


oetiker:
makes sense to me

johnspackman (approved):

derrell:
I'm wondering if this is intuitive behavior. I would expect a drop-down to open and show me an empty list, if the list is empty. If nothing happens, I'd feel that it was broken...

derrell (changes requested):
See my earlier comment about whether this is intuitive behavior.

level420:
@mever what happens currently if I try to open the dropw down list via mouse?

level420:
@mever @derrell I'd like to bring up this PR again. What is needed to have this merged? Could you please discuss? Thank you?

derrell:
I'm still not convinced that it's intuitive behavior. Do others think it is intuitive to have nothing happen when you open an empty list? I prefer to see an empty list show up. Maybe, if others like this new behavior, then a property to enable/disable it would be appropriate?

cboulanger (dismissed):
@derrell I don't know if it is intuitive or not - I've seen both behaviours in webapps, so I tend not to be surprised if either one of them is displayed. A property to control it sounds like a good idea.

level420:
@cboulanger it would be nice if you could turn your approval into a change request. thank you!

cboulanger (changes requested):
A property to control the behaviour as @derrell has proposed sounds like a good idea.

mever:
I do not really mind how the application behaves on this level, both cases are fine to me. I can remember an exception. Therefore I created the original PR https://github.com/qooxdoo/qooxdoo/pull/9373 (which I later split into separate PRs). And now that I think about it, the other part of the split will probably fix that exception (will look into it tomorrow). So that's the important one.

I made this because the error occurred while I was trying to build a custom combobox. The textfield childcontrol was serving as a filter while the list popup was used as result. This field allowed me to search potentially large lists. For example, you could search for "light" and get a list of drinks, like: "light beer", "cola light", etc...

When the textfield changed value using live-update it was sending a HTTP request to the server requesting a filtered list of objects. When the response returned it used was to populate the list. The problem, however, was that it had no visual indication of whether it was searching or whether it was done without result. I think (I need to verify this) that this fix hide the empty list when a search is in progress and show it (also when empty) by calling `open` in the HTTP response handling code.

level420:
@mever is this PR still valid?

johnspackman:
@mever i'm closing this for now, but feel free to reopen etc if you'd like to continue the discussion

PR #9379, opened 2017-08-19
Fixes JSDoc comments, plus a couple of very minor bugs detected by thee TypeScript generator

johnspackman:


cboulanger (commented):
Didn't we say we'd follow JSDoc `@param`'s spec to have the type first and then the variable name?

derrell:
We didn't say that (or at least, we did not agree to that). It's a MASSIVE change that affects both qooxdoo and user code. I agree it's a nice goal, but I don't believe it's practical to make the change unless we support both forms until version 7.0.

johnspackman:
👍  it's a possibility discussed in https://github.com/qooxdoo/qooxdoo-compiler/issues/22 but that should be a separate PR - this is about fixing some compliance problems with the existing standard.  It doesn't affect anything except the meta data which I'm now outputting  from `qooxdoo-compiler` which feeds the TypeScript generator and eventually the API viewer too.  In most cases, the TypeScript interpreter pointed out the problems - except the last commit which the generator picked up and I figured might as well be included.

cboulanger:
Ok - was just wondering why you changed it back. If the python generator doesn't understand both ways (I somehow thought it could), of course!

oetiker (approved):

derrell (approved):

level420 (approved):

level420:
3 approval, more than 3 business days, merging,
thank you @johnspackman

PR #9382, opened 2017-08-24
Strict mode compliance

johnspackman:
Various minor fixes for strict mode compliance, IE remove references to `arguments.callee` and `arguments.caller`

### DIFF
--- a/framework/source/class/qx/test/Bootstrap.js
+++ b/framework/source/class/qx/test/Bootstrap.js
@@ -292,7 +292,7 @@ qx.Class.define("qx.test.Bootstrap",
 
           stopEngine : function()
           {
-            var ret = arguments.callee.base.call();
+            var ret = qx.test.Bmw.prototype.stopEngine.base.call();
             return "brrr " + ret;
           },
 
--- a/framework/source/class/qx/test/Bootstrap.js
+++ b/framework/source/class/qx/test/Bootstrap.js
@@ -292,7 +292,7 @@ qx.Class.define("qx.test.Bootstrap",
 
           stopEngine : function()
           {
-            var ret = qx.test.Bmw.prototype.stopEngine.base.call();
+            var ret = this.base(arguments);
             return "brrr " + ret;
           },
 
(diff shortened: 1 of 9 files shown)


framework/source/class/qx/test/Bootstrap.js
```diff
@@ -292,7 +292,7 @@ qx.Class.define("qx.test.Bootstrap",
 
           stopEngine : function()
           {
-            var ret = arguments.callee.base.call();
+            var ret = qx.test.Bmw.prototype.stopEngine.base.call();
```
derrell: Is this the same as the following?
```
var ret = this.base(arguments);
```

.call() with no first argument means that the context is explicitly excluded. That sounds wrong to start with, unless there is no reference to `this` in the `...stopEngine.base` method...

derrell: Similarly, elsewhere in here. This is *really* old code you're modifying here.

derrell (changes requested):

framework/source/class/qx/test/Bootstrap.js
johnspackman: yes, really old - I guess this is why it was doing some slightly strange things to accomplish the tests; I was aiming to make the smallest changes and just translate the syntax, but I agree it doesn't make it any clearer!  I'll change it

framework/source/class/qx/util/Function.js
```diff
@@ -50,54 +50,44 @@ qx.Bootstrap.define("qx.util.Function", {
      * @return {Function} a wrapper function which <em>shields</em> the given callback function
      */
     debounce: function (callback, delay, immediate) {
+      var fired = false;
+      var intervalId = null;
       var wrapperFunction = function () {
-        arguments.callee.immediate = !!(immediate);
-
         // store the current arguments at the function object
         // to have access inside the interval method
```
derrell: This comment is now incorrect

derrell (changes requested):

cboulanger (approved):

derrell (approved):

oetiker (approved):

cajus (approved):

PR #9384, opened 2017-08-25
Enforce semver version numbers in Manifest.json

cboulanger:
Since we are switching to qooxdoo versions compatible with the [semver standard](http://semver.org/spec/v2.0.0.html), we also need to adapt the validation of the qooxdoo verson required by a library in Manifest.json. At it is now, it allows „trunk“ and „master“ as versions, which is illegal in semver and doesn’t make any sense in any case since it is no indication of any time-independent identifiable version. This PR updates the regular expression used to validate the `info.qooxdoo-version` entry in the manifest json.


tool/pylib/generator/config/Manifest.py
```diff
@@ -104,7 +104,8 @@ def patchLibEntry(self, libentry):
     @classmethod
     def schema_v1_0(self):
         patterns = {
-            "semver": r"^trunk$|master$|^\d+\.\d+(\.\d+)?(-[0-9]+-?)?([-a-zA-Z+][-a-zA-Z0-9.:-]*)?$",
+            # semver validation, see https://github.com/sindresorhus/semver-regex/blob/master/index.js
```
derrell: The `^` and `$` that were in the deleted line are important to match the entire string.

derrell (changes requested):

tool/pylib/generator/config/Manifest.py
cboulanger: Updated the rege accordingly...

oetiker:
the problem of specifying semver compatible version ranges has been explored at length already ... so instead of inventing something new, I would would suggest to rely on existing solutions ... 

for example

https://git.daplie.com/coolaj86/semver-utils

to prevent problems with python generator I would simply leave its version compatibility thing as is and create a new key in the config

cboulanger:
@oetiker This issue is not about semver **range**, but about enforcing the semver **version** it's already pretending to enforce (hence the name of the pattern in the code), but which it isn't because the pattern is not legal semver ( since it allows "trunk" and "master"). Does this address your worries?

oetiker:
for now I have specified something like 5.0 or 5.0.x in my contribs to point out that they would work with all 5.0ish versions ... both would not be semver compatible ... 

master and trunk are neither ... true

but instead of fixing this half way ... how about allowing contribs to specify a semver range

cboulanger:
@oetiker that's the idea, but in a different issue- it will be done in a different config key which is picked up by new compiler. So all you'll
Have to do is to fix the qooxdoo-version value and add a range in the new config-key, tentatively called 'qooxdoo-range'. I will create a separate issue for this..

oetiker:
great

oetiker (approved):

derrell (approved):
LGTM

derrell (approved):
I don't understand the two similar-looking (identical) PRs

cajus (approved):

johnspackman (approved):

patrick-fischer-hirschstein (approved):

level420 (approved):

level420:
6 approvals, merging.
thank you @cboulanger

PR #9388, opened 2017-09-05
Avoid exception on getting the focus target of compound widgets

level420:
Focus able widgets may not have a "textfield" child control resulting in an exception calling getChildControl("textfield"). To avoid this we first have to check if a child control named "textfield" exists.

### DIFF
--- a/framework/source/class/qx/event/handler/Focus.js
+++ b/framework/source/class/qx/event/handler/Focus.js
@@ -1058,7 +1058,7 @@ qx.Class.define("qx.event.handler.Focus",
         // Check compound widgets
         var widget = qx.ui.core.Widget.getWidgetByElement(focusedElement);
         if(widget) {
-          var textField = widget.getChildControl && widget.hasChildControl("textfield") && widget.getChildControl("textfield", true);
+          var textField = widget.hasChildControl("textfield") && widget.getChildControl("textfield", true);
           if (textField) {
             return textField.getContentElement().getDomElement();
           }
--- a/framework/source/class/qx/event/handler/Focus.js
+++ b/framework/source/class/qx/event/handler/Focus.js
@@ -1056,11 +1056,12 @@ qx.Class.define("qx.event.handler.Focus",
           return focusedElement;
         }
         // Check compound widgets
-        var widget = qx.ui.core.Widget.getWidgetByElement(focusedElement),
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

level420:
@ronkie could you please have a look into this PR? Thank you!

framework/source/class/qx/event/handler/Focus.js
```diff
@@ -1056,11 +1056,12 @@ qx.Class.define("qx.event.handler.Focus",
...
         }
         // Check compound widgets
-        var widget = qx.ui.core.Widget.getWidgetByElement(focusedElement),
-          textField = widget && widget.getChildControl && widget.getChildControl("textfield", true);
-
-        if (textField) {
-          return textField.getContentElement().getDomElement();
+        var widget = qx.ui.core.Widget.getWidgetByElement(focusedElement);
+        if(widget) {
+          var textField = widget.getChildControl && widget.hasChildControl("textfield") && widget.getChildControl("textfield", true);
```
johnspackman: as we're sniffing by method name, then it's necessary to sniff for `hasChildControl` also - there is no contract that says that `hasChildControl` will exist if `getChildControl` does.

johnspackman (changes requested):

framework/source/class/qx/event/handler/Focus.js
level420: Hmm. As of the api docs `qx.ui.core.Widget.getWidgetByElement` allways returns an instance of `qx.ui.core.Widget` which always should implement `hasChildControl` and `getChildControl`.
I think we don't need the checks at all.

derrell (approved):

patrick-fischer-hirschstein (approved):

ronkie (approved):
The changes are fine with me. However, as far as I see, getChildControl method does not throw anything itself. Only _createChildControl can throw an exception when control is not supported, but we don't call it when calling getChildControl with "notcreate == true".

In this case we also have a check for textField on line 1062, which seems enough to me.

level420:
Yep! Found one of my classes where I've overridden `getChildControl` without passing the second optional parameter `notcreate`. I'm unsure what I should do now with this PR. Close it?

derrell:
If it turns out that the PR is unnecessary, then yes, I'd say close it.

level420:
Closed because this was an error in my own code.

PR #9392, opened 2017-09-22
Radio group specify property

derrell:
Allow RadioGroup to group by other than the 'value' property.

Prior to this change, widgets to be grouped in a RadioGroup had to have a
'value' property which was monitored for changes to ensure that no more than
one of the grouped widgets was active at any one time; and a 'group'property,
which was set to the RadioGroup object to which the widget was added.

With this change, the two property names can be specified. The RadioGroup
object has two new properties:

- The property groupedProperty, which defaults to 'value', is the property on
  which widgets are grouped.

- The property groupProperty, which defaults to 'group', is the property in
  the widgets which is set by RadioGroup to tell a widget what RadioGroup
  object it is associated with.

It is now possible for a set of widgets to have multiple properties that are
independently maintained as mutually exclusive via multiple RadioGroups, each
responsible for the mutual exclusivity of a particular property. For example,
if a group of widgets can have only one of them with their 'locked' property
set to true, and the same set of widgets can have only one of them with their
'ready' property set to true, then the widgets would be defined with
properties:

    locked :
    {
      init     : null,
      check    : "Boolean",
      apply    : "_applyLocked",
      event    : "changeLocked",
      nullable : false
    },

    // set by the RadioGroup
    lockedGroup :
    {
      init     : null,
      check    : "qx.ui.form.RadioGroup",
      nullable : true
    },

    ready :
    {
      init     : false,
      check    : "Boolean",
      apply    : "_applyReady",
      event    : "changeReady",
      nullable : false
    },

    // set by the RadioGroup
    readyGroup :
    {
      init     : null,
      check    : "qx.ui.form.RadioGroup",
      nullable : true
    },

Then two radio groups can be defined:

    var rgLocked = new qx.ui.form.RadioGroup();
    var rgReady = new qx.ui.form.RadioGroup();

    rgLocked.set(
      {
        groupedProperty     : "locked",
        groupProperty       : "lockedGroup"
      });

    rgReady.set(
      {
        groupedProperty     : "ready",
        groupProperty       : "readyGroup"
      });

Then, widgets which define the 'locked', 'lockedGroup', 'ready', and
'readyGroup' properties can be added to rgLocked and rgReady.

### DIFF
--- a/framework/source/class/qx/test/ui/form/RadioGroup.js
+++ b/framework/source/class/qx/test/ui/form/RadioGroup.js
@@ -79,6 +79,9 @@ qx.Class.define("qx.test.ui.form.RadioGroup",
       composite.destroy();
     },
 
+    /**
+     * @ignore(qx.test.ui.form.RadioGroupTest)
+     */
     testAlteredGroupProperty : function()
     {
       qx.Class.define("qx.test.ui.form.RadioGroupTest",
--- a/framework/source/class/qx/test/ui/form/RadioGroup.js
+++ b/framework/source/class/qx/test/ui/form/RadioGroup.js
@@ -77,6 +77,73 @@ qx.Class.define("qx.test.ui.form.RadioGroup",
       this.assertEquals(this.__radioButtons[0].getModel(), this.__radioGroup.getModelSelection().getItem(0), "Model selection does not work correctly!");
       this.assertTrue(this.__radioGroup.isSelected(this.__radioButtons[0]), "Hidden radio button not selected!");
       composite.destroy();
+    },
+
+    testAlteredGroupProperty : function()
+    {
+      qx.Class.define("qx.test.ui.form.RadioGroupTest",
+      {
+        extend : qx.core.Object,
+
+        properties :
+        {
+          locked :
+          {
+            init     : false,
+            check    : "Boolean",
+            event    : "changeLocked",
+            nullable : false
+          },
+
+          // set by the RadioGroup
+          lockedGroup :
+          {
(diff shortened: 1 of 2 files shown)


level420:
@derrell does your comment related to PR #9367 affect this PR in any way?

level420:
@derrell the radio group test is creating a warning about `qx.test.ui.form.RadioGroupTest` being an unknows symbol. You have to add ` * @ignore(qx.test.ui.form.RadioGroupTest)` to that test. Just to stay clean. This is not the cause for the failed travis test.

```
----------------------------------------------------------------------------
    Executing: test-source::testrunner::source-tests-script
----------------------------------------------------------------------------
>>> Copying application files...
>>> Scanning libraries  
 >>> Collecting classes   
   - Warning: qx.test.ui.form.RadioGroup (118): Unknown global symbol used: qx.test.ui.form.RadioGroupTest
  - Processing explicitly configured includes/excludes...
  - Sorting 1250 classes  
 >>> Generate application
  - Processing 6 locales  
   - Generate packages  
   - Generate loader script
>>> Done (2m00.89)
```

derrell:
It does not appear that there is any negative interaction with #9367.

I've eliminated the warning. Thanks. Hadn't seen it.

cboulanger (approved):

level420 (approved):

cajus (approved):

johnspackman (approved):

level420:
4 approvals, more than 2 business days
merging
thank you @derrell

PR #9394, opened 2017-09-28
Proposed fix for #9393

### LINKED ISSUE #9393: Chrome: blocker does not block keyboard events
Here's how to reproduce it:

 * open [the playground](http://tinyurl.com/y9fk83ow) in google chrome
 * click on the 'Button' (a modal window opens)
 * press "Enter" or "Space"

What happens:
 * the "Button" is executed a second time, opening another modal window

What is expected to happen:
 * nothing, because the button is blocked by the modal window

The problem only exists in Chrome. Firefox does not open the second window. Note, that it only works once.

cajus:
The focus handler receives the "native mouse up" event delayed. As a
result the blocker already passed the focus to the application root.
Then the chrome event drops in and puts the focus back on the button.

closes #9393

### DIFF
--- a/framework/source/class/qx/event/handler/Focus.js
+++ b/framework/source/class/qx/event/handler/Focus.js
@@ -251,9 +251,15 @@ qx.Class.define("qx.event.handler.Focus",
       }
       else
       {
-        try {
-          element.focus();
-        } catch(ex) {}
+        // Fix re-focusing on mousup event
+        // See https://github.com/qooxdoo/qooxdoo/issues/9393 and
+        // discussion in https://github.com/qooxdoo/qooxdoo/pull/9394
+        window.setTimeout(function()
+        {
+          try {
+            element.focus();
+          } catch(ex) {}
+        }, 0);
       }
 
       this.setFocus(element);


cajus:
Looks like the tests nave to be addressed. Or there's another solution for that.

framework/source/class/qx/ui/core/Blocker.js
```diff
@@ -261,7 +261,9 @@ qx.Class.define("qx.ui.core.Blocker",
       this.__focusElements.push(qx.ui.core.Widget.getWidgetByElement(focusHandler.getFocus()));
 
       if (this._widget.isFocusable()) {
-        this._widget.focus();
+        setTimeout(function() {
+          this._widget.focus();
+        }.bind(this), 0);
       }
```
level420: The failing test awaits the focus being set to the widget right after the blocker blocks, but that is not the case anymore due to the delayed focusing.

Does it do any harm if the focusing the widget is called twice? If not we could do instead:
```javascript
      if (this._widget.isFocusable()) {
        this._widget.focus();
        qx.event.Timer.once(this._widget.focus, this._widget, 0);
      }
```
which ensures that the focus is still at the widget after the current code gives the control back to the engine.

cajus: We had discussed that, too. Feels a bit "hackish" ;-)

cajus: But maybe it's not more hackish than the setTimeout...

level420: the question is if the `this._widget` fires the focus event twice in this situation which would make a difference

level420: In the end it may be that we get a `focus` -> `blur` -> `focus` event sequence instead of a single `focus` event

level420:
the question is if `this._widget` fires the focus event twice in this situation which would make a difference.

In the end it may be that we get a `focus -> blur -> focus` event sequence instead of a single `focus` event

cajus:
It will fire that. But that only has an effect on the "active" property of the focus handler. The problem is that chrome sends "old" mouse events for compatibility reason.

framework/source/class/qx/ui/core/Blocker.js
cajus: The interresting thing is that it doesn't differ. If you've just the `this._widget.focus()`, you'll have 6 focus changes in Chrome (due to the *mouseup*). If you add the `setTimeout` path, you've just 4 focus changes, because the *mouseup* doesn't change the initial "Button is focussed".

level420:
I had a deeper look into the routing of native events in `qx.event.handler.Focus.js` especially here
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/Focus.js#L468-L488 
where the native listeners are assigned and here
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/Focus.js#L1000
where the mouse up event is handled and the `__fixFocus` method is called.

Additionally there is a map of focusable elements here
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/Focus.js#L128
which defines for a specific engine/browser which elements can get the focus.

My "feeling" is, that there is a problem as chrome is still detected as
```
qx.core.Environment.get("engine.name")
"webkit"
```
which is not true anymore, as it should be `blink` which may have other specifics which have to be taken into account.

Still this PR seems to be, as you said, "hackish".

cajus:
@level420 if you've more insights in the event handling code, feel free to replace this one by something that does it on low level. Note that firefox also sends that event and triggers re-focusing. But it does it *earlier* than chrome, so it - by accident - works.

I don't know whether these events needs to be "swallowed" by the qx event layer or not. Changing the behaviour there may cause "interresting" issues.

level420:
I don't want to alter something in the event sequence in `qx.event.handler.Focus.js`. I just want to make sure that there is no bug or historically grown incompatibility in there which causes this issue. If we don't find anything there we have to live with the current state of the PR.

cajus:
Yeah. It sounded like you were close before solving it ;-)

level420:
No, far away from a solution, besides removing the native binding to the mouseup event in general ;-) Then it works! Just digging and wanted to give a hint where to search.

cajus:
If the _mouse*_ events are not needed anymore, that would be an option.

level420:
@cajus could you give this solution a try:
https://github.com/level420/qooxdoo/commit/9a29380c964cb7cd3415c80b500922616471d908

This moves the defered focusing via setTimeout where it belongs to. And there is a delayed focusing already present for `mshtml` which worked for long time now.

I've found no negative impact on my main application and the testrunner is happy.

cajus:
@level420 will do so tomorrow...

cajus:
@level420 that one works, too. Maybe it's a cleaner solution. Will you do another PR, or should I change this one to use your solution?

level420:
@cajus please change this PR. thank you.

cajus:
Done. Thanks!

framework/source/class/qx/event/handler/Focus.js
```diff
@@ -251,9 +251,15 @@ qx.Class.define("qx.event.handler.Focus",
...
-        } catch(ex) {}
+        // Fix re-focusing on mousup event
+        // See https://github.com/qooxdoo/qooxdoo/issues/9393 and
+        // discussion in https://github.com/qooxdoo/qooxdoo/pull/9394
+        window.setTimeout(function()
+        {
+          try {
+            element.focus();
+          } catch(ex) {}
+        }, 0);
```
derrell: Is this *solely* a matter of the stack needs to unwind, or does it possibly (sometimes?) require a timeout longer than 0?

derrell (approved):

level420 (approved):

cboulanger (approved):

level420:
3 approvals, more than 2 business days, merging
thank you @cajus

PR #9395, opened 2017-09-29
Add a controller for synchronizing model open property with VirtualTree state

derrell:


framework/source/class/qx/ui/tree/core/OpenCloseController.js
```diff
@@ -0,0 +1,152 @@
...
+    {
+      var bubble = ev.getData();
+      var modelPropRe;
+
+      // generate a regular expression that identifies model changes that
+      // pertain to the open state of a branch in the tree.
+      modelPropRe = new RegExp("\\." + this.getOpenPropertyName() + "$");
+      
+      // open related? sync it back to the node item.
+      if (modelPropRe.test(bubble.name)) {
```
cajus: If this is meant to be generic, then relying on change bubbles may be problematic. Using the JSON marshaller, you can disable bubbling for the automatic properties - and then this construct will never notice that the `openPropertyName` changed. I.e. we only enable bubbling when there's absolutely no other way, because it is very ineffective. In one project, we've a large tree. Disabling event propagation makes initializing the model up to 500% faster.

Just a thought: maybe we can use `qx.event.message.Bus` inside the `VirtualTreeItem` objects as a dispatcher and the method I'm currently commenting as the receiver. The `VirtualTreeItem` could establish a listener to the `openPropertyName` of it's model - and when it changes, it puts it's model as the payload on the message bus topic *treeId.open*. Then we've a single event sink and no need for bubbling. And the listener will not be called on every property change.

Drawback: if you want to use this feature, you need a delegate for i.e. *configureItem* in order to configure the treeId and the openPropertyName.

If this code would be included directly to the VirtualTree class (enabled by a property 'syncOpenState'), a tree-name could be omitted.

derrell: Thanks for the comments, @cajus. I will look into the implementation you suggest. I tried, however, disabling bubbling in the model creation, and this still works. I'm wondering why? See the following gist playground snippet, and notice `false` in the model creation on line 129, which I think should disable change bubbles; and the timer at the end which alters the model: https://gist.github.com/derrell/d5069ab6a981192ac30eff8d969d8565

cajus: I guess that's because you choose item[0] for the test. It's the one we're listening on.

derrell: It works even if I use for the test:

```
qx.event.Timer.once(
  function()
  {
    model.getItem(0).getChildren().getItem(0).setOpen(true);
  },
  this,
  2000);
```

But in any case, let me work on the alternate implementation and assuming I can get it working, we can then choose one or the other. Thanks.

cajus: Sure :+1:. No sure why it works. Either the *true/false* flag for the JSON marshaller is not doing what it says, or I don't know. We've some XML or Protobuf models here, so we don't use the marshaller at all.

Sorry for being a bit offline. The kids... ;-)

derrell:
Closed #9396. The method implemented here appears to be the one true and correct method to rule them all.

cajus (commented):
Maybe we can add tests for the controller?

cboulanger:
Why do the tests fail?

level420:
@derrell could you please have a look into the failing tests? Thank you.

level420 (changes requested):
Please check the failing tests.

cajus:
@derrell this fixes the tests:

```
diff --git a/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js b/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
index f5bb241fb4..bf28cf7d1c 100644
--- a/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
+++ b/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
@@ -22,6 +22,14 @@ qx.Class.define("qx.test.ui.tree.virtual.WidgetProvider",
   implement : qx.ui.tree.core.IVirtualTree,
   include : qx.dev.unit.MMock,
 
+  properties :
+  {
+    openProperty : {
+      check : "String",
+      init : null
+    }
+  },
+
   members :
   {
     model : null,
```

derrell:
Thanks, @cajus. I knew that's what it would be, but I've been so busy I hadn't gotten to it. I'll work on this during the weekend and hopefully even find a way to add a test.

derrell:
Superseded by #9406

PR #9396, opened 2017-09-30
Open close messagebus

derrell:
This is an alternate implementation of #9395 which uses the message bus instead of change bubbles to communicate changes of the model pertaining to opening/closing branches. See #9395 for usage details.

I was able to avoid the downside described by @cajus, of the user having to provide a configureItem delegate. The message name is determined internally, and the open property name is communicated via a property of the tree, as set by the OpenCloseController.

### DIFF
--- a/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
+++ b/framework/source/class/qx/test/ui/tree/virtual/WidgetProvider.js
@@ -463,6 +463,8 @@ qx.Class.define("qx.test.ui.tree.virtual.WidgetProvider",
 
 
     closeNode : function(node) {},
-    closeNodeWithoutScrolling : function(node) {}
+    closeNodeWithoutScrolling : function(node) {},
+
+    getOpenPropertyName : function() { return null; }
   }
 });
--- a/framework/source/class/qx/ui/tree/VirtualTree.js
+++ b/framework/source/class/qx/ui/tree/VirtualTree.js
@@ -16,7 +16,7 @@
 
 ************************************************************************ */
 
-/**
+/*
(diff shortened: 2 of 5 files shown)


derrell:
I can iterate on this work:
- Setting openProperty in VirtualTree can automatically instantiate the OpenCloseController, so it needn't be separately instantiated by the user (and disposed by the user).
- Furthermore, the controller functionality can easily be integrated directly into VirtualTree, avoiding an extra class, if we decide that's appropriate.

I will await comments on this PR as is, first, though, and then iterate based on comments. (So please comment on the above-mentioned iterations as well as what's already in the PR code changes.)

cajus:
@derrell looks very good to me.

For the proposed integration into the virtual tree: are there potential other uses for that class? I mean, is it just virtual-tree related? If so, it could be included.

:+1: for maintaining it automatically when the property is set /removed

derrell:
@level420 With @cajus approval of the general implementation hereof, why don't you remove the request for review for now. Let me get the enhancements I mentioned into place first.

@cajus It is tied to VirtualTree at present, as that's where the linkage of the openProperty is (currently in an internal-use-only property), referenced by both the OpenCloseController and the virtual tree's WidgetProvider. Therefore I don't think it can be used elsewhere. I'll incorporate the controller functionality into VirtualTree.

level420:
done

derrell:
This attempt appears to be a dead end. When branches are closed, their sub-elements' VirtualTreeItems never receive model changes, so bus messages are never issued. Attempts to set an invisible branch to be open when it next appears fail.

Closing this attempt. Reverting back to #9395 which has now been updated.

PR #9397, opened 2017-10-01
Fix Carousel height

jstaehli:
In some situations the height of the Carousel Widget is not correctly set.

Ways to reproduce:

- Minimize and maximize browser window
- Open and close dev tools (F12)
- Change orientation of mobile device (portrait/landscape)

Tested in:

- Safari (iOS 11)
- Chrome 61
- Firefox 56
- Edge
- Internet Explorer 11

Example:

step1

step2

step3

When removing the code line that sets the height "manually" the problem seems to be solved. I have not seen any negative effect when doing so in the browsers mentioned above.


framework/source/class/qx/ui/mobile/container/Carousel.js
```diff
@@ -438,7 +438,6 @@ qx.Class.define("qx.ui.mobile.container.Carousel",
       for (var i = 0; i < this.__pages.length; i++) {
         var pageContentElement = this.__pages[i].getContentElement();
         qx.bom.element.Style.set(pageContentElement, "width", carouselSize.width + "px");
-        qx.bom.element.Style.set(pageContentElement, "height", carouselSize.height + "px");
```
derrell: This implies that carouselSize, as determined by `qx.dom.element.Dimension.getSize` on line 427, is incorrect, but that setting the height isn't necessary. In that case, why was it set in the first place? And if it's not necessary for the height, is that true also of the width? I think maybe we should try to understand what's failing and why, a bit better?

jstaehli: You are right, investigating the issue a bit further makes sense, but I probably need some help of you. What I have observed so far is that it is required to set the width, otherwise a page does not cover the whole screen. It is not required to set the height, because "height: 100%" probably gets inherited at some point (.mobile.root), but I'm not sure if this is the case for all (old) browsers. What I have not found out yet is why the width always gets correctly set while the height does not...

oetiker: my guess is that the problem is the container element referred to on line 427. It has probably the wrong size but it may be invisible. the 'problem' with qx mobile is that not all sizes are determined explicitly by qx (like in desktop) but some things are left to the browser ... the display property would be something to look at ... 

just guessing here ...

level420:
last (real) activity nearly 2 years ago. closing this.
@jstaehli please reopen if the PR is still valid

PR #9400, opened 2017-10-05
fix debounce function to always use the arguments from the latest cal…

peuter:
…l instead of the initial ones + test for that case.

The fix is simple, just move the `args` declaration outside of the wrapper function.

The `testDebounce` TestCase tests that behaviour, without the fix the debounced function would have been called with the initial `true`argument instead of the last calls `false` argument.

### DIFF
--- a/framework/source/class/qx/util/Function.js
+++ b/framework/source/class/qx/util/Function.js
@@ -52,9 +52,10 @@ qx.Bootstrap.define("qx.util.Function", {
     debounce: function (callback, delay, immediate) {
       var fired = false;
       var intervalId = null;
+      var args = null;
       var wrapperFunction = function () {
         // store the current arguments to have access inside the interval method
-        var args = qx.lang.Array.fromArguments(arguments);
+        args = qx.lang.Array.fromArguments(arguments);
 
         // it's necessary to store the context to be able to call
         // the callback within the right scope
--- a/framework/source/class/qx/test/util/Function.js
+++ b/framework/source/class/qx/test/util/Function.js
@@ -1,0 +1,60 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2007-2008 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Fabian Jakobs (fjakobs)
+
+************************************************************************ */
+
+qx.Class.define("qx.test.util.Function",
+{
+  extend : qx.dev.unit.TestCase,
+  include: qx.dev.unit.MMock,
+
+  members :
+  {
+    testDebounce : function()
+    {
+      var test = this.stub();
+
+      var debouncedTest = qx.util.Function.debounce(test, 10);
+
+      debouncedTest(true);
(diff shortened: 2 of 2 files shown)


level420 (approved):

framework/source/class/qx/util/Function.js
```diff
@@ -52,9 +52,10 @@ qx.Bootstrap.define("qx.util.Function", {
     debounce: function (callback, delay, immediate) {
       var fired = false;
       var intervalId = null;
+      var args = null;
       var wrapperFunction = function () {
         // store the current arguments to have access inside the interval method
-        var args = qx.lang.Array.fromArguments(arguments);
+        args = qx.lang.Array.fromArguments(arguments);
```
derrell: What was the incorrect behavior of the original code? The difference here is that args is now retaining closure on debounce instead of only wrapperFunction being retained, but `fromArguments` is called regardless, and nothing uses args outside of wrapperFunction. I'm not seeing a functional change here...???

peuter: Take this example from the provided test case:
```
var debouncedTest = qx.util.Function.debounce(test, 10);
 debouncedTest(true);
 debouncedTest(false);
```
Without the fix the debounced function is called with argument `true`, with the fix it is called with argument `false`.

The wrong behaviour might have been introduced with this change: 9c972d955229e5e86c7928951557195058055013

bus i havent't checked that.

cajus: Adding: That's because `args` is defined inside of the wrapper function and thus is persisted in the callback. There's no chance of getting the last value in there.

cajus (approved):

framework/source/class/qx/util/Function.js
derrell: Maybe it's just too early in the morning. I'm not seeing how it could possibly make a difference. debounce() returns the wrapper function. The first thing the wrapper function does is convert `arguments` to `args` as an array. That should be taking the arguments from the current call to debouncedTest, whether args is defined at line 58 or at line 55. The callback will use (and retain by clousre, if necessary) the current value of args, on each call to debouncedTest.

Since it makes a difference, I am clearly missing something. I don't see any harm in this PR, but would love to figure out why it should improve anything.

cboulanger (approved):

framework/source/class/qx/util/Function.js
cajus: If the `args` is defined inside the wrapper function, it has a different "pointer" every time the wrapper is called. The inner (setInterval) function only has the *initial* "pointer". Moving `args` one level up makes sure that the "pointer" is the same in every subsequent call.

derrell (approved):

framework/source/class/qx/util/Function.js
cajus: @peuter, yes -
 it was introduced by https://github.com/qooxdoo/qooxdoo/commit/9c972d955229e5e86c7928951557195058055013

level420:
@johnspackman could you please have a look into this, as it seems the PR/merge introducing this bug is yours. Thank you.

johnspackman:
@level420 no problem, I'm probably not in a position to today so please can you hang on before merging

johnspackman (approved):
yes this fix is correct, it restores the behaviour to how it was before my commit

cajus:
>2 days. 5 votes, no one against. Thanks @peuter!

Issue #9402, opened 2017-10-05
qx.util.Uri.parseUri() goes wrong when @ in query


mrfeet:
A query string containing the '@' character is interpreted wrong, and everything after the @ is considered to be the host name.

Running this command:
`qx.util.Uri.parseUri("http://localhost/?separator=@")`
results in
>{anchor: "", query: "", file: "", directory: "", path: "", …}
>anchor:""
>authority:"localhost/?separator=@"
>directory:""
>file:""
>host:""
>password:""
>path:""
>port:""
>protocol:"http"
>query:""
>queryKey:{}
>relative:""
>source:"http://localhost/?separator=@"
>user:"localhost/?separator="
>userInfo:"localhost/?separator="

Where there is a value for 'user' and 'query' is empty.

Changing the '@' into 'a' results in something correct:
>{anchor: "", query: "separator=a", file: "", directory: "/", path: "/", …}
>anchor:""
>authority:"localhost"
>directory:"/"
>file:""
>host:"localhost"
>password:""
>path:"/"
>port:""
>protocol:"http"
>query:"separator=a"
>queryKey:{separator: "a"}
>relative:"/?separator=a"
>source:"http://localhost/?separator=a"
>user:""
>userInfo:""

Where 'user' is empty as expected and 'query' contains the query

mrfeet:
The problem is in the regular expressions used for parsing the url. Both the strict and loose one allow too many characters in front of the '@' that is a separator between userid (and optional password) and the domain/ip address.

mrfeet:
I have supplied a pull request with the proper regex:

https://github.com/qooxdoo/qooxdoo/pull/9403

derrell:
I'm not sure this is quite proper. I believe that using an `@` at that point in a URL is not legal; it should be encoded. My basis for this is in this text of RFC3986:

```
2.2.  Reserved Characters

   URIs include components and subcomponents that are delimited by
   characters in the "reserved" set.  These characters are called
   "reserved" because they may (or may not) be defined as delimiters by
   the generic syntax, by each scheme-specific syntax, or by the
   implementation-specific syntax of a URI's dereferencing algorithm.
   If data for a URI component would conflict with a reserved
   character's purpose as a delimiter, then the conflicting data must be
   percent-encoded before the URI is formed.

      reserved    = gen-delims / sub-delims

      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"

[...]

PR #9403, opened 2017-10-05
qx.util.Uri.parseUri() goes wrong when @ in query #9402

mrfeet:


johnspackman:
Following on from the conversation in https://github.com/qooxdoo/qooxdoo/issues/9402, while I agree with @derrell 's concerns about "where do we stop" I think that on balance I would vote for this because IMHO it's better to match browsers, especially when URLs are quite possibly entered by users who do not necessarily have respect for specs.

However - is there some way to reformat that regex?  I find it impossible to read and really difficult to evaluate the change.

oetiker:
yes please if the regexp gets modified, please convert it into a more readable version ... like this for example (this is NOT the same regexp ... it just shows the concept)

```javascript
var urlRegex= new RegExp(''
  + /(?:(?:(https?|ftp):)?\/\/)/.source     // protocol
  + /(?:([^:\n\r]+):([^@\n\r]+)@)?/.source  // user:pass
  + /(?:(?:www\.)?([^\/\n\r]+))/.source     // domain
  + /(\/[^?\n\r]+)?/.source                 // request
  + /(\?[^#\n\r]*)?/.source                 // query
  + /(#?[^\n\r]*)?/.source                  // anchor
);
```

level420:
@mrfeet could you please have a look into the changes requested by @oetiker and @johnspackman ? Thank you.

mrfeet:
Sure I can try. Although it would make it harder to use a site like https://regex101.com/ to actually test the regex...

mrfeet:
Not sure I can simplify it more than this, otherwise the groups are not matching the code anymore:

```
parser: {
         strict: new RegExp(''
			+ /^(?:([^:\/?#]+):)?/.source                                                                                   // protocol
			+ /(?:\/\/((?:(([^:@?]*)(?::([^:@?]*))?)?@)?((?:\[[0-9A-Fa-f:]+\])|(?:[^:\/?#\[\]]*))(?::(\d*))?))?/.source	// "//"(userid(:password)@)(domainname|ipv4address|ipv6address)
			+ /((((?:[^?#\/]*\/)*)([^?#]*))(?:\?([^#]*))?(?:#(.*))?)/.source                                                // (path)(?query)(#anchor)
			),
          loose:  new RegExp(''
			+ /^(?:(?![^:@?]+:[^:@?\/]*@)([^:\/?#.]+):)?/.source														// protocol
			+ /(?:\/\/)?/.source																						// "//"
			+ /((?:(([^:@?]*)(?::([^:@?]*))?)?@)?((?:\[[0-9A-Fa-f:]+\])|(?:[^:\/?#\[\]]*))(?::(\d*))?)/.source			// (userid(:password)@)(domainname|ipv4address|ipv6address)
			+ /(((\/(?:[^?#](?![^?#\/]*\.[^?#\/.]+(?:[?#]|$)))*\/?)?([^?#\/]*))(?:\?([^#]*))?(?:#(.*))?)/.source		// (path)(?query)(#anchor)
					)
}
```

level420:
@mrfeet please change the PR based on your last suggestion. thank you.

johnspackman:
Thank you @mrfeet for taking the time to reformat it, and I take your point about not being able to quickly copy & paste it into an online debugger/formatter.

Perhaps we should just treat regex's like this as binary data and expect it to be unreadable when in code, in which case we would rely on third party tools like https://regex101.com/ (a quick google shows up others too, eg https://www.debuggex.com/ and https://regexper.com/) to decode, debug, and test the expression?

That regexpr.com does visual diagram too which may be more effective than any text formatting we might be able to put in code.  

So despite having asked for reformatting, I'm now leaning towards putting it back the way it was originally in this PR (sorry @mrfeet !!) because out of two imperfect solutions, the first one is at least a plain vanilla regex that anything third party tool can understand.

cajus (approved):
I personally never used a @ inside of URL parmeters. If the specs say it's basically valid, I'm fine with changing the regexp. Although it's hard to see the changes in the diff viewer. Even if there are tools to visualize what's going on, both are complex enough to let you spend some minutes with analyzing them. I'd skip this as time does not allow it in the moment and trust the tests.

johnspackman (approved):

level420 (approved):
I'm trusting the tests.

level420:
3 approvals, more than 2 business days, merging
thank you @mrfeet

PR #9411, opened 2017-10-20
fix warning " Unresolved use of symbol _wrap.$$original" with new qxcompiler

hkollmann:
With the new qxcompiler we get an error message:

 qx.Class: [1803,8] Unresolved use of symbol _wrap.$$original

The compiler transpiled the wrapper function into a named function wrapper and changed the variable wrapper to _wrapper.
In this solution i named the wrapper function so that the compiler did not change the names.

### DIFF
--- a/framework/source/class/qx/Class.js
+++ b/framework/source/class/qx/Class.js
@@ -1745,7 +1745,7 @@ qx.Bootstrap.define("qx.Class",
      */
     __wrapConstructor : function(construct, name, type)
     {
-      var wrapper = function()
+      var wrapper = function wrapFunc()
       {
         var clazz = wrapper;
 


cboulanger (approved):

johnspackman (changes requested):
Thanks for the PR but I don't agree with the change - there is nothing wrong with the code that is in `Class.js`, it's qxcompiler which is at fault here because it is incorrectly identifying `wrapper` as an unresolved symbol (with a strangely different name)

johnspackman:
this is fixed now in `qxcompiler` with https://github.com/qooxdoo/qooxdoo-compiler/commit/7d5bdbcb8a74758ce8209db30f546a36ad537b81, along with fixes for a series of other spurious/inaccurate warnings etc - in fact, a complete compile from scratch should produce no warnings at all now from the current code base.

Note that this is not published to npm yet because I havn't tested it properly, but npm release should be very soon

PR #9414, opened 2017-10-24
revert 7e40ccb. It brings in qx.bom.Selector which disallows non-GUI use

derrell:
I've reopened #9353 because the commit is not backward compatible and prevents non-GUI code from using the ResourceManager. This reverts 7e40ccb to allow non-GUI use, and we need to find a different way of solving the IE problem that #9353 was intending to solve.

### DIFF
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -71,7 +71,7 @@ qx.Class.define("qx.util.ResourceManager",
       if(!registry) {
         return null;
       }
- 
+  
       var ids = [];
       for (var id in registry) {
         if (registry.hasOwnProperty(id)) {
@@ -81,7 +81,7 @@ qx.Class.define("qx.util.ResourceManager",
           ids.push(id);
         }
       }
- 
+  
       return ids;
     },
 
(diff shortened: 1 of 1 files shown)


johnspackman:
can we fix this by adding `@ignore(qx.bom.Selector)` and sniff testing instead of reverting the whole PR?

level420:
@derrell @johnspackman why don't we try to replace the use of `qx.bom.Selector` by native calls, but leave the commit else as it is:

We could use instead
```javascript
    var href;
    // first check if there is a base url set
    if (document.getElementsByTagName('base').length > 0) {
        href = document.getElementsByTagName('base')[0].href;
    } else {
```

derrell:
that would have to be

derrell:
Can't reference document in non-GUI code. In fact, you can't even depend on being able to check window.document, but I suspect that most non-GUI code defines window just for one's own sanity...

```
    var href;
    // first check if there is a base url set
    if (window.document && document.getElementsByTagName('base').length > 0) {
        href = document.getElementsByTagName('base')[0].href;
    } else {
```

derrell:
I'll put together a patch based on that and see if it works (for me, with my code). We may still encounter problems if code doesn't define `window` as a map.

level420:
Well then it should be:
```javascript
    var href;
    // first check if there is a base url set
    if (window && window.document.getElementsByTagName('base').length > 0) {
        href = window.document.getElementsByTagName('base')[0].href;
    }
```

derrell:
Nope. If `window` is undefined, that will crash.

level420:
Ahm yes, then:
```javascript
    var href;
    // first check if there is a base url set
    if (typeof window  !== 'undefined' && window.document && window.document.getElementsByTagName('base').length > 0) {
        href = window.document.getElementsByTagName('base')[0].href;
    }
```

derrell:
Yeah, that's likely the right way to do it.

derrell:
This PR is likely superseded by #9415

level420:
@derrell I'm closing this PR in favor of #9415

PR #9415, opened 2017-10-24
Remove use of qx.bom.Selector in qx.util.ResourceManager

derrell:
ResourceManager is used in non-GUI code. qx.bom.Selector
brings in the Sizzle library, which does a whole bunch of
initialization that requires a GUI environment (expects
'document' to exist).

This patch replaces the use of qx.bom.Selector with explicit
tests for existence of 'window' and 'window.document' and
use of the element API.

### DIFF
--- a/framework/source/class/qx/util/ResourceManager.js
+++ b/framework/source/class/qx/util/ResourceManager.js
@@ -285,9 +285,10 @@ qx.Class.define("qx.util.ResourceManager",
 
           var href;
           //first check if there is base url set
-          var baseElements = qx.bom.Selector.query("base", document);
-          if (baseElements.length > 0) {
-            href = baseElements[0].href;
+          if (typeof window == "object" &&
+              window.document &&
+              window.document.getElementsByTagName("base").length > 0) {
+            href = window.document.getElementsByTagName("base"[0]).href;
           }
 
           // It is valid to to begin a URL with "//" so this case has to


level420 (approved):

framework/source/class/qx/util/ResourceManager.js
```diff
@@ -285,9 +285,10 @@ qx.Class.define("qx.util.ResourceManager",
 
           var href;
           //first check if there is base url set
-          var baseElements = qx.bom.Selector.query("base", document);
-          if (baseElements.length > 0) {
-            href = baseElements[0].href;
+          if (typeof window == "object" &&
```
johnspackman: Please can we make this a `qx.core.Environment` check?  The reason is that we already fake global `window` when generating server apps so that the global/default object *is* `window` and therefore the `typeof window == "object"` will be true, but I doubt if this is documented anywhere.

Also, recently I started faking `window.document` in server apps generated by qxcompiler - ironically, one reason was because of this `qx.bom.Selector` that @derrell is fixing, but AFAICR we also have instances where third party code (eg the Bluebirdjs implementation embedded in `qx.Promise`) will expect `document` to exist and have various methods.  

It is a fine line as to how far we would go when mocking, but it would be helpful if the detection code was isolated into a single environment check.

johnspackman (changes requested):

framework/source/class/qx/util/ResourceManager.js
level420: @johnspackman I see your point. We should maybe enhance `qx.bom.client.Runtime` to not just offer `runtime.name` but also `runtime.document`, `runtime.window` or similar, which should return true if window and document are available as global objects. 

Maybe guarded with some specific tests to be able to distinguish them from fake objects.

A possible test could be to call `document.getElementsByTagName('somenonexistingtagname')` which should succeed and return an empty array.

level420: or
```javascript
typeof document === "object" && document instanceof Document
```

johnspackman: For the moment lets not have `runtime.window` because there is lots of code which assumes that there is a `window` object already, but `runtime.document` would be a good idea.  The boot loader for non-browser can always emit some undocumented values (such as `qx.$$loader.isFakeDocument=true` that allow the `runtime.document` implementation to understand what's going on

ronkie (commented):
Hi,

The changes seem fine to me except that I think there is a misprint.

Instead of 

" window.document.getElementsByTagName("base"[0]).href"

it shoud be 

" window.document.getElementsByTagName("base")[0].href"

level420:
@johnspackman regarding the environment checks for availability of `window` or `document`:

What if somebody explicitly wants to fake the `document` object, including `getElementsByTagName`, to be able to also fake `getElementsByTagName('base')` on a non gui platform?

```javascript
          var href;
          //first check if there is base url set
          if (typeof document == "object" && typeof document.getElementsByTagName == "function") {
              var base = document.getElementsByTagName("base");
              if(typeof base == "object" && base.length > 0) {
                  href = base[0].href;
              }
          }
```

johnspackman:
My thought is that the code would be changed to:

```
          var href;
          //first check if there is base url set
          if (qx.core.Environment.get("runtime.document")) {
              var base = document.getElementsByTagName("base");
              if(typeof base == "object" && base.length > 0) {
                  href = base[0].href;
              }
          }
```

`qx.bom.client.Runtime` would have:

```
    /**
     * Detects whether there is a valid global `document` object with a real document in it
     * @internal
     * @return {Boolean}
     */
    getDocument: function() {
      return typeof document == "object" && 
        !qx.$$loader.hasNoDocument;
    }
```

The boot loader for non-browser apps (which is [here](https://github.com/qooxdoo/qooxdoo/blob/master/tool/data/generator/oo.loader.tmpl.js) for generator or slightly different one [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qxcompiler/app/loader-rhino.tmpl.js) for qxcompiler) would be modified to set `qx.$$loader.hasFakeDocument = true`.  

In [qxcompiler loader](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qxcompiler/app/loader-rhino.tmpl.js#L19-L31) I'm already mocking up methods - and it may be necessary to mock up a `getElementsByTagName` or other methods from time to time in the future so IMHO a specific flag like `qx.$$loader.hasFakeDocument` would be ideal (also the only thing which would be faking document would be the bootloader)

level420:
@johnspackman but what if a third party library introduces a fake `document` object?

johnspackman:
@level420 that's a problem for the person who adds that third party library in; if we wanted to use that third party library, we would have to modify it to not do that, or apply some other work around.  Where *we* need to take care of this is where we have already decided to embed a third party library (like BluebirdJS) and it needs a shim to get it working.

derrell:
The problem here is really separation of concerns. Utility classes should really not ever depend on anything in browser-related variables like window and document. We're seeing this problem here because qx.util.ResourceManager, which is not specific to a browser environment, has browser-specific code.

I'd suggest that rather than put in tests for whether browser-specific classes exist, that we should prefer, if possible, to remove the use of browser-specific variables from non-browser-specific qooxdoo classes. Maybe we do that by calling a function which (elsewhere) uses that test? Or maybe we refactor the non-browser-specific classes into browser-specific portions and non-browser-specific portions?

level420:
@derrell @johnspackman But `qx.util.ResourceManager` is already browser-specific. 

See the beginning of https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/util/ResourceManager.js#L263-L292
Note the `if ((qx.core.Environment.get("engine.name") == "mshtml"))`

So it is only needed for Internet Explorer in case of https requests.

In this case I think we could safely assume that `document` is present!

We should shorten the code to:
```javascript
            var href;
            //first check if there is base url set
            if(document.getElementsByTagName("base").length > 0) {
                href = document.getElementsByTagName("base"[0]).href;
            }
```

I don't think there is a risk with this, as it only would fail if someone is tricking us by faking `mshtml` in a non browser environment.

The problem was only introduced by having the dependency to `qx.bom.Selector.query("base", document)` but the code should never be reached if not running on IE.

level420:
also note that @ronkie reported the issue leading to all this specific to IE served via https!

derrell:
I have a library that is common to both browser-based code and server-side code. The problem that I found was that qx.bom.Selector introduced browser-specific code (Sizzle) in its defer function, meaning that it was not possible to load the library without it trying to run browser-specific code.

If ResourceManager is already browser-specific, there's no need to make it not so. It is then up to the application to ensure that it is not called. (It's not called in my server-side code.) We must just ensure that no dependencies cause browser-specific actions at load time, as did qx.bom.Selector.

level420:
@derrell @johnspackman see https://github.com/qooxdoo/qooxdoo/pull/9417

derrell:
Closing in favor of #9417

PR #9421, opened 2017-11-09
Fix tree open-close controller to handle / properly render model changes with 'open' via push()

derrell:
When the model was changed by using the `push()` method after the initial creation of the model, rather than initially creating the entire tree model, open/close didn't work because the controller's changeBubble handler was only looking at changes to the `open` property, but in this case the changes are to the `children` property.

This implementation feels inefficient for large trees, as it's inspecting all changes instead of only changes to a single property. I don't see a better way to do it, however, and this has the major advantage of actually working. :-) Suggestions welcome!


johnspackman (approved):

cboulanger (approved):

cajus (approved):

level420 (approved):

level420:
4 approvals, more than 2 business days, merging
thank you @derrell

PR #9424, opened 2017-11-13
Migrate to array includes

cajus:
ES6 defined *Array.prototype.includes* which was choosen over the *contains* variant. This PR moves to *includes*.

### DIFF
--- a/framework/source/class/qx/data/Array.js
+++ b/framework/source/class/qx/data/Array.js
@@ -647,10 +647,22 @@ qx.Class.define("qx.data.Array",
     /**
      * Check if the given item is in the current array.
      *
+     * @deprecated {6.0} Please use the include method instead
+     *
      * @param item {var} The item which is possibly in the array.
      * @return {Boolean} true, if the array contains the given item.
      */
     contains: function(item) {
+      return this.includes(item);
+    },
+
+    /**
+     * Check if the given item is in the current array.
+     *
+     * @param item {var} The item which is possibly in the array.
+     * @return {Boolean} true, if the array contains the given item.
+     */
+    includes: function(item) {
       return this.__array.indexOf(item) !== -1;
     },
 
--- a/framework/source/class/qx/lang/Array.js
+++ b/framework/source/class/qx/lang/Array.js
@@ -363,18 +363,18 @@ qx.Bootstrap.define("qx.lang.Array",
       }
     },
 
-
     /**
      * Whether the array contains the given element
+     *
+     * @deprecated {6.0} Please use Array instance include method instead
      *
      * @param arr {Array} the array
      * @param obj {var} object to look for
      * @return {Boolean} whether the arr contains the element
      */
     contains : function(arr, obj) {
-      return arr.indexOf(obj) !== -1;
+      return arr.includes(obj);
     },
-
 
     /**
      * Check whether the two arrays have the same content. Checks only the
--- a/application/demobrowser/source/class/demobrowser/DemoBrowser.js
+++ b/application/demobrowser/source/class/demobrowser/DemoBrowser.js
@@ -347,7 +347,7 @@ qx.Class.define("demobrowser.DemoBrowser",
 
       var currentTags = this._tree.getSelection()[0].getUserData("tags");
       if (currentTags) {
-        playable = playable && !qx.lang.Array.contains(currentTags, "noPlayground");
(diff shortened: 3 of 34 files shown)


framework/source/class/qx/lang/Array.js
```diff
@@ -363,19 +363,19 @@ qx.Bootstrap.define("qx.lang.Array",
...
+     * @deprecated {6.0} Please use Array instance include method instead
+     *
      * @param arr {Array} the array
      * @param obj {var} object to look for
      * @return {Boolean} whether the arr contains the element
      */
     contains : function(arr, obj) {
-      return arr.indexOf(obj) !== -1;
+      return arr.includes(obj);
     },
```
johnspackman: should `qx.lang.Array` also have a `includes` method?

framework/source/class/qx/data/Array.js
```diff
@@ -647,10 +647,22 @@ qx.Class.define("qx.data.Array",
     /**
      * Check if the given item is in the current array.
      *
+     * @deprecated {6.0} Please use the include method instead
+     *
      * @param item {var} The item which is possibly in the array.
      * @return {Boolean} true, if the array contains the given item.
      */
     contains: function(item) {
```
johnspackman: this is often done in `defer`, patching instead of additional function call eg:
```
   defer : function(statics, members) {
     members.contains = members.includes;
```

derrell: The problem with doing it in defer is that the `@deprecated` tag can't be added easily, I think

derrell: I was going to suggest the same thing, but if we do plan to deprecate it (which I'm not convinced we should, since it's existed since the dawn of time), we would want to mark it that way.

johnspackman: it can, what's been used elsewhere is to declare the aliasing method (i.e. `contains`) as just null in `members`, so it can have documentation and jsdoc

framework/source/class/qx/lang/Array.js
cajus: Hmm. As it's meant to be in the prototype, I don't think so.

cboulanger (dismissed):

derrell (dismissed):

framework/source/class/qx/data/Array.js
cajus: @johnspackman I've moved that to *defer*. Lets see if the tests work out...

cajus: Doesn't work as expected. I've reverted it. Feel free to change the PR to use *defer*. Although this doesn't save anything, because the defer block needs to be created anyway.

level420:
I'm not sure, but do we have a polyfill for engines not supporting `Array.prototype.includes`? https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/includes says that Internet Explorer does not support `includes`.

level420 (changes requested):
I'm not sure, but do we have a polyfill for engines not supporting Array.prototype.includes? https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/includes says that Internet Explorer does not support includes.

Should go into: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/lang/normalize/Array.js

cajus:
@level420 the polify is installed. Look at the patch.

level420:
@cajus Oh yes! The polyfill is there! Sorry for this. But it should be implemented according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes with n optional second parameter `fromIndex` which is also present in `indexOf`.

cajus:
You're right. But I guess that's not all, because the behaviour is slightly different, too. Will update it...

cajus:
Updated and added tests which respect the different behaviour of `indexOf(NaN)` vs `includes(NaN)`.

cajus:
@level420 anything I can do here to get some approvals? ;-)

johnspackman (approved):

derrell (approved):

level420:
@cajus sorry, I've been busy these days. It's perfect now.

level420 (approved):

level420:
more than 2 business days, 3 approvals, merging.
thank you @cajus

Issue #9422, opened 2017-11-10
Universal IDs for widgets and other objects


johnspackman:
It is useful to be able to identify widgets by an ID, whether for UI testing via a tool like Selenium or for general coding organisation.  An example of Selenium testing (taken from a traditional web app test) is:

```
describe("Contact form", () => {
  test("lead can submit a contact request", async () => {
    await page.waitForSelector("[data-test=contact-form]");
    await page.click("input[name=name]");
    await page.type("input[name=name]", lead.name);
    await page.click("button[type=submit]");
    await page.waitForSelector(".modal");
  }, 16000);
});
```

Another use for IDs is during development as form of centralising widget creation in a form that can be overridden by derived classes - `qx.ui.core.Widget` has `getChildControl` and `_createChildControlImpl` which are specifically targeted at the construction of UI components, and the IDs are used for selecting appearance.  However, this is targeted at the appearance mechanism and is very low level; it is not suitable for application level identification (and attempts to do so will break because of what the appearance mechanism expects to find there).

## Parent/Child or Owner/Objects
There is benefit in having an ID mechanism for objects which is more application-centric; similar to `getChildControl`/`_createChildControlImpl` this would allow construction of objects to be centralised and allow derived classes to override features.  Note that this should support any `qx.core.Object` and not just widgets.

At an application level, a particular form would have various fields (or other objects) that need to be identified, but the structure does not necessarily bear any relation to the widget `layoutParent`/`children` relationship.  For example, consider a UserProfileEditor which has TextField's for `personName` and `emailAddress`, and a SelectBox for `company`; whether you place them all in one Composite with a Grid layout, or use two or three Composites with VBox/HBox layouts has nothing to do with the application-level structuring of the fields you want to identify.

To avoid confusion, we could call this application-level relationship Owner/Objects, where we add `getObject` and `_createObjectImpl` methods to some class in Qooxdoo (maybe `qx.core.Object` or as a mixin?)

## Universal identification
By enforcing basic rules on names (eg alphanumeric plus "-" and "_"), we can assemble ID paths like "contactForm/edtName" and navigate the hierarchy from a given start point.

This would imply that every instance of a widget has an absolute path based identifier, which would be great for debugging or anything machine based.

[...]

PR #9426, opened 2017-11-17
Universal Object ID

johnspackman:
See https://github.com/qooxdoo/qooxdoo/issues/9422


framework/source/class/qx/core/MObjectId.js
```diff
@@ -0,0 +1,231 @@
...
+    },
+
+    /**
+     * Discards an object from the list of owned objects; note that this does
+     * not dispose of the object, simply forgets it if it exists.
+     * 
+     * @param id {String|Object} the ID of the object to discard, or the object itself
+     */
+    discardOwnedObject: function(id) {
+      if (!this.__ownedObjects) {
```
derrell: What do you think about having this stuff done in a private function, e.g., `__discardOwnedObject`, that also has a `force` argument? Then you could have `addOwnedObject` call `__discardOwnedObject` to discard it from a prior owner. The `force` argument, if `true` would ignore `obj.getOwner()` returning null, and allow the `__changingOwner` member to potentially be set to true.

`discardOwnedObject` would call `__discardOwnedObject` with the `foce` value being false.

The goal, if not obvious, is to allow *changing* ownership with a call to `addOwnedObject` just like a widget can be moved from one container to another by simply calling `add()`.

johnspackman: I did start coding something like that, but because of the additional complication of making sure that IDs are unique within owners when changing owner and/or ID, the processes involved started to get quite complex.  

As the pattern here is to create objects on demand, so there has to be a public `discard` method to undo the on-demand `_createObjectImpl`; also, conceptually it seems that owned objects will normally remain in one owner anyway, so reverting to a situation where it has to be manually removed before adding to another owner seemed acceptable.

On reflection though, I think a couple of points:
* rename `discardOwnedObject` to `removeOwnedObject` for better consistency with the add/remove
* `removeOwnedObject` to allow silently ignore IDs that don't exist in case they have just not been created yet
* and Ill make `removeOwnedObject` remove it from it's current owner

framework/source/class/qx/test/core/ObjectId.js
```diff
@@ -0,0 +1,68 @@
...
+
+   Copyright:
+     2007-2008 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+ * Fabian Jakobs (fjakobs)
```
cajus: Really? ;-)

cajus:
It would be nice to get rid of additional code in final builds. Or just make it *dead* using `qx.Environment.select()` when defining the methods.

framework/source/class/qx/test/core/ObjectId.js
johnspackman: oops, that Fabian gets everywhere ;)

johnspackman:
>It would be nice to get rid of additional code in final builds. Or just make it dead using qx.Environment.select() when defining the methods.

Do you mean the DOM attributes?  I've just done that, but can't see any other candidates - the `getObject` API is intended for accessing (and overriding) objects as a general use case, and which adds IDs and a path based selection mechanism as a bonus rather than just for testing

cajus:
@johnspackman I mean that adding stuff in `qx.core.Object` is included in every qx object. Which might make it bigger and the application slower. Although I don't know how much the impact will be in this case. We solved that here by installing a monkey patch early. In the customer relevant built, the monkey patch is not installed.

johnspackman:
In this case the physical impact on applications is quite low - there won't be an increase in memory, unless it's used in which case the map will cause an increase (although presumably not that big, and in any case it is being used for a reason), and there won't be a performance issue either because although the mixin is added to `qx.core.Object` it doesn't add any code except where it is called.

If this was used for just testing then I'd agree it should be removed in production, but IMHO having Universal IDs like this is really very useful when developing applications because it allows the construction of widgets and other objects to be broken out and overridden.

WRT to testing, having a mechanism for IDs baked into the framework will encourage IDs to be used from the outset, which means that when it comes to adding Selenium unit tests the IDs will already be there.  It's going to hurt to have to add IDs retrospectively so there is an incentive to promote IDs as part of day-to-day coding.

I've just added a comment https://github.com/qooxdoo/qooxdoo/issues/9422#issuecomment-346301649 which tries to explain further what I mean about how useful IDs can be in the general case; this is based on my own mixin for IDs (I just checked and found that I have 655 classes which use this pattern already!)

cboulanger (dismissed):

level420 (approved):

level420 (changes requested):
Just noticed that lint is throwing some errors:
```
Running lint...

  - Warning: qx.core.MObjectId (166, 9): Loop or condition statement without a block as body

  - Warning: qx.core.MObjectId (168, 9): Loop or condition statement without a block as body

Lint warnings are treated as errors! Please fix!
```

level420:
@johnspackman could you please have a look into my change request? thank you.

johnspackman:
@level420 yes - I've kind of put this on hold temporarily because I need to do some tests using the solution that @cajus suggested, because if it works then this PR would be significantly reduced.

johnspackman:
closing this PR in favour of a new one (coming shortly) due to problems rebasing :(

PR #9428, opened 2017-11-21
Added assertEqualsFloat to qx.core.Assert and qx.core.MAssert

mbgonicus:
One thing I really dislike about computers is that their memory and calculation precision is naturally finite. Of course I dropped that brick when I recently wanted to assert that two floats are equal. Couldn't find a function in the qooxdoo framework that allows an assertion of the equality of two floats respecting a precision, thus I wrote it on my own. Hope you like it.

### DIFF
--- a/framework/source/class/qx/core/Assert.js
+++ b/framework/source/class/qx/core/Assert.js
@@ -199,9 +199,8 @@ qx.Bootstrap.define("qx.core.Assert",
     },
 
     /**
-     * Assert that both float values are equal within the given tolerance. This
-     * might be needed because of the natural floating point inaccuracy of
-     * computers.
+     * Assert that both float values are equal. This might be needed because
+     * of the natural floating point inaccuracy of computers.
      *
      * @param expected {Float} Reference value
      * @param found {Float} Found value
@@ -215,6 +214,25 @@ qx.Bootstrap.define("qx.core.Assert",
       qx.lang.Number.equals(expected, found) ||
         this.__fail(msg || "", "Expected '", expected, "' to be equal with '",
         found, "'!"
+      );
+    },
+
+    /**
+     * Assert that both float values are not equal. This might be needed
+     * because of the natural floating point inaccuracy of computers.
+     *
+     * @param expected {Float} Reference value
+     * @param found {Float} Found value
+     * @param msg {String} Message to be shown if the assertion fails.
+     */
+    assertNotEqualsFloat : function(expected, found, msg)
+    {
+      this.assertNumber(expected);
+      this.assertNumber(found);
+
+      !qx.lang.Number.equals(expected, found) ||
+        this.__fail(msg || "", "Expected '", expected,
+        "' to be not equal with '", found, "'!"
       );
     },
 
--- a/framework/source/class/qx/core/Assert.js
+++ b/framework/source/class/qx/core/Assert.js
@@ -199,6 +199,33 @@ qx.Bootstrap.define("qx.core.Assert",
     },
 
     /**
+     * Assert that both float values are equal within the given tolerance. This
+     * might be needed because of the natural floating point inaccuracy of
+     * computers.
+     *
+     * @param expected {Float} Reference value
+     * @param found {Float} Found value
+     * @param precision {Float ? 0.01} Value by which expected and found may
+     *   differ
+     * @param msg {String} Message to be shown if the assertion fails.
+     */
+    assertEqualsFloat : function(expected, found, precision, msg)
+    {
+      this.assertNumber(expected);
+      this.assertNumber(found);
+      if (!qx.lang.Type.isNumber(precision)) {
+        precision = 0.01;
+      }
+
+      Math.abs(expected - found) < precision || this.__fail(
+        msg || "",
+        "Expected '", expected,
+        "' to be equal with '", found, "' within the precision of '",
+        precision, "'!"
+      );
+    },
+
+    /**
      * Assert that both values are identical. (Uses the identity operator
      * <code>===</code>.)
      *
--- a/framework/source/class/qx/core/Assert.js
+++ b/framework/source/class/qx/core/Assert.js
@@ -205,23 +205,20 @@ qx.Bootstrap.define("qx.core.Assert",
      *
      * @param expected {Float} Reference value
      * @param found {Float} Found value
-     * @param precision {Float ? 0.01} Value by which expected and found may
-     *   differ
      * @param msg {String} Message to be shown if the assertion fails.
      */
-    assertEqualsFloat : function(expected, found, precision, msg)
+    assertEqualsFloat : function(expected, found, msg)
     {
       this.assertNumber(expected);
       this.assertNumber(found);
-      if (!qx.lang.Type.isNumber(precision)) {
-        precision = 0.01;
-      }
 
-      Math.abs(expected - found) < precision || this.__fail(
-        msg || "",
-        "Expected '", expected,
-        "' to be equal with '", found, "' within the precision of '",
-        precision, "'!"
+      var diff = Math.abs(expected - found);
+
+      // The maximum of both numbers multiplied by 1e-14 is the relative difference
+      diff < Number.EPSILON ||
+        diff <= Math.max(Math.abs(expected), Math.abs(found)) * 1e-14 ||
+        this.__fail(msg || "", "Expected '", expected, "' to be equal with '",
+        found, "' regarding the precision inefficiencies of floats!"
       );
     },
 
--- a/framework/source/class/qx/core/Assert.js
+++ b/framework/source/class/qx/core/Assert.js
@@ -212,13 +212,9 @@ qx.Bootstrap.define("qx.core.Assert",
       this.assertNumber(expected);
       this.assertNumber(found);
 
-      var diff = Math.abs(expected - found);
-
-      // The maximum of both numbers multiplied by 1e-14 is the relative difference
-      diff < Number.EPSILON ||
-        diff <= Math.max(Math.abs(expected), Math.abs(found)) * 1e-14 ||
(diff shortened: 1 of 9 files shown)


cboulanger (dismissed):

level420 (dismissed):

derrell (dismissed):

cajus (dismissed):

johnspackman (dismissed):

framework/source/class/qx/core/Assert.js
```diff
@@ -198,6 +198,33 @@ qx.Bootstrap.define("qx.core.Assert",
...
+     * @param precision {Float ? 0.01} Value by which expected and found may
+     *   differ
+     * @param msg {String} Message to be shown if the assertion fails.
+     */
+    assertEqualsFloat : function(expected, found, precision, msg)
+    {
+      this.assertNumber(expected);
+      this.assertNumber(found);
+      if (!qx.lang.Type.isNumber(precision)) {
+        precision = 0.01;
```
oetiker: this will not work well for small or very large numbers ... there are established algorithms for this kind of problem ...I can look them up for you if you wish

oetiker (changes requested):

framework/source/class/qx/core/Assert.js
oetiker: http://mathjs.org/docs/datatypes/numbers.html

Equality#

Because of rounding errors in calculations, it is unsafe to compare JavaScript Numbers. For example executing 0.1 + 0.2 == 0.3 in JavaScript will return false, as the addition 0.1 + 0.2 introduces a round-off error and does not return exactly 0.3.

To solve this problem, the relational functions of math.js check whether the relative difference between the compared values is smaller than the configured option epsilon. In pseudo code (without exceptions for 0, Infinity and NaN):

diff = abs(x - y)
nearlyEqual = (diff <= max(abs(x), abs(y)) * EPSILON) OR (diff < DBL_EPSILON)
where:

EPSILON is the relative difference between x and y. Epsilon is configurable and is 1e-14 by default. See Configuration.
DBL_EPSILON is the minimum positive floating point number such that 1.0 + DBL_EPSILON != 1.0. This is a constant with a value of approximately 2.2204460492503130808472633361816e-16;
Note that the relational functions cannot be used to compare small values (< 2.22e-16). These values are all considered equal to zero.

oetiker: http://mathjs.org/docs/datatypes/numbers.html

Equality#

Because of rounding errors in calculations, it is unsafe to compare JavaScript Numbers. For example executing 0.1 + 0.2 == 0.3 in JavaScript will return false, as the addition 0.1 + 0.2 introduces a round-off error and does not return exactly 0.3.

To solve this problem, the relational functions of math.js check whether the relative difference between the compared values is smaller than the configured option epsilon. In pseudo code (without exceptions for 0, Infinity and NaN):

diff = abs(x - y)
nearlyEqual = (diff <= max(abs(x), abs(y)) * EPSILON) OR (diff < DBL_EPSILON)
where:

EPSILON is the relative difference between x and y. Epsilon is configurable and is 1e-14 by default. See Configuration.
DBL_EPSILON is the minimum positive floating point number such that 1.0 + DBL_EPSILON != 1.0. This is a constant with a value of approximately 2.2204460492503130808472633361816e-16;
Note that the relational functions cannot be used to compare small values (< 2.22e-16). These values are all considered equal to zero.

oetiker: there is even Number.EPSILON defined in Many js implementations

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON

mbgonicus:
@oetiker Of course you are right. In a framework this should be done properly. 

I committed a couple of changes:
* introduced qx.lang.Number.equals using the above stated algorithm
* using qx.lang.Number.equals inside qx.core.Assert.assertEqualsFloat
* added qx.core.Assert.assertNotEqualsFloat

framework/source/class/qx/lang/Number.js
```diff
@@ -74,6 +74,27 @@ qx.Class.define("qx.lang.Number",
...
+    equals : function(x, y)
+    {
+      if (qx.core.Environment.get("qx.debug")) {
+        qx.core.Assert.assertNumber(x);
+        qx.core.Assert.assertNumber(y);
+      }
+
+      // 1e-14 is the relative difference
+      return x === y ||
+             Math.abs(x - y) < Number.EPSILON ||
```
oetiker: not all browsers seem to support EPSILON, using 2^-52 as a default is suggested
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON

oetiker (changes requested):

framework/source/class/qx/lang/Number.js
```diff
@@ -91,9 +91,12 @@ qx.Class.define("qx.lang.Number",
...
       }
 
-      // 1e-14 is the relative difference
+      // 2e-52 is the difference between 1 and the smallest floating point
+      // number greater than 1, like Number.EPSILON, but that is not supported
+      // by all browsers.
+      // 1e-14 is the relative difference.
       return x === y ||
-             Math.abs(x - y) < Number.EPSILON ||
+             Math.abs(x - y) < 2e-52 ||
```
oetiker: aah :) sorry ... what I meant is to use 2e-52 IF EPSILON was not defined

oetiker (changes requested):

framework/source/class/qx/lang/Number.js
johnspackman: Could we implement Number.EPSILON as a shim then, in a new `qx.lang.normalize.Number`?

mbgonicus: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON suggests a simple polyfill:

```javascript
if (Number.EPSILON === undefined){
    Number.EPSILON  =  Math.pow(2, -52);
```

But if Number.EPSILON is 2e-52, then why can't we just use the literal constant?

oetiker: since most browsers support it ... so there is no point in ignoring it ...

level420 (dismissed):

framework/source/class/qx/lang/Number.js
```diff
@@ -74,6 +74,32 @@ qx.Class.define("qx.lang.Number",
...
+
+
+    /**
+     * Checks the equality of two numbers regarding the imprecision of floats.
+     *
+     * @param x {Number}
+     * @param y {Number}
+     * @return {Boolean}
+     */
+    equals : function(x, y)
```
oetiker: Sorry to be a pain ... but I think such a call should be called `almostEqual`

at least via 
https://www.google.ch/search?num=100&newwindow=1&dcr=0&q=almost+equal
you find  many other attempts to solve that problem

framework/source/class/qx/test/core/Assert.js
```diff
@@ -153,6 +153,49 @@ qx.Class.define("qx.test.core.Assert",
           this.fireEvent("xyz1");
         });
       }, qx.core.AssertionError);
+    },
+
+    testAssertEqualsFloat : function()
```
oetiker: same problem

framework/source/class/qx/test/lang/Number.js
```diff
@@ -0,0 +1,39 @@
...
+
+************************************************************************ */
+
+qx.Class.define("qx.test.lang.Number",
+{
+  extend : qx.dev.unit.TestCase,
+
+  members :
+  {
+    testEquals : function() {
```
oetiker: same problem

framework/source/class/qx/core/MAssert.js
```diff
@@ -99,6 +99,33 @@ qx.Mixin.define("qx.core.MAssert",
...
+
+    /**
+     * Assert that both float values are equal. This might be needed because
+     * of the natural floating point inaccuracy of computers.
+     *
+     * @param expected {Float} Reference value
+     * @param found {Float} Found value
+     * @param msg {String} Message to be shown if the assertion fails.
+     */
+    assertEqualsFloat : function(expected, found, msg) {
```
oetiker: see further down

framework/source/class/qx/core/Assert.js
```diff
@@ -198,6 +198,44 @@ qx.Bootstrap.define("qx.core.Assert",
...
 
+    /**
+     * Assert that both float values are equal. This might be needed because
+     * of the natural floating point inaccuracy of computers.
+     *
+     * @param expected {Float} Reference value
+     * @param found {Float} Found value
+     * @param msg {String} Message to be shown if the assertion fails.
+     */
+    assertEqualsFloat : function(expected, found, msg)
```
oetiker: see below

oetiker (changes requested):

mbgonicus:
I added a polyfill for Number.EPSILON. Thanks to @cajus !

framework/source/class/qx/lang/Number.js
mbgonicus: I disagree. This is actually to check if two numbers are equal. It just tries to bypass the inefficiencies of a computer to precisely represent float numbers.

It becomes obvious for the example that 0.1 + 0.2 = 0.3. Both sides of that equation are equal, not "almost" equal.

btw: I'd really prefer if you put your criticism here at once and not in dribs and drabs.

cajus (approved):

cboulanger (approved):

oetiker (approved):

level420:
More than 2 business days, 3 approvals, merging
thank you @mbgonicus

PR #9431, opened 2017-11-29
This disables the unit test that causes Travis to break on Headless Chrome

johnspackman:
This is an ugly solution but it seems fruitless to constantly get notification of failed builds because of this one test on one browser.  The problem is isolated to one test case only and it is in a section of code that is very stable

### DIFF
--- a/framework/source/class/qx/test/ui/embed/Iframe.js
+++ b/framework/source/class/qx/test/ui/embed/Iframe.js
@@ -90,6 +90,12 @@ qx.Class.define("qx.test.ui.embed.Iframe",
 
     testSyncSourceAfterDOMMove : function ()
     {
+      // This breaks (very) frequently when run under headless chrome on Travis; we can't
+      //  track down the cause and it works just fine elsewhere.  Disabling this test on 
+      //  Chrome is an effective quick hack until someone can figure it out.
+      if (qx.core.Environment.get("browser.name") == "chrome")
+        return;
+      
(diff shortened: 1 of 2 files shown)


cajus:
I'm already working on a PR for that. It also moves the tests to *saucelabs*, so that we can test against more browsers on more platforms. Currently fighting against travis to make it play nicely.

cboulanger (dismissed):

derrell (dismissed):
This LGTM, but it sounds like @cajus has a more complete fix to use instead?

johnspackman:
I'm more than happy to withdraw this, how far away do you think you are from a full solution @cajus ?  I just had a failure report for editing `.github/CODEOWNERS`, would it be worth committing this in the short term and reverting it with the switch to saucelabs?

cajus:
@johnspackman I hope to get it running soon. But I'm happy with applying this one as a quicky, too.

hkollmann (changes requested):
I would prefer the Change cajus suggested

hkollmann (approved):

cajus (approved):

derrell (approved):

level420 (approved):

cboulanger (approved):

hkollmann:
5 approvals, merged

level420:
@hkollmann thank you for merging. In this case please also close the PR via "Close and comment". Thank you.

oetiker (approved):

Issue #9430, opened 2017-11-27
Min/Max behaviour broken


cajus:
https://github.com/qooxdoo/qooxdoo/pull/9377 introduces a problem where you can't simply set min/max values to the same. We have to do that in various places and from the naming point of view, nothing tells me that *minHeight === maxHeight* isn't allowed.

```
this.set({
  minHeight: value,
  maxHeight: value,
  minWidth: value,
  maxWidth: value
});
```

Depending on the values that are inside the variables already, it fails. You've to inspect the values first to get out of this. That means that you can't set the value (i.e. in the appearance) for sure. It always depends on what's in there in the time of setting. Relatively impractical.

I didn't understand what PR was for - or why it was needed, so I did not vote for it. So I don't know why the assertion was put there. Its a too early place IMHO. Maybe it should be moved back to the *Widget*, because it allows the evaluation on render time. Or it should just be a warning and no assertion at all.

level420:
@cajus as far as I can see in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/core/LayoutItem.js#L751-L763
both asserts are about `minWidth <= maxWidth` and `minHeight <= maxHeight` where a test against lower or equal is done.
I think this is correct.

cajus:
@level420 I didn't say it's wrong. Its just too agressive. Think about a theme A which has a widget with *minWidth* set to *40* and *maxWidth* set to *60*. Now you switch the theme which has different settings. Because *_applyDimensions* is now the place where the check happens, it is a hard constraint between these. It was a soft constraint before, because it was checked at render time.

As a result, you now can by choosing unlucky combinations not set min/max values after each other in the "wrong" order. You'll directly run into the assertion.

level420:
@mever could you please have a look into this issue. I'm in favor of reverting that part of the PR which belongs to the min vs. max checks, moving them form `qx.ui.core.LayoutItem` back to `qx.ui.core.Widget`. Would this change anything regarding the fix for `qx.ui.form.core.VirtualDropDownList`?

mever:
That's unfortunate. I think that the core of the issue is to set qx-properties in order. We can not guarantee that setting qx-properties in a random order will not break consistency. Therefore I fear reverting will not address the real issue.

To clarify, the `set` method on `qx.core.Object` does not respect js-property order (the map argument, which is a js object, is by definition not ordered). The alternative being not able to keep consistent qx-objects while using the `set` method.

[...]

PR #9432, opened 2017-11-30
New feature: transaction mechanism

mever:
Dear Qooxdoo members,

For issue #9430 I am changing the `set` implementation to solve a fundamental problem when setting properties in bulk. The PR I am going to release will probably solve an issue raised by PR #9377.

It occurred to me that Qooxdoo's property system can be very useful to hold integrity constraints between objects. It at least provides facilities for supporting this. Usually property value changes are hard-coded changes in order, in which the programmer is aware of state and the changes to this state.

Features like setting properties in bulk currently are not transactional. Therefore constraint checks can not involve constraints between objects or properties.

For example, when a component has an initial minHeight of 40. Then a programmer, trying to set the component to a min- and maxHeight of 20, will first lower minHeight to 20 and then maxHeight. Doing this in the wrong order may cause a constraint violation (the problem with min- and max height is illustrative, reverting PR #9377 will solve this particular problem at once). Thus, applying properties depend on the initial state (before the bulk operation) of the object(s) in question.

The reason why I created this PR is because I think PR #9377 is part of a bigger problem. In a sense, before PR #9377 the commit was implemented in the render layer of the UI system (not very intuitive, also see the fail fast principle).

This PR allows us to use the property system to keep our object model in a consistent state while still applying changing in bulk without knowing the initial state. Also it would be a nice addition to Qooxdoo :)

This PR isn't ready yet. I made it to receive feedback early on. So let's hear it!

### DIFF
--- a/framework/source/class/qx/core/Transaction.js
+++ b/framework/source/class/qx/core/Transaction.js
@@ -1,0 +1,55 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2017 Martijn Evers, The Netherlands
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Martijn Evers (mever)
+
+************************************************************************ */
+
+/**
+ * Within a transaction, property changes can be applied atomically.
+ */
+qx.Class.define("qx.core.Transaction", {
+  extend : qx.core.Object,
+  members : {
+
+    /**
+     * Sets the property value of given subject.
+     *
+     * Note that any exceptions caused by property constrains are caught in the commit phase
+     * and returned as error object(s).
+     *
+     * @param subject {qx.core.Object} Object to change withing the transaction.
+     * @param property {string} Name of the property to set.
+     * @param value {var} Any value accepted by the property.
+     */
+    setPropertyValue : function(subject, property, value) {
+    },
+
+    /**
+     * Commits the transaction.
+     *
+     * When returned array of errors is not empty the transaction is rolled back.
+     *
+     * @return {Error[]} Empty array on success; otherwise an array of error are returned.
+     */
+    commit : function() {
+    },
+
+    /**
+     * Rollback the transaction.
+     */
+    rollback : function () {
+    }
+  }
+});


mever:
Some background about `qx.core.Transaction`: an instance of this class is instantiated by a module in the `qx.core.Object` like so:
```

  include : qx.core.Environment.filter({
    "module.databinding" : qx.data.MBinding,
    "module.logger" : qx.core.MLogging,
    "module.events" : qx.core.MEvent,
    "module.property" : qx.core.MProperty,
    "module.transactions": qx.core.MTransaction, // <-- not in this PR yet!
    "qx.debug" : qx.core.MAssert
  }),
```

A transaction is started like this:
```
var o = new qx.core.Object();
var tx = o.newTransaction();
(tx instanceof qx.core.Transaction) // <-- true
```

The `set` method of the `qx.core.MProperty` mixin will use this transparently under the hood.

cboulanger:
Sounds plausible! Thank you @mever for thinking in the "bigger picture"

johnspackman (changes requested):
I'd have to see how this works in practice before being able to comment on it, but I think there will be major issues that will be difficult to overcome: 

(1) the property mechanism is pretty hard work to get to grips with already because it is all generated code, and as properties have increased in sophistication this has become really hard to debug problems.  The solution there is to rewrite property set/get/etc to be traditional functions but that is some way off.  I suspect that implementing transactions would depend on making substantial changes to properties.

(2) while I understand that there is an implied order to properties which is not immediately obvious (and often not desirable), this behaviour is baked into the code and changing it so that properties are applied simultaneous will break backwards compatibility.  

(3) implementing `rollback()` is likely to be very tricky indeed, not least because you will require the cooperation of user-supplied apply and transform methods; I suspect that at best this would have very limited use because apply/transform can trigger other property set/get recursively.  

It is impossible to assess this without a detailed proposal, and at the moment I can't even imagine how it *might* work.

mever:
@johnspackman 

(1) If the generated methods for properties are generated in a deterministic manner this shouldn't be a problem. But I agree that the property mechanism is a bit hard to grasp. This is one of the main disadvantages of Qooxdoo (which I usually try to justify to my colleagues by pointing out the many great features of Qooxdoo which come with the package). If you do not mind, I may be able to help you there by bit of refactoring.

(2) Properties are already applied simultaneous with the `set` method (asking users to know if properties have an implied order before bulk `set`'ing them is asking too much of your users. Properties aren't methods. IMHO if they have a place between attributes (public non-function members) and methods it is to relate values in a high-level manner). Alternatively, we could also document that properties are not fit for keeping constraints between objects or properties (which may complicate things a bit for users who are not familiar with the limits of the system).

(3) I was planning to build a execution plan in the `setPropertyValue` method and execute it in the `commit` method, something like this (approximately) in the commit method:
1. Take values from plan.
2. Store values in the property storage layer ($$user_XXX).
3. Run validation layer (checks, transform, etc...) on values.
4. When validation fails no changes happened, validation errors are returned.
5. When validation passes applies are triggered.
6. When applies fail, taken values are restored (which were already successfully applied) and errors are returned.
7. When applies succeed events are fired and no errors are returned.

Rollback will be implemented by disposing the execution plan.

And about backwards compatibility (BC), breaking BC shouldn't be a taboo. Major version numbers are here for a reason. Not willing to break BC results in either having security problems or unpleasant software. Because the world changes and people learn they are insurmountable. That being said, BC is very important. If I will break it, I will announce it.

mever:
Given algorithm in point three is a just to paint a picture (!). Please do not try to find problems with it as it would be wasted effort. I know it is flawed, the real deal will be a commit on this branch for review.

derrell:
I have a similar issue in my current app, but for a different reason. I use the new value of properties as part of the `apply` function in other properties. At present, I implement a `data` property (so I call `setData(<map containing property name/value>) whose `apply` function looks at all of the properties being set, and individually sets the various other properties in the correct order. This is required, primarily, because a JavaScript map is unordered. Were it ordered, I would simply provide the properties to be set in the requisite order.

An alternative PR to yours, which might be more simple to implement, is to add provide an alternative `set` function that can take an array as well as a map. If an array is passed, it would be expected to contain a list of maps, with one or more property name/value pairs in each map. In this way, the properties that must be set "first" would be in the array[0] map; properties that must follow those but possibly precede some others would be in the array[1] map; additional (possibly the final set of) properties in the array[2] map, etc.

I offer this as an alternative to the transaction mechanism which I think would be a nice concept, but side effects in `apply` and `transform` functions are likely to be hard to overcome with `rollback`.

mever:
@derrell thanks for the input (you too @johnspackman ). I though of this but rejected the idea because I use `set` a lot and do not want to know the order (the main reason for using `set`).

For example, consider this function:
```
function x(y) {
  y.set([minHeight, 20, maxHeight, 20]);  // <-- keeping order
}
```
Calling x will succeed when y is in this state: {minHeight: 40, maxHeight: null}
but would still fail when it is configured like this: {minHeight: null, maxHeight: 40}

Obviously you can rewrite the function like this:
```
function x(y) {
  // order here is not necessary
  y.set({minHeight: null, maxHeight: null});
  y.set({minHeight: 40, maxHeight: 40});
}
```
But then we know min- and maxHeight share a constraint. Not all properties are that clear. And when the second line can not be applied y becomes corrupted. Some work needs to be done when modifying more than one property atomically. Obviously this can be done specific to y. But making this a generic feature would certainly make an awesome addition!

I know the following point is highly debatable, but apply and transform should generally be reversible (by providing the old value). IMHO properties are no alternatives to methods. I also think we can not force people to do anything, we can not do anything to prevent a programmer from wreaking havoc on his application state. Instead we should clarify how things work. IMHO properties allow us the specify a relation between an object and a type of value in a declarative way (data-binding causing these declarations to become alive).

It should be documented that `set` is transactional and side effects caused by validation and transformation are not rolled back (like it is of now).

derrell:
I agree with your assumptions. Transform and validation functions should really not have side effects. I'm eager to see your implementation of transactional property setters. I fear the backward-compatability issues will be extreme, but let's see what you come up with and then discuss. Before you go to a huge amount of work here, I suspect that this will be a difficult change to get merged solely because it is so difficult to figure out what the consequences are of changes to that automatically-generated code. But I'm certainly willing to entertain this potential change!

johnspackman:
A refactoring/rewrite of the property implementation would be great 👍 it's overdue and I'd be very interested in collaborating with you on that if you wanted to lead it.

But I think you misunderstand how `.set({...})` works - it does not apply all properties simultaneously, it applies them one at a time.  

I agree that properties do initially appear as discrete data items, but the fact is that there is code attached to them in apply/transform - and this is hugely useful to an application.  Applying a property can easily kick off a cascade of events and other property changes on 'this' object as well as other objects.  You are absolutely right that this breaks down where there is an order required, but that order can be documented in the user's class and then adhered to.

In databases it's possible to have transaction isolation because the database is in total control of what's happening, all changes can be rolled back by just discarding the virtual transactions, but as soon as we call apply it's out of our control.  Apply methods can trigger UI/DOM changes, state changes, server interactions, etc.  Events are tied to apply methods, if a change event is fired, we can know that the apply has been called, so if we can rollback an apply it has to be reflected in the events.  We could re-apply the previous value, but that is not necessarily valid either - how would we handle a failure when re-applying the previous value for example?  How would the code know that this change does not reflect the user editing something that needs to be saved to disk, as opposed to just "undoing" something?

However, these are all problems stemming from the concept of being able to commit and rollback - I do like the idea of having `set()` look across all of the validation methods of all of the properties to be set and check that they all pass before trying to set any of them.  It seems entirely reasonable to assume that a single call to `set()` should either all complete or none complete.  Having a better validation infrastructure would be very cool too.

Leading on from that, I have the same issue that @derrell mentioned in that sometimes several properties need to be set together - it could be useful to be able to group several properties together so that there is an apply method (and maybe event) which is called with the property values, old and new.  When a property is changed via `set` it would be possible to change several at once, but the apply method would always receive *all* property values in the group.

Maybe something like:

```
properties: {
  alpha: {},
  bravo: {},
  charlie: {}
},

propertyGroups: {
  myGroup: {
    apply: "_applyMyGroup",
    event: "changeMyGroup",
    properties: [ "alpha", "bravo" ]
},

members: {
  _applyMyGroup: function(values, oldValues) {
    if (values.alpha == "abc" && oldValues.bravo == "def") 
      /*...*/ ;
  }
```

The effect would be something like @derrell 's `setData` and would also provide ordering for setting the properties

derrell:
@johnspackman 's suggestion of the group apply function would solve the problem I solved with my `data` property, and make it more clear which properties are inter-related. In my case, I well-document in the `data` property's `apply` function exactly what order things need to be done in, but the group property apply would be cleaner still. I like that.

mever:
@johnspackman thanks for the vote of confidence.

I know database transactions have the same problems though. Irreversible side effects are unavoidable when a database supports a scripting language (i.e. using stored procedures). The reason why nobody is bothered by it is because the data model stays in a consistent state. Sending a http request in an apply is more or less the same as doing this in a DB stored procedure. Both can't be rolled back.

The applications we write in Qooxdoo have a long life span (in terms of runtime) and the most serious problem for our users is an inconsistent state (because they need to restart the application). A transaction feature would be really helpful (because we can solve it generically instead of making specific functionality). I know a rollback can have serious consequences when introduced as default `set` feature.  So maybe this isn't a good idea.

I know some applications we build take the reversible property on some of our classes' properties very seriously. Even depend on it. For these systems this would be perfect. But I do not want to force this on `set` and cause problems for others.

Just validation as you described without commit and rollback still need a bit of transactional data for a dimensions check, but may be the answer, take a look at this function:
```
assertDimensions : function() {
  var minWidth = this.getMinWidth();
  var maxWidth = this.getMaxWidth();
  if (minWidth !== null && maxWidth !== null) {
    this.assert(minWidth <= maxWidth, "minWidth is larger than maxWidth!");
  }

  var minHeight = this.getMinHeight();
  var maxHeight = this.getMaxHeight();
  if (minHeight !== null && maxHeight !== null) {
    this.assert(minHeight <= maxHeight, "minHeight is larger than maxHeight!");
  }
}
```
When setting `maxWidth`, `getMinWidth` needs the transaction's version of the data for the check. But maybe it is in the wrong place. It just occurred to me that this is implemented as an `apply`. Whereas this is actually a consistency check and therefore not an `apply` but a `check` or `validation` function.

What if we move this to the `check` or `validate` instead (where it actually should be). And perform the checks in a transaction, something like this:

1. `set` is called.
2. Getters are modified to return values from the `set`'s arguments.
3. Checks and validations are run.
4. Getters are restored.
5. Check and validation errors are handled, returned or thrown.
6. If all checks and validation pass `set` will continue as before, just setting all properties.

Some points, though:
- Checks and validation functions should be simple and only do type checking (check) and validation (e.g. integrity constrains).
-  The property mechanism may be changed a bit to allow validation (checks and validation) to happen once, not twice (once in point 3 and the second time at point 6).
-  When calling setters in the check or validation function we can instruct the property mechanism to throw an exception (or log an error) that asks the user to use `apply` instead.

johnspackman:
WRT transactions, provided that this can be implemented without impact on existing code that would be fine - perhaps some kind of pluggable module that someone could elect to install in their app?  My solution has been to take a design approach which is more like iPhone UI - everything is edited pretty much live, there are no Apply and Undo buttons on forms for example - but I can see that won't work for everybody :)

But for your assertDimensions example with minWidth/maxWidth, I get the part that the validation of minWidth and maxWidth overlap, and therefore the validation needs to happen at the end of all properties being set.  I think that would be solved with the `propertyGroups` example above, eg like this:

```
properties: {
  minWidth: {
    check: "Integer"
  },
  maxWidth: {
    check: "Integer"
  },
  minHeight: {
    check: "Integer"
  },
  maxHeight: {
    check: "Integer"
  }
},

propertyGroups: {
  dimensions: {
    apply: "_applyDimensions",
    validate: "_assertDimensions",
    event: "changeDimensions",
    properties: [ "minWidth", "maxWidth", "minHeight", "maxHeight" ]
  }
},

members: {
  _assertDimensions: function(values, oldValues) {
    if (values.minWidth !== null && values.maxWidth !== null) {
      this.assert(values.minWidth <= values.maxWidth, "minWidth is larger than maxWidth!");
    } 

    if (values.minHeight !== null && values.maxHeight !== null) {
      this.assert(values.minHeight <= values.maxHeight, "minHeight is larger than maxHeight!");
    }
  },
  _applyDimensions: function(values, oldValues) {
    // Action something
  }
}
```

In the current Qooxdoo, it would be necessary to use a common apply method for all 4 properties, which means that for 3 out of 4 times that function will seeing an inconsistent/incomplete set of property values; however, with the property groups idea the apply is being called only once.  This has the same net effect as transactions, because it is about moving the checks to happen at the right time

> 2. Getters are modified to return values from set's arguments.

I'm not sure how I feel about this - I probably need to think about it some more, but it does make me wonder about the possible side effects and what the issues would be in having an inconsistent view of the object.  For example, if you have an apply method for a given property which sets other property or member variables, you would write code that assumes that if the getter returns a value, those other properties or member variables are all consistent with whatever that getter returns.  But as the apply method has not been called, that consistency is not there.  Having a map of "new" values makes it clear where the data comes from and that it is not (yet) permanent.

Also, property apply methods can be asynchronous (they can return a promise) and if we carried that into validation (which seems reasonable and would have the same use case as asynchronous apply methods), then those property values would appear to the rest of the application on a temporary basis, not just inside the validate function.

> When calling setters in the check or validation function we can instruct the property mechanism to throw an exception (or log an error) that asks the user to use apply instead.

Agreed, recursion is not desirable here and an exception would be appropriate.  If the validation functions want to change the values which are being applied, they can modify the map of "new" values.

mever:
If validation can be done asynchronously (which sounds reasonable to me) then changing getters does not seem to be an option.

I think:
```
propertyGroups: {
  dimensions: {
    apply: "_applyDimensions",
    validate: "_assertDimensions",
    event: "changeDimensions",
    properties: [ "minWidth", "maxWidth", "minHeight", "maxHeight" ]
  }
},
``` 
may be easily confused with the current [property groups](http://www.qooxdoo.org/5.1/pages/core/defining_properties.html#defining-property-groups):
```
properties: {
  dimensions: {
    apply: "_applyDimensions",
    validate: "_assertDimensions",
    event: "changeDimensions",
    group: [ "minWidth", "maxWidth", "minHeight", "maxHeight" ]
  }
},
```

If we can reuse the current property group implementation we could do the following: just before the `set` method is done it sets each property with a group list that contains one or more of the properties being set. If the group properties needs values not provided as arguments the values are taken from the object itself.

We probably want to implement `check`, `validate`, `apply` and `event` for these kind of properties as well.

johnspackman:
This is quite old now and there isn't an implementation that we can add into the core, and this PR feels more like an issue for discussion than a PR; @mever please feel free to reopen as an issue

PR #9433, opened 2017-11-30
Fixed crash in VirtualDropDownList#__getAvailableHeight when dropdown has no layout location

milandamen:
When having set a model on a VirtualSelectBox that has not been rendered, and changing the length of the model array in any way, an exception gets thrown in `VirtualDropDownList#__getAvailableHeight`.

Reason:

`qx.ui.list.List` adds a `changeLength` listener to the model, when this gets fired, `qx.ui.list.List#_onModelChange` gets called which fires the `changeModelLength` event. The VirtualDropDownList listens to this event and calls `qx.ui.form.core.VirtualDropDownList#__adjustHeight` when this happens. This method in turn calls `__getAvailableHeight` which cannot deal with a `null` value for `distance`. `distance` will be null if the widget has no bounds.

I've also written a test for this situation, and I've fixed the docblocks on the relevant methods.

### DIFF
--- a/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
+++ b/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
@@ -440,6 +440,10 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
     __adjustHeight : function()
     {
       var availableHeight = this.__getAvailableHeight();
+      if (availableHeight === null) {
+        return;
+      }
+
       var maxHeight = this._target.getMaxListHeight();
       var list = this.getChildControl("list");
       var itemsHeight = list.getPane().getRowConfig().getTotalSize();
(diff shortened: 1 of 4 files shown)


framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -469,6 +469,10 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
     __getAvailableHeight : function()
     {
       var distance = this.getLayoutLocation(this._target);
+      if (!distance) {
```
johnspackman: Returning zero from here means that `__adjustHeight` ([see here](https://github.com/milandamen/qooxdoo/blob/d0f02bdb691e0d622f1ba8a5303ce8da8e43321b/framework/source/class/qx/ui/form/core/VirtualDropDownList.js#L451-L453)) will change the value of the `maxHeight` property.  The recent PR https://github.com/qooxdoo/qooxdoo/pull/9377 has demonstrated that the side effects of changing properties (specifically minHeight/maxHeight) can cause problems.

A better response would be to return a non-numeric value from `__getAvailableHeight` (eg `null` or `undefined`) and make `__adjustHeight` immediately return if that value is detected.

johnspackman (changes requested):

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: Thank you for the suggestion, I have just implemented that.

johnspackman (approved):

level420 (approved):

derrell (approved):

johnspackman (approved):

cajus (approved):

hkollmann (approved):

level420:
6 approvals, more than 3 business days, merging
thank you @milandamen

PR #9434, opened 2017-11-30
This is just a test what happens if we change the package name

hkollmann:


derrell (commented):
If it's just a test, do you really want it merged with master?

cboulanger:
I think @johnspackman has a plan how to roll this out. At the moment, there are a lot of projects that rely on the `qooxdoo` package - I think we have to be careful...

cboulanger:
Just saw your conversation on Gitter! It would be good to have the new setup, in particular having the qooxdoo-sdk package.

hkollmann:
Not yet - i just want to see the travis results

hkollmann:
qooxdoo package is qooxdoo server - generated during build process,. No plans to touch this package in the moment!
The package i think of is qooxdoo-sdk wich should contain the whole Framework.

cboulanger:
We need an issue to remove grunt completely.

hkollmann:
I once tried this. But the whole test enviroment of qooxdoo requires grunt.

hkollmann:
Will come back with more explenation after tests are over

PR #9438, opened 2017-12-01
Migrate tests to saucelabs and use new travis features

cajus:
This Pull Request basically does two things:

 * Use [Saucelabs](https://saucelabs.com) for browser testing
   Because we had re-occuring troubles with travis and the browsers we installed there,
   we're now using an external service which is free for OSS use. Besides the fact that
   we don't need to care about *what is needed* to run the latest and greatest browers,
   we can now do broader browser testing (i.e. Internet Explorer, Edge, Safari). Note
   that I've currently disabled these, because the tests do not run cleanly yet.

   That's something we've to work on, I guess.

 * Use new travis features
   In the past, we had a couple of scripts that alway get run, even if the *branch/tag*
   combination would not allow them to be run. That was i.e. the case for *PRs*. Now
   (finally) they introduced build stages, where stages can be run conditionally. We
   can skip the deployment step if not applicable. No script with various *if/then/else*
   combinations are needed.

   Additionally these stages only continue to build if the previous ones have been
   completed successfully. So - if lint fails, there's no need to do the rest. Or if the
   tests fail, there's no need to do any deployments.

   Travis still runs the coverage tests in parallel, because we don't care about
   *what browser* is used for that. It's runs while the other unit tests are running
   on Saucelabs.

Does it work? Well, the tests itself run, although we've to find a good value for network
timeouts. Sometimes a test runs red, because of network timeouts. Maybe it needs
a little observation before tweaking the values.

*What I couldn't test*: deployment. The script needs to run in the *qooxdoo/qooxdoo*
context to test this. I might need to do some live tweaks on the master branch if it
doesn't work.

Why are the jobs of each stage have cryptic 'names'? That's an open ticket on the
travis side. When it's resolved, we can name the jobs as we want. So the build
overview should be more self explaining then.


cboulanger (dismissed):
Thank you for this!

level420 (changes requested):
There seems to be some authentication problems with saucelab:
```
01 12 2017 14:19:16.255:ERROR [launcher.sauce]: Can not start firefox beta (Windows 10)

  [init({"version":"beta","platform":"Windows 10","tags":[],"name":"Qooxdoo Unit-Tests","tunnel-identifier":"1003.2","record-video":false,"record-screenshots":false,"device-orientation":null,"disable-popup-handler":true,"build":"1003","public":"public","commandTimeout":300,"idleTimeout":90,"maxDuration":1800,"customData":{},"base":"SauceLabs","browserName":"firefox"})] The environment you requested was unavailable.

  Sauce Labs Authentication Error.

You used username 'None' and access key 'None' to authenticate, which are not valid Sauce Labs credentials.
```

cajus:
@level420 yes, need to read some more stuff for PRs. The variables are there, but they're not used.

hkollmann (approved):

cajus:
@level420 the authorization issues are resolved. In PRs we can only use JWT based authentication to Saucelabs, because it would expose the real token otherwise.

Nevertheless, the builds are a bit unstable at the moment (timeouts). Time will show where tweaking is necessary.

hkollmann:
One question: Why do change the file Extension to nothing?

cajus:
@hkollmann it's my favorite choice. As we run under Unix, there's no need for extensions and the *shebang* does the magic. Also it's cleaner to read.

hkollmann:
As Windows user i like extensions :), therefore the question

cajus:
You mean 'Windows is extension centric and doesn't leave you a choice' - so you like extensions ;-) Well, the scripts will most likely never run in any Windows environment, so you can safely ignore the missing extension in this case :-)

johnspackman (approved):

level420 (approved):

derrell (approved):
I know very little about travis. I'll approve this based on others' recommendations, so that it can get merged.

cboulanger (approved):

hkollmann:
5 approves, 3 days. 
Thanks cajus!

PR #9439, opened 2017-12-01
readd missing grunt files for building

hkollmann:
During grunt cleanup this files get lost. They are needed during build process. So i added them again.


hkollmann:
Only needed for build with make. We should deprecate or delete this

PR #9440, opened 2017-12-01
Add needed packages doc

hkollmann:
during setup an enviroment with Windows 10 ubuntu i documented the needed packages for the build process.

### DIFF
--- a/documentation/tech_manual/source/pages/project/release_how_to_build.rst
+++ b/documentation/tech_manual/source/pages/project/release_how_to_build.rst
@@ -15,7 +15,7 @@ What you need
    Sphinx:
       sudo apt-get install python-sphinx   
    PDFLatex:
-      sudo apt-get install texlive
+      sudo apt-get install texlive-full
    nice:
       sudo apt-get install coreutils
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

level420 (approved):

cajus (changes requested):
This is very windows specific stuff and I guess it's not correct, too. At least the qx toolchain is incompatible with python 3.6. It needs python 2.x. In this case (and in case of ordinary ubuntu systems) you don't need any PPAs installed. So that is misleading.

Additionally building releases is done automatically and the documentation does not reflect what we do in the moment. I'd drop it completely.

cboulanger (changes requested):
I defer to @cajus' judgement, since I don't know enough about the matter.

hkollmann:
I will close this PR and open a new one in https://github.com/qooxdoo/qx-dev-manual special for the requierements of an ubuntu subsystem in Windows 10. Would this be ok?
Whe should really delete the admin docs here and maintenance them only in qx-dev-manual

PR #9442, opened 2017-12-04
Fix load time dependency

cajus:
Using the new compiler rises at least this dependency problem
which is really anoying. You just need to move the mouse while
loading the application and you're bugged into the developer
console.

Maybe there are others, but this one seems to be a must have.


cboulanger (approved):

johnspackman (approved):
There will be others that crop up - it's because the generator dependency checker is not 100% deterministic and happens to work in a slightly different (and undocumented) way.

oetiker (approved):

hkollmann (approved):

derrell (approved):

level420 (approved):

hkollmann:
6 approves, 4 days gone.
Thanks @cajus

PR #9443, opened 2017-12-04
Adding a new function to get the full model array with no changes

nwgroup:
Sometimes we need the __fullArr to restore all the model when use some filter

### DIFF
--- a/framework/source/class/qx/bom/client/OperatingSystem.js
+++ b/framework/source/class/qx/bom/client/OperatingSystem.js
@@ -126,6 +126,8 @@ qx.Bootstrap.define("qx.bom.client.OperatingSystem",
       "Win95" : "95",
 
       // OS X
+      "Mac OS X 10_13" : "10.13",
+      "Mac OS X 10.13" : "10.13",
       "Mac OS X 10_12" : "10.12",
       "Mac OS X 10.12" : "10.12",
       "Mac OS X 10_11" : "10.11",


framework/source/class/qx/ui/table/model/Filtered.js
```diff
@@ -105,6 +105,14 @@ qx.Class.define("qx.ui.table.model.Filtered",
     __applyingFilters : null,
     __filterTypes : null,
 
+    /**
+     * Function to get the full array of the filtered model
+     * @return {Array} the full array of model (with no changes)
+    */
+      getFullArray: function ()
+    {
+	return this.__fullArr;
```
derrell: Please fix the indentation.

derrell (changes requested):
Once the indentation is fixed, I have no problem with this PR. I recall, however, that this rarely-used class is highly inefficient. While you're in here playing around with it, if you'd like to review it and possibly offer a (separate) PR that makes it more efficient, that would be really useful.

framework/source/class/qx/ui/table/model/Filtered.js
nwgroup: Hi, we already fix the indentation. We will review it but at the moment the change help us a lot. Thanks!

hkollmann:
Could you please revert the change to .gitignore?

hkollmann (changes requested):
Please revert .gitignore

derrell (changes requested):
The .gitignore changes shouldn't be in this PR.

PR #9445, opened 2017-12-06
Fix script paths

cajus:


cboulanger (dismissed):

derrell (dismissed):

level420 (dismissed):

hkollmann (dismissed):

level420:
@cajus I had a look (as well as @hkollmann) into the failing tests and it seems to be
```
Chrome 62.0.3202 (Windows 10 0.0.0) qx.test.bom.request Script:test: not call ontimeout when request is within timeout limit FAILED
	expected ontimeout to not have been called but was called once
	    ontimeout(): Called fail().
		qx.core.AssertionError.constructor:43:38
		qx.core.Assert.__fail:71:19
```
The timeouts should be raised there as well. Maybe x4? 
```javascript
    "test: not call ontimeout when request is within timeout limit": function() {
      var req = this.req,
          that = this;

      this.spy(req, "ontimeout");

      req.onload = function() {
        that.resume(function() {

          // Assert that onload() cancels timeout
          that.wait(1400, function() {
            that.assertNotCalled(req.ontimeout);
          });

        });
      };

      req.timeout = 1200;
      this.request();
      this.wait();
    },
```

cboulanger (dismissed):

hkollmann:
2 more places with timeouts:
Chrome 62.0.3202 (Windows 10 0.0.0) qx.test.io ImageLoader:testLoadImageFailure FAILED
Chrome 62.0.3202 (Windows 10 0.0.0) qx.test.io ImageLoader:testLoadVectorImageFailure FAILED

hkollmann:
Some more places:
Chrome 63.0.3239 (Windows 10 0.0.0) qx.test.io ImageLoader:testLoadImageSuccess FAILED
Chrome 63.0.3239 (Windows 10 0.0.0) qx.test.io ImageLoader:testLoadVectorImageSuccess FAILED

johnspackman (dismissed):

level420:
@cajus I'm getting the impression that the saucelabs runs are getting interrupted or lowered on their priority in an unpredictable way, maybe on high load situations, or not being a paying customer  or something similar. We may end up having the situation that the testrunner succeeds in the end on saucelabs by taking 30min for a complete test run, just because we raised all those time outs to the max. 
On the other side interactive runs, which I've done a lot yesterday, also take that long.

Could we somehow use a global multiplicator for all this timeouts, maybe some static in `qx.dev.unit.TestCase`  a method `scaleTimeout(timeValue)` which then scales the timeout?

```javascript
scaleTimeout : function(timeValue) {
  return Math.round(timeValue * qx.dev.unit.TestCase.timeoutScaleFactor);
}
```
and then use it like this:
```javascript
this.wait(this.scaleTimeout (8000), function() {
  that.assertEquals(i, completedCount);
});
```
or even modify the `wait` method to do the scaling.

cajus:
@level420 the problem is that there are so many hard coded timeouts. The ones with 500ms are increased and they're working. I don't see how to scale that in a general way. Maybe directly in the testcases.

level420 (approved):
Cool! Thank you @cajus !

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

derrell (approved):

hkollmann:
5 approvals, more then 3 days.
Thank you @cajus

PR #9446, opened 2017-12-06
change modul name and update grunt

hkollmann:
With this patch we can integrate qooxdoo-sdk into qooxdoo-cli as an node plugin.
If you run `node install` you will get qooxdoo-sdk version 5.0.2.

As developer you can  do a
`git link ../qooxdoo`
for linking the current working copy.

The patch is needed because npm link works with the module name.

I checked the current build process and found no issues which depends on the old module name.
After accept i will prepare qooxdoo-cli to use qooxdoo-sdk.

Update for grunt is needed because some submodules of grunt are deprecated since node 7.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,795 @@
+{
+  "name": "qooxdoo-sdk",
+  "version": "6.0.0-alpha",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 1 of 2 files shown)


cboulanger (dismissed):
"qooxdoo" is the npm-name of the qooxdoo-server component anyways, so it wasn't correct in the first place.

derrell:
I don't know what package-lock.json is. I suspect it is a temporary file created by npm or something. Is it really supposed to be a committed file in this PR?

cboulanger:
@derrell Yes. See https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

derrell (approved):

oetiker (approved):

cboulanger (approved):

cajus (approved):

johnspackman (approved):

PR #9448, opened 2017-12-07
Polyfill for Object.is which is missing in IE11

level420:
IE11 and maybe other clients not fully supporting ECMAScript 2015 do not offer `Object.is`. This polyfill is implemented as recommended in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

This is the reason why three tests failed for IE11 in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/core/Property.js

### DIFF
--- a/framework/source/class/qx/lang/normalize/Object.js
+++ b/framework/source/class/qx/lang/normalize/Object.js
@@ -58,6 +58,27 @@ qx.Bootstrap.define("qx.lang.normalize.Object", {
       }
 
       return arr;
+    },
+
+
+    /**
+     * Determines whether two values are the same value.
+     *
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is">MDN web docs: Object.is()</a>
+     *
+     * @signature function(value1,value2)
+     * @param value1 {Object} the first value to compare
+     * @param value2 {Object} the second value to compare
+     * @return {Boolean} indicating whether or not the two arguments are the same value.
+     */
+    is : function(v1, v2) {
+      if (v1 === 0 && v2 === 0) {
+        return 1 / v1 === 1 / v2;
+      }
+      if (v1 !== v1) {
+        return v2 !== v2;
+      }
+      return v1 === v2;
     }
   },
 
@@ -70,6 +91,11 @@ qx.Bootstrap.define("qx.lang.normalize.Object", {
     // values
     if (!qx.core.Environment.get("ecmascript.object.values")) {
       Object.values = statics.values;
+    }
+
+    // is
+    if (!qx.core.Environment.get("ecmascript.object.is")) {
+      Object.is = statics.is;
     }
   }
 });
--- a/framework/source/class/qx/lang/normalize/Object.js
+++ b/framework/source/class/qx/lang/normalize/Object.js
@@ -66,19 +66,20 @@ qx.Bootstrap.define("qx.lang.normalize.Object", {
      *
      * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is">MDN web docs: Object.is()</a>
      *
-     * @signature function(value1,value2)
-     * @param value1 {Object} the first value to compare
-     * @param value2 {Object} the second value to compare
+     * @signature function(x,y)
+     * @param x {Object} the first value to compare
+     * @param y {Object} the second value to compare
      * @return {Boolean} indicating whether or not the two arguments are the same value.
      */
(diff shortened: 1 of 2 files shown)


cboulanger (dismissed):

hkollmann (dismissed):

johnspackman (dismissed):

framework/source/class/qx/lang/normalize/Object.js
```diff
@@ -58,6 +58,27 @@ qx.Bootstrap.define("qx.lang.normalize.Object", {
...
+     * @signature function(value1,value2)
+     * @param value1 {Object} the first value to compare
+     * @param value2 {Object} the second value to compare
+     * @return {Boolean} indicating whether or not the two arguments are the same value.
+     */
+    is : function(v1, v2) {
+      if (v1 === 0 && v2 === 0) {
+        return 1 / v1 === 1 / v2;
+      }
+      if (v1 !== v1) {
```
derrell: This comparison always yields false, doesn't it? If there's some strange edge case where that's not true, it certainly needs documentation here to specify what it is.

level420: @derrell this is not my code. It is the hopfully well tested code recommende by MDN. See:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is

derrell: Disregard that comment. NaN !== NaN.

However, see next comment.

framework/source/class/qx/lang/normalize/Object.js
```diff
@@ -58,6 +58,27 @@ qx.Bootstrap.define("qx.lang.normalize.Object", {
...
+     * Determines whether two values are the same value.
+     *
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is">MDN web docs: Object.is()</a>
+     *
+     * @signature function(value1,value2)
+     * @param value1 {Object} the first value to compare
+     * @param value2 {Object} the second value to compare
+     * @return {Boolean} indicating whether or not the two arguments are the same value.
+     */
+    is : function(v1, v2) {
```
derrell: This function is different than the polyfill suggested at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is. Given that it is exercising edge cases, we run the risk of doing something not quite right if we don't use what is suggested. Is there something that this is intended to accomplish that the polyfill from there doesn't?

level420: Yes! Noticed that now. My code is from the `german` version of MDN which obviously differs from the en-US version. See: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Polyfill
I'll change it to the one suggested on the en-Us version now.

level420: The difference between both versions tells me to allways switch to the en-Us version when searching in MDN, as the german pages seem outdated. :-(
That's odd.

level420:
New polyfill function taken from the en-Us version of the MDN page.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Polyfill vs. https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Polyfill

level420:
Changed it now on MDN and both polyfills are identically now.

derrell (approved):

cboulanger (approved):
Thanks @derrell for looking so closely! I was also assuming that the code on MDN would be correct.

level420:
@cboulanger  the previous code WAS correct, but it was an older version of the MDN polyfill which was present in the slightly outdated german translation of the MDN page related to `Object.is`. There was nothing false with that code. It worked as expected.

My fault was that I went from the google search result to MDN, where I was re routed to the german translated version of the page, went to the polyfill, tested it locally here and created the PR. Afterwards I changed the link in the documentation of the PR to the en-Us version of the MDN page, without noting the differences.

It seems that you can not rely on information which is presented on the international pages of MDN, as not only the translation, but also the code on those may be updated.

level420:
@cboulanger note that now there are no differences in the polyfill of the german and en-Us page anymore, as I edited them on MDN to get par. What bothers me a bit is how simple I could do my edit and how fast it was approved and published. That all happened in a few minutes.

hkollmann (approved):

oetiker (approved):

cajus (approved):

PR #9452, opened 2017-12-12
Fix test qx.test.bom.Attribute for safari/webkit

level420:
Current Safari does not report the maximum value for `maxlength` for input elements anymore which was til some point 524288. Instead it reports, in compliance with W3C [1], the value of -1 as a default for the case that there is no default value for the attribute. The current W3C tests [2] also expect this behaviour.

See 
[1] https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#limited-to-only-non-negative-numbers
[2] http://w3c-test.org/html/semantics/forms/the-input-element/maxlength.html

### DIFF
--- a/framework/source/class/qx/test/bom/Attribute.js
+++ b/framework/source/class/qx/test/bom/Attribute.js
@@ -55,7 +55,6 @@ qx.Class.define("qx.test.bom.Attribute",
 
       this.__maxLengthValues = {
         "mshtml": 2147483647,
-        "webkit": 524288,
         "default": -1
       };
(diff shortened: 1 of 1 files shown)


derrell (approved):

cajus (approved):

cboulanger (approved):

hkollmann (approved):

johnspackman (approved):

hkollmann:
3 days, 5 approvals
Thanks @level420

PR #9453, opened 2017-12-12
Fix for failing qx.test.lang.normalize.Date."test parse()" on current safari

### LINKED ISSUE #9451: Testrunner: test qx.test.lang.normalize.Date."test parse()" fails on current safari
The test fails currently in line https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/lang/normalize/Date.js#L57 because the native `Date.parse()` on safari is able to parse dates not containing any hyphens.

It seems that current safari/webkit Date implementation does parse successfully strings like
```javascript
Date.parse("19700101")
```
into a date object.

This is exactly tested via the failing test with 
```javascript
```
There are 2 possiblitites to handle this:

1. Add a partly polyfill which overrides this behaviour so that it consistently fails on all platforms
2. Adapt the test, so it succeeds on safari but fails on all others, to be able to detect the time where either safari complies to all other browsers or the other browsers comply to safari.

I'd think that the second option is the right one, but I'm not sure.

level420:
It seems that current safari does not return NaN on parsing date strings with missing hyphens. Instead it returns a time value which, when used to set a date object with that value via setTime, returns NaN.

This PR takes care of this speacial (maybe a bug?) behavior to get testrunner to complete successfully on safari again.
It should be reverted if the behavior changes again in future safari versions.

Fixes #9451

### DIFF
--- a/framework/source/class/qx/test/lang/normalize/Date.js
+++ b/framework/source/class/qx/test/lang/normalize/Date.js
@@ -54,7 +54,17 @@ qx.Class.define("qx.test.lang.normalize.Date",
 
       this.assertTrue(isNaN(Date.parse("asdf")), "invalid YYYY (non-digits)");
       this.assertTrue(isNaN(Date.parse("1970-as-df")), "invalid YYYY-MM-DD (non-digits)");
-      this.assertTrue(isNaN(Date.parse("19700101")), "invalid YYYY-MM-DD (missing hyphens)");
+      
+      // safari does not return NaN on parsing date strings
+      // with missing hyphens. Instead it returns a time value
+      // which, when used to set a date object with that value
+      // via setTime, returns NaN
+      if(qx.core.Environment.get("browser.name") == "safari") {
+        this.assertTrue(isNaN((new Date()).setTime(Date.parse('19700101'))), "invalid YYYY-MM-DD (missing hyphens)");
+      }
+      else {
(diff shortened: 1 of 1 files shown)


hkollmann (approved):

cboulanger (approved):

derrell (changes requested):
I wonder if this is the right place to fix this. This test is testing *normalize* functions. Clearly, if you have to look at which browser it is, the function is not normalized. Either we should have a Date normalization class that handles the browser differences, or just delete this test. Is it really accomplishing anything the way it is right now?

hkollmann (changes requested):
After rethinking i think Derrel is right. We test standard behavior here. So we should think about a polyfill for Safari if the parsing is not as expected.

Something like:
usePolyFill if !Date.parse('19970101') = nan
....

johnspackman:
👍 I agree, it would be better if we could normalize `Date.parse` across platforms.

level420:
OK then I's suggest to add an additional validation step for safari into the current polyfill:

If parsing succeeds, e.g. something else then `NaN` is returned from `Date.parse` I'll create an intermediate Date instance and set the parsed time to the new instance via `.setTime()`. Similar to the change in this PR.

Preparing another PR.

level420:
closed in favor of https://github.com/qooxdoo/qooxdoo/pull/9454

PR #9454, opened 2017-12-12
Catch false positive parsing of date strings via Date.parse in safari (and maybe others)

### LINKED ISSUE #9451: Testrunner: test qx.test.lang.normalize.Date."test parse()" fails on current safari

level420:
Check if parsing a date string via the original Date.parse method returns a real time value, which we ensure by using setTime with an intermediate Date object and the parsed time value. 

Safari 11 e.g. parses the date string '19700101' successfully into a time value, but returns NaN if that value is used in setTime.

See:
  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse
  https://github.com/qooxdoo/qooxdoo/issues/9451

Fixes #9451

### DIFF
--- a/framework/source/class/qx/lang/normalize/Date.js
+++ b/framework/source/class/qx/lang/normalize/Date.js
@@ -61,7 +61,24 @@ qx.Bootstrap.define("qx.lang.normalize.Date", {
       // Match input against ISO8601 regular expression
       var captureGroups = /^(\d{4}|[+\-]\d{6})(?:-(\d{2})(?:-(\d{2}))?)?(?:T(\d{2}):(\d{2})(?::(\d{2})(?:\.(\d{3}))?)?(?:(Z)|([+\-])(\d{2})(?::(\d{2}))?)?)?$/.exec(dateString);
       if (!captureGroups) {
-        return Date.originalParse(dateString);
+        //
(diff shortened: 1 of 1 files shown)


derrell (approved):

johnspackman (approved):

cajus (approved):

hkollmann (approved):

cboulanger (approved):

hkollmann:
5 approvals, 3 days over.
Thanks, @level420

PR #9455, opened 2017-12-12
Fix qx.test.bom.webFonts.Manager for "test: load webfont with custom version"

level420:
Previous to this PR only the existence of a rule for the woff-font was tested, only succeeding on browsers which loades woff-fonts.
With this change potentially all types are checked.
Safari seems to load the ttf version and IE the eot version of the font which made this test fail on those browsers.


johnspackman (approved):

cajus (approved):

hkollmann (approved):

cboulanger (approved):

oetiker (approved):

hkollmann:
3 days, 5 approvals,
Thanks @level420!

PR #9457, opened 2017-12-12
add missing grunt stuff

hkollmann:
in august 2017 there was grunt stuff deleted which is needed for creating the docs. This PR reverts the deletion

### DIFF
--- a/tool/grunt/config/application.js
+++ b/tool/grunt/config/application.js
@@ -1,0 +1,76 @@
+/* *****************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2014 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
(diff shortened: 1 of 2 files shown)


johnspackman:
I'm not familiar with this code, but it appears that it's just a wrapper around the generator - that it just automatically imports individual generator jobs into grunt as tasks.  Is that right?  In which case the build process could just call generator directly instead of going via grunt in the first place

hkollmann:
That's an good idea. But not my intention yet. I just want to fix the build process.

cboulanger (approved):
If it is necessary to fix failing tests/cronjobs, sure.

johnspackman (approved):

level420 (approved):
I'm not really familiar with grunt, but I'm trusting @hkollmann here.

hkollmann:
Here the error message from travis build log:

```

Loading "Gruntfile.js" tasks...ERROR

>> Error: Cannot find module './config/application.js'

Warning: Task "api" not found. Use --force to continue.
```

oetiker (approved):

cajus (approved):

PR #9460, opened 2017-12-13
Fix 3 testrunner tests for ie and edge

level420:
1. qx.test.Browser.testForIn for edge: This tests the hash map element order where it seems that edge has the same behavior as all other modern browsers have: sort by hash key.
2. qx.test.ui.embed.Iframe.testSyncSourceAfterDOMMove for IE and edge: IE and edge deliver an extra blank at the end of body.innerText which made the test fail. The extra blank is now deleted specifically for edge and ie.
3. qx.test.io.request.XhrWithRemote."test: progress phases when abort after loading": breaks on win 10 for edge, ie, chrome and firefox. Disabled this test by the added detection for not having win 10 ("noWin10") 
4. qx.test.util.DateForma.testInvalidDate for edge: edge parses even invalid dates and calculates  from the "overflowing" days and months the next "logical" date. In the test the invalid date "2011-11-32" is parsed as "2011-12-02". Disabled this test for edge.

### DIFF
--- a/framework/source/class/qx/dev/unit/MRequirements.js
+++ b/framework/source/class/qx/dev/unit/MRequirements.js
@@ -257,6 +257,18 @@ qx.Mixin.define("qx.dev.unit.MRequirements", {
 
 
     /**
+     * Checks if the application is running on Windows 10
+     *
+     * @return {Boolean} <code>false</code> if operating system is Windows 10
+     */
+    hasNoWin10 : function()
+    {
+      var isWin10 = (qx.core.Environment.get("os.name") === "win" && qx.core.Environment.get("os.version") === "10");
+      return (isWin10 ? false : true);
+    },
+
+
+    /**
      * Checks if the application is not running in a Google Chrome browser on Linux
      *
      * @return {Boolean} <code>true</code> if the browser is not Google Chrome on Linux
--- a/framework/source/class/qx/test/util/DateFormat.js
+++ b/framework/source/class/qx/test/util/DateFormat.js
@@ -20,6 +20,8 @@ qx.Class.define("qx.test.util.DateFormat",
 {
   extend : qx.dev.unit.TestCase,
 
+  include: qx.dev.unit.MRequirements,
+
   members :
   {
 
@@ -122,6 +124,13 @@ qx.Class.define("qx.test.util.DateFormat",
 
     testInvalidDate : function()
     {
+      // Note:
+      //   * Edge parses even invalid dates and calculates
+      //     from the "overflowing" days and months the
+      //     next "logical" date. In the example below
+      //     the date parsed is "2011-12-02".
+      this.require(["noEdge"]);
+
       var invalidDate = new Date("2011-11-32");
       var dateFmt = new qx.util.format.DateFormat();
       this.assertNull(dateFmt.format(invalidDate));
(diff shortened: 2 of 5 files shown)


johnspackman (dismissed):

level420:
Please hold on. Pushing more fixes.

framework/source/class/qx/dev/unit/MRequirements.js
```diff
@@ -256,6 +256,18 @@ qx.Mixin.define("qx.dev.unit.MRequirements", {
...
 
+    /**
+     * Checks if the application is running on Windows 10
+     *
+     * @return {Boolean} <code>false</code> if operating system is Windows 10
+     */
+    hasNoWin10 : function()
+    {
+      var isWin10 = (qx.core.Environment.get("os.name") === "win" && qx.core.Environment.get("os.version") === "10");
+      return (isWin10 ? false : true);
```
derrell: `isWin10` is already a boolean. It doesn't need the trinary expression. Could just `return ! isWin10;`

framework/source/class/qx/test/util/DateFormat.js
```diff
@@ -849,6 +858,10 @@ qx.Class.define("qx.test.util.DateFormat",
     dfDE.dispose();
     dfUS.dispose();
 
+  },
+  
+  hasNoEdge: function() {
+    return !(qx.core.Environment.get("browser.name") == "edge");
```
derrell: Any reason not to do `return (qx.core.Environment.get("browser.name") != "edge");` instead? The `!=` operator is easier to read than a negated expression.

derrell (changes requested):

framework/source/class/qx/dev/unit/MRequirements.js
level420: @derrell please view the other methods in the file. they are all written similar e.g the method `hasNowWin7`.

framework/source/class/qx/test/util/DateFormat.js
level420: same here. adopted from https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/test/io/request/XhrWithRemote.js#L227-L229

derrell:
Sigh. Yeah, consistency is important. They should all be fixed.

derrell (approved):

level420:
Yep, they should maybe be fixed at some point. But for the tests it is not so important I think. 
This PR is for bringing the failing testrunner tests for edge down to 0 to be able to add it as a saucelabs test platform.

johnspackman (approved):

derrell:
Indeed. That's why I approved it as is. Fixing it globally is a different, probably large, PR.

cajus (approved):

hkollmann (approved):

derrell:
Regarding `testForin`, why is this even being tested. Maps are by definition unordered, and you can never, ever, ever count on the order of retrieval. We have that as a known issue in @cboulanger 's Dialog widget (at least in the old version that I'm using), which assumes that retrieving the map keys will occur in the order in which they were generated in a map literal or map element assignment. I really, really hope we don't have any such assumptions within the qooxdoo core, and we should probably remove this test so no one thinks that it's reasonable to assume an order.

level420:
I think we should, if appropriate, delete the `testForin` in a separate PR. The intention of all these PRs is to have safari and edge testrunner succeed again.

derrell:
SGTM.

cboulanger (approved):

johnspackman:
5 approvals, merging. thank you @level420 !

PR #9462, opened 2017-12-13
Fix webfont rule removal for IE and edge

level420:
In IE and edge even if the rule was added with font-style first and font-weight second, it is not guaranteed that the attributes remain in that order. Therefore we check for both versions, style first, weight second and weight first, style second.
Without this fix the rule isn't found and removed reliable. 

This also makes the test qx.test.bom.webfonts.Manager."test: do not create rule for invalid font" succeed again on IE and edge


cajus (approved):

hkollmann (approved):

cboulanger (approved):

johnspackman (approved):

oetiker (approved):

oetiker:
5 days 5 votes

PR #9463, opened 2017-12-13
Testrunner: enqueue qx.test.bom.media.MediaTestCase.testPlayEvent as the first test to make it succeed on safari

level420:
On safari this test fails, if we are running the complete testsuite and the test is not the first test in qx.test.bom.media.MediaTestCase. By naming it with a blank behind test it is enqueued as the first test and the test run succeeds.
 
Note that if the tests in qx.test.bom.media.MediaTestCase are started seperately or this test is restarted after it failed, then it succeeds.
This is some weird timing or caching issue which I could not solve otherwise.

### DIFF
--- a/framework/source/class/qx/test/bom/media/MediaTestCase.js
+++ b/framework/source/class/qx/test/bom/media/MediaTestCase.js
@@ -189,7 +189,19 @@ qx.Class.define("qx.test.bom.media.MediaTestCase",
       this.wait();
     },
 
-    testPlayEvent: function()
+    // On safari this test fails, if we are running the
+    // complete testsuite and the test is not the first
+    // test in qx.test.bom.media.MediaTestCase. By naming
+    // it with a blank behind test it is enqueued as the first
+    // test and the test run succeeds.
+    //
+    // Note that the tests in qx.test.bom.media.MediaTestCase
+    // are started seperately or this test is restarted after 
+    // it failed, then it succeeds.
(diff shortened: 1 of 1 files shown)


cajus (approved):

framework/source/class/qx/test/bom/media/MediaTestCase.js
```diff
@@ -189,7 +189,19 @@ qx.Class.define("qx.test.bom.media.MediaTestCase",
       this.wait();
     },
 
-    testPlayEvent: function()
+    // On safari this test fails, if we are running the
```
derrell: This is awful that it's required. We'd better add lines here:
```
//
// KLUDGE ALERT!
//
```

level420: @derrell feel free to add whatever and wherever you think the comment should be changed or moved. You can do that by editing the PR online. I know that this is not an optimal solution, but it’s th only fix I could find.

cboulanger (approved):

derrell (approved):
Random failures are miserable to work on. If this solves it, at least it's documented why it's named that way. It is a serious kludge, but there are other things to work on. :-(

hkollmann (approved):
awfull...

johnspackman (approved):

johnspackman:
5 approvals, merging.  thank you @level420 !

PR #9464, opened 2017-12-14
Prepare alpha release of NPM module qooxdoo-sdk 

hkollmann:
With this we are able to publish the alpha release to npm
I changed readme.rst to README.md so that it will be regonized by npm.
I removed uneeded grunt stuff from the framework.

Tests are done with `npm pack` and `npm link` of this package to qooxdoo-cli. All tests are running.
packed package size is about 16MB.

If we want to have automatic   publishing via travis we must found a way to increase the version number in the package json to something like 6.0.0-alpha.xx.  This change must be commited back in the repository. I do not know if this is working with the current settings.


hkollmann:
`npm version prerelease` will do the job to change version to 6.0.0-alpha.x and do an commit.
We must check if this could works with the current repo settings.
@oetiker ?

johnspackman (dismissed):

package.json
```diff
@@ -12,5 +12,14 @@
...
+  },
+  "files": [
+    "version.txt",
+    "LICENSE",
+    "CHANGELOG.md",
+    "README.md",
+    "framework/Manifest.json",
+    "framework/source",
+    "tool/data/cldr"
+  ]
```
cboulanger: > "Certain files are always included, regardless of settings:
> package.json
> README
> CHANGES / CHANGELOG / HISTORY
> LICENSE / LICENCE
> NOTICE
> The file in the "main" field
> README, CHANGES, LICENSE & NOTICE can have any case and extension.

see https://docs.npmjs.com/files/package.json#files

hkollmann: I know. I just wanted to document the Inklusion.

cboulanger:
So is the module called qooxdoo-sdk or qooxdoo-framework? Thinking about it, qooxdoo-framework is the better name because it is not the SDK what is needed for qooxdoo-compiler, just the framework. Fittingly, you have placed the package.json inside the framework folder, and not in the SDK folder in ../

hkollmann:
package.json in framework is'nt delivered. It's only needed for the karma stuff to run the tests  on travis.
The package name wich is used in npm at the moment for the whole sdk is qooxdoo-sdk.

I think changing the name to qooxdoo-framework is an good idea. We must then change qooxdoo-cli to use this name.

I can try to move the package definition into the framework folder. We then must change the cli accordingly.

cboulanger:
yes, sorry, forget my last sentence. package.json needs to be in the top folder. but that's not a problem . Just publish the meta-files in the top folder plus the framework/ folder.

hkollmann:
That's what i'm doing

hkollmann:
shell gurus!

Could someone write a script which recursively Scans for package.json files and run npm install in that directory?


grunt Task does this:
```
   function() {
      var rootDir = shell.pwd();
      var isPackageJson = function(file) {
        return file.match(/package\.json$/) && !file.match("node_modules");
      };
      var pkgJsonPaths = shell.find('.').filter(isPackageJson);

      pkgJsonPaths.forEach(function(filePath) {
        var basePath = path.dirname(filePath);
        grunt.log.subhead("Installing " + basePath);
        shell.cd(basePath);
        shell.exec('npm install');
        shell.cd(rootDir);
      });


```
With this we could remove the grunt dependency from root package.json.

Thanks for your help!

derrell:
How about something like this?

```
#!/bin/bash

files=$(find . -name package.json -print)

for f in ${files}; do
    dir=$(dirname $f)
    (cd ${dir} && npm install )
done
```

hkollmann:
Should exclude node_modules dir. Will test tomorrow.

derrell:
That's no fair, adding new requirements after I've already submitted the bill! :-)

```
#!/bin/bash

files=$(find . -name package.json -print | grep -v node_modules)

for f in ${files}; do
    dir=$(dirname $f)
    (cd ${dir} && npm install )
done
```

hkollmann:
This requirement was in the js script😀

hkollmann:
Thank you very much!

derrell:
Sure. You're welcome.

johnspackman (dismissed):

cboulanger (approved):
Thank you very much for your work on this (and for removing the grunt dependency!)

derrell (approved):

.travis.yml
```diff
@@ -26,8 +27,7 @@ before_install:
 
 install:
 - export PATH=`pwd`/texlive/bin/x86_64-linux:$PATH
-- npm install
-- grunt setup
+- .travis/prepare_npm
```
cajus: I guess the script can be omitted if you just write:

`- find . -name package.json ! -path '*/node_modules/*' -execdir npm install \;`

A big nitpicker, though ;-)

oetiker:
@hkollmann hmmm I don't think anyone gets to push to master directly

johnspackman (approved):

.travis.yml
hkollmann: You could be right 
But i prefer the script:
1. it can be locally tested
2. i use it in the local build script run_local_build.sh to test
3. it‘s my first bash script 😉

If you could agree could you do the merge please?

hkollmann: OK, got it. wonderful find. read the man page. Could you live with the script? It echos the working dir.

cajus: I didn't block the PR because of that - that was just a hint that the script isn't needed in this case. The single `find` is more robust against spaces in the path and it spawns much less processes than the script with it's loop. Both facts are not relevant in this case, so I'm fine with the script.

cajus (approved):
I can't see whether this change will work without running everything manually before. Most of the parts are not executed for PRs on travis, so we'll have to move it to *master* to see if it works finally. 

@oetiker I don't like the idea of no one being able to push to the repository directly. It's not possible to do a quick fix for deployment if it's completely broken in this case. We always need to go the PR way and the PR checks will not do anything useful in this case. Despite a 30 minutes+ delay. Maybe it's possible to allow GPG signed direct pushes in these case?

oetiker (approved):

oetiker:
@cajus I just looked through the settings again ... what is blocked are 'force pushes' and the rest of the rules seem to affect only merging of PRs ... so it may be that actually anyone who can merge PRs is able to push (without force) to master directly ... might be worth trying ... the description of the options in these settings are not very elaborate.

cajus:
I tried earlier to correct the script path and that was blocked.

oetiker:
@cajus protection only extends to particular branches, so if we create a branch for the npm alpha releases, we should be good

oetiker:
5 votes in 3 days

Issue #8104, opened 2014-01-31
Tab no longer works when modal window looses the focus (BZ#8074)



qx-bug-importer:
_Jeroen (Jeroen.Smit) wrote:_

I encountered a bug in the combination of focus handling and the blocker of a modal window.

Below I included a small piece of code that shows the bug when running the code in e.g. Chrome on the playground. When the window is shown and you click in the first textfield you can use TAB to go from one textfield to the other and SHIFT + TAB to go back. As soon as you click outside of the window and click back in the textfield you can edit but TAB no longer works. Other special keys like Escape and Enter are no longer propagated to the window too.

Debugging the Qooxdoo code I found that
- Initially TAB is handled by the FocusHandler (method __onKeyPress)
- After clicking out of the window and back in the TAB is caught by method __stopTabEvent in Blocker.js

I think that the order of registered events changes at runtime by the focus loss.

------ Code example for playground ------

var win = new qx.ui.window.Window(&quot;First Window&quot;);
win.setWidth(300);
win.setHeight(200);
win.setShowMinimize(false);

win.setLayout(new qx.ui.layout.Canvas());

var tf1 = new qx.ui.form.TextField(&quot;test1&quot;);
win.add(tf1, {left:10, top:10});

var tf2 = new qx.ui.form.TextField(&quot;test2&quot;);
win.add(tf2, {left:10, top:40});

this.getRoot().add(win, {left:20, top:20});

win.setModal(true);

win.open();

adrelino:
Is this finally fixed now? Took me and @marcelr a lot of time to trace this loosing tabbing issue down to a bug in qooxdoo. Or is the suggestion to not use modal windows when tabbing is desired?

level420:
Reopening this issue

level420:
@adrelino @marcelr please have a look into PR https://github.com/qooxdoo/qooxdoo/pull/9466 and report back if it solves the issue for you. Thank you.

level420:
Fixed via #9466

PR #9466, opened 2017-12-18
Issues #9449, #8104: Prevent blocker stealing tab handling of the current focus root

### LINKED ISSUE #9449: Tab key not working if you click outside a modal window and then return to the window
in desktop:
when you open a window with controls in it, if you press tab key it will navigate correctly, but if you click outside the window and then click inside the window, the cursor remains on the textfield and will not move to the next control.
I notices also, not only the Tab key not working, also Space key not working when you press it on buttons.

level420:
If the widget attached to the blocker is also attached to the `qx.ui.core.FocusHandler` as a focus root, activating the blocker after this widget was activated, leads to the focus handler re-activate the widget behind the blocker, loosing tab handling for widget which is visually in front. Hence we prevent activating the blocker in this situation.

Fixes #9449, #8104

### DIFF
--- a/framework/source/class/qx/ui/core/Blocker.js
+++ b/framework/source/class/qx/ui/core/Blocker.js
@@ -115,6 +115,14 @@ qx.Class.define("qx.ui.core.Blocker",
      * events, this is useful to block keyboard input on other widgets.
      * Take care that only one blocker instance will be kept active, otherwise your
      * browser will freeze.
+     * 
+     * Setting this property to true is ignored, if the blocker is attached to a
+     * widget with a focus handler, as this would mean that the focus handler
+     * tries to activate the widget behind the blocker.
+     * 
+     * fixes:
+     *     https://github.com/qooxdoo/qooxdoo/issues/9449
+     *     https://github.com/qooxdoo/qooxdoo/issues/8104
      */
     keepBlockerActive :
     {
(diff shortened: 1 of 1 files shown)


oetiker (approved):
sounds reasonable

cajus (approved):

cboulanger (approved):

johnspackman (approved):

hkollmann (approved):

derrell (approved):
I don't much care for big block comments in the middle of an expression / condition, so I'd prefer that the comment be moved just above the `if` statement. I don't want to block this for that reason, but if you don't object, please move the comment.

level420:
@derrell will do this in a subsequent PR, as adding it now would restart the whole review request, discarding what was approved now.

hkollmann:
Could we go the short way for merge or must we wait for @zaucker?
If not i can publish the sdk after merge

cajus:
What about renaming the SDK to *qx-sdk*? Or do you want to use the current name as a test?

hkollmann:
Let‘s start with qooxdoo-sdk. Otherwise we have to change Package.json again

derrell:
@level420 Yeah, I thought about the whole starting-over thing. Subsequent PR sounds good.

@hkollmann We have enough approvals to merge, but it has not been the requisite time period. I intend to propose a change to the process to allow an immediate merge (no wait time) if *all* core team members have approved, but that is not currently in place, so the addition of zaucker's approval wouldn't speed things up at present.

zaucker (approved):

hkollmann:
With new rule " Allow fast-track approvals" we can merge.
Thanks, @level420!

PR #9468, opened 2017-12-18
Remove texlive and introduce 'git fetch' limit

cajus:
This is an attempt to reduce the required *travis* disk space on the deploy job.

### DIFF
--- a/.travis/deploy
+++ b/.travis/deploy
@@ -74,7 +74,7 @@ git commit -m "Refresh site at ${rev}"
 git push -q upstream HEAD:master
 
 # Do a regular checkout and make a dummy commit
-git clone -q git@github.com:qooxdoo/qooxdoo.github.io.git tmp
+git clone --depth 10 -q git@github.com:qooxdoo/qooxdoo.github.io.git tmp
 cd tmp
 echo $rev > revision
 git add revision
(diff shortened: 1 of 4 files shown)


johnspackman (dismissed):

cboulanger (dismissed):

level420 (dismissed):

derrell (dismissed):

hkollmann (changes requested):
line 77, git clone -q git@github.com:qooxdoo/qooxdoo.github.io.git tmp is a candidate for --depth either....

cajus:
@hkollmann definitely.

hkollmann (approved):

.travis/deploy
```diff
@@ -74,7 +74,7 @@ git commit -m "Refresh site at ${rev}"
 git push -q upstream HEAD:master
 
 # Do a regular checkout and make a dummy commit
-git clone -q git@github.com:qooxdoo/qooxdoo.github.io.git tmp
+git clone --depth 10 -q git@github.com:qooxdoo/qooxdoo.github.io.git tmp
```
derrell: I know what it means, but in this particular case, what are these `--depth` options accomplishing? What's so far down the tree that we don't need it, and if we don't need it, why is it there?

derrell (approved):
See my comment on line 77 (also applies to line 35). If it's just cruft, we should create a separate PR to remove it.

cboulanger (approved):

oetiker (approved):

johnspackman (approved):

hkollmann:
3 days, 5 approvals. Thanks @cajus

zaucker (commented):
Sounds ok to me.

PR #9478, opened 2018-01-07
return promise in start function 

hkollmann:
so that it can be used in async/await context

### DIFF
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -143,7 +143,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
 
     /**
      * Start loading scripts. This may only be called once!
-     *
+     * @return {Promise} a promise which will be resolved after load of all scripts.
      */
     start: function() {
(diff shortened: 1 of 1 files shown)


derrell (dismissed):

cboulanger (changes requested):
Can you document the return value type?

hkollmann:
of Course ...

cboulanger (approved):

level420:
It seems that the travis run fails for some internal problems. I like most the following message in the logs:
```
Encrypted environment variables have been removed for security reasons.
```

level420:
@hkollmann does promisifying the start method somehow change the non promise usage of the script loader? I'm not as familiar with promises as I should be, so my main concern is that it changes something for existing apps not using promises.

hkollmann:
No you can use the start method as before an just ignore the result

johnspackman (approved):

level420 (approved):

oetiker (approved):

cajus (approved):

level420:
5 approvals, more than 3 business days, merging
thank you @hkollmann

PR #9480, opened 2018-01-14
Improved error messages when databinding fails on missing target setter

cboulanger:
This has bugged me for years, but I never got around fixing it. During a binding, if a target setter method is missing, it will throw a cryptic error message (`target["set" + qx.lang.String.firstUp(lastProperty)] is not a function`), with no indication where and why this error was thrown. This PR provides a more descriptive error message which allows to search the code for databinding involving the target object and property.

### DIFF
--- a/framework/source/class/qx/data/SingleValueBinding.js
+++ b/framework/source/class/qx/data/SingleValueBinding.js
@@ -658,12 +658,10 @@ qx.Class.define("qx.data.SingleValueBinding",
           target["reset" + qx.lang.String.firstUp(lastProperty)]();
         } else {
           // fallback if no resetter is given (see bug #2456)
-          try{
-            target["set" + qx.lang.String.firstUp(lastProperty)](null);
-          } catch (e) {
-            throw new Error("Binding '" + lastProperty + "' on target " + target + " failed: " + e);
+          if( typeof target["set" + qx.lang.String.firstUp(lastProperty)] != "function") {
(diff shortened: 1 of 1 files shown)


cajus:
Looks like the try/catch breaks validation in combination with the `onSetFail` option.

cajus (changes requested):
See comment. The `throw Error()` does not respect errors that get catched later on (i.e. ValidationError).

cboulanger:
Doesn't the option `onSetFail` refer to setting a _value_? This concerns the process of setting up the _binding_, so I wouldn't expect `onSetFail` to report the error. The problem,too, is that at that point, the value of the `option` parameter is unavailable, one would have to store it in a private property, but that looks like a hack.

cboulanger:
... the same seems to concern Validation errors - it looks to me that at the time of setting up the binding, values aren't validated, but only at the time of the actual value transfer. Or am I missing something here?

cboulanger:
One test fails on an issue unrelated to the PR: https://travis-ci.org/qooxdoo/qooxdoo/jobs/328772627#L6874

hkollmann:
Sometimes a restart helps!

hkollmann (approved):

derrell (approved):

cajus (approved):

oetiker (approved):
great!

johnspackman (approved):

level420 (approved):

johnspackman:
6 approvals, merging, thank you @cboulanger

PR #9491, opened 2018-02-08
Promise support in Events

johnspackman:
Promise support in Events

While events can appear at first glance to be asynchronous, they form part of an API which is (currently) inherently synchronous; EG when multiple handlers are attached to a widget, they are required to complete fully and synchronously before moving on to the next.

Since implementing promises, properties have been extended to support promises as the return value from apply methods, enabling the apply to delay triggering events until some asynchronous work (EG server round trip) has completed - this PR completes the implementation of promises by allowing event handlers to do the same.

With this PR, the event processing model is unchanged except that any event handler may optionally return a promise, which will effectively suspend the event dispatch process until that promise is resolved.  The suspension will affect situations where there are multiple handlers of the same type on a widget, as well as the when there are a sequence of events triggered by another event (EG a single mouse can trigger multiple pointer and gesture events).

Promises are only used if the event handler chooses to create a promise, meaning that the overhead of adding promises is avoided in the general case.

The most obvious use case for this is handling user interactions which require server round trips - for example, a File Explorer user interface could need to check with the server during drag and drop operations, but drag and drop requires a series of related events (`dragstart` -> `dragover` -> `droprequest` -> `drop`) and without promises it is impossible to do any form of asynchronous work during that sequence.

Backwards compatibility is only an issue in code which is calling several `fireEvent` methods in sequence, and those methods are expected to complete synchronously, and an event handler is added which returns a promise.  From the point of view of existing applications, this will not cause an issue unless the coder adds code to return a Promise from an event handler, in which case he/she will be aware of the implications.  The framework itself does not need to return promises so will not introduce any issues (any instances where the framework issues events in sequence would need to be updated however, and should already be part of this PR).

### DIFF
--- a/application/uitests/source/class/uitests/DragAndDropApplication.js
+++ b/application/uitests/source/class/uitests/DragAndDropApplication.js
@@ -54,8 +54,10 @@ qx.Class.define("uitests.DragAndDropApplication", {
         return new qx.Promise((resolve, reject) => {
           if (cbxCancelDragstart.getValue())
             e.preventDefault();
-          if (cbxRejectDragstart.getValue())
-            return reject();
+          if (cbxRejectDragstart.getValue()) {
+            reject();
+            return;
+          }
           
           if (!e.getDefaultPrevented()) {
             // Register supported types
--- a/application/uitests/source/class/uitests/DragAndDropApplication.js
+++ b/application/uitests/source/class/uitests/DragAndDropApplication.js
@@ -1,0 +1,156 @@
+/* ************************************************************************
+
+   Copyright:
+
+   License:
+
+   Authors:
+
+ ************************************************************************ */
+
+/**
+ * This is the main application class of your custom application "dragndrop"
+ * 
+ * @asset(qx/icon/Tango/16/actions/document-print.png)
+ */
+qx.Class.define("uitests.DragAndDropApplication", {
+  extend: qx.application.Standalone,
+
+  members: {
+    main: function() {
+      this.base(arguments);
+
+      if (qx.core.Environment.get("qx.debug")) {
+        qx.log.appender.Native;
+        qx.log.appender.Console;
+      }
+
+      var doc = this.getRoot();
+      var tb = new qx.ui.toolbar.ToolBar();
+      var btn = new qx.ui.toolbar.Button("Click Me", "qx/icon/Tango/16/actions/document-print.png");
+      tb.add(btn);
+      doc.add(tb, { left: 10, top: 10, right: 10 });
+      
+      var root = new qx.ui.container.Composite(new qx.ui.layout.HBox(10));
+      doc.add(root, { left: 10, top: 100, right: 10, bottom: 10 });
+
+      var compSource = new qx.ui.container.Composite(new qx.ui.layout.VBox(6));
+      root.add(compSource);
+      compSource.add(new qx.ui.basic.Label("Simple Source"));
+      var lstSource = new qx.ui.form.List().set({ draggable: true, selectionMode: "multi" });
+      compSource.add(lstSource);
+
+      for (var i = 0; i < 20; i++)
+        lstSource.add(new qx.ui.form.ListItem("Item " + i, "icon/16/places/folder.png"));
+
+      var cbxCancelDragstart = new qx.ui.form.CheckBox("Cancel dragstart").set({ value: false });
+      compSource.add(cbxCancelDragstart);
+      
+      var cbxRejectDragstart = new qx.ui.form.CheckBox("Reject dragstart").set({ value: false });
+      compSource.add(cbxRejectDragstart);
+      
+      lstSource.addListener("dragstart", function(e) {
+        this.debug("lstSource: dragstart");
+        return new qx.Promise((resolve, reject) => {
+          if (cbxCancelDragstart.getValue())
+            e.preventDefault();
+          if (cbxRejectDragstart.getValue())
+            return reject();
+          
+          if (!e.getDefaultPrevented()) {
+            // Register supported types
+            e.addType("value");
+            e.addType("items");
+  
+            // Register supported actions
+            e.addAction("copy");
+            e.addAction("move");
+          }
+          
+          resolve();
+        });
+      });
+
+      lstSource.addListener("droprequest", function(e) {
+        this.debug("lstSource: droprequest");
+        return new qx.Promise((resolve, reject) => {
+          this.debug("Related of droprequest: " + e.getRelatedTarget());
+
+          var action = e.getCurrentAction();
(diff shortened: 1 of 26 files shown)


oetiker (dismissed):
wow!

application/uitests/source/class/uitests/DragAndDropApplication.js
```diff
@@ -0,0 +1,156 @@
...
+
+      var cbxCancelDragstart = new qx.ui.form.CheckBox("Cancel dragstart").set({ value: false });
+      compSource.add(cbxCancelDragstart);
+      
+      var cbxRejectDragstart = new qx.ui.form.CheckBox("Reject dragstart").set({ value: false });
+      compSource.add(cbxRejectDragstart);
+      
+      lstSource.addListener("dragstart", function(e) {
+        this.debug("lstSource: dragstart");
+        return new qx.Promise((resolve, reject) => {
```
derrell: This may be a bigger discussion than just this PR... Although I'm greatly looking forward to core code being able to use ES6, and it'll be fine for user code to use ES6 along with the new compiler, we're only deprecating, not removing the generator for 6.0, so users should still be able to run and create tests which use the generator. I think we should continue to avoid ES6 in core code until 7.0.

application/uitests/source/class/uitests/DragAndDropApplication.js
```diff
@@ -0,0 +1,156 @@
...
+          if (cbxCancelDragstart.getValue())
+            e.preventDefault();
+          if (cbxRejectDragstart.getValue())
+            return reject();
```
derrell: `return` with a value here; without a value later. `reject()` doesn't actually return anything. This should change to:
```
reject();
return;
```

framework/source/class/qx/event/Utils.js
```diff
@@ -0,0 +1,328 @@
...
+      return qx.event.Utils.ABORT;
+    },
+    
+    /**
+     * Equivalent to `.catch()`; note that unlike promises, this method must be called *before*
+     * `.then()` so that it is able to handle rejections when promises are not in use; this is 
+     * because `Promise.catch` only catches rejections from previous promises, but because promises 
+     * are *always* asynchronous the `.catch` goes at the end.  For synchronous, this is nt possible 
+     * so `Utils.catch` must go before `Utils.then` 
+     *
```
derrell: This similarity in names but difference in how it must be used is very likely to cause people problems. Maybe it should have a different name?

derrell (changes requested):
Wow. Huge PR. Looks useful. I added a few comments / change requests. The changes are way too big to catch problems. This will require a lot of testing for regressions.

framework/source/class/qx/event/Utils.js
johnspackman: I wondered about that, and I'm open to suggestions - the dilemma is that it is modelled after promises, which is conceptually convenient at times but not at others.  In many cases this class is only for use as an internal class, so what do you think about if the comments at the top should make that clearer?

application/uitests/source/class/uitests/DragAndDropApplication.js
johnspackman: oops, forgot to lint that part...

johnspackman: ah, on second thoughts this `uitests` folder is designed to produce multiple applications, each of which can be run as independent test applications; the generator only supports a single application model.  The idea behind this `uitests` folder is to create tests that can easily be operated by selenium/etc when it arrives (which should also be in 6.x, soon after I can answer the outstanding question in https://github.com/qooxdoo/qooxdoo/pull/9426)

Also, we've been moving recently towards integrating the compiler into the qooxdoo core repo (i.e. so that it can ship with qooxdoo alongside the generator), but even if that doesn't make it into the 6.0 release the compiler can still be installed automatically thanks to `npm` and `node` is already a requirement for 5.x

Can we then support the compiler/ES6 in specific places like framework tests  provided that it is not in the framework source code itself?

johnspackman:
PS @derrell with regards to regression testing: I completely agree; while I've had an earlier version of this in testing for months there were some bugs which cropped up (the fixes for which are included here).  I've added this PR into my production releases and it should start getting some real use starting in the next few days

framework/source/class/qx/event/Utils.js
derrell: Are you able to trap, somehow, when `catch` is used in the "wrong" place, i.e., at the end instead of before `when`? If so, I think it's fine to name it this way along with that comment. If not... hmmm... rename `catch` to be `catchMustBeBeforeWhen`? Sigh.

johnspackman: re trapping wrong catch: it's not easy because it can come down to the use case and it's preserving a synchronous model **until** promises are invoked and then it switches to asynchronous which turns the exception handling upside down.  I could trap that `catch` was called after the rejection, but that's not necessarily invalid - also, I'm cautious that we will start to "go down the rabbit hole" of creating some kind of well defined and generic AlmostButNotQuitePromise when what's really needed is just some helper methods that are abstract enough to enable unit testing.

I resisted calling the class `qx.event.FastPromise` or similar to avoid making it look like anything other than internal helper methods for specific tasks, we could go further with names like `qx.event.Utils.psuedoThen` and `qx.event.Utils.psuedoCatch`?  

But then how far do we go, given that it is impossible to create a fully promise compliant solution, especially if the code is more clearly marked as internal?

derrell: Document the crap out of it, and let it be, I guess.

application/uitests/source/class/uitests/DragAndDropApplication.js
derrell: I suppose. But for 6.0, there should be no uses of ES6 in the framework source code, and we should aggressively ensure that.

derrell (approved):

cboulanger (approved):

hkollmann (approved):
Sounds good

level420 (approved):

level420:
@hkollmann you've assigned this PR to yourself. Would you like to merge?

hkollmann:
There is missing one approval 😞

hkollmann:
@oetiker could you approve this ?

johnspackman:
@hkollmann @oetiker approved it, the only change was the changed return in commit https://github.com/qooxdoo/qooxdoo/pull/9491/commits/93461ed4327ec66e4abc2ed76cbda9ca74741248

level420:
@hkollmann I think you can merge

cboulanger:
Cool! Great work, @johnspackman!

PR #9492, opened 2018-02-15
oo loader template could fail to initialize userAgent

derrell:
In the React Native environment, window.navigator is defined, but
window.navigator.userAgent is not. This patch ensures that userAgent gets
set even if window.navigator was previously set. For consistency, it also
sets window.navigator.product and window.navigator.cpuClass if either is
unset, regardless of whether only those individual members were unset or if
window.navigator was originally unset.

### DIFF
--- a/tool/data/generator/oo.loader.tmpl.js
+++ b/tool/data/generator/oo.loader.tmpl.js
@@ -7,7 +7,7 @@ if (!window.navigator.userAgent) {
   window.navigator.userAgent = "Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_6_4; de-de) AppleWebKit/533.17.8 (KHTML, like Gecko) Version/5.0.1 Safari/533.17.8";
 }
 if (!window.navigator.product) window.navigator.product = "";
-if (!window.navigator.cpuClass) window.navigator.ecpuClass = "";
+if (!window.navigator.cpuClass) window.navigator.cpuClass = "";
 
 if (!window.qx) window.qx = {};
 
--- a/tool/data/generator/oo.loader.tmpl.js
+++ b/tool/data/generator/oo.loader.tmpl.js
@@ -2,11 +2,12 @@
 
 if (!this.window) window = this;
 
-if (!window.navigator) window.navigator = {
-  userAgent: "Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_6_4; de-de) AppleWebKit/533.17.8 (KHTML, like Gecko) Version/5.0.1 Safari/533.17.8", 
-  product: "", 
-  cpuClass: ""
-}; 
+if (!window.navigator) window.navigator = {};
+if (!window.navigator.userAgent) {
+  window.navigator.userAgent = "Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_6_4; de-de) AppleWebKit/533.17.8 (KHTML, like Gecko) Version/5.0.1 Safari/533.17.8";
+}
+if (!window.navigator.product) window.navigator.product = "";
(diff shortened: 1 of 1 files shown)


hkollmann (dismissed):

cboulanger (dismissed):

johnspackman (dismissed):

level420 (approved):

hkollmann (approved):

tool/data/generator/oo.loader.tmpl.js
```diff
@@ -2,11 +2,12 @@
...
 if (!this.window) window = this;
 
-if (!window.navigator) window.navigator = {
-  userAgent: "Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_6_4; de-de) AppleWebKit/533.17.8 (KHTML, like Gecko) Version/5.0.1 Safari/533.17.8", 
-  product: "", 
-  cpuClass: ""
-}; 
+if (!window.navigator) window.navigator = {};
+if (!window.navigator.userAgent) {
+  window.navigator.userAgent = "Mozilla/5.0 (Macintosh; U; Intel Mac OS X 10_6_4; de-de) AppleWebKit/533.17.8 (KHTML, like Gecko) Version/5.0.1 Safari/533.17.8";
```
oetiker: since this 'bug' is about react-native, how about actually detecting if we are running under react native?

others have run into this problem too it seems https://github.com/facebook/react-native/issues/1331

https://github.com/facebook/react-native/blob/da3424c929220304ff01032a9bf473fb251375f1/Libraries/Core/InitializeCore.js#L148

so with

`global.navigator.product === 'ReactNative'`

oetiker (changes requested):

tool/data/generator/oo.loader.tmpl.js
level420: @oetiker I think this PR is the right way for a generic initialization. From what I read under react `window.navigator.product` should have the value `ReactNative` as it will be set by react itself.
So this PR initializes all questionable attributes if not already set by an arbitrary runtime.

level420: Or do you mean that we should initialize `window.navigator.userAgent` with some value containing the string `ReactNative`?

derrell: It seems bad form to assume that if an object exists, all of its members that exist in some environments will always exist in others as well. I think the change here is actually the best way of ensuring, in the general case, that all expected members now exist, without making any assumptions about what existed initially. We could *then* check for specific environments and do environment-specific things if necessary. (I don't see any being necessary in this case, though.)

johnspackman: I'd argue that this (ie the current PR implementation) is a more flexible "feature detection"-style approach instead of detecting an actual platform

level420:
@oetiker could you please check if your concerns are still there

tool/data/generator/oo.loader.tmpl.js
cajus: `window && window.process && window.process.type` can detect if we're running in *electron*. Maybe adding some generic env info about 'where am I running in' is helpful.

oetiker: what irks me is that we set an actual user agent string which might cause confusion further down the line ...

oetiker: but yes this is not a new features ... so ... I guess the PR is ok

oetiker (approved):

level420:
5 approvals, more than 3 business days, merging.

level420:
thank you @derrell

PR #9493, opened 2018-02-19
Fixes #9488: Dispose the previous layout in qx.ui.form.List on orientation change

### LINKED ISSUE #9488: qx.ui.form.List._applyOrientation leaks previous layout instance
When changing the orientation of a list widget from horizontal to vertical or vice-versa:

https://github.com/qooxdoo/qooxdoo/blob/7fa76d1550a4186a6d8f3c54d50e961c34914bcb/framework/source/class/qx/ui/form/List.js#L272

the previous layout instance (qx.ui.layout.VBox or qx.ui.layout.HBox) is not disposed.

level420:
See https://github.com/qooxdoo/qooxdoo/issues/9488

### DIFF
--- a/framework/source/class/qx/ui/form/List.js
+++ b/framework/source/class/qx/ui/form/List.js
@@ -267,23 +267,27 @@ qx.Class.define("qx.ui.form.List",
     // property apply
     _applyOrientation : function(value, old)
     {
+      var content = this.__content;
+
+      // save old layout for disposal
+      var oldLayout = content.getLayout();
+      
       // Create new layout
       var horizontal = value === "horizontal";
       var layout = horizontal ? new qx.ui.layout.HBox() : new qx.ui.layout.VBox();
 
       // Configure content
-      var content = this.__content;
       content.setLayout(layout);
       content.setAllowGrowX(!horizontal);
       content.setAllowGrowY(horizontal);
 
       // Configure spacing
       this._applySpacing(this.getSpacing());
-    },
-
-    // property apply
-    _applySpacing : function(value, old) {
-      this.__content.getLayout().setSpacing(value);
+      
+      // dispose old layout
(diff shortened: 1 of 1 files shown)


oetiker:
I thought we do not have to dispose of things anymore unless they have created DOM elements which have to be removed...?

hkollmann (approved):

level420:
@oetiker When digging down you'll find in the end the following destructor in `qx.ui.layout.Abstract`: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/layout/Abstract.js#L257-L259 where we could end having a circular ref from the widget which sets the layout to the layout with holds the widget instance in `this.__widget`. This reference should be broken.
There is  similar code in the framework e.g. in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/container/SlideBar.js#L273-L315

oetiker:
@level420 ok!

oetiker (approved):

derrell (approved):

level420:
btw: the failing tests are unrelated.

johnspackman:
Even with circular references, this shouldn't be necessary on 6.x - the garbage collector will be able to handle that kind of reference and not cause a leak. There's certainly no harm in adding this in, especially if it's backported to 5.x (not sure we need to do that though)

johnspackman (approved):

hkollmann:
Sometimes a new start of the tests helps!

level420:
@hkollmann thank you for restarting the tests. I've done that 2 times before you've done it with no luck!

cboulanger (approved):

hkollmann:
5 approvals, 6 days, merged.
Thanks, @level420!

PR #9497, opened 2018-02-26
removes last instances of arguments.callee

johnspackman:


cboulanger (approved):

hkollmann (approved):

derrell (approved):

framework/source/class/qx/ui/table/model/Simple.js
```diff
@@ -62,8 +62,8 @@ qx.Class.define("qx.ui.table.model.Simple",
      */
     _defaultSortComparatorAscending : function(row1, row2)
     {
-      var obj1 = row1[arguments.callee.columnIndex];
-      var obj2 = row2[arguments.callee.columnIndex];
+      var obj1 = row1[qx.ui.table.model.Simple._defaultSortComparatorAscending.columnIndex];
+      var obj2 = row2[qx.ui.table.model.Simple._defaultSortComparatorAscending.columnIndex];
```
level420: @johnspackman maybe it's to obvious, but I can't find it: where get's the current column index 
 assigned to `qx.ui.table.model.Simple._defaultSortComparatorAscending.columnIndex`?
I'm having several custom sorting functions which are not defined as static members, but as normal members of classes with tables as children.

level420: I'm then using `qx.ui.table.model.Simple.setSortMethods` with the table model instance to assign an arbitrary sorting function for a specific column.

derrell: It's terrible code. (And unfortunately, possibly even mine, from 10 years ago.)

It's set on line 328, as `comparator.columnIndex = columnIndex;`. So it's setting a property of the function, which changes on each call. Yuck! I had to convince myself that it was actually uniquely setting this for each column. It turns out it is (I'm pretty sure), because it's doing so right in the `sortByColumn` function.

johnspackman: this PR is a bit of a quick-fix, i.e. if someone had the time and wanted to reimplement it I'd be happy to withdraw this PR

derrell: How about someone entirely reimplements Table...

level420: @derrell sure, the famous table rewrite ;-) But meanwhile: what about changing the implementation of `qx.ui.table.model.Simple.setSortMethods` by including the column index in a closure:
```javascript
    setSortMethods : function(columnIndex, compare)
    {
      var methods;
      if (qx.lang.Type.isFunction(compare))
      {
        methods =
          {
            ascending  : function(row1, row2)
            {
              return compare(row1, row2, columnIndex);
            },
            descending : function(row1, row2)
            {
              return compare(row2, row1, columnIndex);
            }
          };
      }
      else
      {
        methods =
          {
            ascending  : function(row1, row2)
            {
              return compare.ascending(row1, row2, columnIndex);
            },
            descending : function(row1, row2)
            {
              return compare.descending(row1, row2, columnIndex);
            }
          };
      }
      this.__sortMethods[columnIndex] = methods;
    },
```
and changing the signature of the sorting methods to:
```javascript
    _defaultSortComparatorAscending : function(row1, row2, columnIndex)
    {
      var obj1 = row1[columnIndex];
      var obj2 = row2[columnIndex];
      if (qx.lang.Type.isNumber(obj1) && qx.lang.Type.isNumber(obj2)) {
        var result = isNaN(obj1) ? isNaN(obj2) ?  0 : 1 : isNaN(obj2) ? -1 : null;
        if (result != null) {
          return result;
        }
      }
      return (obj1 > obj2) ? 1 : ((obj1 == obj2) ? 0 : -1);
    },
```

derrell:
Actually, I think `this.__sortColumn` was added for this purpose, and seems to be used in the Remote model... 

Can't add another argument to the sort functions, as that changes the API and everyone's own custom sort functions would have to change.

level420:
Yes we would break backwards compatibility. That's not good while it would be a clear and simple solution. I think I'll derive from `qx.ui.table.model.Simple` and use it this way.

Another idea would be to bind the sort functions to the model instance. that would allow to use `this` in the function. It would allow to call `this.getSortColumnIndex()`.

level420:
@derrell and please remember: compatibility breaks anyway with qxcompiler, as those `arguments.callee` calls do not work anymore in the sort functions.

level420:
one more thing: if we change the function signature from `sortFunction(row1, row2)` to `sortFunction(row1, row2, columnIndex)` sorting function which ignore the third parameter could use `arguments.callee.columnIndex` as they did before, as long as we do not change that ugly `comparator.columnIndex = columnIndex` assignment.

Or you can use `my.sorting.function.containing.class.myCustomSortingMethod.columnIndex`. It will work.

derrell:
Existing apps, not using qxcompiler, can continue to use arguments.callee. I would be highly opposed to changing the API here. It's simply not necessary. We can fix our own code (as has already been done), but shouldn't require changes to apps' code.

level420:
But foreign code will still work!

derrell:
I had thought about your option to bind the sort function to `this`, except then arguments.callee would no longer work, I think, because of the bind, thus breaking app code. `bind()` returns a new function. It could maybe be done in a backward-compatible fashion by doing the bind *first*, and *then* assigning `columnIndex` to the bound function. Would require some good testing to ensure it didn't break BC.

hkollmann (commented):
https://github.com/qooxdoo/qooxdoo/pull/9499 Is the cleaner solution!

johnspackman:
closing in favour of https://github.com/qooxdoo/qooxdoo/pull/9499

PR #9499, opened 2018-02-26
Remove arguments.callee in qx.ui.table.model.Simple and introduce a new third parameter columIndex for sort functions

level420:
This is an alternative implementation for @johnspackman s proposal in PR #9497 

It preserves compatibility but adds a new, more clear interface for sort methods.

### DIFF
--- a/framework/source/class/qx/test/ui/table/model/Simple.js
+++ b/framework/source/class/qx/test/ui/table/model/Simple.js
@@ -125,6 +125,72 @@ qx.Class.define("qx.test.ui.table.model.Simple",
       
       table.destroy();
       model.dispose();
+    },
+    
+    testCustomSortFunctionArgumentsCalleeColumnIndexDeprecated : function()
+    {
+      var stringValues = this.getStringValues();
+      var rowCount = 20;
+      
+      // table
+      var model = new qx.ui.table.model.Simple();
+      model.setColumns([ "ID", "String" ]);
+      model.setData(this.createRandomRows(rowCount));
+
+      // custom individual ascending descending integer sort function for column 0
+      model.setSortMethods(0, {
+        ascending : function(row1, row2) {
+          var columnIndex = arguments.callee.columnIndex;
+          var int1 = row1[columnIndex];
+          var int2 = row2[columnIndex];
+          return (int1 > int2) ? 1 : ((int1 == int2) ? 0 : -1);
+        },
+        descending : function(row2, row1) {
+          var columnIndex = arguments.callee.columnIndex;
+          var int1 = row1[columnIndex];
+          var int2 = row2[columnIndex];
+          return (int1 > int2) ? 1 : ((int1 == int2) ? 0 : -1);
+        } 
+      });
+      
+      // custom single string sort function for column 1
+      model.setSortMethods(1, function(row1, row2) {
+        var columnIndex = arguments.callee.columnIndex;
+        var string1 = row1[columnIndex];
+        var string2 = row2[columnIndex];
+        return string1.localeCompare(string2);
+      });
+      
+      var table = new qx.ui.table.Table(model);
+
+      // test sorting column 1
+      
+      // sort descending
+      model.sortByColumn(1);
+      var data = model.getData();
+      this.assertTrue(data[0][1] == stringValues[stringValues.length-1]);
+      this.assertTrue(data[data.length-1][1] == stringValues[0]);
+
+      // sort ascending
+      model.sortByColumn(1, true);
+      this.assertTrue(data[0][1] == stringValues[0]);
+      this.assertTrue(data[data.length-1][1] == stringValues[stringValues.length-1]);
+
+      
+      // test sorting column 0
+      
+      // sort descending
+      model.sortByColumn(0);
+      this.assertTrue(data[0][0] == rowCount);
+      this.assertTrue(data[data.length-1][0] == 1);
+
+      // sort ascending
+      model.sortByColumn(0, true);
+      this.assertTrue(data[0][0] == 1);
+      this.assertTrue(data[data.length-1][0] == rowCount);
+      
+      table.destroy();
+      model.dispose();
     }
 
   }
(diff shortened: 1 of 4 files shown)


derrell (changes requested):

johnspackman (dismissed):

framework/source/class/qx/test/ui/table/model/Simple.js
```diff
@@ -48,6 +48,84 @@ qx.Class.define("qx.test.ui.table.model.Simple",
...
+        descending : function(row2, row1, columnIndex) {
+          var int1 = row1[columnIndex];
+          var int2 = row2[columnIndex];
+          return (int1 > int2) ? 1 : ((int1 == int2) ? 0 : -1);
+        } 
+      });
+      
+      // custom single string sort function for column 1
+      model.setSortMethods(1, function(row1, row2, columnIndex) {
+        var string1 = row1[columnIndex];
```
derrell: It just occurred to me... maybe we should `assert(columnIndex === arguments.callee.columnIndex)` here, to prove backwards compatability?

derrell (changes requested):

framework/source/class/qx/test/ui/table/model/Simple.js
level420: You're right. But I think it is better to have a separate test for the `arguments.callee.columnIndex` way.

I've done that here locally and found that it fails for the case that a single comparator function is used and not a map having separate functions for ascending and descending sorting.

I think that this would also have failed for the version before my PR. Please have a look into the unchanged code at https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/table/model/Simple.js#L384-L387 
As you can see the ascending function gets wrapped into a function by changing the order of row1 and row2 in the parameters.
Later on when the sort is prepared, the columnIndex attribute to the function object is set here: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/table/model/Simple.js#L328

But this does not work for the formerly in `setSortMethods` wrapped descending function as within the innert function you have no access to `arguments.callee.columnIndex` as this belongs to the outer function.

A fix for the old version would be to set the column index for the inner function in the original version within the wrapper like this:
```javascript
        methods =
          {
            ascending  : compare,
            descending : function(row1, row2)
            {
              compare.columnIndex = arguments.callee.columnIndex;
              return compare(row2, row1);
            }
          };
```
or for the new version it would be:
```javascript
        methods =
          {
            ascending  : compare,
            descending : function(row1, row2, columnIndex)
            {
              compare.columnIndex = arguments.callee.columnIndex;
              return compare(row2, row1, columnIndex);
            }
          };
```
which would lead to having `arguments.callee` again in the framework code.

Or we can argue that having the non working version is backwards compatible to the old one.

derrell: What a mess. Yes, I see that it's been broken for years and no one has noticed it. Either they're always passing two functions (the better way to do it anyway), or they just haven't noticed.

I'm torn between "fix the bug in the old code" and "the old code was uncomprehendable and fixing the bug affects BC so maybe they should just use the new way if they are going to alter their code anyway"...

Since it's clearly a bug, I think we should fix it, as you've indicated here.

level420: Well the new method with the third parameter `columnIndex` fixes it already, but not for the backwards compatibility way via `arguments.callee.columnIndex`.
So you argue to add that line
```javascript
compare.columnIndex = arguments.callee.columnIndex;
```
to the wrapper?
But all this started as an attempt to remove all occurences of `arguments.callee` in the framework code.
I'm undecided what to do.

level420: Yes it works the old way if the sort function uses the `arguments.callee.columnIndex` method including having the same bug as it had before in the case of giving a single function to `setSortMethods`.
But it fixes the bug if the new method with the third parameter `columnIndex` is used in the sort functions for all methods either single sort function or map with two separate sort functions for ascending and descending sort.

derrell: Ok. I think if you add the test you wrote that shows that (possibly with a huge comment saying that we're retaining the existing bug), we'll be good to go.

level420: I think I'll add the line this way:
```javascript
// assure backwards compatibility for sort functions using
// arguments.callee.columnIndex and to fix a bug where
// this method did not work for the case where a single comparator
// function was used 
// See discussion in https://github.com/qooxdoo/qooxdoo/pull/9499#pullrequestreview-99655182
if(arguments.callee !== undefined) {
  compare.columnIndex = arguments.callee.columnIndex;
}

Comment (derrell) [reply framework/source/class/qx/test/ui/table/model/Simple.js]: Really good.

**Review by derrell — CHANGES_REQUESTED** [review]

File: framework/source/class/qx/test/ui/table/model/Simple.js (hunk 1 above)
Comment (johnspackman) [reply framework/source/class/qx/test/ui/table/model/Simple.js]: AIUI the old code is so broken that if someone was using it, then it would not work?  In which case, then surely BC is not an issue?

Also, you could check and emit a warning, eg 
```
if (qx.core.Environment.get("qx.debug") &&
    compare.toString().match(/arguments.callee.columnIndex/)) {
  this.debug("Warning:: .....");
}
```

level420: @johnspackman I've chosen to implement the solution from comment  https://github.com/qooxdoo/qooxdoo/pull/9499#discussion_r170922372 if we can live with that.

This fixes the bug with the `arguments.callee` way. If `arguments.callee` is not present the old way sort functions will break anyway be throwing errors on `arguments.callee.columnIndex` access.

level420:
@derrell test and fix added

derrell (approved):

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

cajus (approved):

cajus:
6 Votes. Merging. Thanks @level420!

petitben:
Just my 2 cents here but why adding a third parameter ?

We could simply transform the anonymous functions by named ones and call it directly in replacement of arguments.callee ?

OK, we then lose the IE8- compatibility (because of memory-management related issue) but... Who still cares about IE8- ?

level420:
@petitben please open a ticket for this and add some example implementation to it. I'm not sure if I understand your proposal right. Thank you.

PR #9505, opened 2018-03-09
Fix non working non strict mode compatibility fix in qx.ui.table.model.Simple

level420:
We don't need to take columnIndex from arguments.callee, as we get it anyway from wrapper around the sort function created in method sortByColumn, as the third parameter! It is enough for backwards compatibility to make arguments.callee.columnIndex available in the user supplied sort function.

### DIFF
--- a/framework/source/class/qx/ui/table/model/Simple.js
+++ b/framework/source/class/qx/ui/table/model/Simple.js
@@ -404,9 +404,8 @@ qx.Class.define("qx.ui.table.model.Simple",
                * See discussion in 
                * https://github.com/qooxdoo/qooxdoo/pull/9499#pullrequestreview-99655182
                */ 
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

derrell (commented):
Is this a replacement for the other PR, or in addition to? In the other PR, we also fixed the problem where the column index was not properly passed if automatic creation of the reverse sorting function was used.

hkollmann (approved):

hkollmann:
It's a fix for the fix because arguments.callee do not work in strict mode and it's uneeded.

derrell (approved):

johnspackman (approved):

cajus (approved):

oetiker (approved):

Issue #9506, opened 2018-03-15
Bug when breaking target property chain


RemiHeugue:
I've found a bug which happen sometimes when you break a target property chain.

For example if you do

```js
object1.bind("foo", object2, "foo.bar.baz");
```
with `object2.foo` not null and then set `object2.foo` to null, you get the error `GlobalError: Cannot read property 'constructor' of null`.

After some research, in certain cases, when setting a property to null, qooxdoo correctly remove the chain of listeners but then try to add them back on the new value, which is null in this case.

Playground example:
http://tinyurl.com/yb3c56hy

I have a fix that seems to work so I'll do a PR ;)

PR #9507, opened 2018-03-15
~ Fix Bug when breaking target property chain #9506

RemiHeugue:
Fix for #9506.

Break the loop if the property chain have been broken.

### DIFF
--- a/framework/source/class/qx/data/SingleValueBinding.js
+++ b/framework/source/class/qx/data/SingleValueBinding.js
@@ -495,6 +495,9 @@ qx.Class.define("qx.data.SingleValueBinding",
               target = target["get" + firstUpPropName]();
             }
             targets[j] = target;
+            if (!target) {
+              break;
+            }
 
             // check if a listener already created
             if (listeners[j] == null) {


framework/source/class/qx/data/SingleValueBinding.js
```diff
@@ -495,6 +495,9 @@ qx.Class.define("qx.data.SingleValueBinding",
               target = target["get" + firstUpPropName]();
             }
             targets[j] = target;
+            if (!target) {
+              break;
```
derrell: I'm not sure of the full context of what's going on here, but is `break` the appropriate thing to do, or is `continue`? It's inside a loop of various property names. Do we want, upon finding one of them to not exist, to stop handling the others?

johnspackman: Surely `break` is correct - this loop is re-establishing the listeners, going from left to right, so for `foo.bar.baz` and where `foo.bar` is null, you have to break because it makes no sense to subsequently try and bind to `.baz`

johnspackman (approved):

framework/source/class/qx/data/SingleValueBinding.js
derrell: Now I understand the context here, and yes, of course.

derrell (approved):

cboulanger (approved):

level420 (approved):

hkollmann (approved):

Issue #9429, opened 2017-11-27
Appearance property "base" works incorrect, if appearance from parent theme has string value.


mfedyashev:
For example, appearance for "datefield" from theme "qx.theme.modern.Appearance" has value:

> "datefield" : "combobox",

If create custom theme and extend "qx.theme.modern.Appearance", then in custom theme create apearance with the same name "datefield" and use property "base: true":

`"datefield": {
      base: true,
      style: function (states) {
          return {
              textColor: "red"
          };
      }
 }`

In this case the result appearance does not have properties from parent theme.

Qooxdoo Playground example: http://tinyurl.com/y8hj6b82

cajus:
Didn't you mean to change the textColor of *datefield/textfield*?

mfedyashev:
No, I didn't mean textColor. You can use empty style function, for example:
`style: function (states) { return { }; }`
I mean, that datefield appearance in custom theme does not inherite properties from parent theme.

cajus:
I can't find a problem in your description. What I ment with my comment was that there *is no textColor* on the level of *datefield* that is used somewhere, so it will not have any effect.

Maybe the problem is different: `base: true` does not work if the parent class has no own top level styling. In the modern theme, _datefield_ is just directly aliased to _combobox_. So you might need to add an _include: 'combobox'_ to your definition to make it work.

mfedyashev:
Problem is not about textColor, I know that textColor on the level datefield does not have any effect. It just was bad example.
Issue is the Date field with custom theme does not have border like the Date field with moderh theme.
So, as User, I set "base: true" and expect that custom datefield will include all property from Modern datefield. But, actually, it does not work.

Yes, _include: 'combobox'_ can help in this case, but, I think, it just workaround.

cajus:
It reacts like the comments in the appearance manager suggest. So that was done this way by intention of the original authors. I don't know why they did it this way.

mfedyashev:
The function _qx.theme.manager.Appearance.styleFrom_ uses common **this.__aliasMap** for all themes. For my custom theme datefield appearance is Entry, so when _styleFrom_ calculates this appearnce, item _"datefield": "datefield"_ is added to alias map. Then _styleFrom_ calculates datefield appearance from Modern theme, but it does not resolve new id, it just gets existing alias from map. But alias _"datefield": "datefield"_ is not correct for Modern theme. For Modern theme it should be **"datefield": "combobox"**

[...]

PR #9509, opened 2018-03-20
Appearance property "base" works incorrect, if appearance from parent theme has string value. #9429

ronkie:
https://github.com/qooxdoo/qooxdoo/issues/9429

We suggest to store aliases and calculated style caches per theme, so that when styleFrom is called for an appearance, it first picks up an id that this appearance has in a current theme, and therefore styles for it, and then, when base is encountered, it picks up styles from the parent theme, which are assigned through alias in that parent theme.

This way appearance datefield/textfield will pick up both styles defined in current theme, and also styles from parent theme that are specified in combobox/textfield, as it is aliased to that one.

### DIFF
--- a/framework/source/class/qx/theme/manager/Appearance.js
+++ b/framework/source/class/qx/theme/manager/Appearance.js
@@ -199,10 +199,7 @@ qx.Class.define("qx.theme.manager.Appearance",
 
       // Resolve ID
       var aliasMap = this.__aliasMap;
-      var resolved = aliasMap[id];
-      if (!resolved) {
-        resolved = aliasMap[id] = this.__resolveId(id, theme, defaultId);
-      }
+      var resolved = aliasMap[id] = this.__resolveId(id, theme, defaultId);
 
       // Query theme for ID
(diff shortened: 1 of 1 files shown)


cboulanger:
Are the failing checks related to the PR or due to external issues?

hkollmann:
seems to be an external problem:

Chrome 66.0.3359 (Windows 10.0.0) qx.test.bom.media Audio:test Play Event FAILED
	Asynchronous Test Error: Timeout reached before resume() was called.
	Chrome 66.0.3359 (Windows 10.0.0) qx.test.bom.media Video:test Play Event FAILED

hkollmann (approved):

hkollmann (approved):

cboulanger (approved):

level420 (approved):

derrell (approved):

level420:
plenty of time passed, 4 approvals, merging.
thank you @ronkie

level420:
@oetiker you added the voting label?

PR #9511, opened 2018-03-26
Merge property groups for appearance and decoration entries

Adolfrizer:
Our team found out property groups for appearance and decoration entries don't merge correctly.

For appearance entries, `styleFrom` method adds local style properties after all others. We can have, for example, 
```
{
padding: [1, 2, 3, 4],
paddingLeft: 5
}
```
as the result of this method. Qooxdoo applies this style properties in order of addition by operator `in` so if `padding: [1, 2, 3, 4]` property was moved from `base` or `include` entry and `paddingLeft: 5` was moved from local style the last one will be applied after the previous one. We think it's incorrectly to rely on it so we offers to expand group properties to merge they correctly. As the result we will have:
```
{
paddingTop: 1,
paddingRight: 2,
paddingBottom: 3,
paddingLeft: 5
}
```
instead of:
```
{
padding: [1, 2, 3, 4],
paddingLeft: 5
}
```

For decoration entries,` resolve` method adds local style properties _before_ all others. Then qooxdoo also applies this style properties in order of addition by operator `in`. So if we have `colorTop: red` property as local style property and `color: green` property as included property we will have
```
{
colorTop: red,
color: green
}
```
as the result and `color: green` property will be applied after `colorTop: red` property. We offers the same solution: group properties expansion. As the result we will have:
```
{
colorTop: red,
colorRight: green,
colorBottom: green,
colorLeft: green
}
```
For this we just store group members.

### DIFF
--- a/framework/source/class/qx/theme/manager/Appearance.js
+++ b/framework/source/class/qx/theme/manager/Appearance.js
@@ -286,38 +286,6 @@ qx.Class.define("qx.theme.manager.Appearance",
         // Create new map
         result = {};
 
-        // handler for group properties
-        var groupHandler = function(result, groupName, groupStyle) {
-          // there is a a group property "width" from qx.ui.decoration.Decorator
-          // and a single property "width" from qx.ui.core.LayoutItem
-          // so we can call this function for the single property because there are the same group property
-          if (groupName === "width") {
-            // just copy this like a single property
-            result[groupName] = qx.lang.Object.clone(groupStyle, true);
-            return;
-          }
-          var handledGroupStyle = qx.lang.Object.clone(groupStyle, true);
-          var groupConfig = qx.core.Property.$$groups[groupName];
-          var group = groupConfig.group;
-          // we need to get a four element list
-          if (handledGroupStyle !== undefined) {
-            handledGroupStyle = qx.lang.Array.fromShortHand(qx.lang.Type.isArray(handledGroupStyle) ? handledGroupStyle : [handledGroupStyle]);
-          } else {
-            handledGroupStyle = [];
-            for (var i = 0; i < group.length; i++) {
-              handledGroupStyle[i] = undefined;
-            }
-          }
-          // expand the group property (recursively) and copy it to result
-          for (var i = 0; i < group.length; i++) {
-            if (qx.core.Property.$$groups[group[i]]) {
-              groupHandler(result, group[i], handledGroupStyle[i]);
-            } else {
-              result[group[i]] = qx.lang.Object.clone(handledGroupStyle[i], true);
-            }
-          }
-        };
-
         // Copy base data, but exclude overwritten local and included stuff
         if (entry.base)
         {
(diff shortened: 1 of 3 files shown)


framework/source/class/qx/theme/manager/Appearance.js
```diff
@@ -286,6 +286,38 @@ qx.Class.define("qx.theme.manager.Appearance",
         // Create new map
         result = {};
 
+        // handler for group properties
```
johnspackman: This `groupHandler` function seems to be the same implementation as the one in Decoration.js#L228; please can you refactor this to a single function, possibly one that includes the `if (qx.core.Property.$$groups[baseIncludeKey]) {` bit as well?

johnspackman (changes requested):
Although there's a possibility that this might cause some backwards compatibility issues, I'd support @Adolfrizer's proposal because IMHO the current implementation is really counter intuitive and arguably a bug

framework/source/class/qx/theme/manager/Appearance.js
Adolfrizer: Done. Thank you!

Adolfrizer:
@level420 Yes, I've just found the test failure reason. `qx.theme.modern.Appearance` has an entry uses `padding` as an array:

```
"splitbutton/arrow" : {
  alias : "button",
  include : "button",

  style : function(states, superStyles) {
    return {
      icon : "decoration/arrows/down.png",
      padding : [superStyles.padding[0], superStyles.padding[1] - 6],
      marginLeft : 1
    };
  }
}
```
but we expanded this array to `paddingTop`, `paddingRight`, etc.

So, we understood our solution is bad :( We think we should just change the order in which decoration properties are added. We still think it's bad to rely on the operator `in` but we haven't any ideas yet.

Do you have some ideas, maybe?

derrell:
You are correct that it's really bad to depend on the order of `in` since that order is not defined by the language. My suggestion would be that we make explicit that an individual property takes precedence over that property in a group, so in your example:

```
{
padding: [1, 2, 3, 4],
paddingLeft: 5
}
```
The `paddingLeft` parameter would take precedence over the array element in `padding` for left. This means that the map should be scanned first, handling groups, then scanned again, handling non-groups. Although you can still end up with conflicts, at least the resolution is well-defined.

Adolfrizer:
@derrell Theoretically it's possible to have a situation when included style has, for example, the `paddingLeft` property, and local style has the `padding` array. In this case it's wrong to apply the previous one after the last one.

level420:
@Adolfrizer the problem is that properties are defined as object attributes, e.g. 
```javascript
{
  paddingLeft: 8,
  padding: [ 1 2 3 4]
}
```
and that the sequence given in the source file is not preserved by the interpreter. That means that iterating them with for-in or Object.keys etc may retrieve them in an unpredictable order. 

See https://stackoverflow.com/a/5525820 .

As a consequence the style definitions in an appearance can not act the same way as css where order is preserved.

PR #9512, opened 2018-03-26
Allow drop-downs to be wider than their parent

milandamen:
New property `allowGrowDropDown` in AbstractVirtualBox and VirtualDropDownList that enables the drop-down to become wider than their parent (_target), with a width based on the items in the model.

This feature is disabled by default.

### DIFF
--- a/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
+++ b/framework/source/class/qx/ui/form/core/VirtualDropDownList.js
@@ -129,7 +129,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
     __ignoreListSelection : false,
 
 
-    /** @type {qx.data.Array} The initial selection array. */
     __defaultSelection : null,
 
 
@@ -343,7 +342,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @protected
      */
     _onChangeModel : function(event) {
-      "use strict";
       if (this.getAllowGrowDropDown()) {
         this._recalculateMaxListItemWidth();
       }
@@ -361,7 +359,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @protected
      */
     _onChangeModelLength : function (event) {
-      "use strict";
       if (this.getAllowGrowDropDown()) {
         this._recalculateMaxListItemWidth();
       }
@@ -376,7 +373,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @param event {qx.event.type.Data} The change event.
      */
     _onChangeDelegate : function(event) {
-      "use strict";
       this.getSelection().removeAll();
     },
 
@@ -478,7 +474,6 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      */
     _adjustWidth : function()
     {
-      "use strict";
       var width = this._target.getBounds().width;
       var uiList = this.getChildControl('list');
       if (this.getAllowGrowDropDown()) {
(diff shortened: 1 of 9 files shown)


framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -484,6 +550,51 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
...
+      "use strict";
+      var maxWidth = 0;
+      var list = this.getChildControl("list");
+      var model = list.getModel();
+      if (model && model.length) {
+        model.forEach(function (item) {
+          var length;
+          if (typeof item === "string") {
+            // Approximately 7 pixels per character
+            length = item.length * 7;
```
oetiker: hmmm, that seems like a very bold assumption... I would rather like to see you measure the actual width of the string in the correct font.

oetiker (changes requested):

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: The problem is that because it is a virtual dropdown, the widgets get generated whenever they're needed; so I can't easily get the width. Is there a way to get the width of the string in a certain font?

level420: you should use `qx.bom.Label.getHtmlSize` for calculating the real width based on the styles of the content element. See also how it is done in https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/basic/Label.js#L382

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -407,28 +458,43 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
 
     /**
      * Adjust the drop-down to the available width and height, by calling
-     * {@link #__adjustWidth} and {@link #__adjustHeight}.
+     * {@link #_adjustWidth} and {@link #_adjustHeight}.
      */
-    __adjustSize : function()
+    _adjustSize : function()
     {
```
level420: Plese explain why you changes `__adjustSize`, `__adjustHeight` and `adjustWidth` to protected.

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
```diff
@@ -326,6 +376,7 @@ qx.Class.define("qx.ui.form.core.VirtualDropDownList",
      * @param event {qx.event.type.Data} The change event.
      */
     _onChangeDelegate : function(event) {
+      "use strict";
```
level420: Please do not change to strict mode in the framework. This may have implications which lead to failing code. For example I'm not really sure if this works when using the generator, as in strict mode `arguments.callee` is not available anymore. So someone deriving from this class and using the `this.base(arguments` syntax to call the superclass method will fail with generator, as `this.base` relies on `arguments.callee` being available.

level420 (changes requested):
In general: 
please don't use strict mode within the in the framework (yet).
only changes in the PR which are related to the subject
add some description on why the interface changes, in this case why some methods are changed from private to protected.
Thank you

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: This is only the case if you try to use `this.base` in a method that has strict mode; it doesn't matter if the superclass method has strict mode unless that superclass method also uses `this.base`. I will remove it for now.

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: So that you can adjust the size handling in a subclass; I thought it might be useful to someone, though I'm not taking advantage of this myself.

milandamen:
I've processed your comments.

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
level420: OK. Its just that we want to have the PRs only contain changes related to the subject/enhancement or the bug fix. Let's see what the other say.

level420 (approved):

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
milandamen: I see, thank you for your patience in explaining it to me.

derrell: I'm ok with the change to protected, if the ultimate commit message with PR merging mentions it.

johnspackman (approved):

derrell (approved):

framework/source/class/qx/ui/form/core/VirtualDropDownList.js
cboulanger: I don't mind either.

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

zaucker (approved):

PR #9518, opened 2018-04-02
Add properties from decoration entries in another order

Adolfrizer:
The problem is the same as in my https://github.com/qooxdoo/qooxdoo/pull/9511 PR.

For decoration entries, `resolve` method adds local style properties _before_ all others. Then qooxdoo applies this style properties in order of addition by operator `in`. So if we have `colorTop: red` property as local style property and `color: green` property as included property we will have
```
{
colorTop: red,
color: green
}
```
as the result and `color: green` property will be applied after `colorTop: red` property.

I changed the order in which decoration properties are added. At the beginning I create the map of included properties which is added recursively starting from the deepest `include` and then I copy it to result map. After this I add local properties to result map so we have the correct order of properties.

### DIFF
--- a/framework/source/class/qx/theme/manager/Decoration.js
+++ b/framework/source/class/qx/theme/manager/Decoration.js
@@ -215,22 +215,52 @@ qx.Class.define("qx.theme.manager.Decoration",
       }
 
       // check for inheritance
-      var currentEntry = entry;
-      while (currentEntry.include) {
-        currentEntry = theme.decorations[currentEntry.include];
-        // decoration key
-        if (!entry.decorator && currentEntry.decorator) {
-          entry.decorator = qx.lang.Object.clone(currentEntry.decorator);
+      if (entry.include) {
+        var included = {
+          style: {}
+        };
+
+        var handleIncludedEntry = function(entry) {
+          if (entry.include) {
+            handleIncludedEntry(entry.include);
+          } else {
+            var decorationEntry = theme.decorations[entry];
+            // decoration key
+            if (decorationEntry.decorator) {
+              included.decorator = qx.lang.Object.clone(decorationEntry.decorator);
+            }
+
+            // styles key
+            if (decorationEntry.style) {
+              for (var key in decorationEntry.style) {
+                included.style[key] = qx.lang.Object.clone(decorationEntry.style[key], true);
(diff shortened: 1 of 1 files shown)


level420 (approved):

hkollmann (approved):

framework/source/class/qx/theme/manager/Decoration.js
```diff
@@ -215,22 +215,52 @@ qx.Class.define("qx.theme.manager.Decoration",
...
+            }
+
+            // styles key
+            if (decorationEntry.style) {
+              for (var key in decorationEntry.style) {
+                included.style[key] = qx.lang.Object.clone(decorationEntry.style[key], true);
+              }
+            }
+          }
+        };
```
derrell: This all looks fine. For your consideration, though... You have the opportunity here -- and I don't see it being done -- to order each level's groups before non-groups, which would allow `{colorTop:"red", color:"green}"` to be handled in a sane fashion. If you were to handle the group first, and dictate the order rather than simply using a map, you'd end up with
```
[
  { colorTop : "green" }, 
  { colorBottom : "green" },
  { colorLeft : "green" },
  { colorRight : "green" },
  { colorTop : "red" }
]
```
which would overwrite `colorTop : "green"` with `colorTop : "red"` which seems like a sane and consistent ordering...

There's somewhat more work to do here. After creating the array, you'd be able to then iterate it and create the map for backward compatibility of the rest of the code that expects a map.

derrell (approved):
Approving as is, but see my comment with a recommended potential improvement.

cboulanger (approved):

framework/source/class/qx/theme/manager/Decoration.js
Adolfrizer: In my https://github.com/qooxdoo/qooxdoo/pull/9511 I've already tried to expand group properties and overwrite them with non-group ones but there are some parts of code that expect group properties and get their items by ``[i]`` so we need to save a set of properties we had before the merger. I think we can store this array as additional information for the result of merger and convert it to the map if we need to apply/set this result.

level420:
plenty of time passed, 4 approvals, merging.
thank you @Adolfrizer

PR #9519, opened 2018-04-03
Split the travis testrunner tests into 3 separate stages

level420:
Split the tests which were previously run in 1 stage into 3 different stages.

The split is done into

- `Chrome-unit-tests`: ChromeSL,ChromeBetaSL
- `Firefox-unit-tests`: FirefoxSL,FirefoxBetaSL
- `Edge-unit-tests`: EdgeSL

Additionally I've added the lint run and the coverage run into one stage called `Lint-and-coverage`

As a result we now have 4 stages which are run in sequence.

The benefit is that each stage takes from 10 to 15 min run time and with the hope that those stages do not fail as often as the one monolithic ~ 30 min run. And we are able to restart only that stage which fails which means that those 15 min stages rerun but not the whole 30 min monolithic stage.
Additionally the stages are now named by the tested browser platform which allows quicker identification of the failing platform.

The disadvantege is that we may wait ~ 10 min longer for travis to finish, but only for the seldom situation where the tests run completely at the first attempt which happened rarely recent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ jobs:
     - stage: Edge-Unit-Tests
       script: ../.travis/test EdgeSL
     - stage: Coverage
-    - script: ../.travis/coverage
+      script: ../.travis/coverage
 
     - stage: Deploy
       script: ../.travis/deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,27 +28,25 @@ install:
 - cd framework
 
 stages:
-  - name: Lint
+  - name: Lint-and-Coverage
   - name: Chrome-Unit-Tests
   - name: Firefox-Unit-Tests
-  - name: Edge-Unit-Tests
-  - name: Coverage
+  - name: Edge-Safari-Unit-Tests
   - name: Deploy
     if: repo = qooxdoo/qooxdoo AND (NOT type IN (pull_request)) AND ( branch = master OR tag =~ ^v[0-9] )
 
 jobs:
   include:
-    - stage: Lint
+    - stage: Lint-and-Coverage
       script: ../.travis/lint
+      script: ../.travis/coverage
 
     - stage: Chrome-Unit-Tests
       script: ../.travis/test ChromeSL,ChromeBetaSL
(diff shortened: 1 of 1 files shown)


level420:
@cajus It seems that the long running approx. 30min run is failing due to some general timeout. something like "no answer since xx min from yy". Would it harm how it is done in the PR (with my latest commit)?

level420:
Please note that @hkollmann restarted the complete travis checks for 4 other PRs so we have to wait a long time before the tests of this PR are run.

cajus:
These failures have nothing to do with travis AFAIK. Travis just starts the connection tunnel to Saucelabs. This may fail from time to time, because saucelabs may have a problem. And this will fail permanently after april 17th.

Restarting all PRs at the same time may lead to other problems: we've a limited concurrency on saucelabs. So if you start many of them, the concurrency is exhausted and the tests wait in the queue. This may lead to the timeouts, too - because the tests run some minutes.

level420:
I'm still interested in the results with the separated stages. Currently the lint job is running while the other jobs are waiting.

cajus:
You can just lower the concurrency to have the same effect. Moving stuff that logically belongs to one stage into seperate ones is a misuse of the concept IMHO ;-)

So the only difference will be that "if someone starts many PRs at the same time, it will fail later".

level420:
One difference of the different stages would be that we could, in case of a failing stage, rerun only that stage. From what I'm currently observing one unit test stage takes ~ 14min. The previous test monolith took ~ 30min, and has a bigger chance to fail during it's longer run time.

level420:
We have now a total run time for the checks of ~ 1 hour where we had 44 min before. Now trying to run lint and coverage within one stage and add safari to the edge state.

oetiker (approved):

hkollmann (approved):

johnspackman (approved):

cboulanger (approved):

derrell (approved):

cajus (approved):

zaucker (approved):

PR #9521, opened 2018-04-05
PLEASE DO NOT APROVE YET: attempt to speed up builds by caching the qooxdoo directory

level420:
cache qooxdoo directory to speed up builds


level420:
The first simple attempt had no influence on the build time needed. closing.

Issue #9524, opened 2018-04-20
regression: properties from included decorators do not get propagated


level420:
And it seems that the last merged PRs #9518 and/or #9509 introduced a regression:

I have the following decorator definition in my theme:

```javascript
"input-base" : {
  style : {
    radius: 2,
    color : "border-input",
    width: 1,
    innerWidth: 1,
    innerColor: "#f5f5f5"
  }
},

"input" : {
  include: "input-base",
  style : {
    gradientStart : ["#eaeaea", 0],
    gradientEnd : ["#fbfbfb", 100]
  }
},
```

What I suppose to happen is, that the deocorator `input` inherits `width` and `innerWidth` from `input-base`, but it does not happen.

@Adolfrizer and @ronkie could you please check if your PRs broke something? Thank you.

level420:
@Adolfrizer It's definitely https://github.com/qooxdoo/qooxdoo/pull/9518 I've backed out the commit here locally and inheritance works again.

I'm not sure, but this may be due to the shortcut `width: 1` where `width` is defined http://www.qooxdoo.org/devel/api/#qx.ui.decoration.MSingleBorder~setWidth!method_public as being an array, but it also seems to accept the shortcut with a single number `x` which presumably means `[x, x, x, x]`. But this has to be verified.

Adolfrizer:
I've just verified this decorator definition and got the correct result: _the decorator input inherits width and innerWidth from input-base_. I debugged the `resolve` method and took a look at properties by qooxdoo plugin for Chrome - inherited properties exist. 

And my changes (https://github.com/qooxdoo/qooxdoo/pull/9518), as far as I understand, were not related to the shortcut and group properties - I tried to expand them in the previous PR (https://github.com/qooxdoo/qooxdoo/pull/9511) which wasn't merged.

level420:
@Adolfrizer thank you for investigating. It remains the fact that if I revert commit d4d3a359c84cbe9a4262d64dab9a8a57ba4dc953 the decorator above works again with my theme. But maybe the patch unveils a bug in my theme?

I'll let this issue open until I find the time to debug my theme. Let's see if there are other people running into this issue.

BTW: what do you mean with "qooxdoo plugin for Chrome"? Sounds interesting.

Adolfrizer:
@level420 I can attach my test project if needed :)

This plugin
[...]

PR #9525, opened 2018-04-23
Fix for #9524: fix recursive included decorator style propagation

level420:
Replaced the faulty recursive decorator include algorithm.

See #9524.

### DIFF
--- a/framework/source/class/qx/theme/manager/Decoration.js
+++ b/framework/source/class/qx/theme/manager/Decoration.js
@@ -214,54 +214,28 @@ qx.Class.define("qx.theme.manager.Decoration",
         entry.style = {};
       }
 
-      // check for inheritance
-      if (entry.include) {
-        var included = {
-          style: {}
-        };
-
-        var handleIncludedEntry = function(entry) {
-          if (entry.include) {
-            handleIncludedEntry(entry.include);
-          } else {
-            var decorationEntry = theme.decorations[entry];
-            // decoration key
-            if (decorationEntry.decorator) {
-              included.decorator = qx.lang.Object.clone(decorationEntry.decorator);
-            }
-
-            // styles key
-            if (decorationEntry.style) {
-              for (var key in decorationEntry.style) {
-                included.style[key] = qx.lang.Object.clone(decorationEntry.style[key], true);
-              }
-            }
-          }
-        };
-
-        handleIncludedEntry(entry);
-
-        var result = {
-          style: {}
-        };
-
-        if (entry.decorator || included.decorator) {
-          result.decorator = qx.lang.Object.clone(entry.decorator || included.decorator);
+      // handle recursive decorator includes
+      var recurseEntryInclude = function(currentEntry) {
+        // follow the include chain to the topmost decorator entry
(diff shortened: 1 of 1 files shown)


cboulanger (dismissed):

hkollmann (dismissed):

framework/source/class/qx/theme/manager/Decoration.js
```diff
@@ -214,54 +214,28 @@ qx.Class.define("qx.theme.manager.Decoration",
...
-        if (entry.decorator || included.decorator) {
-          result.decorator = qx.lang.Object.clone(entry.decorator || included.decorator);
+      // handle recursive decorator includes
+      var recurseEntryInclude = function(currentEntry) {
+        // follow the include chain to the topmost decorator entry
+        if(currentEntry.include && theme.decorations[currentEntry.include]) {
+          recurseEntryInclude(theme.decorations[currentEntry.include]);
+        }
+        
+        // if the docorator is not set yet, clone the topmost
```
derrell: if the *decorator* is not set...

derrell (changes requested):
Very minor. Just a typo. Sorry.

oetiker (dismissed):

Adolfrizer:
You changed the order of addition properties so an operator `in` will iterate them in this order:
`local style (input-focused) -> input-base -> input`
but the correct order is:
`input-base -> input -> local style (input-focused)`
I added local style after all others

level420:
Please do not approve currently. Still in discussion. Thank you.

level420:
@Adolfrizer Please have again a look into the PR. I've fixed one existing problem and simplified the algorithm even more.

The "trick" is, that recursion starts with the current decorator definition, in the example `input-focused` in function `recurseDecoratorInclude` where, if an `include` property exists, it recursively calls the function with the `input` decorator which recursively calls the function with the `input-base` decorator.

As the recursion ends here, the styles get copied first over from `input-base` then from `input` and last from `input-focused`. This should be the right order, beginning from the topmost include down to the current decorator.

Additionally I've removed the unneeded clone of the `decoration` entry which seems to me to be a relic from ancient code, as not decorator entry has an attribute called `decorator`.

As only the `style` object/hash is used when creating the decorator, there is no need for `entry.styles`.

level420:
And also note that the PR will not include the debug message anymore. This is just to illustrate that copying the styles works in the right order. 

The `name` parameter of the function `recurseDecoratorInclude` could also be used to track if we have circular references in the decorators, thus things like `dec1.include = "dec2"` and `dec2.include = "dec1"`.

Adolfrizer (approved):
I think it's OK. I verified it locally

cboulanger (approved):

hkollmann (approved):

johnspackman (approved):

cajus (approved):

derrell (approved):

oetiker (approved):

johnspackman:
full approval, merging thank you @level420 !

hkollmann:
@level420 @Adolfrizer : Can we close this issue?

PR #9536, opened 2018-05-31
Fixed basic column model disposing cell editor factory when it shouldn't in setter

milandamen:
When setCellEditorFactory was called when the factory of the column was already set to a non-default factory, and when then that same factory was set on the column model again, the factory would be disposed.

This fix makes this setter more in-line with regular setters (in that the state doesn't change when setting the same value) and it fixes the dispose problem described above.


johnspackman (approved):

cboulanger (approved):

derrell (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

level420:
6 approvals, cajus in vacation, merging
thank you @milandamen

PR #9539, opened 2018-06-14
Add promisify; remove `this` from static methods

derrell:
- Add new static method qx.Promise.promisify
- Remove the use of `this` from all static methods in qx.Promise. The problem with `this` when used in static methods is that if the developer using a method does something like, 
```
var all = qx.Promise.all
all(xxx);
```
then `this` will have the wrong value, and the code will bork.

### DIFF
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -772,6 +772,24 @@ qx.Class.define("qx.Promise", {
     },
 
 
+    /**
+     * Returns a new function that wraps a function that is in node.js
+     * style. The resulting function returns a promise instead of taking a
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

johnspackman (approved):

level420 (approved):

cajus (approved):

oetiker (commented):
can you make the tests pass ?

oetiker (approved):

level420:
@oetiker the tests for firefox failed as usual on the first attempt. I've restarted the tests now. the failing tests are not related to the PR.

hkollmann (approved):

PR #9540, opened 2018-06-18
Added File to Xhr whitelist because File inherits from Blob

milandamen:
See https://developer.mozilla.org/en-US/docs/Web/API/File

I've tested this with Chrome 67.0.3396.87 and Firefox 60.0.2.

### DIFF
--- a/framework/source/class/qx/bom/request/Xhr.js
+++ b/framework/source/class/qx/bom/request/Xhr.js
@@ -881,7 +881,7 @@ qx.Bootstrap.define("qx.bom.request.Xhr",
       this.__disposed = this.__send = this.__abort = false;
 
       // Initialize data white list
-      this.__dataTypeWhiteList = [ "ArrayBuffer", "Blob", "HTMLDocument", "String", "FormData" ];
+      this.__dataTypeWhiteList = [ "ArrayBuffer", "Blob", "File", "HTMLDocument", "String", "FormData" ];
     },
 
     /**


oetiker (approved):
makes sense

hkollmann (approved):

level420 (commented):
The comment where `this.__dataTypeWhiteList` is evaluated states that it may be that `File` is omitted in order to get par with browser functionality. If I read/interprete the code correctly if `File` is not listed in the whitelist it gets converted to string for other browsers.
@milandamen could you please also test this for IE, Edge and Safari if this works there as expected?
See https://github.com/milandamen/qooxdoo/blob/b7730516f6fd26e0bf8cf1f61f17499200f26a24/framework/source/class/qx/bom/request/Xhr.js#L425-L431

level420 (changes requested):
See my last comment.

milandamen:
Sadly I don't have access to IE, Edge or Safari. The File API is supported by all major browsers, check out this page: https://developer.mozilla.org/en-US/docs/Web/API/File#Browser_compatibility

When `File` is not listed in the whitelist, the only data that gets sent is the literal `[object File]` as the request body content.

oetiker:
how about checking if it exists and only add it if it does ?

level420 (approved):
After looking a bit arround I'm changing my mind and approve this.

As the whitelist only contains strings, it should work automatically even on those browsers/clients not supporting `File` at all.

The type of object instance which ist tested against the whitelist is retreived via `qx.Bootstrap.getClass(someinstance)` which should always retreive the type of the given instance on the given client and which never should be "File" if there is no "File" support.

johnspackman (approved):

cboulanger (approved):

derrell (approved):

PR #9541, opened 2018-06-18
Improved jsDoc documentation

milandamen:


cboulanger (dismissed):

framework/source/class/qx/ui/core/MChildrenHandling.js
```diff
@@ -38,7 +38,7 @@ qx.Mixin.define("qx.ui.core.MChildrenHandling",
     /**
      * Returns the children list
      *
-     * @return {LayoutItem[]} The children array (Arrays are
+     * @return {qx.ui.core.LayoutItem[]} The children array (Arrays are
```
johnspackman: It's not necessary to provide the absolute path here, because by default the class name refers to the current package, so the original version is correct

johnspackman (changes requested):
Thanks for the changes, but I think it would be better to stick with the existing convention where package names can be resolved automatically if the class is in the same package.  It's used extensively and saves loads of unnecessary typing.

framework/source/class/qx/ui/core/MChildrenHandling.js
derrell: My opinion differs here. It is far easier to read a full class name, than to have to figure out what package you're in and append the truncated name. The API viewer may take the current package name as the default and show it as the full name, but viewing source code is far easier than using the API viewer, so making the source code easily readable should be a priority IMO.

oetiker: I thought that the convention was that if it is just the 'name' then it points to the fact that you will most likely get a class inheriting from this one ... in any event, I think it would be good to have a consistent convention an not just starting to change things at one end ...

milandamen: I did a search on the Qooxdoo codebase regarding the usage of (non) full-path class names. These are the results I got:

jsDoc tag | Full path occurrences | Non full-path occurrences
--- | --- | ---
@param | 1052 | 676
@return | 700 | 165

Grep commands used:
Param full path: `grep -r '@param.\+{qx.' . | wc -l`
Param non-full path: `grep -r '@param.\+{[^qx]' . | grep -vE '{Object|{Function|{String|{var|{Integer|{Number|{Map|{Boolean|{Array|{RegExp|{Class|{object|{string|{number|{integer|{boolean|{array|{map|{class|{function|{document|{Error|{arguments|{Float|{Color|{Date|{File|{IArguments|{Iterable|{null' | grep -vE '@link qx' | wc -l`
Return full path: `grep -r '@return.\+{qx.' . | wc -l`
Return non-full path: `grep -r '@return.\+{[^qx]' . | grep -vE '{Object|{Function|{String|{var|{Integer|{Number|{Map|{Boolean|{Array|{RegExp|{Class|{object|{string|{number|{integer|{boolean|{array|{map|{class|{function|{document|{Error|{arguments|{Float|{Color|{Date|{File|{IArguments|{Iterable|{null' | grep -vE '@link qx' | wc -l`

I agree with @derrell; in code its easier to read what class is meant when the class is written using the full package path. This means I don't have to click through on function jsDocs to figure out what class is used, I can just read it. I've also seem some cases where the referenced class is not in the same package as the current class.
Another advantage is that IDEs can do a better job of code completion.

johnspackman: Sounds like a job for a perl script :)  I don't have an objection to the change but IMHO if we're going to do this there's not a lot of point in doing it randomly in a few places

oetiker: I agree ... let's do it across the board

oetiker (dismissed):

framework/source/class/qx/dev/unit/TestResult.js
```diff
@@ -98,8 +98,8 @@ qx.Class.define("qx.dev.unit.TestResult",
     /**
      * Run a test function using a given test result
      *
-     * @param testResult {TestResult} The test result to use to run the test
-     * @param test {TestSuite|TestFunction} The test
+     * @param testResult {qx.dev.unit.TestResult} The test result to use to run the test
+     * @param test {qx.dev.unit.TestSuite|TestFunction} The test
```
derrell: This is missing the full class of TestFunction. It should say,
`qx.dev.unit.TestSuite|qx.dev.unit.TestFunction`.

framework/source/class/qx/dev/unit/TestResult.js
```diff
@@ -123,7 +123,7 @@ qx.Class.define("qx.dev.unit.TestResult",
     /**
      * Run the test
      *
-     * @param test {TestSuite|TestFunction} The test
+     * @param test {qx.dev.unit.TestSuite|TestFunction} The test
```
derrell: `qx.dev.unit.TestSuite|qx.dev.unit.TestFunction`

framework/source/class/qx/dev/unit/TestResult.js
```diff
@@ -318,7 +318,7 @@ qx.Class.define("qx.dev.unit.TestResult",
      *
      * @param eventName {String} Name of the event
      * @param exceptions {Error[]} The exception(s), which caused the test to fail
-     * @param test {TestSuite|TestFunction} The test
+     * @param test {qx.dev.unit.TestSuite|TestFunction} The test
```
derrell: `qx.dev.unit.TestSuite|qx.dev.unit.TestFunction`

derrell (changes requested):

level420 (approved):

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

derrell (approved):

cajus (approved):

PR #9542, opened 2018-06-19
Add more documentation (and an example) for promisify

derrell:
@hkollmann requested an example, provided herein. Also, I documented the method of providing a context to the callback function.

### DIFF
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -781,11 +781,51 @@ qx.Class.define("qx.Promise", {
      * expect its first argument to be an error if non-null. If the first
      * argument is null, the second argument (optional) will be the success
      * value.
-     * 
+     *
+     * Example:
+     *
+     * Assume there is a member method in myApp.Application such as the
+     * following:
+     * <pre><code>
+     *   issueRpc : function(method, params, callback)
+     *   {
+     *     ...
+     *   }
+     * </code></pre>
+     *
+     * where the signature of <code>callback</code> is:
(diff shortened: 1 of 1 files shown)


johnspackman (dismissed):

hkollmann (dismissed):
Thanks!

oetiker (dismissed):

framework/source/class/qx/Promise.js
```diff
@@ -781,11 +781,51 @@ qx.Class.define("qx.Promise", {
...
+     *
+     * where the signature of <code>callback</code> is:
+     * <pre><code>
+     *   function callback(e, result)
+     * </code></pre>
+     *
+     * The <code>issueRpc</code>method could be converted to be called using
+     * promises instead of callbacks, as shown here:
+     * <pre><code>
+     *   let app = qx.core.Init.getApplication();
```
level420: As this examples should also be valid when using the old generator and should work in non ES6 browsers, I think it would be better to use `var` here instead of `let`.

level420 (changes requested):

framework/source/class/qx/Promise.js
level420: And it may be that there are no arrow functions in the (still) supported browsers.

cboulanger (approved):

level420 (approved):

hkollmann (approved):

level420:
documentation only PR. 3 approvals and 3 business days.
merging.
thank you @derrell

PR #9547, opened 2018-07-18
[Fixed] trying to remove listener where given id is undefined

mever:
This PR fixes an exception where given argument to removeListenerById is undefined. I got this when trying to destroy a widget. With this fix all seems to work now.


oetiker (approved):

johnspackman:
I'd like to know more about exactly what the cause or your original issue was - when I've come across this myself, it's always as a side effect of another problem (often to do with lists and controllers), and triggered by recursive binding (eg part way through binding a widget, an event causes the model or selection of the original widget to change, causing the binding to be redone part way through).

IMHO there are definitely opportunities to improve error detection in bindings because when it goes wrong it is hard to track down what on earth is really going on, but this also makes me cautious about applying fixes which may be only masking the "real" problem.

mever:
You are totally right. I am debugging now.

mever:
On line 257 of SingleValueBinding.js there is a similar check. When do the lengths of  `sources` and `listenerIds` differ? When can I expect this to happen?

johnspackman:
Do you mean [here](https://github.com/mever/qooxdoo/blob/ac7a31c70ea6c6798a1b466a561d433153802cfc/framework/source/class/qx/data/SingleValueBinding.js#L255-L260) ?

That looks like a redundant check, otherwise it would be a symptom of a resource leak (ie an un-freed listener) because it is not possible to release a listenerId without the object to release it from.

mever:
Yes. Debugging the SingleValueBinding class is a real pain I must admit. My time is very limited so with some luck I can dig into it on Sunday  when I continue.

hkollmann:
@mever: Did you have found some time to continue this issue?

cajus (approved):

derrell (approved):

level420 (approved):

framework/source/class/qx/data/SingleValueBinding.js
```diff
@@ -1185,14 +1185,14 @@ qx.Class.define("qx.data.SingleValueBinding",
         // go threw all added listeners (source)
         for (var i = 0; i < id.sources.length; i++) {
           // check if a source is available
-          if (id.sources[i]) {
+          if (id.sources[i] && id.listenerIds[i]) {
```
level420: @mever it would maybe help to add a this.error in the situation where somehow `sources` gets "out of sync" from `listenerIds`. we hesitate to merge this as the situation could be caused by a bug somewhere else which is then silenced by this pr. A this.error would help here as it offers a stack trace.

level420 (changes requested):

cboulanger:
@mever could you update the PR as per @level420's suggestion? Thank you.

cboulanger (approved):

level420 (approved):

johnspackman (approved):

level420:
merging. thank you @mever for the initial PR which was hijacked by me ;-)

PR #9550, opened 2018-07-25
Adds support for a special "null" value in `qx.ui.form.SelectBox` and `qx.data.controller.List`

johnspackman:
`qx.ui.form.SelectBox` already supports empty selection, whether or not it is controlled by a `qx.data.controller.List` - however, an empty selection results in the UI displaying the first
 value in the drop down, so the user thinks that something is selected when there is none.  There is no easy way to add a corresponding "null" value when using a controller.

This PR adds properties to `qx.data.controller.List` to allow a special "null" value to be created at the top of the list; it also allows SelectBox to use rich appearance in its list items

### DIFF
--- a/framework/source/class/qx/data/controller/List.js
+++ b/framework/source/class/qx/data/controller/List.js
@@ -666,6 +666,7 @@ qx.Class.define("qx.data.controller.List",
      * @param index {Number} The index of the ListItem.
      */
     _bindListItem: function(item, index) {
+      // -1 is the special, "null" value item.  Nothing to bind, just fix the display and model
       if (index < 0) {
         item.setLabel(this.getNullValueTitle()||"");
         item.setIcon(this.getNullValueIcon());
@@ -1060,6 +1061,8 @@ qx.Class.define("qx.data.controller.List",
       }
 
       this.__lookupTable = [];
+      
+      // -1 is a special lookup value, to represent the "null" option 
       if (this.isAllowNull())
         this.__lookupTable.push(-1);
       for (var i = 0; i < model.getLength(); i++) {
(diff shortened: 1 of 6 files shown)


cboulanger (dismissed):
Very good addition, thank you!

framework/source/class/qx/data/controller/List.js
```diff
@@ -190,6 +190,42 @@ qx.Class.define("qx.data.controller.List",
...
+     * Whether a special "null" value is included in the list
+     */
+    allowNull :
+    {
+      apply: "_applyAllowNull",
+      event: "changeAllowNull",
+      init: false,
+      nullable: false,
+      check: "Boolean"
+    },
```
derrell: Since this adds nullValueTitle and nullValueIcon properties, could allowNull be elided, and references to it assume that a null value is allowed if either (or both) nullValueTitle or nullValueIcon is set?

framework/source/class/qx/data/controller/List.js
```diff
@@ -987,6 +1060,8 @@ qx.Class.define("qx.data.controller.List",
       }
 
       this.__lookupTable = [];
+      if (this.isAllowNull())
+        this.__lookupTable.push(-1);
```
derrell: -1 could probably use an explanatory comment

derrell (changes requested):

framework/source/class/qx/data/controller/List.js
johnspackman: I know what you mean, and I did consider that, but I like the ability to turn this on and off explicitly - icons are almost secondary and IMHO the explicit clarity is preferable.

framework/source/class/qx/data/controller/List.js
johnspackman: done

derrell (approved):
I'd prefer it the other way -- eliding allowNull -- but not with a strong enough basis to argue. Developer's preference rules.

oetiker (approved):

hkollmann (approved):

level420 (approved):

cajus (approved):

PR #9554, opened 2018-08-01
Theme prep function

sqville:
Needed a way to include code into framework classes (i.e. Image, Atom, etc.) in order to support advanced themeing and icon features such as before and after pseudo modifications, and utilizing qx.ui.basic.Image as a means of using embedded SVG for graphics and icons. This gives theme developers the ability to mixin code into framework classes without requiring end users doing it in their application code

### DIFF
--- a/framework/source/class/qx/Theme.js
+++ b/framework/source/class/qx/Theme.js
@@ -58,7 +58,7 @@ qx.Bootstrap.define("qx.Theme",
      *   widgets : {},
      *   appearances : {},
      *   meta : {},
-     *   boot : {function(){}}
+     *   boot : function(){}
      * });
      * </pre>
      *
--- a/framework/source/class/qx/Theme.js
+++ b/framework/source/class/qx/Theme.js
@@ -57,7 +57,8 @@ qx.Bootstrap.define("qx.Theme",
      *   fonts : {},
      *   widgets : {},
      *   appearances : {},
-     *   meta : {}
+     *   meta : {},
+     *   boot : {}
      * });
      * </pre>
      *
@@ -116,6 +117,11 @@ qx.Bootstrap.define("qx.Theme",
 
       for (var i=0, a=config.patch, l=a.length; i<l; i++) {
         this.patch(theme, a[i]);
+      }
+      
+      // Run boot code
+      if (config.boot) {
+      	config.boot();
       }
     },
 
@@ -321,7 +327,8 @@ qx.Bootstrap.define("qx.Theme",
         "appearances" : "object", // Map
         "meta"        : "object", // Map
         "include"     : "object", // Array
-        "patch"       : "object"  // Array
+        "patch"       : "object", // Array
+        "boot"		  : "function" // Function
       },
 
       "default" : null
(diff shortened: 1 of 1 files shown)


johnspackman (dismissed):

cboulanger (dismissed):
Sounds good. Could this help me with my dialog contrib's problem that the app author has to manually include the contrib's icon theme into the application theme (see https://github.com/cboulanger/qx-contrib-Dialog#install-icon-theme)?

framework/source/class/qx/Theme.js
```diff
@@ -57,7 +57,8 @@ qx.Bootstrap.define("qx.Theme",
      *   fonts : {},
      *   widgets : {},
      *   appearances : {},
-     *   meta : {}
+     *   meta : {},
+     *   boot : {}
```
derrell: It appears that this example is wrong, and doing as the example suggests would crash the program. The `boot` key is expected to point to a function, not a map as shown here.

framework/source/class/qx/Theme.js
```diff
@@ -321,7 +327,8 @@ qx.Bootstrap.define("qx.Theme",
         "appearances" : "object", // Map
         "meta"        : "object", // Map
         "include"     : "object", // Array
-        "patch"       : "object"  // Array
+        "patch"       : "object", // Array
+        "boot"		  : "function" // Function
```
derrell: Being anal retentive here... The colons all align, except for this added one. Since there's already another required change, let's align the colons.

derrell (changes requested):

oetiker:
How does this whole concept work in the light of on the fly theme switching ?

framework/source/class/qx/Theme.js
sqville: Any recommendations on how to best show? So far i have this: "boot : {function(){}}"

framework/source/class/qx/Theme.js
sqville: Yup. Agreed. I took out the tab I had in there and replace with all single spaces like the others

sqville:
Works with theme switching just fine. I tested on a separate, local copy of qx.5.0.2’s widgetbrowser where I added this to the qx.theme.Modern Theme definition file. The include code get’s executed and the theme works as designed.

sqville:
No. I hit this same issue with my FontAwesome icon font implementation. I had to instruct the end user to mix in my theme files into the current theme (see example below):

 

// Add FontAwesome specific Appearance, Decoration and Font entries to the current theme 

        qx.Theme.include(sqv.theme.clean.Appearance, sqv.fonticon.FontAwesome.Appearance);

        qx.Theme.include(sqv.theme.clean.Decoration, sqv.fonticon.FontAwesome.Decoration);

        qx.Theme.include(sqv.theme.clean.Font, sqv.fonticon.FontAwesome.Font);

 

Since dialog.Dialog is your key class you could add @require <your contrib theme classes> to the top, and qx.Theme.include() code to the class’ construct function. I’m essentially adding a “construct” function to my Theme. You have one built in as part of your Dialog project.

johnspackman (approved):

cboulanger (approved):

level420 (approved):

derrell (approved):

hkollmann (approved):

oetiker (approved):

PR #9555, opened 2018-08-02
Fixes #9551: blur and deactivate widget before activating the blocker

### LINKED ISSUE #9551: Blocker does not block active/focused widget when blocker is attached to Composite
qx.ui.core.Blocker does not block active/focused widget when blocker is attached to qx.ui.container.Composite which contains that widget.

**To Reproduce**
1. Go to playground example: http://tinyurl.com/y862cwg7 
2. Type something into text field, then Press Enter. It will activate the Blocker which is attached to Composite that contains text field
3. See that you can still type into the widget, which is not correct. If now to click outside the text field to make it loose focus, it will get blocked as expected.

**Expected behavior**
Text field should get blocked immediately, so user cannot type anything into there anymore.

**Desktop:**
 - OS: Mac OSX, Windows 10
 - Tested in latest Chrome

ronkie:
blur and deactivate widget before activating the blocker

### DIFF
--- a/framework/source/class/qx/ui/core/Blocker.js
+++ b/framework/source/class/qx/ui/core/Blocker.js
@@ -115,11 +115,11 @@ qx.Class.define("qx.ui.core.Blocker",
      * events, this is useful to block keyboard input on other widgets.
      * Take care that only one blocker instance will be kept active, otherwise your
      * browser will freeze.
-     * 
+     *
      * Setting this property to true is ignored, if the blocker is attached to a
      * widget with a focus handler, as this would mean that the focus handler
      * tries to activate the widget behind the blocker.
-     * 
+     *
      * fixes:
      *     https://github.com/qooxdoo/qooxdoo/issues/9449
      *     https://github.com/qooxdoo/qooxdoo/issues/8104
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/core/Blocker.js
```diff
@@ -263,13 +263,19 @@ qx.Class.define("qx.ui.core.Blocker",
...
+        activeWidget = qx.ui.core.Widget.getWidgetByElement(focusHandler.getActive()),
+        focusedWidget = qx.ui.core.Widget.getWidgetByElement(focusHandler.getFocus());
 
-      this.__activeElements.push(qx.ui.core.Widget.getWidgetByElement(focusHandler.getActive()));
-      this.__focusElements.push(qx.ui.core.Widget.getWidgetByElement(focusHandler.getFocus()));
+      this.__activeElements.push(activeWidget);
+      this.__focusElements.push(focusedWidget);
 
-      if (this._widget.isFocusable()) {
-        this._widget.focus();
```
oetiker: wouldn't this also automatically cause any other widget to blur?

ronkie: The problem is that when blocker is attached to a composite, this piece of code is not doing anything, because composite is not focusable.

hkollmann (commented):
could we have the var in single lines? IMHO is better readable.

var focusHandler = qx.event.Registration.getManager(window).getHandler(qx.event.handler.Focus);
var  activeWidget = qx.ui.core.Widget.getWidgetByElement(focusHandler.getActive());
var  focusedWidget = qx.ui.core.Widget.getWidgetByElement(focusHandler.getFocus());

oetiker (dismissed):

ronkie:
Put variables into single lines, as requested by @hkollmann

hkollmann (approved):

oetiker (approved):

derrell (approved):

cboulanger (approved):

level420 (approved):

cajus:
This PR breaks *readonly* Textfields in combination with Blockers.

PR #9557, opened 2018-08-08
Adds `qx.Promise.allOf`

johnspackman:
This PR adds `qx.Promise.allOf`, which is similar to `qx.Promise.all` except that `.allOf` transforms the property values of an object from promises into real values, whereas `.all` transforms the elements of an array (values which are not a promise are transferred unchanged)

This is useful in that it allows changing these kind of constructs:

```
let valueA;
let valueB;
myObj.doA()
  .then(tmp => {
    valueA = tmp;
    return myObj.doB();
  })
  .then(tmp => {
    valueB = tmp;
    return myOtherObj.doC();
  })
  .then(valueC => {
    console.log("a=" + valueA + ", b=" + valueB + ", c=" + valueC);
    /* ... do stuff ... */
  });
```

into this more concise form:

```
qx.Promise.allOf({
  valueA: myObj.doA(),
  valueB: myObj.doB(),
  valueC: myOtherObj.doC()
}).then(obj => {
  console.log("a=" + obj.valueA + ", b=" + obj.valueB + ", c=" + obj.valueC);
  /* ... do stuff ... */
});

``

### DIFF
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -536,8 +536,10 @@ qx.Class.define("qx.Promise", {
           }
         }
         return qx.Promise.all(arr)
-          .then(arr => {
-            arr.forEach((item, index) => value[names[index]] = item);
+          .then(function(arr) {
+            arr.forEach(function(item, index) { 
+              value[names[index]] = item; 
+            });
             return value;
           });
       }
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -520,7 +520,7 @@ qx.Class.define("qx.Promise", {
     /**
      * Returns a promise that resolves when all of the promises in the object properties have resolved, 
      * or rejects with the reason of the first passed promise that rejects.  The result of each property
-     * is placed back in the object, replacing the promise. 
+     * is placed back in the object, replacing the promise.  Note that non-promise values are untouched.
      * 
      * @param value {var} An object
      * @return {qx.Promise}
@@ -548,7 +548,9 @@ qx.Class.define("qx.Promise", {
     
     /**
      * Returns a promise that resolves when all of the promises in the iterable argument have resolved, 
-     * or rejects with the reason of the first passed promise that rejects.
+     * or rejects with the reason of the first passed promise that rejects.  Note that non-promise values 
+     * are untouched.
+     * 
      * @param iterable {Iterable} An iterable object, such as an Array
      * @return {qx.Promise}
      */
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: framework/source/class/qx/lang/Type.js [hunk 1]
```diff
@@ -64,6 +64,18 @@ qx.Bootstrap.define("qx.lang.Type",
...
+     
+     /**
+      * Whether the value is iterable, ie supports the iterable protocol
+      * 
+      * @signature function(value)
+      * @param value {var} Value to check
+      * @return {Boolean} Whether the value is iterable
+      */
+     isIterable : function(value) {
+       return value !== null && (qx.Bootstrap.isArray(value) || Symbol.iterator in Object(value));
```
cboulanger: `Symbol.iterator` might be a problem if we need to stay E5 compatible in the main framework code. I suppose the generator would not recognize it. Or have you tested?

cboulanger (dismissed):

derrell (dismissed):
Looks useful.

framework/source/class/qx/Promise.js
```diff
@@ -517,6 +517,35 @@ qx.Class.define("qx.Promise", {
...
+     */
+    allOf: function(value) {
+      function action(value) {
+        var arr = [];
+        var names = [];
+        for (var name in value) {
+          if (value.hasOwnProperty(name) && value[name] instanceof qx.Promise) {
+            arr.push(value[name]);
+            names.push(name);
+          }
```
oetiker: this will silently ignore stuff non-promises ... at least in debug mode, there might be a warning warranted ...

framework/source/class/qx/test/Promise.js
```diff
@@ -59,6 +59,33 @@ qx.Class.define("qx.test.Promise", {
...
 
+    /**
+     * Tests the qx.Promise.allOf method
+     */
+    testAllOf: function() {
+      var t = this;
+      var obj = {
+          a: new qx.Promise(),
+          b: new qx.Promise(),
+          c: new qx.Promise()
```
oetiker: you are not testing the case where no promise is provided in the input

oetiker (changes requested):

framework/source/class/qx/Promise.js
johnspackman: that's deliberate - the `qx.Promise.all` does the same, in that non-promises are not changed, the intention is that the property values (or in `.all`'s case, the array elements) are transformed so that there are no Promises left.

oetiker (dismissed):

framework/source/class/qx/Promise.js
oetiker: ok ... maybe add a note to the documentaion

oetiker (approved):

hkollmann (approved):

derrell (approved):

cboulanger (approved):

level420 (approved):

PR #9558, opened 2018-08-13
use other separator than " for classname because it breaks the generated

peuter:
code of properties added by mixins.

The Json marshaller uses `"` as separator for the property names in the classname, which generated classes with names like `qx.data.model.other"test`. If you include a Mixin with a property that uses checks. the code uses the classname for error messages, e.g. `var msg = "Invalid incoming value for property 'modified' of class 'qx.data.model.other"test'"` and this is no valid code.

See http://tinyurl.com/y8g8sq2j for a complete example.

This PR suggests `|`as separator, but any other would be fine too, if there are arguments against this one or for another one.

### DIFF
--- a/framework/source/class/qx/data/marshal/Json.js
+++ b/framework/source/class/qx/data/marshal/Json.js
@@ -62,6 +62,21 @@ qx.Class.define("qx.data.marshal.Json",
       this.$$instance.toClass(data, includeBubbleEvents);
       // return the model
       return this.$$instance.toModel(data);
+    },
+    
+    /**
+     * Legacy json hash method used as default in Qooxdoo < v6.0.
+     * You can go back to the old behaviour like this:
+     * 
+     * <code>
+     *  var marshaller = new qx.data.marshal.Json({
+     *   getJsonHash: qx.data.marshal.Json.legacyJsonHash
+     *  });
+     * </code>
+     */
+    legacyJsonHash: function (data, includeBubbleEvents) {
+      return Object.keys(data).sort().join('"')
+        + (includeBubbleEvents===true ? "♥" : "");
     }
   },
 
--- a/framework/source/class/qx/data/marshal/Json.js
+++ b/framework/source/class/qx/data/marshal/Json.js
@@ -83,6 +83,9 @@ qx.Class.define("qx.data.marshal.Json",
      */
     __jsonToHash : function (data, includeBubbleEvents)
     {
+      if (this.__delegate && this.__delegate.getJsonHash) {
+        return this.__delegate.getJsonHash(data, includeBubbleEvents);
+      }
       return Object.keys(data).sort().join('|')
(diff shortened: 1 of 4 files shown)


cboulanger:
This has always bothered me for aesthetic reasons so I'd be happy to change it. Since it is a purely internal convention that shouldn't be relied on by any end-user code, in theory, it should be possible to change it without side effects. However, since you cannot rule out that somebody is using some hack that relies on the class names, you should probably make this configurable, with a default of "|".

peuter:
We can just add something to `qx.data.marshal.IMarshalerDelegate` to make it overridable by the user and default ti the current PR`s solution. So everyone has the change to go back to the old version or implement something own.

framework/source/class/qx/data/marshal/Json.js
```diff
@@ -83,7 +83,10 @@ qx.Class.define("qx.data.marshal.Json",
      */
     __jsonToHash : function (data, includeBubbleEvents)
     {
-      return Object.keys(data).sort().join('"')
+      if (this.__delegate && this.__delegate.getJsonHash) {
+        return this.__delegate.getJsonHash(data, includeBubbleEvents);
+      }
+      return Object.keys(data).sort().join('|')
```
oetiker: Since the hash function gives complete freedom I would not change the current default as it may have side effects in applications where people have been looking at the generated strings in their application code ...

peuter: But the existing code leads to errors as shown in the example. Might not be relevant for existing user code, because they would have recognized the error, but for new ones. I consider this as a bug, which needs to be fixed. I wouldn't stick to backwards compability here, because it is a very rare use case. And users can use the delegate to go back to the old behaviour, if they really need to.

oetiker: true ... ok

oetiker (dismissed):

cboulanger (dismissed):

derrell (dismissed):
Approved for 6.0 release. Not approved for 5.x.

level420:
@peuter could you please have a look into the tests, as there are tests which explicitly check for " as a class separator. They need to be adapted and ideally extended to test the case where some other character is used as a class separator. Thank you.
Additionally your branch is out of date, but it seems that you could hit the "Update branch" button in this PR (above the "Squash and merge" button) to sync your branch.

peuter:
Tests are ok now!

oetiker (approved):

cboulanger (approved):

cajus (approved):

level420 (approved):

johnspackman (approved):

hkollmann (approved):

derrell (approved):
Approved for 6.0. Not appropriate for merging to 5.x, as it is a BC change.

PR #9563, opened 2018-08-24
travis improvements

oetiker:
* make travis run stage scripts in parallel
* add chrome test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,18 +39,25 @@ stages:
 jobs:
   include:
     - stage: Lint-and-Coverage
+      name: Lint
       script: ../.travis/lint
     - # parallel
-      script: ../.travis/coverage
+      name: Coverage
+      script: ../.travis/coverage      
     - stage: Local-Browser-Tests
+      name: Firefox
       script: ../.travis/test Firefox
     - # parallel
+      name: Chrome
       script: ../.travis/test Chrome_travis_ci      
     - stage: SL-Brower-Tests
+      name: SL Chrome
       script: ../.travis/test ChromeSL,ChromeBetaSL
     - #
+      name: SL Firefox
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -44,14 +40,17 @@ jobs:
...
     - stage: Lint-and-Coverage
       script: ../.travis/lint
+    - # parallel
       script: ../.travis/coverage
-    - stage: Firefox-Unit-Tests-Local
+    - stage: Local-Browser-Tests
       script: ../.travis/test Firefox
-    - stage: Chrome-Unit-Tests
+    - # parallel
+      script: ../.travis/test Chrome_travis_ci
```
level420: The reason why (it was me) the tests were run in stages one after the other was that the tests sometime fail with timeouts caused by slow running containers at travis or saucelab. If the first test in firefox succeeded and the chrome test failed afterwards it was sufficent to restrat the chrome test via the travis web interface until it succeeded. If you are running firefox and chrome in parallel the probability that one of the test fails and you have to rerun the stage with both tests again is higher. The drawback is that the test will take a longer time if the tests run sequentially in stages.

level420: I also suspect that the travis test have cpu quota which potentially lead to container run interruptions or lower cpu priority if you run tests in parallel, at least if you are a non paying customer.

level420 (dismissed):

level420 (changes requested):
See comments above

.travis.yml
oetiker: you can restart the individual jobs within a stage

oetiker:
these tests also 'enable' the lint tests ... in the old setup they were in .travis but since you can not have 2 script keys in the same hash they were inactive!

level420 (approved):

oetiker:
merging this to get the CI back into play

Issue #9569, opened 2018-09-10
new `null` value feature in list controller does not work with modelSelection


peuter:
unfortunately the playground isn't up-to-date, so I can only provide some "pseudo"-code to describe the problem:

If you select an "non-null" item in the selectbox, the model selection changes. If you select the "null"-value again, the model selection does not change.

The problem is this line: https://github.com/qooxdoo/qooxdoo/blob/97418ba9a62fed5eb00efbbac5190fdde66ed24b/framework/source/class/qx/ui/form/MModelSelection.js#L77

because the "null" item has not model, and is not added to the modelSelection. Then you have `modelSelection.length = 0`and `data.length = 1` and nothing happens, although the model selection should be resetted.

@johnspackman Do you have any suggestion how to solve the problem.

The code above would show that there is no model selection change in case of re-selecting the "null"-item.

```
// Create a button
var model = new qx.data.Array();
model.push(qx.data.marshal.Json.createModel({name: 'Test 1', id: 1}))
model.push(qx.data.marshal.Json.createModel({name: 'Test 2', id: 2}))
var box = new qx.ui.form.SelectBox();
var controller = new qx.data.controller.List();
controller.set({
  allowNull: true,
  nullValueTitle: '-'
})

// Document is the application root
var doc = this.getRoot();

// Add button to document at fixed coordinates
doc.add(box,
{
  left : 100,
  top  : 50
});

// Add an event listener
box.getModelSelection().addListener("change", function(e) {
  console.log('model selection changed:', e.getData())
});
```

level420:
@peuter playground should be (nearly) up to date: http://www.qooxdoo.org/devel/playground/ based on this commit: https://github.com/qooxdoo/qooxdoo/commit/6606c75

It seems that the last merge did not push a new playground version, as it had a timeout in the firefox tests. I'm investigating and be back.

johnspackman:
@peuter please can you try that PR in your code?  the file https://github.com/johnspackman/qooxdoo/blob/fix-list-controller-null-selection/uitests/selectbox/source/class/selectbox/Issue9569Application.js has your sample code from above

peuter:
@johnspackman Your fix works for me.

PR #9570, opened 2018-09-10
Fixes issue #9569 where `modelSelection` does not work properly with `qx.data.controller.List.allowNull`

johnspackman:
Note that the fix for bug #3852 conflicted with the fix for this bug, so there's a little more code changes here than might be expected.  The UI test incorporates tests against regression of #3852 as well as fix for this issue


cboulanger (dismissed):

derrell (dismissed):

hkollmann:
Could you fix this:  - Warning: qx.data.controller.MSelection (223, 11): Loop or condition statement without a block as Body

Thanks, Henner

hkollmann (approved):

cajus (approved):

derrell (approved):

oetiker (approved):

level420 (approved):

PR #9571, opened 2018-09-11
double timeout delays

oetiker:


cboulanger (approved):

oetiker:
merging now to see if the higher timeout fixes the saucelab tests

level420:
@oetiker the failing travis jobs in saucelabs for chrome are all like "server does not respond" which means that the travis container serving the content did not respond I suspect:
```
11 09 2018 08:31:42.127:ERROR [launcher.sauce]:Heartbeat to chrome latest (Windows 10) failed
  [title()] Error response status: 13, , UnknownError - An unknown server-side error occurred while processing the command. Selenium error: Invalid message: ERROR Internal Server Error
```

PR #9573, opened 2018-09-11
make iconfonts work with latest qxcompiler in --target=build mode

oetiker:
It seems that something in the qxcompiler dependencies (babel?) has changed so that the large
`if else` structure in `Image.js` gets damaged in the compilation output when running in build mode. Switching to a `switch case` structure I was able to make icon fonts work again. 

As the code is much nicer like this anyway, I am submitting a patch. But as the `if else` structures do show up elsewhere in the code, further investigation into why this breaks would certainly be warranted. https://github.com/qooxdoo/qooxdoo-compiler/issues/207

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -396,24 +396,21 @@ qx.Class.define("qx.ui.basic.Image",
       var tagName;
       var clazz = qx.html.Image;
 
-      if (mode == "font")
-      {
+      switch (mode) {
+        case "font":
         clazz = qx.html.Label;
         scale = true;
         tagName = "div";
-      }
-      else if (mode == "alphaScaled")
-      {
+          break;
+        case "alphaScaled":
         scale = true;
         tagName = "div";
-      }
-      else if (mode == "nonScaled")
-      {
+          break;
+        case "nonScaled":
         scale = false;
         tagName = "div";
-      }
-      else
-      {
+          break;
+        default:
         scale = true;
         tagName = "img";
       }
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/basic/Image.js
```diff
@@ -396,26 +396,23 @@ qx.Class.define("qx.ui.basic.Image",
...
+          scale = true;
+          tagName = "div";
+          break;
+        case "nonScaled":
+          scale = false;
+          tagName = "div";
+          break;
+        default:
+          scale = true;
+          tagName = "img";
```
derrell: Even though not strictly required here, the default block should have a trailing `break;` statement.

derrell (changes requested):

cboulanger (dismissed):

derrell (approved):

hkollmann (approved):

level420 (approved):

cboulanger (approved):

cajus (approved):

johnspackman (approved):

johnspackman (changes requested):
On second thoughts, this should not be approved just for the sake of a bug in qxcompiler (especially as I think that may be fixed now).  Are there reasons *other than* https://github.com/qooxdoo/qooxdoo-compiler/issues/207 ?

derrell:
Although certainly subjective, the `switch` statement feels to me like a cleaner implementation than a long series of `if/else` statements...

johnspackman:
> Although certainly subjective, the `switch` statement feels to me like a cleaner implementation than a long series of `if/else` statements...

Fair enough, i'll remove my request for change

johnspackman (approved):

PR #9577, opened 2018-10-10
make sure theme colors get applied to iconFonts

oetiker:
The problem is that when the textcolor would get applied, it does not as at that time the source property is not yet set for themed fontIcons.

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -510,6 +510,10 @@ qx.Class.define("qx.ui.basic.Image",
       // Detect if the image registry knows this image
       if (ResourceManager.isFontUri(source)) {
         this.__setManagedImage(contentEl, source);
+        var color = this.getTextColor();
+        if (qx.lang.Type.isString(color)) {
+          this._applyTextColor(color, null);
+        }
       }
       else if (ResourceManager.has(source)) {
         var highResolutionSource = this._findHighResolutionSource(source);


johnspackman (approved):

framework/source/class/qx/ui/basic/Image.js
```diff
@@ -508,6 +508,10 @@ qx.Class.define("qx.ui.basic.Image",
       // Detect if the image registry knows this image
       if (ResourceManager.isFontUri(source)) {
         this.__setManagedImage(contentEl, source);
+        var color = this.getTextColor();
+        if (qx.lang.Type.isString(color)) {
+          this._applyTextColor(color, null);
+        }
```
derrell: Is that the only relevant property that hasn't yet been set? It seems that if text color hasn't been set, it's likely that other properties too have not been set...

oetiker: The problem is only with textColor as the _applyTextColor method checks if we are working with a fontIcon ... but that test fails at creation time as the source gets set later than the textColor when theming ... or rather randomly ... I think we had that problem before ... that properties depend on each other and thus the orther they get applied matters ... this patch works around this by just re-applying the text color once the styling call comes

hkollmann (approved):

derrell (approved):

cboulanger (approved):

level420 (approved):

PR #9578, opened 2018-10-15
use promises only if qx.promise is true

### LINKED ISSUE #9576: "qx.promise" Environment setting has no effect
the "qx.promise" flag has been introduced to be able to remove the complete `qx.Promise`-Class and its dependencies to save some KB in the build size.

This does not work anymore. At least the DynamicScriptLoader uses `qx.Promise` without checking if its enabled: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/util/DynamicScriptLoader.js#L149

There might be other places where the "qx.promise"-flag should be checked.

peuter:
closes #9576

### DIFF
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -143,7 +143,7 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
 
     /**
      * Start loading scripts. This may only be called once!
-     * @return {Promise?} a promise which will be resolved after load of all scripts if promises are enabled.
+     * @return {Promise?} a promise which will be resolved after load of all scripts if promise support is enabled; nothing (undefined) if promises are not enabled.
      */
     start: qx.core.Environment.select('qx.promise', {
       "true": function() {
--- a/framework/source/class/qx/util/DynamicScriptLoader.js
+++ b/framework/source/class/qx/util/DynamicScriptLoader.js
@@ -143,24 +143,36 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
 
     /**
      * Start loading scripts. This may only be called once!
-     * @return {Promise} a promise which will be resolved after load of all scripts.
+     * @return {Promise|null} a promise which will be resolved after load of all scripts if promises are enabled or null if not.
      */
-    start: function() {
-      return new qx.Promise(function(resolve, reject) {
-        this.addListenerOnce("ready", resolve, this);
-        this.addListenerOnce("failed", function(e) {
-          reject(new Error(e.getData()));
+    start: qx.core.Environment.select('qx.promise', {
+      "true": function() {
+        return new qx.Promise(function (resolve, reject) {
+          this.addListenerOnce("ready", resolve, this);
+          this.addListenerOnce("failed", function (e) {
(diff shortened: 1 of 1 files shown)


oetiker (dismissed):

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -143,24 +143,36 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
 
     /**
      * Start loading scripts. This may only be called once!
-     * @return {Promise} a promise which will be resolved after load of all scripts.
+     * @return {Promise|null} a promise which will be resolved after load of all scripts if promises are enabled or null if not.
```
derrell: It appears this documentation is incorrect. I don't see anyplace that returns `null`; rather, line 167 simply does `return;` so the return value is `undefined`, not `null`. Either the documentation here should be corrected, or line 167 should be changed to `return null;`.

derrell (changes requested):
see embedded comments

derrell (dismissed):

johnspackman (dismissed):

framework/source/class/qx/util/DynamicScriptLoader.js
```diff
@@ -143,24 +143,36 @@ qx.Class.define("qx.util.DynamicScriptLoader", {
 
     /**
      * Start loading scripts. This may only be called once!
-     * @return {Promise} a promise which will be resolved after load of all scripts.
+     * @return {Promise?} a promise which will be resolved after load of all scripts if promises are enabled.
```
cboulanger: Shouldn't it document what it returns if promises are NOT enabled? Sp that the API consumer knows how to code against this?

peuter: But it doesn't return anything if promises are NOT enabled. It only returns something (a promise) if promises are enabled and that is documented.

derrell: Since it was unclear to one person, it may be unclear to others, too. Maybe change the documentation for that return value to,
```
a promise which will be resolved after load of all scripts if promises are enabled, if promise support is enabled; nothing (undefined) if promises are not enabled.
```

hkollmann (dismissed):

framework/source/class/qx/util/DynamicScriptLoader.js
cboulanger: I don't insist - it wasn't clear to me that the JSDOC "{Promise?}" means "{Promise|undefined}" but if it does, that should be enough.

cboulanger (dismissed):

framework/source/class/qx/util/DynamicScriptLoader.js
peuter: Well I think @derrel is right: it is better to mention the `undefined` return explicitly.

cboulanger (approved):

johnspackman (approved):

cajus (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

derrell (approved):

PR #9579, opened 2018-10-16
fix blob requests

peuter:
Handling ob blob requests does not work in qooxdoo, because the response can be found in transports response property, responseText is empty in these cases.

Blobs can be requested by manually setting the transports responseType:

```javascript
var req = new qx.io.request.Xhr(image);
req.getTransport().responseType = 'blob';
```

### DIFF
--- a/framework/source/class/qx/io/request/Xhr.js
+++ b/framework/source/class/qx/io/request/Xhr.js
@@ -300,7 +300,7 @@ qx.Class.define("qx.io.request.Xhr",
      * @return {String|Object} The parsed response of the request.
      */
     _getParsedResponse: function() {
-      var response = this._transport.responseText,
+      var response = this._transport.responseType === 'blob' ? this._transport.response : this._transport.responseText,
           contentType = this.getResponseContentType() || "",
           parsedResponse = "";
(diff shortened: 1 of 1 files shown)


cboulanger (commented):
Using a simple property feels a bit non-qooxdoo-ish to me. Wouldn't it be more in line with the qooxdoo way to introduce a managed property (even if a simple property does the job just fine)?

peuter:
That would imply a complete refactoring of `qx.bom.request.Xhr` (https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/request/Xhr.js#L192). That would be an option for another PR and not part of a bugfix.

cboulanger:
@peuter Forget my comment, I didn't pay attention to the fact that the transport  uses simple properties. Sorry for the noise!

cboulanger (approved):

hkollmann (approved):

level420 (approved):

cajus (approved):

oetiker (approved):

derrell (approved):

PR #9581, opened 2018-10-22
Fixed table removeSelectionInterval not correcting selection when removing rows

milandamen:
**Problem:**
When removing rows in the table, the selection would not update correctly.

**Reproduction:**
1. Have a table with 3 rows.
2. Select all rows (selection: `{minIndex: 0, maxIndex: 2}`)
3. Remove the first row of the table
4. Get the selected ranges: `table.getSelectionModel().getSelectedRanges()` => `{minIndex: 2, maxIndex: 2}`

Even though there are now only 2 rows instead of 3, the selection still points to the third row.
The selection model has no knowledge of changing rows.

**Solution:**
Make the selection model aware if a selection is removed because of the removal of one or more rows. If this is the case, calculate the selection correctly; moving/cropping selections where needed.

### DIFF
--- a/framework/source/class/qx/test/ui/table/selection/Model.js
+++ b/framework/source/class/qx/test/ui/table/selection/Model.js
@@ -1,0 +1,111 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2004-2018 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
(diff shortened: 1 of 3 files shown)


level420 (approved):

oetiker (approved):

derrell (approved):

hkollmann (approved):

cboulanger (approved):

johnspackman (approved):

level420:
Plenty of time passed. 6 approvals. merging.
Thank you @milandamen

PR #9585, opened 2018-10-26
the exclude of "qx.event.GlobalError" leads to errors during testing …

hkollmann:
…with current master. (generate npm-package-test).

The method qx.event.GlobalError.observeMethod is needed in different places.
So i removed the exclude


level420 (approved):
if this is needed then we need it ;-)

johnspackman (approved):

johnspackman (changes requested):
Have you checked what including GlobalError does to the dependency graph?  I have been using server for many years now and not needed this - I'm also using Rhino, which has a 64Kb limit on compiled code, so unintentionally including lots more files is a big problem (eg if I accidentally cause a reference to a `qx.ui.*` class, that's catastrophic)

hkollmann:
This Change is only needed in the current version. In 5.x branch testing is working without errors. Simce that we have changes in the Bootloader code?

I compared the build with global error handler and without. The difference are 2 more classes in build:
"qx.event.GlobalError" and "qx.core.GlobalError".

qx.event.Global error is referenced in different places of qx-oo, e.g. qx.util.DeferredCallManager, __timeout:     __timeout : qx.event.GlobalError.observeMethod(function().

This is the place where testing result's in an error during bootstrap. The other places will get an error if some exceptio happen.
[script.zip](https://github.com/qooxdoo/qooxdoo/files/2526647/script.zip)

I append both version as zip so that you can have a look.

cboulanger (approved):

johnspackman (approved):

oetiker (approved):

derrell (approved):

level420:
lot of time passed. 5 approvals. merging.
thank you @hkollmann

PR #9588, opened 2018-11-05
Fixed setValue of VirtualSelectBox not resetting selected item

milandamen:
**Problem**
When you call setValue(null) on the VirtualSelectBox, it does not reset its selection (and thus not resets value) when using the initial initSelection array.

The form resetter calls setValue(null) on the field (if the initial value is null).

**Reproduction**
1. Make a VirtualSelectBox with items set on its model
2. Select a random item
3. Call setValue(null) on the selectbox
4. The selection is still on the same item

**Solution**
Let the setValue(null) act as resetter of the field, to be consistent with the other form fields.

### DIFF
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -121,7 +121,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
      */
     setValue : function(selected) {
       if (null === selected) {
-        this.resetSelection();
+        this.getSelection().removeAll();
         return null;
       }
 
@@ -140,7 +140,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
 
 
     resetValue : function() {
-      this.getSelection().removeAll();
+      this.setValue(null);
     },
 
 


hkollmann (approved):

level420 (approved):

cboulanger (approved):

oetiker:
hmmm somehow I have a bad feeling with this ... since a selectbox without selection should return an empty array and not null ... null could be a valid selection ...

milandamen:
Well it all depends on what you define your "null-value" of the selectbox to be as.
In a selectbox you can only select one item, so it would be logical (to me) that the "value" of the selectbox would be either the item you have selected or the first item in your model if the selection is reset.

Remember, if the selectbox has items in its model, the selectbox *always* has a non-null value (unless that has changed recently). A selectbox without a model will return null.

If you `this.getSelection().removeAll()` it actually does not empty out the selection, it selects the first item in your model because a selectbox uses the always-one selection mode.

johnspackman:
what does `qx.ui.form.SelectBox`/`List` do?

milandamen:
`qx.ui.form.SelectBox` uses `qx.ui.core.MSingleSelectionHandling` as implementation for `setValue`.
`qx.ui.form.List` uses `qx.ui.core.MMultiSelectionHandling` as implementation for `setValue`

For `MSingleSelectionHandling`:

1. MSingleSelectionHandling#setValue(null)
2. MSingleSelectionHandling#resetSelection() (this is NOT the same as resetSelection on the `VirtualSelectBox`)
3. SingleSelectionManager#resetSelected() (which selects the first available item)

For `MMultiSelectionHandling`:

1. MMultiSelectionHandling#setValue(null)
2. qx.ui.core.selection.Abstract#clearSelection() (which selects the first available item, if mode is 'one')

The reason that `resetSelection` on `VirtualSelectBox` is different than on `MSingleSelectionHandling` is because `VirtualSelectBox` has a `selection` property containing a `qx.data.Array`. If `resetSelection` is called on the `VirtualSelectBox`, that property gets reset to the initial selection array; it does not clear the selection.

johnspackman:
Ah, ISWYM - so you mean that `resetSelection` does not change the property value, so it has zero effect.  And `resetSelection` is called by `setValue(null)`, which is how the form resetter resets the field, so the form resetter also has no effect.

This change is to access the `selection` property value (which is a `qx.data.Array`) and empty it.

johnspackman (approved):

milandamen:
Exactly. This PR makes sure the "value" (aka the selected item) of the VirtualSelectBox is reset to a default (first item in the model).

derrell (approved):

oetiker (approved):
ok

cajus (approved):

cajus:
Full quorum - merging. Thank you @milandamen .

PR #9594, opened 2018-11-29
Provides Universal Widget ID implementation

johnspackman:
Please see https://github.com/qooxdoo/qooxdoo/issues/9422 and also the previous PR at https://github.com/qooxdoo/qooxdoo/pull/9426 (abandoned because of rebasing issues)


uitests/objectid/Manifest.json
```diff
@@ -0,0 +1,34 @@
...
+    "authors" : 
+    [
+      {
+        "name" : "First Author (uid)",
+        "email" : "first.author@some.domain"
+      }
+    ],
+
+    "version" : "master",
+    "qooxdoo-versions": ["6.0.0-alpha"]
```
cboulanger: I don't think this is checked anywhere, but you can already update it to 6.0.0

cboulanger (dismissed):

hkollmann:
Could you have a look for the lint and coverage Errors? Thanks!

framework/source/class/qx/html/Element.js
```diff
@@ -787,6 +795,17 @@ qx.Class.define("qx.html.Element",
...
     */
+    
+    updateObjectId: function() {
+      // Copy Object Id
+      if (qx.core.Environment.get("module.objectid")) {
+        var id = null;
+        if (this.__widget && this.__widget.getObjectId()) {
+          id = qx.core.Id.getAbsoluteIdOf(this.__widget, false) || null;
+        }
+        this.setAttribute("data-object-id", id, true);
```
oetiker: maybe namespace this with "data-qx-object-id" ?

framework/source/class/qx/core/Id.js
```diff
@@ -0,0 +1,166 @@
...
+          return obj;
+        }
+      }
+      
+      switch(id) {
+        case "application":
+          return qx.core.Init.getApplication() || undefined;
+      }
+      
+      return; // undefined
```
derrell: For the benefit of editors that flag as an error `return` with a value mixed with `return` without a value in a single function, please `return undefined;` here instead.

framework/source/class/qx/core/Id.js
```diff
@@ -0,0 +1,166 @@
...
+     * ancestors has a null `objectId`.
+     * 
+     * This will also return null if the top-most ancestor is not one of the globals registered
+     * with `registerObject` or a known global (such as the application); however, by passing
+     * `false` as the `mustBeRegistered` parameter, this restriction is removed and a path will
+     * still be returned (it's just that in this case, you will have to know the the root object
+     * to query and `qx.core.Id.getObject` will not work on that path)  
+     * 
+     * @param obj {qx.core.Object} the object
+     * @param mustBeRegistered {Boolean} whether the
```
derrell: explanation of parameter is truncated

framework/source/class/qx/core/Id.js
```diff
@@ -0,0 +1,166 @@
...
+     * @return {String} full path to the object
+     */
+    getAbsoluteIdOf: function(obj, mustBeRegistered) {
+      var segs = [];
+      var application = qx.core.Init.getApplication();
+      while (obj) {
+        var id = obj.getObjectId();
+        if (!id) {
+          this.error("Cannot determine an absolute Object ID because one of the ancestor ObjectID's is null");
```
derrell: Would it be useful here, in the error message, to state which child's parent had a null object id?

framework/source/class/qx/core/Id.js
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+          if (mustBeRegistered === false) {
+            break;
+          }
+          this.error("Cannot determine a global absolute Object ID because the topmost object is not registered");
+          return null;
+        }
+        obj = owner;
+      }
+      var path = segs = segs.join("/");
```
derrell: `segs` is a local variable. It's not used after this line, in this method, so need not be assigned to.

framework/source/class/qx/core/Id.js
```diff
@@ -0,0 +1,166 @@
...
+     */
+    register: function(obj, id) {
+      if (!this.__registeredObjects) {
+        this.__registeredObjects = {};
+        this.__registeredIdHashes = {};
+      }
+      if (!id) {
+        id = obj.getObjectId();
+      }
+      this.__registeredObjects[id] = obj;
```
derrell: Why doesn't this create the same problem we had previously, with the registry, of preventing automatic garbage collection because the object reference is retained here?

framework/source/class/qx/core/MObjectId.js
```diff
@@ -0,0 +1,234 @@
...
+     * 
+     * @param id {String} ID of the object
+     * @return {qx.core.Object?} the created object
+     */
+    _createObjectImpl: function(id) {
+      return;  // Return undefined
+    },
+    
+    /**
+     * Adds an object as owned by this object
```
derrell: I don't understand this comment. What does this method actually do?

framework/source/class/qx/core/MObjectId.js
```diff
@@ -0,0 +1,234 @@
...
+        obj.__changingOwner = false;
+      }
+      this.__ownedObjects[id] = obj;
+    },
+
+    /**
+     * Discards an object from the list of owned objects; note that this does
+     * not dispose of the object, simply forgets it if it exists.
+     * 
+     * @param id {String|Object} the ID of the object to discard, or the object itself
```
derrell: Calling an object `id` is a bit confusing. How about calling this variable `data` and then set local variables `id` and `obj` appropriately?

uitests/objectid/source/class/objectid/Application.js
```diff
@@ -0,0 +1,89 @@
...
+
+      var root = this.getRoot();
+      
+      root.add(new qx.ui.basic.Label("Use the DOM Elements inspector to look at the widgets, " +
+      		"check out the 'data-object-id' attributes.<br>\n" +
+      		"<br>\n" +
+      		"Click the 'Run Test' button and then try this code in console:")
+        .set({ rich: true }), { left: 100, top: 20 });
+      root.add(new qx.ui.form.TextArea().set({ 
+        value: "var edtName = $(\"input[data-object-id='application/exampleEditor/edtName']\");\n" +
```
derrell: Is this some ES6 syntax? What does this `$()` phrase do?

derrell (commented):
This will be very useful, I think, but I'm not seeing how it's actually used in a real application. Does each object need to be explicitly requested to have an ID, or is one assigned automagically? And importantly, we appear to lose the automatic garbage collection that we regained with the Registry changes a while back, but maybe I'm missing something?

cboulanger:
@derrell I am working on a prototype app for integrating [TestCafé](https://devexpress.github.io/testcafe/) and [Puppeteer](https://pptr.dev) UI testing frameworks, they depend on an ID system that we will finally have with this PR. 

Other than that will it be useful in decoupling business logic, for example, in plugins that cannot or should not reference objects directly. 

As to automation: that shouldn't be in this PR which should be as minimal as possible, as we have seen in our previous discussions that the use cases differ so radically that no one solution will be useful for everybody. I think it is the developer's task to automatically assign IDs to object in any way that s/he finds useful for the particular app. That doesn't mean that some best practice emerges which we can then put into the core.

derrell:
Hmmm... I think we've had this discussion before. If IDs don't get added to every object automatically, this does not fulfill the one significant use case I've had for it: to determine, when something goes wrong, what object is at issue. To accomplish that, it sounds like you're telling me that I'd have to manually add an ID to every object I instantiate? (It would seem the same issue exists for automating GUI testing based on IDs. How can you automate that if every application-instantiated widget doesn't have an ID?)

cboulanger:
@derrell Yes we discussed that before. You will have to add a solution to your use case yourself, but you can now do it on too of @johnspackman's "primitive"  There is no other way we can get this issue forward, because your use case will differ from everybody else's use case. :-)

cboulanger:
As to memory leak: I don't know how much of an issue this still is, but the feature is optional and can be turned off via environment variable. So if you need it for testing, you can disable it in the production build. If you need it to reference particular business object only, you won't have the problem. It probably is only an issue if you massively and automatically assign those IDs for use in production builds.

level420:
@johnspackman travis tests are failing and I think the error is related to the PR:
`Firefox 63.0.0 (Linux 0.0.0) qx.test.core ObjectId:testGetObject FAILED
	obj.toHashCode is not a function
		qx.core.Id:3971
		qx.test.core.ObjectId:7`

framework/source/class/qx/html/Element.js
johnspackman: done

framework/source/class/qx/core/Id.js
johnspackman: Because this is only a list of top level objects, not an exhaustive list of *all* objects.

EG the idea is that objects with public IDs are 'scoped" within their parent, so that the path "A/B/C/D" means that the object "A" is registered globally and it has a pointer to B which has a pointer to C etc - and this means that B, C, and D are not in a global list and hence do not trigger the memory leaks that the ObjectRegistry lead to.

It is true that in order to garbage collect, you need to remove A from the global list (or if you just wanted to remove C and D, you'd remove C from B) - but this benefits from a cascade effect.

framework/source/class/qx/core/MObjectId.js
johnspackman: this is adding to heirarchy, eg using the example above, if `this === B` then `this.addOwnedObject(C)` means that `qx.core.Id.getObject("A/B/C") === C`

uitests/objectid/source/class/objectid/Application.js
oetiker: this is jquery syntax

johnspackman: it's jquery - it just an example to illustrate how something other than Qooxdoo can navigate the objects inside a Qooxdoo application (when unit testing this would be Selenium doing the query, it's just that jQuery is easier to explain for testing)

johnspackman:
@derrell regarding assigning IDs, the easiest way is to create IDs by overridding `_createObjectImpl` so that the hierarchy of IDs is built up automatically, but obviously this is not suitable for existing code without extensive rework (or necessarily always convenient anyway), so there is the `addOwnedObject` method to allow you to associate objects with IDs in the hierarchy.

This does mean that you are choosing the ID to assign, but then that's pretty much essential in order for you to be able to intuitively recognise what DOM element relates to what line of code anyway.  But what's different here is that because there is a hierarchical "scoping" of IDs, it is easy to maintain those IDs and have them unique.

There's an example of this pattern here https://github.com/qooxdoo/qooxdoo/issues/9422#issuecomment-346301649 - I find that it is often useful and an improvement in readability to code using this pattern because your `_createObjectImpl` method becomes a clear, almost declarative section, of UI widgets and controllers and other objects.  My editors highlight the `case "some-id-string"` statement which picks them out visually, I can arrange them in any order I like (rather than in the order they needed to be added to a container widget), and derived classes can replace or augment the objects returned.

johnspackman:
> @johnspackman travis tests are failing and I think the error is related to this PR:
> `Firefox 63.0.0 (Linux 0.0.0) qx.test.core ObjectId:testGetObject FAILED obj.toHashCode is not a function qx.core.Id:3971 qx.test.core.ObjectId:7`

@level420 darn, thats a qxcompiler issue with nested classes - working on that now; I've just pushed a new commit with the response to comments, but it'll fail until I get qxcompiler updated

derrell (approved):
This implementation looks fine. I'm not at all thrilled that it requires manual registration of objects; I would much prefer an implementation that simply added hierarchical IDs to the DOM nodes automatically, which would be a huge aid to debugging. That said, it seems this implementation is useful to others, and can possibly be adapted later to accomplish these other goals.

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

cboulanger:
Thanks @johnspackman, this will be very useful! 

@derrell, maybe you should write up a description of what you need in a separate issue, and then we can figure out whether this should be implemented based on this id system or separately. Thinking about it, if I understand you correctly, maybe a single Mixin to `qx.ui.core.Widget` could do what you need.

derrell:
@cboulanger Actually, it may be that with the qx-compiler implementation of `$$createdAt` in each element, I may already have what I need. I didn't recall that that had been implemented. Now I just need to get my older app moved over to qx-compiler, and I should be all set. We'll see. Thanks, though!

PR #9600, opened 2018-12-05
fixes #9598

### LINKED ISSUE #9598: Regression with Drag&Drop
**Describe the bug**
When the drag target passes over a possible drop target enroute to a different drop target, and is dropped on the latter, the `drop` event is fired on the former instead of the latter as expected.

This regression was introduced with commit 4160586f5784633b377d6afb81cd11a7ac40ec67
@johnspackman This was your commit.

**To Reproduce**
In the following playground sample, drag the "Drag Me" label _over_ "First Window" then continue on to "Second Window". When you release the mouse while over "Second Window", the alert will tell you that the `drop` event was actually wrongly fired on "First Window". If, on the other hand, you drag the round-about way, around the windows so that the drag never passes over "First Window" and goes immediately to "Second Window" where it is dropped, the correct object receives the `drop` event.

http://tinyurl.com/y7k8a2zq

**Expected behavior**
A drop on "Second Window" would be expected to fire a `drop` event on 'Second Window", not on "First Window" as it is doing with the regression commit through master.

johnspackman:
see https://github.com/qooxdoo/qooxdoo/issues/9598

### DIFF
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -768,18 +768,21 @@ qx.Class.define("qx.event.handler.DragDrop",
 
             // new drop target detected
             if (droppable && droppable != self.__dropTarget) {
+              var dropLeaveTarget = self.__dropTarget;
+              
               self.__validDrop = true; // initial value should be true
               self.__dropTarget = droppable;
 
               var innerTracker = {};
               qx.event.Utils.catch(innerTracker, function () {
+                self.__dropTarget = null;
                 self.__validDrop = false;
               });
               
               // fire dragleave for previous drop target
-              if (self.__dropTarget) {
+              if (dropLeaveTarget) {
                 qx.event.Utils.then(innerTracker, function () {
-                  return self.__fireEvent("dragleave", self.__dropTarget, self.__dragTarget, false, e);
+                  return self.__fireEvent("dragleave", dropLeaveTarget, self.__dragTarget, false, e);
                 });
               }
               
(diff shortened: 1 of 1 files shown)


cboulanger (dismissed):

hkollmann (dismissed):

cajus (dismissed):

framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -768,21 +768,25 @@ qx.Class.define("qx.event.handler.DragDrop",
...
 
               var innerTracker = {};
               qx.event.Utils.catch(innerTracker, function () {
                 self.__validDrop = false;
               });
+              
+              // fire dragleave for previous drop target
+              if (self.__dropTarget) {
+                qx.event.Utils.then(innerTracker, function () {
+                  return self.__fireEvent("dragleave", self.__dropTarget, self.__dragTarget, false, e);
```
derrell: This _looks_ like the `dragleave` is now issued on the wrong target, isn't it? At this point, `self.__dropTarget` has already been set to the new, droppable target, but then this code issues the `dragleave` on that new droppable target. Is it instead necessary to save `self.__dropTarget`'s value near line 771, and use the saved value here?

johnspackman: yes, you're right again - 2secs, just going to double check the original before i push again

derrell (approved):

hkollmann (approved):

level420 (approved):

cajus (approved):

cboulanger (approved):

oetiker (approved):

PR #9601, opened 2018-12-05
fixes an issue with resolving IDs

johnspackman:
fixes an issue with resolving IDs, where if the object does not have any owner its absolute ID will never be found, even if it has an ID and is registered as top level;

also validation of ids and comment fixes

### DIFF
--- a/framework/source/class/qx/core/MObjectId.js
+++ b/framework/source/class/qx/core/MObjectId.js
@@ -42,13 +42,12 @@ qx.Mixin.define("qx.core.MObjectId", {
     },
 
 
-    /** The ID of the object.  */
+    /** {String} The ID of the object.  */
     objectId : {
       init: null,
-      check : "String",
+      check : function(value) { return value === null || (typeof value == "string" && value.indexOf('/') < 0); },
       nullable : true,
-      apply : "_applyObjectId",
-      transform : "_transformObjectId"
+      apply : "_applyObjectId"
     }
   },
 
@@ -72,16 +71,6 @@ qx.Mixin.define("qx.core.MObjectId", {
       }
     },
     
-    /**
-     * Transform/validate objectId
-     */
-    _transformObjectId: function(value) {
-      if (value && value.indexOf('/') > -1) {
-        throw new Error("Object IDs cannot have '/' characters");
-      }
-      return value;
-    },
-
     /**
      * Apply objectId
      */
--- a/framework/source/class/qx/core/MObjectId.js
+++ b/framework/source/class/qx/core/MObjectId.js
@@ -76,8 +76,9 @@ qx.Mixin.define("qx.core.MObjectId", {
      * Transform/validate objectId
      */
     _transformObjectId: function(value) {
-      if (value && value.indexOf('/') > -1)
+      if (value && value.indexOf('/') > -1) {
         throw new Error("Object IDs cannot have '/' characters");
+      }
       return value;
     },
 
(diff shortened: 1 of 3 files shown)


hkollmann (dismissed):

uitests/objectid/source/class/objectid/Application.js
```diff
@@ -53,6 +53,14 @@ qx.Class.define("objectid.Application", {
       this.getObject("exampleEditor/ctlrTitle").setModel(model);
 
       root.add(this.getObject("exampleEditor/container"), { left: 100, top: 250 });
+      
+      var button_panel = new qx.ui.toolbar.ToolBar();
+      button_panel.add(new qx.ui.toolbar.Button("Push Me"));
+      root.add(button_panel, { left: 100, top: 300 });
+      qx.core.Id.getInstance().register(button_panel, "buttons");
+      button_panel.setObjectId("buttons");
```
cboulanger: Wouldn't it make sense to set the object id implicitly when registering the object? Specifying it twice seems redundant and might be a source of bugs. It would also save one line of code.

johnspackman: the idea is that you can register shortcuts, so you could register A and C as top level, and then this code would be true: `qx.core.Id.getObject(“A/B/C/D/E”) === qx.core.Id.getObject(“C/D/E”)`
but when registering an object globally, you probably want to give it a different ID in that context, eg something more meaningful at a global level - so you have the option to override the ID for just globally, eg:
```
 qx.core.Id.register(C, “currentUser”);
/* .. snip .. */
qx.core.Id.getObject(“currentUser/edtName”).setValue(“George”);
qx.core.Assert.assertTrue (qx.core.Id.getObject("A/B/C") === qx.core.Id.getObject("currentUser"));
```
in other words, as the global objects set via `qx.core.Id.register()` are pointers into the ID heirarchy, the fact that you are registering something should not change the thing that you are registering

also, you could express your code as this:
```
       button_panel.setObjectId("buttons");
      qx.core.Id.getInstance().register(button_panel);
```
and `register()` will use the object’s objectId rather than assuming you need to override.

cboulanger (dismissed):

hkollmann (dismissed):

cboulanger (dismissed):

framework/source/class/qx/core/MObjectId.js
```diff
@@ -63,16 +64,26 @@ qx.Mixin.define("qx.core.MObjectId", {
...
     _applyOwner : function(value, oldValue) {
       if (!this.__changingOwner) {
         throw new Error("Please use API methods to change owner, not the property");
       }
     },
+    
+    /**
+     * Transform/validate objectId
+     */
+    _transformObjectId: function(value) {
```
derrell: This method isn't actually doing any transforms. Would it not be cleaner to have a customer `check` function that checked for string and no slashes, than to do it in a non-transforming `transform` function?

derrell (changes requested):
My comment is solely a readability fix. If there's reason to do it as currently implemented, I'd be fine with it.

framework/source/class/qx/core/MObjectId.js
johnspackman: yes i agree - funny, i've never used that feature before!

derrell (approved):

hkollmann (approved):

cboulanger (approved):

oetiker (approved):
ship it!

level420 (approved):

cboulanger:
Four approvals, merging, thank you @johnspackman !

Issue #9599, opened 2018-12-04
width/height mixup in high-res images


jpriebe:
On line 519 of Image.js:

```
  var imageWidth = ResourceManager.getImageHeight(source);
  var imageHeight = ResourceManager.getImageWidth(source);
```

This causes images with high dpi versions (@2x versions) to have their height/width flipped, leading to some very strange results (unless, of course, the image is perfectly square, in which case you would never know).

johnspackman:
@jpriebe Good catch!  Would you like to submit a PR 😉 and we'll integrate it? 👍

jpriebe:
Sure.  I was just being lazy.   Any plans for a 5.0.3 release in the near future?

johnspackman:
great, thank you.  don't know about 5.0.3 just yet - we have collected a set of essential fixes together before and back ported them into a release but we typically recommend using master (aka 6.0.0 release candidate).  i've asked the team what they think

cboulanger:
Fixed with #9602

PR #9602, opened 2018-12-05
Fixing issue #9599 - width/height mixup in high-res images

jpriebe:
does just what the subject line says.  :-)

### DIFF
--- a/framework/source/class/qx/ui/basic/Image.js
+++ b/framework/source/class/qx/ui/basic/Image.js
@@ -516,8 +516,8 @@ qx.Class.define("qx.ui.basic.Image",
       else if (ResourceManager.has(source)) {
         var highResolutionSource = this._findHighResolutionSource(source);
         if (highResolutionSource) {
-          var imageWidth = ResourceManager.getImageHeight(source);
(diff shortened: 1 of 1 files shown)


derrell (dismissed):
My only concern here is that this _should_ have been commonly-used code, so why was it never discovered before? Are we missing something insidious?

johnspackman:
@derrell i would guess that 2DP images are rarer, and that they're normally used for icons so they're normally square

johnspackman (dismissed):

hkollmann (dismissed):

cboulanger (dismissed):

level420 (dismissed):

jpriebe:
Sorry, guys.  I just screwed this up by pushing another set of changes for a related issue (tables don't support high-res images).  Let me try to clean up the mess.

jpriebe:
OK, I rebased the 2nd commit out of the branch.

cajus (approved):

johnspackman (approved):

oetiker (approved):

cboulanger (approved):

hkollmann (approved):

PR #9604, opened 2018-12-07
Allows to customize the label suffix (default " :") and the required …

cboulanger:
…suffix (default: red asterisk) in the form renderer by refining the values of two new properties.

This time, without the merge commits!

### DIFF
--- a/framework/source/class/qx/ui/form/renderer/AbstractRenderer.js
+++ b/framework/source/class/qx/ui/form/renderer/AbstractRenderer.js
@@ -52,29 +52,31 @@ qx.Class.define("qx.ui.form.renderer.AbstractRenderer",
   },
 
   properties :
+  {
+    /**
+     * A string that is appended to the label if it is not empty.
+     * Defaults to " :"
+     */
+    labelSuffix :
     {
-      /**
-       * A string that is appended to the label if it is not empty
-       */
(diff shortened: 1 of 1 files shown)


oetiker (dismissed):
great ... that one has been bothering me too :)

hkollmann (dismissed):

framework/source/class/qx/ui/form/renderer/AbstractRenderer.js
```diff
@@ -51,6 +51,31 @@ qx.Class.define("qx.ui.form.renderer.AbstractRenderer",
...
+      labelSuffix :
+        {
+          check : "String",
+          init : " :",
+          event : "changeLabelSuffix",
+          nullable : true
+        },
+
+      /**
+       * A string that is appended to the label if it is not empty
```
derrell: This documentation string is the same as the one above. Both properties have a bit more (and different) detail about what the property actually does.

derrell (changes requested):

derrell (approved):

hkollmann (approved):

level420 (approved):

oetiker (approved):

johnspackman (approved):

cajus (approved):

Issue #211, opened 2016-04-29
Add more modern theme 


woprandi:
It miss a modern theme/icons for qooxdoo (ie flat theme). 
This example seems to be the best candidate for that : http://sqville.com/examples/qxthemes/clean/index.html

johnspackman:
+1

oetiker:
@johnspackman you can just click on the little face on the issue itself to +1 it

johnspackman:
aha :)

slachtar:
The guys at wisej.com provide a very nice theme builder and there's very nice flat themes as examples.

johnspackman:
is it possible to build a theme that is reusable independent of the wisej api?

slachtar:
the theme builder generates a .theme file which is a json file with all
qooxdoo theme entries (fonts, images, colors, apperances).
I've to test it first.

On Fri, Apr 29, 2016 at 3:40 PM, John Spackman <notifications@github.com>
wrote:

> is it possible to build a theme that is reusable independent of the wisej
> api?
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/qooxdoo/qooxdoo/issues/211#issuecomment-215739742>
>



-- 
Slah Lachtar
Tel. (+216) 98 221 575

iceteagroup:
Hi, I just saw this.

We (iceteagroup) will be happy to contribute back to the qooxdoo community both the theme builder and the theme roller classes.

The theme builder is a .NET executable that we can modify to work without wisej. But can make the exe available for free. 

The "theme roller" javascript classes can be added to qooxdoo or provided as an addition. They take the theme json file and convert it to qooxdoo theme classes.

However, I don't know how we can contribute. Maybe we will simply make those available on our website.

iceteagroup:
Oh, forgot to add that we can also contribute all the themes we designed.

level420:
@iceteagroup That would be great! 
Regarding the contributions: please  set your contribution under the MIT license and use the following text passage in the header for this:

```
   License:
     MIT: https://opensource.org/licenses/MIT
```

qooxdoo will change to MIT license soon.

oetiker:
now the real kicker would be if the theme builder was written in qooxdoo ... seems a bit odd that this is in .NET ...

level420:
@oetiker yes, but having the wisej theme builder is about 1000% better than having at the moment ... well ... nothing!

[...]

PR #9608, opened 2018-12-11
Clean theme widgetbrowser version

sqville:
This is the widgetbrowser version of the Clean theme. It has been tested using the top 3 major browsers on both Windows 10 and the latest MacOS. This version works with the widgetbrowser app with two exceptions. qx-inset class gets demoted for some odd reason for textfields, and theme switching breaks when going from Clean to any other theme. I will continue to address these issues. I am submitting this PR as a means to get this process started, and to get help on these two outstanding issues. This PR directly addresses Enhancement Request #211  

**New\Unique feature of this theme:**
All decorator images, with the exception of treevirtual, have been replaced with code (pure qx, qx + CSS) or embedded SVG using the template class. Prime examples are: 

- arrows and radiobuttons (pure qx)

- checkbox and table select column button (qx + CSS)

- datefield and tree (embedded SVG)

**Design decisions to consider are as follows:**
1. Mixins are used to add new, themable fields to framework classes - See "qx.theme.Clean" for the exact list
2. "qx.ui.popup.Popup" is **patched** in order to include two new states, bottom and top placement, in order to know if a selectbox's popup is above or underneath the control (i recommend updating the framework itself in order to enable this feature, but it is here for now)
3. iconProps property is added to Atom so that Button appearance entries can change it's image based on button states (hover, focus, etc)
4. "html" property has been added to Image so that SVG (or HTML) can be added and manipulated by the Image's parent container (i.e. Atom and HeaderCell). SVG code is centralized in the theme's "qx.theme.clean.Image" class as static entries
5. "freeStyleCss" is added to Decorator to allow for the addition of psudo classes (after and before) and custom CSS. Custom CSS is centralized in the theme's "qx.theme.clean.Image" class as static entries

Working demo can be found here: http://sqville.com/examples/widgetbrowser2018/index.html

I am committed to making this theme acceptable to all. Thank you for your consideration.

Chris (sqville)


jpriebe:
Great work, Chris!  Qooxdoo team, please incorporate this!  We need a fresher theme for qooxdoo!

cboulanger:
Hi! Thsi is really great work that will add significantly to the visual presentation of the framework. So we are very happy. Would you consider to [provide the theme as a contrib library](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/docs/cli/contrib.md)? The problem with integrating code into the core is that that means the core development team takes responsibility for maintaining it. Since yours is a complex piece of code with a lot of thinking behind it, we are afraid we cannot do that. We'll be happy to integrate the theme in the demos etc. Thanks again!

sqville:
Thank you, and yes, happy to provide the theme as a contrib library. I read through the instructions, a few times now, and the process is still unclear to me. Looks like I will have to trial-and-error my way through this to a contrib version. Question too. Does this mean that I need to change the namespace of my classes? i.e. from "qx.theme.Clean" to something like "**sqv**.theme.Clean"

hkollmann:
Have you seen the travis Fails?

oetiker:
@sqville you can look at this one for inspiration https://github.com/ITISFoundation/qx-osparc-theme

cboulanger:
@sqville "sqv.theme.Clean" would be a good name, yes. We have also talked about recommending that contrib creators use the pattern (in your case) "gh.sqville.theme.Clean" to indicate that the contrib is hosted on github under the given username, but it is totally up to you how to name it.

sqville:
Thanks team! I addressed all the Travis Fails. I will make the switch to a contrib over the holidays and announce it's readiness over gitter, and anywhere else you all recommend.

sqville:
Added as a contrib - https://github.com/sqville/sqv-qxthemes
Hope i set it up correctly

oetiker:
closed PR since this is now a contrib

PR #9611, opened 2018-12-17
Fixed removing listeners from binding chain with incomplete binding (see #9547)

milandamen:
This PR fixes the problem reported in #9547 in a better way that fixes the underlying issue rather than covering it up.

**Situation**
Given a `source` object with a property named `a`
Given a class `A` that does not have a property named `b`
Given a `target` object

**Act**
Bind from `source` using path `a.b` to target `target`
Set the property `a` of `source` to an instance of A. This instance does not have a property with name `b`.
Remove the binding from `source`

**Result**
An exception will be thrown. It is trying to remove a listener that does not exist from the instance of `A`.

**Expected result**
The binding is removed from `source` without errors.

**Problem**
When creating a binding, a list `sources` and list `listenerIds` is made.
For each part in the binding chain, a source and a corresponding listener id is created.
If any other object in the chain is null, the sources and listener ids are only set for the objects that do exist.

When creating our binding in the example, the lists become like this:

- sources: `[ source ]`
- listenerIds: `[ __chainListener func ]`

When we set an object to property `a` of our `source` object, the `__chainListener` function is called. It will modify the `sources` and `listenerIds` lists and will try to make a listener from instance of `A` to our target, but it does not succeed because the instance of `A` does not have a property named `b`, so it will quit.

The resulting lists will look like this:

- sources: `[ source, A ]`
- listenerIds: `[ __chainListener func ]`

As you can see, `sources` has 2 items while `listenerIds` has only 1 item.
This is because a listener was not made because `A` does not have a property with name `b`.

Now, when you try to remove this whole binding, it will loop over all `sources` and will try to remove the listeners, but because there is no listener for the 2nd source, it will crash.

This fix will make sure that `sources` and `listenerIds` are kept in-sync; they will always have the same amount of items.


cboulanger (approved):

derrell (approved):
Very nice description of the problem!

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

level420 (approved):
@milandamen good catch!

cajus (approved):

cajus:
Merging. Sorry for my _approve_ took so long... Thanks @milandamen !

PR #9614, opened 2018-12-18
Add qx prefix to objectid

### LINKED ISSUE #9612: New ObjectId system breaks classes that have property "owner" or "objectId", breaking backwards compatibility
**Problem 1**

The new Universal Widget ID system breaks all classes that have a property named `owner` or `objectId`. Why are these primary properties of an object? These should not be primary properties, or the new object id system should be disabled by default; because this basically breaks any application with such.

When a class of mine has a property named `owner` it will fail to define the class, because:
`Uncaught Error: Could not refine property "owner" without a "refine" flag in the property definition! This class: SomeClass, original class: qx.core.Object.`

**Reproduce**
`qx.data.marshal.Json.createModel({owner: null})`

**Problem 2**
When disabling the "module.objectid" environment, the application will crash on initialization:
`Uncaught Error: Overwriting member "_createObject" of Class "qx.core.Id" is not allowed!`

In `qx.core.Id` it is trying to overwrite the member function in `qx.core.MObjectId`.

Relevant PR: #9594

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/9612

### DIFF
--- a/framework/source/class/qx/core/Id.js
+++ b/framework/source/class/qx/core/Id.js
@@ -21,7 +21,6 @@
  */
 qx.Class.define("qx.core.Id", {
   extend: qx.core.Object,
-  include: [ qx.core.MObjectId ],
   type: "singleton",
   
   members: {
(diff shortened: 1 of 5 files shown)


cboulanger (approved):

derrell (approved):

level420 (approved):

hkollmann (approved):

oetiker (approved):

cboulanger:
5 approvals, merging. Thank you @johnspackman !

PR #9615, opened 2018-12-23
Rename methods to avoid name conflicts

johnspackman:
This is (hopefully) a last PR on this subject, made to try and ensure that there are no possible side effects or name collisions with existing code belonging to users.  The naming convention has been changed so that the "object" part is always "Qx Object", meaning that we talk about properties like `qxObjectId` and `getQxObject` etc

One change that I've made here which is different to recent discussions on gitter with @cboulanger and @derrell is that `qx.core.MObject.getObject` has been renamed to `.getQxObject` and *not* `.getObjectByQxObjectId` - the reason for this is that the design pattern of `_createObjectImpl` is specifically intended to allow user interface widgets (and other objects) to be routinely accessed by `.getQxObject`, and for this to be used so frequently brevity seems essential, but this PR still used the "QX" convention

For example, see https://github.com/johnspackman/qooxdoo/blob/add-qx-prefix-to-objectid-2/uitests/objectid/source/class/objectid/ExampleEditor.js#L11-L14.

### DIFF
--- a/framework/source/class/qx/core/Id.js
+++ b/framework/source/class/qx/core/Id.js
@@ -31,16 +31,16 @@ qx.Class.define("qx.core.Id", {
     /*
      * @Override
      */
-    _createObject: function(id) {
+    _createQxObject: function(id) {
       // Create the object, but don't add it to the list of owned objects
-      var result = this._createObjectImpl(id);
+      var result = this._createQxObjectImpl(id);
       return result;
     },
     
     /*
      * @Override
      */
-    _createObjectImpl: function(id) {
+    _createQxObjectImpl: function(id) {
       if (this.__registeredObjects) {
         var obj = this.__registeredObjects[id];
         if (obj !== undefined) {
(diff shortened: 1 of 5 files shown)


derrell (approved):

cboulanger (approved):
I don't like it *at all* for aesthetic and semantic reasons, but I approve for reasons of necessity and consistency.

johnspackman:
@cboulanger What do you mean? Isn't this what you asked me to do?




Sent from my phone







On Sun, Dec 23, 2018 at 7:20 PM +0000, "Christian Boulanger" <notifications@github.com> wrote:












@cboulanger approved this pull request.



I don't like it at all for aesthetic and semantic reasons, but I approve for reasons of necessity and consistency.

johnspackman:
ah ok! ISWYM.  Happy Christmas :)

level420 (approved):

hkollmann (approved):

oetiker (approved):

oetiker:
oops sorry missed that one

cboulanger:
5 approvals, merging, thanks @johnspackman !

Issue #9623, opened 2019-02-05
Firefox 65 table scroll by keyboard


mappopo:
In Firefox 65 is no longer possible to change the row selection by keyboard.

The behaviour can be replicated in playgroud http://www.qooxdoo.org/current/demobrowser/#table~Table.html 
Select a row with a mouse click and try use keyboard arrows up and down to move selection. It takes no effect. With other browser (eg. Edge or Chrome) the focus and selection successful changes.

zaucker:
Works for me on

   http://www.qooxdoo.org/devel/demobrowser/#table~Table.html

and

    http://www.qooxdoo.org/devel/playground/#Table-ria

with Ubuntu and FF 65

Cheers,
Fritz

On Tue, 5 Feb 2019, mappopo wrote:

> In Firefox 65 is no longer possible to change the row selection by keyboard.
>
> The behaviour can be replicated in playgroud http://www.qooxdoo.org/current/demobrowser/#table~Table.html
> Select a row with a mouse click and try use keyboard arrows up and down to move selection. It takes no effect. With other browser (eg. Edge or Chrome) the focus and selection successful changes.

-- 
Oetiker+Partner AG              tel: +41 62 775 9903 (direct)
Fritz Zaucker                        +41 62 775 9900 (switch board)
Aarweg 15                            +41 79 675 0630 (mobile)
CH-4600 Olten                   fax: +41 62 775 9905
Schweiz                         web: www.oetiker.ch

cboulanger:
Works on MacOS 10.14.2, FF 65.

mappopo:
I use windows

hkollmann:
Yes it is. Most of us use git repos version

cboulanger:
I see - I can confirm, http://www.qooxdoo.org/current/demobrowser/#table~Table.html is broken for Firefox 65 on the Mac. Looks like we need a backport of the fix?

level420:
@mappopo did it work with firefox 64?

lucapivato:
https://www.fxsitecompat.com/en-CA/docs/2018/non-printable-keys-no-longer-fire-keypress-event/

> An exception is the Enter key; pressing Enter with no modifiers, Shift+Enter or Ctrl+Enter will fire the keypress event as before, which is invalid in the current spec but compatible with other browsers.

This genius change may be coming to Chrome too...

level420:
@lucapivato yes, currently investigating that and why it still works in master.

[...]

PR #9628, opened 2019-02-15
#9623 added tab key to emulateKeyPress 

ofurrer:
Since Firefox 65 the Tab key is considered an unprintable key.  Therefore a simulated "keypress" event is necessary.


level420:
@odahme thank you for the hint. Currently working on a more general PR for the dead keys in firefox 65.

Closing this PR in favour of #9629

odahme:
Thank you for the quick response.

Issue #9626, opened 2019-02-12
Deprecation of keypress events according to UI event spec https://w3c.github.io/uievents/#keypress


level420:
Since Firefox with version 65 deprecates keydown events for non printable characters, we should discuss how deal with this in the framework:

- https://www.fxsitecompat.com/en-CA/docs/2018/non-printable-keys-no-longer-fire-keypress-event/
- https://www.fxsitecompat.com/en-CA/docs/2018/keydown-and-keyup-events-are-now-fired-during-ime-composition/

There also may be changes in future versions of the other browsers which will hit us from time to time in the cases the are tightening more to the specs.

level420:
See also issue #9623 which started this discussion

level420:
As @marcelr wrote on gitter, it seems that some methods like @qx.event.type.KeySequence.getKeyIdentifier@ and @qx.event.type.KeySequence.getKey@ do not retreive the correct key anymore if used with @keydown@ events.

oetiker:
I think this whole thing also raises a conceptual question for us ... for many users of qooxdoo we are providing a 'sane' way to creating web apps with a stable API, shielding users largely from browser differences and other hard to navigate portions of the web-world.

Every now and then the 'real world' does bleed through though and people are faced with having to change their applications to cater to browser changes. I think we should try to shield them as much as possible as long as they are using qx objects.

If a change is necessary it should come with a major qooxdoo release as all breaking changes do.

johnspackman:
> @oetiker said: I wonder if we could put something in place to warn authors about this problem ...maybe have the compailer complain if it sees keypress eventhandlers being assigned ? 

Possibly, but I think it would be much easier (and more reliable) to implement by overriding `addListener` in `qx.ui.core.Widget`.

@oetiker my thoughts exactly.  AIUI keyboard events was one of the big normalisations that Qooxdoo did because in the good ol' bad ol' days browsers had markedly different details for keyboard events.  Having to track keyup/down manually is a pain, and seems like a textbook case for encapsulation in the framework.

Has anyone seen a rationale for browsers deprecating `keypress` now?  It seems pretty dumb, because `keyup`/`keydown` are really quite low level for most application use cases...

level420:
Some more reading about `keypress` vs `keydown`:
https://www.mutuallyhuman.com/blog/2018/03/27/keydown-is-the-only-keyboard-event-we-need/

level420:
[...]

PR #9629, opened 2019-02-15
Consistent behaviour for keypress events in general and firefox version 65

level420:
To have a consistent behaviour between browsers, we need to also fire artificial `keypress` events when the meta keys are pressed, except the `SHIFT` key as in this situation only `keydown`/`keyup` is fired.
Also fire artificial `keypress` events for keys `BACKSPACE`, `TAB` and `ESCAPE` for gecko/firefox version 65 and up.

See #9626 #9623


cboulanger (dismissed):

framework/source/class/qx/event/handler/Keyboard.js
```diff
@@ -347,7 +347,14 @@ qx.Class.define("qx.event.handler.Keyboard",
...
           // non-printable, backspace or tab
-          if (qx.event.util.Keyboard.isNonPrintableKeyCode(keyCode) || this._emulateKeyPress[keyCode]) {
+          if (qx.event.util.Keyboard.isNonPrintableKeyCode(keyCode) || this._emulateKeyPress[keyCode] ||
+              /*
+               * We need an artificial keypress event for every keydown event.
+               * Newer browsers do not fire keypress for a regular charachter key (e.g when typing 'a')
+               * if it was typed with the CTRL, ALT or META Key pressed during typing, like
+               * doing it when typing the combination CTRL+A
+               */
+              domEvent.ctrlKey || domEvent.altKey || domEvent.metaKey) {
```
derrell: This concept looks good. Please, however, don't insert comments right in the middle of an expression. That makes the code difficult to read. I'd suggest something like the following, inserted at line 349, and move that comment to before line 349:
```
let isModified = domEvent.ctrlKey || domEvent.altKey || domEvent.metaKey);
```

(And then, of course, change the `if` statement to use `isModified`)

derrell (changes requested):

level420:
FYI: the fix needed for `qx.ui.table.Table` here https://github.com/qooxdoo/qooxdoo/pull/216 would not have been necessary if we had this fix.

The problem there was, that at some point chrome did not fire a keypress event anymore if you had hold down the `CTRL` key. This resulted in the `CTRL` key being detected, but not the `A` key afterward.
With this fix you get first a `keypress` event for the `CTRL` key and then a `keypress` for the `A` key.

derrell (approved):

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

PR #9630, opened 2019-02-17
try to fix deploy:

hkollmann:
- move to github token instead of ssh
   - autoincrement version.txt during build


.travis/deploy
```diff
@@ -1,37 +1,35 @@
...
 
+node -v
+npm -v
+git config user.email "travis@travis-ci.org"
+git config user.name "Travis CI"
+NEW_VERSION=$(npm version prerelease -m "[ci skip] new npm version: %s")
+echo "new version: $NEW_VERSION"
+echo $NEW_VERSION > version.txt
+GIT_URL="$(git config remote.origin.url)"
+GIT_URL=${GIT_URL/https:\/\//https:\/\/$GH_TOKEN@}
```
oetiker: what is the @ for ?

oetiker (approved):

oetiker:
worth a try ...

level420 (approved):
Whatever makes it work again: do it!

cboulanger (approved):

johnspackman (approved):

cajus:
I've no time to check it in the moment. I guess the easiest change would be to just call the travis cli tool to re-add the ssh key. If it does work now with the deploy key, and we're OK with the token stored unencrypted on the travis side (which makes only a very small difference to what we do in the moment), go for it!

cajus (approved):

derrell (approved):

cboulanger:
merged, thank you @hkollmann and sorry @derrell, I was on mobile and didn't see that you were missing!

Issue #9627, opened 2019-02-15
Bug in MEvent.addListenerOnce() for multiple times call with same args


fortooon:
Hi,
I found that MEvent.addListenerOnce() has the bug. If I set same handler to one object to process same event multiple times, in result after fire current event only first handler will be removed, others will be a live forever.  Reason is that `type + this.$$hash` key in  `listener.$$wrapped_callback[type + this.$$hash] = callback;` is not unique in such case.

Probably,  there is needed `__lastUnique` helper variable to generate wright key like in  qx.event.Manager.js

**To Reproduce**
You can run simple example based on http://www.qooxdoo.org/current/demobrowser/#widget~Button.html and see in browser console
```
qx.Class.define("demobrowser.demo.widget.Button",
{
  extend : qx.application.Standalone,

  members :
  {
    main: function()
    {
      this.base(arguments);

      var box = new qx.ui.layout.HBox();
      box.setSpacing(10);

      var container = new qx.ui.container.Composite(box);
      container.setPadding(20);
      this.getRoot().add(container, {left:0,top:0});

      this.addNormalButtons(container);
      
    },

    addNormalButtons : function(container)
    {
      var btn1 = new qx.ui.form.Button("Button A", "icon/22/apps/media-video-player.png");
      container.add(btn1);
      btn1.addListenerOnce("execute", this.onExecute, this);
      // second handler will be a live forever
      btn1.addListenerOnce("execute", this.onExecute, this);
    },    

    onExecute : function()
    { 
      var t1 = new Date();
      console.log("onExecute", t1.getTime());      
    }
  }
});
```

**Expected behavior**
On second button click `console.log` (`onExecute`) must not be called. But on this example in browser console we can see that every other click ob button will be handled.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser chromium
 - Version 71.0.3578.98

qooxdoo version : 5.0.2

level420:
@fortooon Thank you for reporting this issue. It would be great if you could offer a PR to solve the issue.

fortooon:
@level420  please, check

hkollmann:
@fortooon :Is this still an issue in the current master?

hkollmann:
Just did a test: It is.

hkollmann:
@fortooon : You have an branch wich will fix this. Could please create an PR for qooxdoo/qooxdoo? You created a PR request for fortooon/qooxdoo in February....

[...]

PR #9631, opened 2019-02-18
#9627 store listener unique ids added  in addListenerOnce

fortooon:
store listener unique ids added  in addListenerOnce to remove all listeners correctly.

Bug in MEvent.addListenerOnce() for multiple times call with same args #9627


fortooon:
please, review

level420:
@fortooon thank you for the PR, but we only accept PRs against qooxdoo master. Would you be so kind to rework your PR starting from master? Thank you.

Closing this PR

fortooon:
OK, I will do.

PR #9632, opened 2019-02-19
store listener unique ids added in addListenerOnce

fortooon:
store listener unique ids added in addListenerOnce to remove all listeners correctly.

Bug in MEvent.addListenerOnce() for multiple times call with same args #9627


fortooon:
Sorry, master has a code that differs from branch_5_0_X.  I need to be more careful later.

hkollmann:
Why did you close this? It should come to the current master.

cboulanger:
@fortooon Can you resolve the conflicts and have a look at the failing tests? 
Thank you!

oetiker:
@fortooon ping

hkollmann:
found a much simpler solution

Issue #9643, opened 2019-03-21
Since Firefox 66, event.getKeyIdentifier() returns wrong identifiers for some keys


ronkie:
Firefox 66 has changed keyCode and charCode of the keypress event. For more info see:

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/66

Because of that, letters from p to z result in incorrect key identifiers. For example, "r" results in "F3".
It can lead to problems when developers rely on getKeyIdentifier() method of qx.event.type.KeySequence event.

1. Go to qooxdoo playground (make sure to have Firefox 66):
http://www.qooxdoo.org/current/playground/
2. Try to change some code by typing 'v'
3. See that no v symbol is typed and console opens instead.

Expected behavior:
Character 'v' is typed.

Browser:
Firefox 66.

gnikolaidis:
Thank you! 97b3ea6 applied cleanly on Qooxdoo 5.0.2 and solved our many problems with users reporting that they are unable to type in 'R' on their new Firefox 66.

hkollmann:
Done  with #9644

PR #9624, opened 2019-02-05
cherry picked PRs from master to branch_5_0_x

level420:
This started with a simple attempt to fix #9629, but escalated into a mass cherry-picking PR from master, trying to succeed the testrunner runs on chrome and firefox.

PR #9644, opened 2019-03-22
Firefox 66 handles keypress key/charCode the same way as other browsers

ronkie:
Firefox 66 handles keypress key/charCode the same way as other browsers. Fix for #9643

### DIFF
--- a/framework/source/class/qx/event/handler/Keyboard.js
+++ b/framework/source/class/qx/event/handler/Keyboard.js
@@ -446,10 +446,18 @@ qx.Class.define("qx.event.handler.Keyboard",
 
       "gecko" : function(domEvent)
       {
-        var charCode = domEvent.charCode;
-        var type = domEvent.type;
+        if(qx.core.Environment.get("engine.version") < 66) {
+          var charCode = domEvent.charCode;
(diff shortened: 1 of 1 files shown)


cboulanger (approved):
Thank you.

johnspackman (approved):

hkollmann (approved):

derrell (approved):

oetiker (approved):

level420 (approved):

miketaylr:
Thanks for fixing this!

@level420 Is there any ETA when https://github.com/qooxdoo/qooxdoo/releases/tag/v5.0.3-beta will be released as stable? (If a 5.0.3 is planned).

cboulanger:
There is one issue that hasn't been resolved: https://github.com/qooxdoo/qooxdoo/issues/9655#issuecomment-482163616

We're also still working on a back port of other issues that might or might not be included into 5.03:

https://github.com/qooxdoo/qooxdoo/pull/9624

Since no-one in the developer team still uses v5, we're happy about testing and input on the issue. 

However, the best way forward is to try and use master (v6) and see if it works with your app.

PR #9625, opened 2019-02-06
Add support for hi-res images to AbstractImage cellrenderer

hkollmann:
Based on https://github.com/qooxdoo/qooxdoo/pull/9610. Moved findHigh… ResolutionSource and _getHighResolutionSource to qx.util.ResourceManager.

PR #9648, opened 2019-03-25
use image size for Boolean table cell render, up to max (and default) 11x11

### LINKED ISSUE #9646: Regression with images in Table
Reference commit 0937f6c9ef8bf14ca05a9f778001b8d9290b0804: Add support for hi-res images to AbstractImage cellrenderer (#9625)

The correct  behavior and new, incorrect can be seen in these two images. The incorrect behavior was introduced with the above commit.

derrell:
Fixes #9646. An image in the Boolean cell renderer was traditionally placed in an 11x11px box regardless of actual image size, and placed starting at the top-right corner of that 11x11px box. With the recent changes proposed in #9610 and merged in #9625, images smaller than 11x11 used in the Boolean cell renderer were stretched to fit the 11x11 box, causing a backward-compatibility break.

This change attempts to retrieve the actual image size, and use that size for the Boolean indication. If the actual image size can't be determined, as might be the case with an unmanaged image, the original 11x11 size is used.

### DIFF
--- a/framework/source/class/qx/ui/table/cellrenderer/Boolean.js
+++ b/framework/source/class/qx/ui/table/cellrenderer/Boolean.js
@@ -131,12 +131,39 @@ qx.Class.define("qx.ui.table.cellrenderer.Boolean",
     // overridden
     _identifyImage : function(cellInfo)
     {
-      var imageHints =
+      var w;
+      var h;
+      var rm;
+      var id;
+      var ids;
+      var imageHints;
+
+      // Retrieve the ID
+      rm = qx.util.ResourceManager.getInstance();
+      ids = rm.getIds(this.__iconUrlTrue);
+
+      // If ID was found, we'll use its first (likely only) element here.
+      if (ids)
       {
-        imageWidth  : 11,
-        imageHeight : 11
+        id = ids[0];
+
+        // Get the natural size of the image
+        w = rm.getImageWidth(id);
+        h = rm.getImageHeight(id);
+      }
+
+      // Create the size portion of the hint.
+      //
+      // The traditional (fixed) size of the image was 11x11px. Use that if we
+      // weren't able to retrieve the actual size of the image, and never
+      // exceed that size.
+      imageHints =
+      {
+        imageWidth  : w ? Math.min(w, 11) : 11,
+        imageHeight : h ? Math.min(h, 11) : 11
       };
 
+      // Add the URL portion of the hint
       switch(cellInfo.value)
       {
         case true:


cboulanger (approved):

hkollmann (approved):

framework/source/class/qx/ui/table/cellrenderer/Boolean.js
```diff
@@ -131,12 +131,39 @@ qx.Class.define("qx.ui.table.cellrenderer.Boolean",
...
+      }
+
+      // Create the size portion of the hint.
+      //
+      // The traditional (fixed) size of the image was 11x11px. Use that if we
+      // weren't able to retrieve the actual size of the image, and never
+      // exceed that size.
+      imageHints =
+      {
+        imageWidth  : w ? Math.min(w, 11) : 11,
```
oetiker: this means there can never be a boolean image larger than 11 pixel in size ?

derrell: It means that if the image size is larger than 11x11, we revert to using the prior behavior of specifying 11x11. I believe this is the best backward-compatibility solution. Previously, it was hard-coded at 11x11, so with this change, images smaller than 11x11 render correctly even with the addition of the #9610 changes.

oetiker: as a theme developer, this seems very wrong to me ... having icons forcefully rescaled to 11x11 pixel ... one shudders ... I would suggest du dop this whole thing and just check that our existing themes look ok .... apart from that, let people shoot themselves. silently resizing seems extremely wrong.

cboulanger: But isn't @derrell argument that we just need to make sure things work as expected while accommodating @jpriebe's patch? As far as I understand the icon was force resized to 11x11 before, too. That might be a bad idea, but at least is B/C. Or did I misunderstand something?

derrell: @cboulanger is exactly correct. The code is currently broken. Images smaller than 11x11 used to work fine, but are now being stretched and look awful. This code is trying to make things work like they did before, without introducing new, non-BC changes.

It is trivial to subclass the Boolean cell renderer and override that one method, if a different behavior is desired. Anyone who needed larger images previously would have already done that.

cboulanger: @oetiker Would it be ok to accept this state for v6 and put the correction on the list of breaking changes for v7.0?

oetiker: sure as you wish ...

oetiker (approved):

johnspackman (approved):

level420 (approved):

cboulanger:
5 approvals, merging, thank you, @derrell !

PR #9649, opened 2019-03-27
Knobless splitpane

ignapas:
Allows the user to set a ``splitterVisible`` property in the splitpane to remove the splitter, still being able to resize the panes.
It is possible to set the visibility of the splitter in the theme:
```
"splitpane/splitter": {
   style: function(states) {
      return {
         visible: false
      };
   }
}
```
or programatically:
```
const splitpane = new qx.ui.splitpane.Pane("horizontal");
splitpane.getChildControl("splitter").setVisible(false);
```
invisible-splitter
In the screencast it might look that there is a splitter at the beginning, but it is actually just the padding for the inner content widget if that view.

### DIFF
--- a/framework/source/class/qx/ui/splitpane/Pane.js
+++ b/framework/source/class/qx/ui/splitpane/Pane.js
@@ -102,6 +102,13 @@ qx.Class.define("qx.ui.splitpane.Pane",
       init  : "horizontal",
       check : [ "horizontal", "vertical" ],
       apply : "_applyOrientation"
+    },
+
+    displaySplitter :
+    {
+      init: true,
+      check: "Boolean",
+      apply: "_applyDisplaySplitter"
     }
   },
 
@@ -187,7 +194,7 @@ qx.Class.define("qx.ui.splitpane.Pane",
       // is removed.
       splitter.addListener("resize", function(e) {
         var bounds = e.getData();
-        if (bounds.height == 0 || bounds.width == 0) {
+        if (this.getDisplaySplitter() && (bounds.height == 0 || bounds.width == 0)) {
           this.__blocker.hide();
         } else {
           this.__blocker.show();
@@ -269,6 +276,9 @@ qx.Class.define("qx.ui.splitpane.Pane",
       this.__setBlockerPosition();
     },
 
+    _applyDisplaySplitter : function(value, old) {
+      this.getChildControl("splitter").getChildControl("knob").setVisibility(value ? "visible" : "excluded");
+    },
 
     /**
      * Helper for setting the blocker to the right position, which depends on
@@ -299,11 +309,11 @@ qx.Class.define("qx.ui.splitpane.Pane",
         }
         var left = bounds && bounds.left;
 
-        if (width) {
+        if (width || !this.getDisplaySplitter()) {
           if (isNaN(left)) {
             left = qx.bom.element.Location.getPosition(splitterElem).left;
           }
-          this.__blocker.setWidth(offset, width);
+          this.__blocker.setWidth(offset, width || 6);
           this.__blocker.setLeft(offset, left);
         }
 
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/splitpane/Pane.js
```diff
@@ -102,6 +102,17 @@ qx.Class.define("qx.ui.splitpane.Pane",
       init  : "horizontal",
       check : [ "horizontal", "vertical" ],
       apply : "_applyOrientation"
+    },
+
+    /**
+     * The visibility of the splitter.
+     * Allows to remove the splitter in favor of other visual separation means like background color differences.
+     */
+    splitterVisible :
```
oetiker: how about calling it

`impliedSplitterMode`

then the logic of the code might be a bit easier to follow.

> **implied**
> /ɪmˈplʌɪd/
> adjective
> suggested but not directly expressed; implicit.
> "she was aware of his implied criticism"

oetiker (changes requested):

framework/source/class/qx/ui/splitpane/Pane.js
cboulanger: Hm - I would find that more difficult to understand than simply "whether the splitter is visible" - but maybe that's just me 😄

hkollmann: I would agree with @cboulanger

ignapas: I think is more important to have an accurate, self-describing name... plus that way it allows to store a simple ``Boolean``

framework/source/class/qx/ui/splitpane/Pane.js
```diff
@@ -299,11 +313,11 @@ qx.Class.define("qx.ui.splitpane.Pane",
         }
         var left = bounds && bounds.left;
 
-        if (width) {
+        if (width || !this.getSplitterVisible()) {
           if (isNaN(left)) {
             left = qx.bom.element.Location.getPosition(splitterElem).left;
           }
-          this.__blocker.setWidth(offset, width);
+          this.__blocker.setWidth(offset, width || 6);
```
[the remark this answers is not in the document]
cboulanger: @oetiker can you be more specific? I would think in the exact opposite direction (if I understand you correctly): make the 6px width part of the theme so that it can be overridden.

derrell:
Both that pixel value and `splitterVisible` should, I think, be properties with `themeable : true` so they can be modified in a theme.

framework/source/class/qx/ui/splitpane/Pane.js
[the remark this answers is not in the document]
oetiker: the size of the blocker is calculated from the width of the splitter plus the offset ... so in this setup, the offset will come from the theme while the width might be overridden by "6" if it happens to be 0 which seems odd ... and also this change goes into effect regardless of the setting of `splitterVisible`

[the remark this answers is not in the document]
ignapas: > the offset will come from the theme

But the offset is not a themeable property, it is a setting of the Pane. The width (or height) could be 0, coming from the theme. I think the original implementation is thought in a way that it could be possible to disable the resizing of the panes, setting them to 0 like that, so I wanted to respect that, unless the splitterVisible property is set to false, then understanding that the user wants to hide the splitter, but keep using the resize feature.

cboulanger (approved):

hkollmann (approved):

derrell (approved):

oetiker (approved):

level420 (approved):

schnepe2:
The "fix" introduces a (runtime-)warning, as it is "overriding" a generated property metghod!

`008327 Conflicting property method qx.ui.splitpane.Splitter.isVisible with qx.ui.core.Widget`

hkollmann:
@schnepe2 : Could you write an issue or better a PR with a fix? This PR is closed so normally hidden.

schnepe2:
Done (#9719)

Issue #9301, opened 2017-03-17
"[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive." verbose message in Chrome


firewave:
When changing the console in Chrome to verbose it will display this message. It originates from qx.event.handler.Mouse._initWheelObserver().

I am using qooxdoo 5.0.2 with Chrome 58 beta and 59 canary 64-bit on Windows 10.

johnspackman:
Passive events require that the event is guaranteed to not call `preventDefault()` on the event, but `qx.event.handler.Mouse` will use this event handler to fire the `mousewheel` event - this is a generic event and can be trapped by any application code; because the framework is not able to know in advance whether the application will choose to call `preventDefault()`, we don't know whether using a passive event will be acceptable to the application.

I'm interpreting this message as a suggestion from Chrome, rather than a warning that in future the event will be mandatory passive or something.  The main goal is more responsive UI but it's also so new that people are still figuring out how to best handle it (see [2]) in comparison with other mechanisms.

I'm not against supporting it, just cautious about the best way to handle it.

[1] http://stackoverflow.com/a/37721906

[2] http://stackoverflow.com/questions/41740082/scroll-events-requestanimationframe-vs-requestidlecallback-vs-passive-event-lis

level420:
@firewave Did you try what happens with your application if you set the passive flag to the mouse wheel event listener? Does it work as expected? This would be very interesting. Please report back.

level420:
@johnspackman Interresting SO article [2] on using `requestIdleCallback` instead of `requestAnimationFrame`. It would be wort a try on how qoodoo behaves if we change to `requestIdleCallback`. https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/client/CssAnimation.js#L198-L213

johnspackman:
Yes, I think you might be right 👍   I just read [3] which gives a better rationale as to why to support it.

It's an interesting question as to how often an app would want to prevent the scroll event from being handled; if scroll events are passive by default, we would be able to detect `preventDefault()` on passive events and alert the developer to the environment setting.

[3] https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

level420:
Additional article related to `requestIdleCallback`: https://developers.google.com/web/updates/2015/08/using-requestidlecallback

level420:
[...]

PR #9650, opened 2019-03-27
Added support for the passive option when adding native event handlers.

lucapivato:
#9301

### DIFF
--- a/framework/source/class/qx/bom/client/Event.js
+++ b/framework/source/class/qx/bom/client/Event.js
@@ -217,22 +217,27 @@ qx.Bootstrap.define("qx.bom.client.Event",
     /**
      * Checks whether the browser supports passive event handlers.
      */
-	getPassive: function () {
-      return eval("(function(){" +
-        "var passiveSupported = false;" +
-        "try {" +
-        "  var options = {" +
-        "      get passive() {" +
-        "        passiveSupported = true;" +
-        "      }" +
-        "    };" +
-        "    window.addEventListener('test', options, options);" +
-        "    window.removeEventListener('test', options, options);" +
-        "  } catch (err) {" +
-        "    passiveSupported = false;" +
-        "  }" +
-        "  return passiveSupported;" +
-      "})()");
+    getPassive: function () {
+      var passiveSupported = false;
+      try {
+        var options = Object.defineProperties(
+          {}, {
+            passive: {
+              get: function () {
+                // this function will be called when the browser
+                // attempts to access the passive property.
+                passiveSupported = true;
+              }
+            }
+          }
+        );
+        window.addEventListener("test", options, options);
+        window.removeEventListener("test", options, options);
+      }
+      catch (err) {
+          passiveSupported = false;
+      }
+      return passiveSupported;
     },
 
   },
--- a/framework/source/class/qx/bom/client/Event.js
+++ b/framework/source/class/qx/bom/client/Event.js
@@ -212,7 +212,31 @@ qx.Bootstrap.define("qx.bom.client.Event",
       catch(ex) {};
       
       return (hasAuxclick ? true : false);
-    }
+    },
+
+    /**
+     * Checks whether the browser supports passive event handlers.
+     */
+    getPassive: function () {
+      var passiveSupported = false;
+      try {
+        var options = {
+            get passive() {
+              // this function will be called when the browser
+              // attempts to access the passive property.
+              passiveSupported = true;
+            }
+          };
+
+          window.addEventListener("test", options, options);
+          window.removeEventListener("test", options, options);
+        } catch (err) {
+          passiveSupported = false;
+        }
+        return passiveSupported;
+      }
+    },
+
   },
 
   defer : function(statics) {
@@ -226,5 +250,6 @@ qx.Bootstrap.define("qx.bom.client.Event",
     qx.core.Environment.add("event.hashchange", statics.getHashChange);
     qx.core.Environment.add("event.mousewheel", statics.getMouseWheel);
     qx.core.Environment.add("event.auxclick", statics.getAuxclickEvent);
+    qx.core.Environment.add("event.passive", statics.getPassive);
   }
 });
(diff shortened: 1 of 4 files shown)


framework/source/class/qx/bom/client/Event.js
```diff
@@ -212,7 +212,31 @@ qx.Bootstrap.define("qx.bom.client.Event",
...
+    },
+
+    /**
+     * Checks whether the browser supports passive event handlers.
+     */
+    getPassive: function () {
+      var passiveSupported = false;
+      try {
+        var options = {
+            get passive() {
```
oetiker: I guess the generator will hate this ...

lucapivato:
Seems to be the oy way to check browser supoort for the passive option.

lucapivato:
The main issue is that chrome now sets passive to true by default and e.preventDefault() throws an error so it needs to be explicitly set to false for scrolling and wheel events attached to window.

framework/source/class/qx/bom/client/Event.js
cboulanger: @johnspackman Could we create an environment variable or does one exist that is set only for the compiler, and then `qx.core.Environment.select` the code so that the generator will ignore it? Or doesn't this work this way?

cboulanger: @lucapivato For qooxdoo v6, we still need to be compatible with the python generator, which does not understand modern javascript. Have you tried compiling a test app with the generator yet? If it chokes on the getter syntax, maybe you'd have to use eval() to do the test.

lucapivato: I will change it to use eval(""). I don't use the generator for the qooxdoo library in wisej, I have a single bundled minified visual studio project that builds qx.js.

lucapivato:
Don't know how to revert/delete this pull request...

lucapivato:
I have no idea what this things is doing... now it shows the latest file in my pull request on my forked repo but not here.

Anyway,

this is the eval change:

    /**
     * Checks whether the browser supports passive event handlers.
     */
	getPassive: function () {
      return eval("(function(){" +
        "var passiveSupported = false;" +
        "try {" +
        "  var options = {" +
        "      get passive() {" +
        "        passiveSupported = true;" +
        "      }" +
        "    };" +
        "    window.addEventListener('test', options, options);" +
        "    window.removeEventListener('test', options, options);" +
        "  } catch (err) {" +
        "    passiveSupported = false;" +
        "  }" +
        "  return passiveSupported;" +
      "})()");
    },

  },

johnspackman:
I'm not keen on that use of `eval` - surely there's a way to achieve this without using ES6 and without `eval`?  My thought is that qxcompiler would have translated the ES6 into ES5 before it hit the browser, and that ES5 code presumably implemented the test correctly?

Also, this code would usually be placed in a test which `qx.core.Environment.get` could determine (and cache, to avoid lots of unnecessary exceptions triggering the debugger)

cboulanger:
@johnspackman For me, this is a welcome fix to get rid of those annoying Chrome warnings. Certainly, eval() is not ideal, but qooxdoo uses it internally all over the place in the class and property system, and this would allow to avoid breaking the generator (in theory at least, are you sure @oetiker that the generator chokes on the setter syntax?). The eval() can go as soon as we work on v7 and need no longer support the generator.

cboulanger:
@johnspackman For me, this is a welcome fix to get rid of those annoying Chrome warnings. Certainly, eval() is not ideal, but qooxdoo uses it internally all over the place in the class and property system, and this would allow to avoid breaking the generator (in theory at least, are you sure @oetiker that the generator chokes on the setter syntax?). The eval() can go as soon as we work on v7 and need no longer support the generator.

derrell:
I agree that (a) we can't yet use ES6-specific features in qooxdoo proper, and (b) the use eval is undesirable, even though it's used elsewhere in the core since early in qooxdoo's history. I'd accept (b) if that's the only way to implement it, but I'd like to know exactly how qxcompiler transpiles that snippet, to see if it has found a cleaner way to do it.

johnspackman:
there's a bug @derrell 's found where qxcompiler isn't transpiling that `get property` syntax :(  but heres how it will do it (without eval or ES6) when that bug is fixed:

https://babeljs.io/docs/en/6.26.3/babel-plugin-transform-es5-property-mutators

derrell:
So for this particular issue, it seems that the fix to not use the ES6-specific code in framework/source/class/qx/bom/client/Event.js would be something like this:
```
    getPassive: function ()
    {
      var passiveSupported = false;
      try
      {
        var options =
          Object.defineProperties(
            {},
            {
              get : function passive()
              {
                // this function will be called when the browser
                // attempts to access the passive property.
                passiveSupported = true;
              }
            });

        window.addEventListener("test", options, options);
        window.removeEventListener("test", options, options);
      }
      catch (err)
      {
        passiveSupported = false;
      }

      return passiveSupported;
    },
```

lucapivato:
Is there anything I need to do on t he pending pull request? The ES5 syntax would be

```
function getPassive() {
  var passiveSupported = false;
  try
  {
    var options = Object.defineProperties(
      {},
      {
        passive : {             
          get: function()
          {
            // this function will be called when the browser
            // attempts to access the passive property.
            passiveSupported = true;
          }
        }
      }
    );
    window.addEventListener("test", options, options);
    window.removeEventListener("test", options, options);
  }
  catch (err)
  {
    passiveSupported = false;
  }

  return passiveSupported;
}

```

See https://jsfiddle.net/b0wcjn5e/

cboulanger:
@lucapivato Can you create a new PR with this solution, if GitHub doesn't play nicely with the current PR?

lucapivato:
@cboulanger looks like github updated the original PR with  the Object.defineProperties code.

hkollmann (approved):

level420 (approved):

cboulanger (approved):

johnspackman (approved):

oetiker (approved):

PR #9651, opened 2019-03-28
Doc. fix: toNativeObject returns an Object

odeimaiz:


hkollmann (dismissed):

oetiker (dismissed):

framework/source/class/qx/util/Serializer.js
```diff
@@ -122,7 +122,7 @@ qx.Class.define("qx.util.Serializer",
      * @param dateFormat {qx.util.format.DateFormat?} If a date formater is given,
      *   the format method of this given formater is used to convert date
      *   objects into strings.
-     * @return {String}
+     * @return {Object}
```
derrell: Actually, it returns a whole bunch of types. It can return `null`, `Array`, `String`, or `Object` depending on what type is being converted.

derrell (changes requested):

framework/source/class/qx/util/Serializer.js
odeimaiz: Would then {var} be more adequate?

derrell: Yes, either that, or `{null|Array|String|Object}`. A bit of text describing when each applies would be great, too.

derrell (dismissed):

hkollmann (dismissed):

oetiker (dismissed):

cboulanger (dismissed):

level420 (approved):

derrell (approved):

hkollmann (approved):

johnspackman (approved):

Issue #9653, opened 2019-03-28
Table Boolean checkbox is not displayed


marcelr:
In
> qx.ui.table.cellrenderer.AbstractImage.js #256

the line
>  "background-size": backgroundSize,

breaks displaying a checked checkbox or unchecked checkbox from

> qx.ui.table.cellrenderer.Boolean.js

Removing
>  "background-size": backgroundSize,

fixes the issue.

Wrong:
![grafik](https://user-images.githubusercontent.com/956536/55162790-d7c56f80-5168-11e9-816d-264a86347ff8.png)

Correct:
![grafik](https://user-images.githubusercontent.com/956536/55162882-09d6d180-5169-11e9-8c1c-5a641f26e267.png)

Manual fix in debugger:
![grafik](https://user-images.githubusercontent.com/956536/55162939-27a43680-5169-11e9-961e-aca3eaf9f7f9.png)

I'm using qooxdoo master from today.
Same for Firefox (here 66.0.1) oder google-chrome.

This is the background png
![grafik](https://user-images.githubusercontent.com/956536/55167056-9042e180-5170-11e9-9f0c-e01a2d3a705e.png)

derrell:
@jpriebe #9648 fixes the problem we discussed, where the image is too small and gets stretched. This one reported here is different. The actual image to be displayed is 11x11, but it's in a combined-image file. If you'd take a look, we'd appreciate it!

hkollmann:
This seems to be a Problem of using combined images.  Simple images are working correct.

hkollmann:
Done    with #9660

PR #9654, opened 2019-03-28
Table Boolean checkbox is not displayed #9653

marcelr:
See issue


hkollmann:
preferred solution is https://github.com/qooxdoo/qooxdoo/pull/9660

PR #9660, opened 2019-04-03
Use background-size only when hi-res image is used.

Holdwyn:
fixes issue #9653 
background-size was added while implementing support for high res images (to avoid clipped images?), so I guess it is safe to only add the exact size when a high res source is found.

### DIFF
--- a/framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
+++ b/framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
@@ -232,15 +232,15 @@ qx.Class.define("qx.ui.table.cellrenderer.AbstractImage",
     _getContentHtml : function(cellInfo)
     {
       var content = "<div></div>";
-
-      // background size is critical for high-resolution images
-      var backgroundSize = this.__imageData.width + "px " + this.__imageData.height + "px";
+      var backgroundSize = "auto auto";
 
       var srcUrl = this.__imageData.url;
       if (this.__imageData.url !== null)
       {
         var highResolutionSource = qx.util.ResourceManager.getInstance().findHighResolutionSource (this.__imageData.url);
         if (highResolutionSource) {
+          // background size is critical for high-resolution images
+          backgroundSize = this.__imageData.width + "px " + this.__imageData.height + "px";
           srcUrl = highResolutionSource;
         }
       }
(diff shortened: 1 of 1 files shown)


johnspackman (dismissed):

oetiker (dismissed):

framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
```diff
@@ -232,15 +232,15 @@ qx.Class.define("qx.ui.table.cellrenderer.AbstractImage",
     _getContentHtml : function(cellInfo)
     {
       var content = "<div></div>";
-
-      // background size is critical for high-resolution images
-      var backgroundSize = this.__imageData.width + "px " + this.__imageData.height + "px";
+      var backgroundSize = "auto auto";
```
hkollmann: The original code did not use background-size.

See here: https://github.com/qooxdoo/qooxdoo/commit/0937f6c9ef8bf14ca05a9f778001b8d9290b0804#diff-8fdc282bf7f5c55d884a72d38bd81948

hkollmann:
Second idea: The problem only occurs with combined icons. What's about checking if it's a combined image with qx.util.ResourceManager.getInstance().getCombinedFormat(this.__imageData.url) === "" and using background-size only in this case?

framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
Holdwyn: Yes but background-size: auto auto; is the default value. The code would become more complicated building two different style objects (one with and one without background-size).

Holdwyn:
> Second idea: The problem only occurs with combined icons. What's about checking if it's a combined image with qx.util.ResourceManager.getInstance().getCombinedFormat(this.__imageData.url) === "" and using background-size only in this case?

Does this work with the url of the image? The function has resource id as parameter.

hkollmann:
Is the same id as used in findHighResolutionSource: http://www.qooxdoo.org/current/api/#qx.util.ResourceManager

framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
hkollmann: OK

hkollmann:
Why did you close this PR?

Holdwyn:
> Why did you close this PR?

by mistake :blush:

hkollmann (approved):

cboulanger (approved):

level420 (approved):

derrell (approved):

johnspackman (approved):

Issue #9663, opened 2019-04-17
Move NPM packages to scoped names, remove separate server package


cboulanger:
Before release of v6, we should move the NPM packages to[ scope names](https://docs.npmjs.com/about-scopes):

 - `@qooxdoo/framework` replaces `qooxdoo-sdk` (the current "framework") and `qooxdoo` (the server package with a misleading name) and then contains all of the classes that can be compiled
 - `@qoxdoo/compiler` replaces `qxcompiler`

Reasons for not shipping a separate "qx.Server" package:
 -  it is conceptually easier to deal with a single distribution and reduces maintenance effort
 - even though most of the framework folder cannot be used on the server, there might be more blurring of the server-client boundary in the future
 - the size of `node_modules` does not really matter anymore these days

hkollmann:
We had this discussion before: https://github.com/qooxdoo/qooxdoo/issues/9471
We should close #9471

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo/pull/9671

PR #9665, opened 2019-04-18
ready for review: move npm package to scoped name @qooxdoo/framework

hkollmann:
See #9663

### DIFF
--- a/.travis/deploy_npm
+++ b/.travis/deploy_npm
@@ -11,7 +11,6 @@ FRAMEWORK_VERSION=$(cat $BASE_DIR/version.txt)
 TAG_VERSION="$FRAMEWORK_VERSION"
 REV=$(git rev-parse --short HEAD)
 FRAMEWORK_DATE=$(date +%Y%m%d)
-GENERATE="python $BASE_DIR/tool/bin/generator.py"
  
 # Adjust the framework version for non TAG builds
 if [ "$TRAVIS_TAG" = "" ]; then
@@ -41,13 +40,5 @@ exe npm -v
 # fill .npmrc with access token
 echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
 # publish to npm
-exe npm publish 
-
-# build qooxdoo server package
-cd $BASE_DIR/component/standalone/server 
-exe $GENERATE distclean
-exe $GENERATE build
-exe $GENERATE npm-package-copy
-exe $GENERATE npm-package-test
-exe $GENERATE npm-package-publish
+exe npm publish --access public
 
--- a/.travis/deploy_npm
+++ b/.travis/deploy_npm
@@ -37,6 +37,13 @@ exe npm -v
 # for token authentication we need newest npm!
 exe npm i npm@latest -g
 exe npm -v
+
+# get qooxdoo compiler, must be changed to @qooxdoo/compiler later
+npm install qxcompiler 
+# compile framework
+npx qx compile -v
+
+
 # fill .npmrc with access token
 echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
 # publish to npm
--- a/compile.json
+++ b/compile.json
@@ -19,7 +19,7 @@
   "applications": [
     {
       "class": "qx.application.Native",
-      "name": "qx_node",
+      "name": "qx_server",
       "type": "library",
       "environment": {
         "qx.debug": false
--- a/compile.json
+++ b/compile.json
@@ -1,0 +1,52 @@
+{
+  "targets": [
+    {
+      "type": "build",
+      "outputPath": "lib",
+      "minify": "off",
+      "babelOptions": {
+        "targets": "node >= 10"
+      }
+    }
+  ],
+  "defaultTarget": "build",
+  "locales": [
+    "en"
+  ],
+  "libraries": [
+    "./framework"
+  ],
+  "applications": [
+    {
+      "class": "qx.application.Native",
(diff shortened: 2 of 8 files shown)


cboulanger:
If we want to integrate qx.Server, don't we need to script something to achieve that (I assume that's currently done by the python scripts that you have removed).

hkollmann:
There is missing the compilation of the qx library wich is used by qxcompiler.

hkollmann:
This small changes allow to compile a server package with qooxdoo compiler and use this inside the compiler after some compiler changes: https://github.com/qooxdoo/qooxdoo-compiler/pull/373

oetiker (dismissed):
Can we make it so that 'old users' of qooxdoo-sdk get notified in some way about the name change ?

hkollmann:
We can deprecate the old npm packages.

compile.json
```diff
@@ -0,0 +1,52 @@
...
+  "defaultTarget": "build",
+  "locales": [
+    "en"
+  ],
+  "libraries": [
+    "./framework"
+  ],
+  "applications": [
+    {
+      "class": "qx.application.Native",
```
cboulanger: I wonder about the name "qx.application.Native" - wouldn't "qx.application.Server" be more fitting? What would "Native" mean in this context?

cboulanger:
Great work, thanks so much. 

For everything to be consistent, we also need to update `Manifest.json/requires`, so that it is now "@qooxdoo/compiler" and "@qooxdoo/framework" instead of "qooxdoo-sdk" and "qooxdoo-compiler" - although the old values need to be accepted, too  so that code doesn't break.

cboulanger:
We then should write a blog post to announce the latest changes.

compile.json
hkollmann: The class here is needed by the compiler and is just an dummy which will be loaded during startup. 
The native application is the app for non GUI apps and do nothing.

hkollmann:
> Great work, thanks so much.
> For everything to be consistent, we also need to update Manifest.json/requires, so that it is now "@qooxdoo/compiler" and "@qooxdoo/framework" instead of "qooxdoo-sdk" and "qooxdoo-compiler" - although the old values need to be accepted, too  so that code doesn't break.
> Can you add that?

That's should be part of the qxcompiler changes. Yes i can add it there.

cboulanger (dismissed):

.travis/deploy_npm
```diff
@@ -38,16 +37,15 @@ exe npm -v
...
+npx qx compile -v
+
+
 # fill .npmrc with access token
 echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
 # publish to npm
-exe npm publish 
-
-# build qooxdoo server package
-cd $BASE_DIR/component/standalone/server
```
johnspackman: why are we removing this - has it been replaced?

hkollmann: Yes. Server package becomes part of @qooxdoo/framework compiled by qx:


# get qooxdoo compiler, must be changed to @qooxdoo/compiler later 
npm install qxcompiler  
# compile framework 
npx qx compile -v

johnspackman (dismissed):

level420 (dismissed):

cboulanger:
> We can deprecate the old npm packages.

Seems quite easy: https://docs.npmjs.com/cli/deprecate

oetiker (dismissed):

derrell (changes requested):
When there's only one, the translation file is typically not committed to git since it is autogenerated from the source when the compiler is run. There is nothing in a framework package that can be run by itself without running the compiler, right? Is there a reason that translation files (.po) need be included here in the package?

hkollmann:
They have been there for the old toolchain.

hkollmann:
Should i revert the changes from node toolchain?

cboulanger:
Since the aren't any translations in the file, isn't this the equivalent of "no translation file exists"? Does the generator expect un untranslated .po file and throw if it is not there? Otherwise, I'd be in favor of not including it.

hkollmann:
I can revert it - but i will not delete it. I do not know what is happening with generate.py if it‘s deleted and i am not longer working with it.

hkollmann:
I reverted the en.po file to it‘s original version - so you must review again

derrell (approved):

level420 (approved):

derrell (approved):

johnspackman (approved):

oetiker (approved):

cboulanger:
How do we handle deprecation of the "qooxdoo-sdk" framework? As discussed in the compiler repo, we probably should create a branch off the latest master before merging this PR, and then modifying this branch to display the deprecation notice in its readme.md.

cboulanger (approved):

hkollmann:
I will deprecate the npm package after @qooxdoo/framework is working. IMHO we do not need a branch in qooxdoo because we have only changes in manifest.json.

Von meinem iPad gesendet

hkollmann:
Do you have a good name for the latest master?

cboulanger:
@hkollmann I'd call it `v6_0_0_alpha` since the new package will (eventually) be ^beta - does that make sense?

hkollmann:
Should be the last Legacy 5.0.x package. I thought master contains the 6.0.x version 

Von meinem iPhone gesendet

hkollmann:
@cboulanger what´s about 5.0.x_legacy_branch ?

cboulanger:
5 Approvals, enough days have passed, merging. Thank you @hkollmann

PR #9668, opened 2019-05-06
Allow access to the whole userData map

TechplexEngineer:
As a convenience allow the user to access a copy of the userData map


framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,25 @@ qx.Class.define("qx.core.Object",
     /**
      * Load user defined data from the object
      *
-     * @param key {String} the key
+     * @param key {String?} the key
+     *   If undefined, pass the user a copy of the entire map
```
derrell: It's likely that someone would prefer to pass `null` or `false` here to explicitly indicate that the entire map is requested. The code already allows that. The documentation should probably say something like, `If not a string, return a copy of the entire map.`

framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,25 @@ qx.Class.define("qx.core.Object",
...
       }
-      var data = this.__userData[key];
+      if (typeof key !== 'undefined')
+      {
+        data = this.__userData[key];
+      }
+      else
+      {
+        // give the caller a copy to protect our internal copy
+        data = JSON.parse(JSON.stringify(this.__userData));
```
derrell: What if the user has stored something here that can't be JSON stringified? I agree that a copy of the user data should be returned, but I see cases where it's not possible. I'm not sure of the right solution to this.

derrell (changes requested):

framework/source/class/qx/core/Object.js
derrell: Or better yet, `If falsy, return a copy of the entire map."

TechplexEngineer: I am concerned that someone might use something that is falsy as a key in the map. Consider
```
obj.setUserData(0, 'test');
```

TechplexEngineer: Now that I've changed that code I think it was better before. In the setUserData() function there is nothing to require that the keys of the map are strings.

I posit that the check should be `if (typeof key === "undefined") ... return map`

derrell: I would venture that that's an oversight in the code, as the documented behavior is that the key must be a string... and using a number as a key in a map implicitly converts it to a string.

Well in any case, how about we at least allow `undefined` or `null` to mean return the map (and document it as such)?

framework/source/class/qx/core/Object.js
oetiker: maybe a second parameter could specify if a copy is to be provided or not ... and then the copying could not only happen on full export but also on a single value

cboulanger:
@TechplexEngineer Would you update the PR as per @oetiker's suggestion? Then we could quickly move towards merging it.

framework/source/class/qx/core/Object.js
TechplexEngineer: I believe this has been resolved in the latest changes.

framework/source/class/qx/core/Object.js
TechplexEngineer: Parameter added in d3e4d0b

TechplexEngineer:
@cboulanger I have made the requested change.

framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,31 @@ qx.Class.define("qx.core.Object",
     /**
      * Load user defined data from the object
      *
-     * @param key {String} the key
+     * @param key {String?} the key
+     *   If key is not a string, pass the user the entire map
+     * @param bCopy {Boolean?}
+     *   If bCopy is truthy pass the user a deep clone
      * @return {Object} the user data
```
cboulanger: Can you update the documentation of the return value? Something like
"Returns null if no user data exists or if the given key has not been set, the user data map as an object if requested, or the value of the user data with the given key"
either in the JSDoc header or the `@return` tag.

TechplexEngineer: updated as requested

cboulanger (approved):

johnspackman (approved):

hkollmann (approved):

framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,34 @@ qx.Class.define("qx.core.Object",
     /**
      * Load user defined data from the object
      *
-     * @param key {String} the key
-     * @return {Object} the user data
+     * @param key {String?} the key
+     *   If key is not a string, pass the user the entire map
```
oetiker: why check for it being a string, so if one enters a number the full data is returned .. that seems odd ... shouldn't it be that only if 'null' is entered that the full data is returned.

TechplexEngineer: Even though the docstring specifies that the parameter is a string? Changes made.

oetiker: then you would have to error out if someone entered something other than a string ... I think with this change we are good.

oetiker (approved):

framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,30 @@ qx.Class.define("qx.core.Object",
...
+     *   not been set, the user data map as an object if requested, or
+     *   the value of the user data with the given key
      */
-    getUserData : function(key)
+    getUserData : function(key, bCopy)
     {
-      if (!this.__userData) {
-        return null;
+      var data;
+      if (this.__userData == null) {
```
hkollmann: are your shure that you mean this.__userData in this place? IMHO it should be key...

TechplexEngineer: Oops good catch! Fixed.

framework/source/class/qx/core/Object.js
```diff
@@ -230,15 +230,30 @@ qx.Class.define("qx.core.Object",
...
+     *   not been set, the user data map as an object if requested, or
+     *   the value of the user data with the given key
      */
-    getUserData : function(key)
+    getUserData : function(key, bCopy)
     {
-      if (!this.__userData) {
-        return null;
+      var data;
+      if (key == null) {
```
derrell: If we care, since this is `==` rather than `===`, this will also match if key is `""` (the empty string) or `0` or `false`. The latter two aren't likely a problem. I'm not sure whether an empty string key is considered a valid key or instead should return the whole map. I'll approve anyway, but others can may consider this.

derrell (approved):

framework/source/class/qx/core/Object.js
hkollmann: What's about  (key === null)  ||  (key === undefined)  instead? Then it would be absolutely clear what you mean and will not break code if someone has used 0 as key.

derrell: I think that's how I'd probably do it.

hkollmann: And the check   if (!this.__userData) { is necesary, see faild test.

hkollmann (changes requested):

cboulanger:
@TechplexEngineer Can you fix the code so the tests pass?

TechplexEngineer:
I don't want this to block the upcoming release.

cboulanger:
@TechplexEngineer You wouldn't. Even though we'd like to release ASAP, it will take a few more weeks or even months before we're ready. You're PR is welcome and certainly almost ready...

PR #9670, opened 2019-05-09
add some ignores so that qxcompiler is happy again

hkollmann:
see https://github.com/qooxdoo/qooxdoo-compiler/issues/392

### DIFF
--- a/framework/source/class/qx/Promise.js
+++ b/framework/source/class/qx/Promise.js
@@ -56,6 +56,15 @@
  *  </pre>
  * 
  */
+
+/* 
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

cboulanger (approved):

cboulanger:
merged, thank you @hkollmann

cboulanger:
Sorry, I didn't pay attention to the fact that this is the qooxdoo repo and not the compiler repo! I hope this is ok since it is about documentation only. We need to reinstate the merging rules.

Issue #9674, opened 2019-05-15
Theming doesn't support background images with spaces in their names


mrfeet:
*When creating a custom theme and referencing background images with a space in their name, these images are not shown.

The problem is in qx.bom.element.Background.js, that doesn't take spaces in image urls into account:

var backgroundImageCssString;
if (this.__isBase64EncodedImage(backgroundImageUrl)) {
backgroundImageCssString = "url('" + backgroundImageUrl + "')";
} else {
backgroundImageCssString = "url(" + backgroundImageUrl + ")";
}

So only for base64 urls the url is quoted. But this quoting also needs to be done when the url contains a space.

So my suggestion is to remove the check for base64encoding and always include the quotes.

I will supply a pull request for this.

hkollmann:
Done with #9675

PR #9675, opened 2019-05-15
Theming doesn't support background images with spaces in their names

mrfeet:
#9674

### DIFF
--- a/framework/source/class/qx/bom/element/Background.js
+++ b/framework/source/class/qx/bom/element/Background.js
@@ -80,23 +80,6 @@ qx.Class.define("qx.bom.element.Background",
 
 
     /**
-     * Checks if the given image URL is a base64-encoded one.
-     *
-     * @param url {String} image url to check for
-     * @return {Boolean} whether it is a base64-encoded image url
-     */
-    __isBase64EncodedImage : function(url)
-    {
-      var String = qx.lang.String;
-
-      // only check the first 50 characters for performance, since we do not
-      // know how long a base64 image url can be.
-      var firstPartOfUrl = url.substr(0, 50);
-      return firstPartOfUrl.startsWith("data:") && String.contains(firstPartOfUrl, "base64");
-    },
-
-
-    /**
      * Compiles the background into a CSS compatible string.
      *
      * @param source {String?null} The URL of the background image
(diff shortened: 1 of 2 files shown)


johnspackman (dismissed):

cboulanger (dismissed):

derrell (dismissed):
I'd sure like to know the reason that data URLs were being treated differently before. It's as if non-data URLs couldn't be quoted, so only data URLs were. This has the potential to break existing code if there are cases were quoting the URL causes problems.

johnspackman:
this may explain it: https://stackoverflow.com/questions/2168855/is-quoting-the-value-of-url-really-necessary

Apparently "older browsers" request once with quotes and once without

cboulanger:
Since we don't support older browsers anymore and the standard seems to be to recommend quotes, the PR seems sensible. 

One of the posts says

"The unquoted version is only supported for legacy reasons and needs special parsing rules (for escape sequences, etc.), thus being cumbersome and not supporting url-modifiers."

hkollmann (dismissed):

oetiker (dismissed):

hkollmann:
You Need to fix the tests:

https://travis-ci.org/qooxdoo/qooxdoo/jobs/532813235:


Firefox 66.0.0 (Linux 0.0.0) qx.test.bom.element Background:testGetStyles FAILED

	Do not use "'" for image urls!: Expected 'url(foo/bar/baz.gif)' but found 'url('foo/bar/baz.gif')'!

		qx.core.AssertionError.constructor:326

		qx.core.Assert.__fail:1179

		qx.core.Assert.assertEquals:4808

		qx.type.BaseString.assertEquals:779

		qx.test.bom.element.Background.testGetStyles:7

		qx.dev.unit.TestFunction:1450

		qx.dev.unit.TestResult.run:5892

		qx.dev.unit.TestFunction.run:1287

		.runTest://travis-job-1a6258c3-1d7f-486b-b9c2-81d8884a92f0:9876node_modules/karma-qooxdoo/qooxdoo-adapter.js

.

Firefox 66.0.0 (Linux 0.0.0) qx.test.bom.element Background:testCompile FAILED

	Compiled CSS string is not valid!: Expected 'background-image:url(foo/bar/baz);background-position:0 0;background-repeat:no-repeat;' but found 'background-image:url('foo/bar/baz');background-position:0 0;background-repeat:no-repeat;'!

		qx.core.AssertionError.constructor:326

		qx.core.Assert.__fail:1179

		qx.core.Assert.assertEquals:4808

		qx.type.BaseString.assertEquals:779

		qx.test.bom.element.Background.testCompile:7

		qx.dev.unit.TestFunction:1450

		qx.dev.unit.TestResult.run:5892

hkollmann:
@mrfeet : Is it possible for you to fix the tests?

hkollmann:
@mrfeet : if it's not possible for you to fix - should i commit to your repos?

mrfeet:
I’ll give it a go!

hkollmann (approved):

cboulanger (approved):

derrell (approved):

oetiker (approved):

level420 (approved):

level420:
Full quorum. mergeing
thank you @mrfeet

PR #9677, opened 2019-05-27
fix ignores for @qooxdoo/compiler

hkollmann:
tthis will remove `qx.dev.unit.Sinon: [4335,23] Unresolved use of symbol setTimeout.clock`


cboulanger (dismissed):

cboulanger:
@johnspackman Would there be a way that the compiler could ignore these kind of symbols by default (global vars, NodeJS API), so that this doesn't have to be done manually each time?

hkollmann:
Please not. In Webapps node vars are forbidden. I just integrated a global ignores section in compile.json for this purpose.

oetiker (dismissed):

level420 (dismissed):

johnspackman (dismissed):

cboulanger (dismissed):

framework/source/class/qx/dev/unit/Sinon.js
```diff
@@ -74,8 +70,10 @@ qx.Bootstrap.define("qx.dev.unit.Sinon",
 
 
 /**
- * @ignore(module, require, global, process.*, setImmediate)
+ * @ignore(module, require, global, process.*, setImmediate.*)
```
derrell: `setImmediate` is a function, not an object with members. Why does it need `.*` here? Seems wrong.

framework/source/class/qx/dev/unit/Sinon.js
```diff
@@ -74,8 +70,10 @@ qx.Bootstrap.define("qx.dev.unit.Sinon",
...
  * @ignore(msSetImmediate.*)
+ * @ignore (process.*) 
+ * @ignore (setTimeout.*)
```
derrell: Ditto.

derrell (changes requested):

cboulanger (approved):

framework/source/class/qx/dev/unit/Sinon.js
hkollmann: Done

derrell (commented):
I just looked carefully at this file to see what's going on. Firstly, msSetImmediate is not used anywhere in the file, so that `@ignore` can be removed entirely. 

Secondly, it's referencing `setTimeout.clock`. I've never heard of `setTimeout` having a `clock` property, and neither has the interweb. What the heck is this code doing?!?

hkollmann:
I really do not know. It's from the qooxdoo used sinon package. I only want to get away the warnings during compile.

derrell (approved):
This file needs review to figure out what it's doing. In the mean time, the proposed change doesn't hurt, and will remove the warning.

johnspackman (approved):

oetiker (approved):

level420 (approved):

level420:
full quorum. merging
thank you @hkollmann

PR #9678, opened 2019-05-30
A first try at handling async tests

oetiker:
This is the first try at handling async tests ... it works (somewhat) but unfortunately asserts in async tests cause odd behavior. while they do get handled properly, they ALSO spill an error message into the console. As well as confusing the tests system into thinking the test had already completed while it had not.

`test02`  demonstrates the problem. also, see https://github.com/qooxdoo/qxl.testtapper/pull/7

```javascript
function resolveAfter1Seconds() {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve('resolved');
    }, 1000);
  });
}

qx.Class.define("qxl.testtapperdemo.test.Test09AsyncFunction", {
  extend: qx.dev.unit.TestCase,
  members: {
    "test01: async function": async function () {
      let value = await resolveAfter1Seconds();
      this.assertEquals(value,'resolved',"Timeout Resolved");
    },
    "test02: async function fail": async function () {
      let value = await resolveAfter1Seconds();
      this.assertEquals(value,'unresolved',"Timeout Resolved");
    },
    "test03: async function exception": async function () {
      let value = await resolveAfter1Seconds();
      unknownFunctionCall();
    }
  }
```


framework/source/class/qx/dev/unit/TestFunction.js
```diff
@@ -109,16 +109,30 @@ qx.Class.define("qx.dev.unit.TestFunction",
...
         try {
-          inst[method]();
+          ret = inst[method]();
         } catch (ex) {
           throw ex;
         }
+        if (ret && ret.constructor && ret.constructor.name === "Promise") {
+          ret
+          .then(
+            ()=>inst.resume()
```
level420: It seems that the linter job in travis fails on the arrow function syntax.
It may be odd, but could you please change this to explicit function declarations?
Or should the linter be adapted to accept ES6?

level420 (changes requested):

hkollmann:
We can not adopt the phyton based linter- so we should change to function

Von meinem iPhone gesendet

framework/source/class/qx/dev/unit/TestFunction.js
oetiker: fixed

oetiker:
if inside an async function there is an await line, this will not cause the usual `this.wait()` call to be issued, and then when the whole thing fails l8r, due to an assert the context is gone since one would have to call `this.resume` explicitly after 'suspending' a function with `wait` ... a bit at a loss here as to what should be done.

oetiker:
maybe I should approach this differently, actually going async all the way for the appropriate functions, instead of trying to fit them into the normal 'mould' ... adding async support to testtapper

cboulanger:
I am afraid you will probably have to do that. I had a similar issue once and tried to get around it with .then(), but it turned out not feasible, so I had to rewrite the logic in an async way...

level420 (approved):

hkollmann (approved):

johnspackman (approved):

derrell (approved):

cboulanger (approved):
@oetiker Can you update the PR?

hkollmann:
full approval, merged, Thanks @oetiker !

oetiker:
> @oetiker Can you update the PR?

yes ... I will get back to this ...

PR #9679, opened 2019-06-01
Fix deploy to npm

hkollmann:


cboulanger:
Shouldn't we add ourselves to the list of authors? And shouldn't we use this opportunity to move to 6.0.0-beta?

Manifest.json
```diff
@@ -91,7 +91,7 @@
   },
   "requires": {
     "@qooxdoo/compiler": "^1.0.0-beta",
-    "@qooxdoo/framework": "6.0.0-alpha"
+    "@qooxdoo/framework": "6.0.0-beta"
```
cboulanger: Does the framework require itself here?

oetiker (dismissed):

Manifest.json
hkollmann: Not the framework. But the schema requires this. And the Compiler requires a Manifest.json. Without you get an error.

cboulanger (dismissed):

derrell (dismissed):

Manifest.json
cboulanger: Ok, understood. But it should probably be `"@qooxdoo/framework": "^6.0.0-beta"`  (vith ^), so that you don't have to update it when we release 6.0.0...

hkollmann: Version will be set to current version during deployment. See tool/admin/bin/bumpqxversion.py

cboulanger: Duh. Of course! 😄

johnspackman (dismissed):

hkollmann:
Version in requires must match the Version wich is used in current compoiler. this is not equal to the Version of the lib

cboulanger (approved):

level420 (approved):

johnspackman (approved):

oetiker (approved):

derrell (approved):

cboulanger:
full quorum, merging. Thank you @hkollmann !

PR #9680, opened 2019-06-01
documentation merge

oetiker:


cboulanger (approved):

hkollmann (approved):

cboulanger:
I welcome merging @oetiker's new documentation into a qooxdoo repository - thanks again for having created it! - and I really like the ability to version the documentation with the code. Just two things that make me wonder if it's the right way to go:

1. we should have a lower quorum for merging documentation PRs since 5 approvals is time consuming and not really necessary for documentation updates. However, it seems not be possible to have two different quorum rules within one repository. On the other hand, it is a good thing if the 5 approval rule is enforced automatically. 

2. For each PR, the full qooxdoo test suite is run, despite the fact that documentation changes do not need such a run at all. You can suppress the tests with `[ci skip]` in the commit message, but that's easily forgotten and doesn't seem to work if you have several commits. The documentation should have a different kind of test, namely one that rebuilds the documentation website and checks for broken links. 

I am fine with merging, but I really wonder if a separate repo wouldn't be a better idea, even if synchronizing versions becomes much more difficult.

oetiker:
@cboulanger for 1. we could lower the configured requirement back down again and for 2. we can add code to the travis build script to detect when there is only a documentation update and skip building all.

cboulanger (approved):

cboulanger:
Ok, see https://github.com/qooxdoo/qooxdoo/issues/9683

derrell (approved):

johnspackman (approved):

cboulanger:
@oetiker can you lower the approval requirements so that we can merge this PR? @level420 seems to be busy...

oetiker:
done

cboulanger:
More than 2 approvals (documentation), merged, thank you @oetiker !

PR #9681, opened 2019-06-01
Xhr with promise

voger:
This is a first attempt to make qx.io.request.Xhr to return a qx.Promise. On promise resolution it returns the xhr object as a parameter to then's callback. On rejection it throws a qx.type.BaseError with the phase as a comment and the "error code: error message" as failMessage, if the request fails with statusError or an empty string.


voger:
My editor's formatting has changed more of the files that I wanted to change. I will see to fix that and create a new PR.

PR #9682, opened 2019-06-02
Add ability to return a qx.Promise

voger:
This is a first attempt to make `qx.io.request.Xhr` to return a `qx.Promise`. 

On promise resolution it returns the xhr object as a parameter to then's callback. 

On rejection it throws a qx.type.BaseError with the phase as a comment. If the request fails with `statusError`, the `qx.type.BaseError` object has "error code: error message" as failMessage. If it fails for some other reason, e.g. `timeout` the `failMessage` is just an empty string.


hkollmann:
Great!

Would it be possible to implement this method in qx.io.request.AbstractRequest? 
So qx.io.request.Jsonp would have the same benefit?

And could you suround the method with 

      if (qx.core.Environment.get("qx.promise")) {

Your method...

      } else {
        this.skip("Skipping because qx.promise==false");
      }

Because promise support can be disabled through enviroment setting.

Thanks for sharing code again!

hkollmann:
Question to all: I like sendSync better than sendWithPromise. What do you think?

zaucker:
sendP() is used other places.

Fritz Zaucker
Platanen 42
4600 Olten
+41 62 212 7712
+41 79 675 0630

voger:
@hkollmann thanks for the feedback. I will try to implement it as AbstractRequest and I will surround it with the conditional.

I opted for the name `sendWithPromise` because `qx.io.request.Xhr` already has a property named `async` that controls "Whether the request should be executed asynchronously", and I believe it would be confusing as to what `sendAsync` refers to. 

Also, I would like some feedback as to what is would be best to be returned in each case. Right now it returns the xhr object on resolution and a qx.type.BaseError() with a string like "statusError: 404: Not found" on rejection. Are maybe there better options?

cboulanger:
What about instead of adding a method, you create a new subclass `AsyncRequest` (since XHR is really old terminology), which doesn't have the `sendAsync` property and of which the `send` method returns a promise by default? You could put some of the common methods in a mixin or into the abstract class.

voger:
@cboulanger In that case the `qx.io.request.Xhr` API would stay the same. It wouldn't have the ability to return a promise. Also `qx.io.request.Jsonp` wouldn't share the ability to return the promise either.  

Another idea I just got:

Right now the `qx.io.request.AbstractRequest#send` method doesn't return anything. What if I change it so it returns the promise? I know that it modifies working API but since nobody expected a return value from that method, how would it affect already working code?

hkollmann:
Good idea. Should´nt be a breaking change. 

Von meinem iPad gesendet

hkollmann:
Promise only if async is true!

cboulanger:
@voger You're right, it makes more sense to add Promise support to all of the request types.

I like the Promise as return value but since it would create useless Promise objects to any code that works with events, maybe an extra property is in order, such as  a boolean `returnPromise`-property?

voger:
@cboulanger yes indeed. It would create a useless object for someone who doesn't need it. 


I'm thinking that adding another check to the `send()` function would add another asterisk in it's use making it a bit more complicated  to reason about. It would return a promise if promises are enabled and if the property is set to true, otherwise it would return `undefined`(?). In case of synchronous communication it would return a settled promise. I am afraid all this would increase usage complexity for the `send()` function.

Maybe simply adding a method that returns a promise would be much simpler to use. What do you think?

cboulanger:
@voger Probably you are right and the most straightforward approach is to leave `send()` as it is as a relic of the ancient Promises-less ways and just use a slightly more verbose method name to move into the future. 😄

cboulanger (dismissed):

johnspackman (approved):

voger:
@hkollmann I will move it to AbstractRequest but I will need few days so I can also asjust the tests properly.

johnspackman (dismissed):

hkollmann (changes requested):
Should be moved to AbstractRequest. 
@voger  will do this. So this should block merging up to that.

oetiker:
I have added a label to make it clear that this one should not be merged yet

derrell (changes requested):
I think this is being moved to the abstract class. Awaiting that change.

voger:
Yes. I moved it already to the abstract class, added ability to reject the promise on abort it and I am adding a listener for `parserError`. I believe today or tomorrow I will push the updates.

hkollmann (approved):
That's it! Thanks @voger !

cboulanger (approved):

derrell (approved):

voger:
Can we please postpone merging this? I think I found a bug. If the user attaches another `finally()` function in the chain, then the cleanup `finally()` code is not called. I am working on a fix for this.

hkollmann (changes requested):
@voger founds a bug. Hold back to resolve this

voger:
I rewrote this function so it desn't depend on a call to `finally()` for cleanup.

* Now every event listener performs the cleanup if it can settle the promise. 
* Also the event listener doesn't listen for separate error events anymore. We have a single listener for `changePhase`, and depending on the phase, it acts upon or ignores the event. This minimizes the number of event listeners that need to be registered/removed every time.
* `parseError` is handled as a special case because it is not a phase and can't be listened with `changePahse`

cboulanger:
@voger once the tests pass, is this ready for merging?

voger:
@cboulanger Yes. It is ready.

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

cboulanger:
Full quorum, merging. Thank you, @voger !

Issue #9683, opened 2019-06-03
Skip travis tests when a PR/commit concerns only documentation and check markdown instead


cboulanger:
When a PR is created or merged that contains only changes to documentation, the Travis job should skip all unit tests and run a script instead that checks the validity of the markdown markup and of all links in the documentation.

cboulanger:
Here are some tools that might be useful for "testing" the documentation 👍 
- https://www.npmjs.com/package/markdownlint
- https://www.npmjs.com/package/markdown-link-check
- https://github.com/remarkjs/remark-lint
- https://www.npmjs.com/package/eslint-plugin-markdown
- https://textlint.github.io/

cboulanger:
[This script](https://github.com/dev-id/Magic-Spoiler/blob/8111a06ab6682e020169991d5e2aa4fa503d787f/preflight.sh) uses the undocumented [`travis_terminate`](https://medium.com/@manjula.cse/how-to-stop-the-execution-of-travis-pipeline-if-script-exits-with-an-error-f0e5a43206bf) feature. We could run the documentation lister and then exit with a the result of the lint.

cboulanger:
Implemented with #9694

PR #9687, opened 2019-06-05
Update travis to skip tests for documentation commits

cboulanger:
This will skip tests for commits/PRs that contain "documentation" or "docs" either in the commit message or in the branch name. This is necessary so that documentation updates do not have to wait for completely needless tests to finish.


cboulanger:
Unfortunately, it doesn't seem to be possible for Travis to check the paths of the files that are part of the commit. This would have allowed a completely automatic solution. With this, you still have to make sure to include "docs" or "documentation" in either the branch name or all commit messages of the commits that are pushed.

derrell (dismissed):
I'm concerned about this. I think we're going to end up with false positives on commit messages... But I'm not sure what to do with it if it's not otherwise possible to simply look at the paths. Sigh.

cboulanger:
We could also check the branch name only and not look at the commit message, if that seems safer to you.

johnspackman:
sorry if this is a dumb question because I simply don't understand the travis syntax, but does this mean that if I have a commit message which is "fixed bug with blah and updated documentation to match" I would hit this travis rule?

I've no problem with these kind of parsing filters, but maybe we should use a specific syntax - eg "[travis-docs-only]"

cboulanger:
@johnspackman You're right, that kind of commit message would match the filter. Maybe we require a more specific code word like the one you suggested. I'll fix it.

oetiker (changes requested):
another option would be to use 

https://github.com/ITISFoundation/osparc-simcore/blob/master/ops/travis/helpers/test_for_changes

this has the advantage that it actually tests which files have changed. relying on comments seems a bit error-prone and might lead to tests not running by accident.

you can see the script in use here

https://github.com/ITISFoundation/osparc-simcore/blob/master/ops/travis/unit-testing/frontend

and ultimately like this in the `.travis.yml`

```yml
# frontend testing ----------------------------------------------------------------------
    - stage: build / unit-testing
      name: front end testing
      language: node_js
      node_js:
        - "10"
      addons:
        apt:
          packages:
            - libgconf-2-4

      before_install:
        - bash ops/travis/unit-testing/frontend before_install
      install:
        - bash ops/travis/unit-testing/frontend install
      before_script:
        - bash ops/travis/unit-testing/frontend before_script
      script:
        - bash ops/travis/unit-testing/frontend script
      after_success:
        - bash ops/travis/unit-testing/frontend after_success
      after_failure:
        - bash ops/travis/unit-testing/frontend after_failure

      deploy:
        # https://docs.travis-ci.com/user/deployment/pages/
        provider: pages
        target-branch: master
        # Set in travis-ci.org dashboard, marked secure
        github-token: $GH_PAGES_TOKEN
        skip-cleanup: true
        keep-history: false
        local-dir: itisfoundation.github.io
        repo: ITISFoundation/itisfoundation.github.io
        verbose: true
        on:
          repo: ITISFoundation/osparc-simcore
          branch: master
```

cboulanger:
@oetiker How does this work with stages? Can we go with my solution first and then replace it with something better later- the current situation is bad, we need to wait hours for simple doc changes...

cboulanger:
@oetiker I like your solution because we'll be able to detect if the changes concern documentation *only* without the committer having to provide special meta information in the branch name or commit message. But as I said, I suggest we put a simple solution in place now until we find the time to write a more adequate one.

level420 (approved):

johnspackman (approved):

hkollmann (approved):
Lets commit this yet and add a better solution later!

derrell (approved):

oetiker (approved):

cboulanger:
This might be useful:
https://github.com/RReverser/react/commit/233af7e2dff4c79619f5cda6943388ff9e21581b

cboulanger:
Or this: https://github.com/dev-id/Magic-Spoiler/blob/8111a06ab6682e020169991d5e2aa4fa503d787f/preflight.sh

cboulanger:
Let's continue over here: #9683

PR #9691, opened 2019-06-11
This is a test for the new pullapprove system.

cboulanger:
If it is working, only two approvals should be required for merging.


cboulanger:
Forgot title tag to skip test

PR #9692, opened 2019-06-11
[travis-docs-only] This is a test for a documentation PR

cboulanger:
If it is working, only two approvals should be required for merging (as per .pullapprove.yml) and no tests should be skipped (as per .travis.yml).

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,32 +1,69 @@
 About
 =====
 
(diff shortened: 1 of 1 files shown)


cboulanger:
Wrong labeling, `[travis-docs-only]` must be in the commit message, not in the title. Sorry.

PR #9694, opened 2019-06-11
Test for a documentation PR

cboulanger:
[travis-docs-only]

Expected:
- test suite should not run
- only two approvals should be necessary

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -49,6 +49,9 @@ pullapprove_conditions:
   - condition: "'WIP' not in title"
     unmet_status: pending
     explanation: "Work in progress - review not yet required"
+  - condition: "'do not merge' not in label"
+    unmet_status: pending
+    explanation: "PR should not be merged yet - review not yet required"
 
 notifications:
   - when: pull_request.opened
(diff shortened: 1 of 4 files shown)


derrell (approved):

cboulanger:
The Travis test is stuck. I wonder why this is. Maybe it is because in the current scenario, non of the travis stages run, since none of their conditions are met:

https://github.com/qooxdoo/qooxdoo/blob/master/.travis.yml#L31-L39

But it seems strange to me that it would just stall, instead of erroring or passing.

derrell:
It looks like it may be deadlocked. Travis is waiting on pullapprove, and pullapprove is waiting on travis.

cboulanger:
We might have to migrate Travis integration from "GitHub Services" to "GitHub Apps" for this to work: https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/ . I think only @oetiker can do this?

cboulanger (commented):
(deleted)

johnspackman (approved):

hkollmann (approved):

oetiker (approved):

cboulanger:
This implements issue #9683 now. However, it fails with a connection error to saucelabs.

level420 (approved):

PR #9701, opened 2019-06-15
[travis-docs-only] adds documentation for https://github.com/qooxdoo/qooxdoo-compiler/pull/448

johnspackman:


hkollmann (changes requested):
We should have an hint that this is just an example! Sass is working out of the box in current Compiler so this is not needed  for mobile apps!

oetiker (approved):

johnspackman:
@hkollmann it says that this is a "working example" three times and also that it is "meant as an example only" ...  I'll make it clear that sass is already working.

hkollmann (approved):

cboulanger:
Tests fail because my documentation PR needs to be merged first which fixes most dead internal links: https://github.com/qooxdoo/qooxdoo/pull/9700.
BTW, `[travis-docs-only]` is no longer necessary, it will automatically detect PRs that commit only to `docs/*`

derrell (approved):

PR #9706, opened 2019-06-17
add CapsLock detection

peuter:
Sometimes you need to know if caps lock is active, e.g. to show the user a hint when he is typing in a password field with caps lock activated. This PR adds the caps lock state to the modifiers


cboulanger (approved):

oetiker (approved):

hkollmann (approved):

johnspackman (approved):

level420 (approved):

derrell (approved):

lucapivato:
Sorry for  the late comment. This change can break existing code when checking e.getModifiers() == 0 or != 0 when CapsLock is locked. It broke some of our code checking for the arrow keys and no modifiers.

Also, if the new feature is extended to "NumLock" and "ScrollLock" (also supported by getModifierState) you always get != 0 when NumLock is pressed which is the most common state.

In our build I have added instead getKeyLockState() with 3 masks for caps-lock, num-lock and scroll-lock.

Let me know if you'd like me to create PR to change it or if you want to keep it like this it's also fine for me.

oetiker:
Good point, since the other keys are not "Locking" I guess adding a separate call might be warranted.

hkollmann:
@lucapivato please do so!

PR #9707, opened 2019-06-17
qooxdoo npm source package should live under source.

hkollmann:
like all other packages.

This will copy framework/source to source while building package.
See https://github.com/qooxdoo/qooxdoo-compiler/issues/450

### DIFF
--- a/.travis/deploy_npm
+++ b/.travis/deploy_npm
@@ -40,6 +40,8 @@ exe npx qx --version
 exe npm update --no-save @qooxdoo/compiler 
 exe npx qx --version
 exe npx qx compile -v
+exe mkdir source
+exe cp -rf ./framework/source/* ./source
 
 
 # fill .npmrc with access token
--- a/package.json
+++ b/package.json
@@ -18,12 +18,14 @@
     "Manifest.json",
     "lib/qx_server",
     "lib/resource",
-    "framework/source",
+    "source/class",
+    "source/resource",
+    "source/translation",
     "tool/data/cldr",
     "docs"
   ],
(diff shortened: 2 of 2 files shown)


cboulanger (approved):

level420 (approved):

oetiker (approved):

johnspackman:
why does it copy instead of move?  EG would this mean that the npm package has two `./source` and `./framework/source` ?

hkollmann:
cp instead of move let the checkout intact. no package has only ./source folder. This is managed  in package.json

johnspackman (approved):

.travis/deploy_npm
```diff
@@ -40,6 +40,8 @@ exe npx qx --version
 exe npm update --no-save @qooxdoo/compiler 
 exe npx qx --version
 exe npx qx compile -v
+exe mkdir source
```
derrell: Is `source` guaranteed not to exist? If not, this should be `exe mkdir -p source` so that it creates it if it doesn't eixist,and doesn't barf if it does like the current could would do.

derrell (approved):
I added a comment that should be reviewed before merging. If the answer to my question is, "It can't happen," then go ahead and merge.

.travis/deploy_npm
cboulanger: @hkollmann what do you think ? I guess since this runs only on travis, it should be "clean"...

hkollmann:
It's on travis so it should be clean. @derrell But thanks for the hint - i will add it in the next iteration.,

PR #9718, opened 2019-06-27
Lower the approval quorum for all markdown files regardless of in which folder they live

cboulanger:


johnspackman (approved):

zaucker (approved):

.pullapprove.yml
```diff
@@ -5,8 +5,8 @@ groups:
   # documentation and tests: only 2 approvals necessary
   documentation:
     conditions:
-      - "files.include('docs/*') or files.include('documentation/*')"
-      - "not (files.exclude('docs/*') and files.exclude('documentation/*'))"
+      - "files.include('docs/*') or files.include('documentation/*') or files.include('*.md')"
+      - "not (files.exclude('docs/*') and files.exclude('documentation/*') and files.exclude('*.md'))"
```
johnspackman: I take it that this does *not* mean that if a *.md file is edited as well as other files (eg .js) ?

oetiker: the second line afaik takes care of the case where other files are in the mix

oetiker (approved):

.pullapprove.yml
cboulanger: Yes, the condition means: Mark this PR as belonging to the "documentation" group (which has a lower approval quorum) if and only if
-  the PR contains at least one file which is either in documentation folders or has the `.md` ending
- AND the PR contains NO file for which these conditions do not apply. 

That means: 
1) if there is ONLY documentation, use fast-track procedure (2 approvals)
2) if there is ANY non-documentation, use normal procedure (5 approvals)

cboulanger: I also had some difficulty to wrap my head around the boolean logic here but I tested it with a javascript equivalent.

hkollmann (approved):

derrell (approved):

PR #9722, opened 2019-06-27
fixes tutorial - it should not be necessary to specify the framework,…

johnspackman:
… but doing so will break under current qxcompiler

### DIFF
--- a/docs/tutorial/twitter/tutorial-part-1.md
+++ b/docs/tutorial/twitter/tutorial-part-1.md
@@ -7,7 +7,7 @@ The first step is to get a working qooxdoo application where we can start our de
 $ mkdir tweets
 $ cd tweets
 $ npm init
-$ npm install --save-dev @qooxdoo/compiler @qooxdoo/framework
+$ npm install --save-dev @qooxdoo/compiler
 $ npx qx create tweets --type desktop --noninteractive --out=.
 ```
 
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

hkollmann (approved):
Framework is part of the compiler!

johnspackman:
@hkollmann yes, but adding it in to the application as a dependency should not break the app.  IMHO its a bug https://github.com/qooxdoo/qooxdoo-compiler/issues/474

oetiker (approved):

cboulanger:
this circular dependency is bad. The compiler obviously must hava a dependency on the framework, but isn't the reverse only true as one of the "devDependencies"? I.e. the framework works with the python toolchain in default (until v.7) and the compiler is currently, even if we all treat it as the default toolchain, an add-on from the perspective of backwards-compatibility. It is needed to compile the NPM package only. So it could be moved to "devDependencies" Or am I wrong?

Everything will be much better once we merge the two repos.

hkollmann:
I am on the way to delete this reference in qooxdoo. I have more problems during building with this as it benefits.
In the Moment i am testing the build with my fork - will make a PR if everything works.

Testing in fork is much easier because it works without approval chain

cboulanger:
Merging, thank you @johnspackman !
Sometimes it is necessary to add a random tag just to send an event to pullaprove to update the status.

PR #9727, opened 2019-06-30
Publish docs and apps on website

cboulanger:
This publishes the new docs and the new demo apps in the `6.0.0-beta/docs` and `6.0.0-beta/apps` subfolders of the homepage, respectively.

### DIFF
--- a/.travis/build-site-v6
+++ b/.travis/build-site-v6
@@ -1,0 +1,24 @@
+#!/bin/bash
+#
+# This builds the new v6 website documentation and apps
+#
+
+set -o errexit -o nounset
+
+exe() { echo "\$ ${@/eval/}" ; "$@" ; }
+
+TARGET=deploy/$(cat ./version.txt)
+mkdir -p $TARGET
+
+[ -d ./node_modules/@qooxdoo/compiler ] || npm i --no-save @qooxdoo/compiler
+
+# build apps (unless already built) & copy to website target
+[ -d ./apps/resource ] || exe npx qx compile compile.json --target build  --feedback=0
+exe cp -a ./apps $TARGET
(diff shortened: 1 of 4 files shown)


derrell (approved):

.travis/build-site-v6
```diff
@@ -0,0 +1,24 @@
...
+
+TARGET=deploy/$(cat ./version.txt)
+mkdir -p $TARGET
+
+[ -d ./node_modules/@qooxdoo/compiler ] || npm i --no-save @qooxdoo/compiler
+
+# build apps (unless already built) & copy to website target
+[ -d ./apps/resource ] || exe npx qx compile compile.json --target build  --feedback=0
+exe cp -a ./apps $TARGET
+exe find $TARGET -type f -name '*.map' -delete
```
hkollmann: Should'nt we put this into a compile.js file so that it works for npm_deploy too?

hkollmann (commented):
Do we have a doc page with a list off all apps? Otherwise they are not reachable yet...

hkollmann (approved):

.travis/build-site-v6
cboulanger: Yes, this could be DRYed up with `$TARGET` being a CLI parameter - but let's test it "live" first, then optimize later.

johnspackman (approved):

cboulanger:
@hkollmann Not yet, but I will link to the demo apps from everywhere in the documentation once the PR is merged, and we can also add a special page in the docs.

.travis/build-site-v6
hkollmann: aggreed

oetiker (approved):

zaucker (approved):

PR #9729, opened 2019-07-01
Update links to apps in the documentation

cboulanger:
This replaces all links to the applications on the qooxdoo website in the documentation with a pseudo-protocol "apps://" prefix, which will be dynamically replaced by the docsify engine with links to the applications.

This is necessary because "real" relative links will break the markdow link validation because they refer to files/"headings" that do not exist at validation time. 

In the NPM-packaged version, these applications are pre-built. If you work with the qooxdoo source, they will be built when executing `qx serve` in the root of the qooxdoo framework folder


johnspackman (approved):

hkollmann (approved):

derrell (approved):

zaucker (approved):

oetiker (approved):
great

PR #9737, opened 2019-07-06
fix sendWithPromise() doesn't send

### LINKED ISSUE #9736: qx.io.request.AbstractRequest sendWithPromise() doesn't send
**Describe the bug**
I used the new sendWithPromise() function in a project and waited forever to settle. It turns out it wasn't sending the request.

**To Reproduce**
Use `sendWithPromise()` in a `qx.io.request.Xhr`. Wait for the result's to come. The request is not even sent. No trace of the communication in browser's network tab in the developer tools.

**Expected behavior**
Returned promise should settle after a while.

voger:
Fixes #9736 

Also refactored the `sendWithPromise()` related tests so they are not nested inside an if condition.

### DIFF
--- a/framework/source/class/qx/io/request/AbstractRequest.js
+++ b/framework/source/class/qx/io/request/AbstractRequest.js
@@ -431,7 +431,7 @@ qx.Class.define("qx.io.request.AbstractRequest",
     */
     sendWithPromise: function(context) {
       if (qx.core.Environment.get("qx.promise")) {
-        var context =  (context !== undefined && context !== null) ?
+        context = (context !== undefined && context !== null) ?
           context : this;
 
         // save this object's context
@@ -481,20 +481,22 @@ qx.Class.define("qx.io.request.AbstractRequest",
           var errorListener = req.addListener("error", function(e) {
             listeners.forEach(req.removeListenerById.bind(req));
             var failMessage = "Request failed.";
-            var err = new qx.type.BaseError("error");
+            var err = new qx.type.BaseError("error", failMessage);
             reject(err);
           }, this);
+          listeners.push(errorListener);
 
           req.send();
         }, context)
 
-        .finally(function(){
+        .finally(function() {
           if (req.getReadyState() !== 4) {
             req.abort();
           }
         });
 
         return promise;
+        // eslint-disable-next-line no-else-return
       } else {
         // fail loudly
         throw new qx.type.BaseError("Error", "Environment setting qx.promise is set to false.");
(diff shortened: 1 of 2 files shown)


derrell (approved):

cboulanger (approved):

voger:
Please don't merge it yet. I noticed it doesn't detect network errors, CORS errors etc. I am writing a new commit and I will test it more with my project. Sorry about that.

hkollmann:
I added the label "do not merge" for you!

johnspackman (approved):

voger:
It is ready now. The tests all pass and I used it in a project. It works OK. It doesn't listen to the `changePhase` event any more but to the various events that are emited during the request cycle. This provides more reliable error handling.

cboulanger:
@voger Could you fix the following two lint warnings:
```
  - Warning: qx.io.request.AbstractRequest (481, 15): Declared but unused variable or parameter: 'errorListener'
  - Warning: qx.io.request.AbstractRequest (483, 17): Declared but unused variable or parameter: 'failMessage'
```
Thank you! (It is always good to run `qx lint` before pushing).

voger:
@cboulanger I am working on it now. Thanks for the qx lint advice. I wasn't aware of it.

hkollmann (approved):

framework/source/class/qx/io/request/AbstractRequest.js
```diff
@@ -433,7 +431,7 @@ qx.Class.define("qx.io.request.AbstractRequest",
     */
     sendWithPromise: function(context) {
       if (qx.core.Environment.get("qx.promise")) {
-        var context =  (context !== undefined && context !== null) ?
+        context = (context !== undefined && context !== null) ?
```
derrell: More typically, this would be written as something like `context = context || this;` which is much simpler and easier to see what's going on. Is there a reason not to do that in this case?

derrell (approved):

derrell (changes requested):
See inline comment

cboulanger (approved):

framework/source/class/qx/io/request/AbstractRequest.js
voger: There is no particular reason for that. I was searching for what is the preferred way to set default function parameters in qooxdoo and it seemed more specific that way. 


I changed it and added two more tests for context binding. 


Another way would be using ES6 default function parameters so it would  be 
```javascript
sendWithPromise: function(context = this) {
```
but I have not seen it being used so far in framework's source so I was reluctant to use it.

level420 (approved):
@voger 👍 for integrating the case in the tests that promises are disabled or for other reasons unavailable

oetiker (approved):

framework/source/class/qx/io/request/AbstractRequest.js
cboulanger: Since we still support the python compiler, which understands ECMA5 only, newer features are not yet allowed. So you did the right thing.

voger:
@level420 Thanks. This was @hkollmann's idea.

derrell (approved):

PR #9746, opened 2019-07-18
make this compileable by qx

hkollmann:
qx do not knows the synonym qxWeb for qx.Bootstrap


derrell (approved):

johnspackman:
Will this break the qxWeb build?  Not that I have any love for it, but I think we're going to deprecate it in 6.x and not dispose of it until 7?  

However, as there's no development in qxWeb (or likely to be) if we could just archive a qxWeb distro and tie it to a branch or tag (just in case someone really needed to make a patch) that would get my vote...

hkollmann:
No web build will not break. All tests are running. And qxWeb is just set to qx.Bootstrap internal

johnspackman (approved):

cboulanger (approved):

level420 (approved):

oetiker (approved):

PR #9752, opened 2019-07-22
Update package documentation

cboulanger:


hkollmann (approved):

docs/cli/commands.md
```diff
@@ -137,6 +137,13 @@ is used.
 
 If `compile.json` does not exist, `qx lint` tries to use `.eslintrc`.
 
+If you need to turn of a rule (generally discouraged but sometimes necessary),
```
derrell: s/of/off/

docs/cli/packages.md
```diff
@@ -144,7 +144,26 @@ Options:
 ### Install a library
 
 You can then install any package from this list by executing `qx package
-install <URI>`. The URI takes the form `github_user/repository[/path]`: the first
+install <URI>`. It has the following options:
+
+```
+  --release, -r    Use a specific release tag instead of the tag of the latest 
+                   compatible release
+  [string]
```
derrell: Is this `[string]` in the wrong place? Should it be underneath or just after `compatible release`?

derrell (changes requested):

derrell (approved):

PR #9755, opened 2019-07-23
Add top-level symbolic link to release "docs" folder

cboulanger:


johnspackman (approved):

.travis/deploy_website
```diff
@@ -43,6 +43,8 @@ cp -a ../build "$TARGET"
 if [ $CURRENT = 1 ]; then
   rm -rf current &> /dev/null
   ln -s "$TARGET" current
+  rm docs &> /dev/null
```
derrell: `rm -f docs` (or `rm -rf docs` if docs can be a directory) is probably more appropriate. There's no need for the redirect to /dev/null when using `-f`.

derrell (approved):
Approved, but recommend the inline change.

hkollmann (approved):

.travis/deploy_website
cboulanger: I just looked it up: The rm utility removes symbolic links, not the files referenced by the links, i.e. the `-r` option is superfluous.

cboulanger: It's more complicated than that: https://superuser.com/questions/382314/does-rm-rf-follow-symbolic-links `rm -r current` will only delete the symlink, `rm -r current/` would also delete the content of the link target. Since the syntax here is without a trailing slash, we're safe.

derrell: Mostly, I was suggesting to use `-f` and delete the redirection to `/dev/null`. If `docs` will always be a symlink and all you want to do is remove the synlink before creating a new one, then `rm -f docs` would be the appropriate thing to do. I only mentioned `-rf` for the case where `docs` was actually a directory (not a symlink to a directory) and you wanted to remove it in preparation for creating the symlink to the other directory.

cboulanger: Yes, that's how I fixed it now - thanks!

oetiker (approved):

oetiker:
oops merged accidentally :

cboulanger:
@oetiker No, that's not possible ... pullapprove enforces the approval rules strictly.

level420 (approved):

PR #9756, opened 2019-07-24
fixes warnings

johnspackman:
fixes some warnings in line with changes in https://github.com/qooxdoo/qooxdoo-compiler/pull/493 and https://github.com/qooxdoo/qooxdoo-compiler/pull/504


level420 (approved):

cboulanger (approved):

hkollmann (approved):

hkollmann (changes requested):
old lint needs the deleted ignores!

oetiker (approved):

johnspackman:
@hkollmann fixed

hkollmann (approved):
That's it!

cboulanger:
Would such a PR, which does not change the operative code itself, qualify for being categorized as "Hotfix" or a yet-to-be created category of "Minor change" that could go with fewer than 5 approvals? I think we should be able to move faster with these kind of changes...

derrell (approved):

derrell:
I don't think a change such as this should be fast-tracked. We fast-track hotfixes because they are time-critical. What defines a "minor change" though? I'd just keep it as is for anything that doesn't require extreme speed.

PR #9757, opened 2019-07-25
Make /docs link to current master documentation

cboulanger:
at least temporarily, until we have our first release

### DIFF
--- a/.travis/deploy_website
+++ b/.travis/deploy_website
@@ -43,9 +43,11 @@ cp -a ../build "$TARGET"
 if [ $CURRENT = 1 ]; then
   rm -f current
   ln -s "$TARGET" current
-  rm -f docs 
-  ln -s "$TARGET"/docs docs
 fi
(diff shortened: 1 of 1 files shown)


derrell (approved):

hkollmann (approved):

cboulanger:
The [travis job](https://github.com/qooxdoo/qooxdoo/blob/master/.travis/deploy_website) was [succesful](https://api.travis-ci.org/v3/job/563529038/log.txt), but there is no trace of the top-level `docs/` folder - see https://github.com/qooxdoo/qooxdoo.github.io . I wonder why. The travis is output is hard to read because it does not use @hkollmann 's trick to output the command itself, so it is difficult find out what happens where.

cboulanger:
Oh - I see - the symlink is wrong! https://github.com/qooxdoo/qooxdoo.github.io/blob/master/docs

PR #9774, opened 2019-08-14
allows the objectId mechanism to return objects which are not Qooxdoo objects

johnspackman:


cboulanger (approved):

hkollmann (commented):
if (!(obj instanceof qx.core.Object)) {

IMHO this excluded object wich are not qx.core.Object which is the base class of qooxdoo..

So my question: Is the title wrong or the implementation?

framework/source/class/qx/core/MObjectId.js
```diff
@@ -206,6 +209,18 @@ qx.Mixin.define("qx.core.MObjectId", {
       if (!this.__ownedQxObjects) {
         this.__ownedQxObjects = {};
       }
+      
+      if (!(obj instanceof qx.core.Object)) {
```
cboulanger: @hkollmann is right ! Good catch - this looks like the object is not saved IF it is a qx.core.Object. Or did I miss something?

johnspackman: no - that code is saying that if the object is not a `qx.core.Object` then to take a special case of recording it; if it *is* an instance of `qx.core.Object` then the code will continue from line 224 (which is unchanged from master).

The issue this PR is solving is that the code previously expected that you will *always* return an object which supports methods like `getQxObjectId` and `toHashCode` - while this is the case 90% of the time, it is also useful to be able to support non-Qooxdoo objects. And as there is already the option to specify the ID when adding, it's an easy fit to support it.

cboulanger: Ok, you're right - I missed line https://github.com/qooxdoo/qooxdoo/blob/0ecd322cc77e1e5812e8b94d39a977468ae73641/framework/source/class/qx/core/MObjectId.js#L251

hkollmann (approved):

derrell (approved):

oetiker (approved):

level420 (approved):

PR #9777, opened 2019-08-20
Added warning when an object in the property chain has a property but no corresponding event

milandamen:
When binding a source object to a target using a property chain, and a source in that chain has the correct property, but does not have an event defined for it, this PR will make sure a warning gets logged. This saves a lot of time debugging.

This problem surfaced because I was missing an event for a property that was used in a VirtualTree. The VirtualTree uses the labelPath that was set, to bind the label of the item to the VirtualTreeItem label. It uses a property chain like: `[0].labelPath` where `labelPath` would be set by your own code.

### DIFF
--- a/framework/source/class/qx/data/SingleValueBinding.js
+++ b/framework/source/class/qx/data/SingleValueBinding.js
@@ -150,19 +150,29 @@ qx.Class.define("qx.data.SingleValueBinding",
       try {
         // go through all property names
         for (var i = 0; i < propertyNames.length; i++) {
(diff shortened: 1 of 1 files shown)


level420 (approved):
Looks helpful

johnspackman (approved):

oetiker (approved):

derrell (approved):

cboulanger (approved):

hkollmann (approved):

Issue #9513, opened 2018-03-26
Add application lifecycle events


cboulanger:
For libraries and other type of code that does not have control over the main application, or which is designed to be only loosely coupled, it is useful to be able to observe the state of the application lifecycle.

The main application instance should therefore fire (at least) the following events:

- `beforeMain`
- `afterMain`
- `beforeFinalize`
- `afterFinalize`
- `beforeTerminate`

`afterMain` and `beforeFinalize` are probably identical.

level420:
The question is at which point the event handlers get called. Does the event queue run between `beforeMain` and `afterMain`? Or are they both handled within the same event queue run which would mean that we have for both events the application in the "after main" state when running the event handler? Or do we need some synchronous mechanism here (e.g. "callbacks")?

But in general I think this would be a good addition.

cboulanger:
If we want to be on the safe side, we fire a `beforeX` and `afterX` event for each of the interface methods of [`qx.application.IApplication`](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/application/IApplication.js).

The firing of event should probably go into [`qx.core.BaseInit`](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/core/BaseInit.js), but I'll have to do some research first.

cboulanger:
Sorry wrong button, reopened....

level420:
Still I think it is worth to think about the fact that all fired events will be handled in one queue run. So I think that the execution sequence will be:

- fire beforeMain event
- call main
- fire afterMain event
- fire beforeFinalize event
- call finalize --> calls qx.ui.core.queue.Manager.flush() leads to the event queue being processed
- call beforeMain event handler
- call afterMain event handler
- call beforeFinalizeEventHandler
- fire afterFinalize event handler
- ... some queue flush later ...
- call afterFinalizeEventHandler

cajus:
`qx.bom.Lifecycle` already does some of them.

cboulanger:
Important points, thanks! Will have a look at the queue manager and `qx.bom.Lifecycle`. Will take me until after Easter though...

cboulanger:
Revisiting this after a long time. We can already do
```javascript
qx.bom.Lifecycle.onReady(()=>{
  console.log("Application is ready");
});
```
[...]

PR #9778, opened 2019-08-22
Adds an event which is fired only after the application has fully initialized

johnspackman:
Previously, this was the "ready" event, but that event is what triggers the app to be created and means that there is not a reliable means to know when the app has been finalized.  This event does that.


cboulanger (approved):
See also #9513

derrell (approved):

hkollmann (approved):

level420 (approved):

oetiker (approved):

PR #9782, opened 2019-08-26
ECMAScript BigInt environment checks

schnepe2:
As I found that the ECMAScript `BigInt` object[1] is getting supported by the current major desktop browsers[2], I think environment checks for it could be nice.

While using `BigInt` in my code, I found that `toLocaleString` does not work on current FF[3], so I added an additional check for that, too.

------
[1] ECMAScript NEXT
[2] https://kangax.github.io/compat-table/esnext/#test-BigInt
[3] Firefox Quantum 68.0.2 (32-bit)

### DIFF
--- a/framework/source/class/qx/bom/client/EcmaScript.js
+++ b/framework/source/class/qx/bom/client/EcmaScript.js
@@ -303,8 +303,36 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
     getStringTrim : function() {
       return typeof String.prototype.trim === "function";
     },
-    
-    
+
+
+    /**
+     * Checks if 'BigInt' type is supported.
+     * @internal
+     * @ignore(BigInt)
+     * @return {Boolean} <code>true</code>, if BigInt is available.
+     */
+    getBigInt : function() {
+      return typeof BigInt != "undefined";
+    },
+
+
+    /**
+     * Checks if 'toLocaleString' is supported on the BigInt object and whether
+     * it actually works
+     * @internal
+     * @ignore(BigInt)
+     * @ignore(BigInt.prototype.toLocaleString)
+     * @return {Boolean} <code>true</code>, if the method is supported and
+     *   works at least rudimentary.
+     */
+    getBigIntToLocaleString : function()
+    {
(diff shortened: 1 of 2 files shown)


cboulanger (approved):

framework/source/class/qx/bom/client/EcmaScript.js
```diff
@@ -303,8 +303,36 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
...
+     * it actually works
+     * @internal
+     * @ignore(BigInt)
+     * @ignore(BigInt.prototype.toLocaleString)
+     * @return {Boolean} <code>true</code>, if the method is supported and
+     *   works at least rudimentary.
+     */
+    getBigIntToLocaleString : function()
+    {
+      return typeof BigInt != "undefined"                          // BigInt type supported...
```
level420: Please user the more exact ```!==``` comparison here. thank you.

level420 (changes requested):
see inline comments.

framework/source/class/qx/bom/client/EcmaScript.js
```diff
@@ -303,8 +303,36 @@ qx.Bootstrap.define("qx.bom.client.EcmaScript",
...
+
+
+    /**
+     * Checks if 'BigInt' type is supported.
+     * @internal
+     * @ignore(BigInt)
+     * @return {Boolean} <code>true</code>, if BigInt is available.
+     */
+    getBigInt : function() {
+      return typeof BigInt != "undefined";
```
level420: Please user the more exact ```!==``` comparison here. thank you.

framework/source/class/qx/bom/client/EcmaScript.js
schnepe2: ...which does not make a difference in this case, but O.K. I'll change that.

johnspackman (approved):

level420 (approved):
Thank you @schnepe2

derrell (approved):

oetiker (approved):

hkollmann (approved):

schnepe2:
Thank you

PR #9783, opened 2019-08-29
Fix bind length from data array (BZ#8169) #8197

RemiHeugue:
Fix issue #8197

When you bind on the length of an array and the array instance is replaced by another in the object holding the array the bind isn't correctly updated to the new length but bind on array[0] instead.

This is because __chainListener incorrectly assume that if qx.Class.implementsInterface(source, qx.data.IListData) is true, it is always binding on an item.

### DIFF
--- a/framework/source/class/qx/data/SingleValueBinding.js
+++ b/framework/source/class/qx/data/SingleValueBinding.js
@@ -373,7 +373,7 @@ qx.Class.define("qx.data.SingleValueBinding",
         // if its the last property
         if (j == context.propertyNames.length - 1) {
           // if its an array
-          if (qx.Class.implementsInterface(source, qx.data.IListData) && context.arrayIndexValues[i] !== "") {
+          if (qx.Class.implementsInterface(source, qx.data.IListData) && context.arrayIndexValues[j] !== "") {
             // set the initial value
             var itemIndex = context.arrayIndexValues[j] === "last" ?
               source.length - 1 : context.arrayIndexValues[j];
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/data/SingleValueBinding.js
```diff
@@ -373,7 +373,7 @@ qx.Class.define("qx.data.SingleValueBinding",
         // if its the last property
         if (j == context.propertyNames.length - 1) {
           // if its an array
-          if (qx.Class.implementsInterface(source, qx.data.IListData)) {
+          if (qx.Class.implementsInterface(source, qx.data.IListData) && context.arrayIndexValues[i] !== "") {
```
cboulanger: It looks like the variable should be `j`, not `i` - the linter says `i` is an unknown global symbol. I wonder why it works at all for you...

cboulanger (changes requested):

framework/source/class/qx/data/SingleValueBinding.js
RemiHeugue: Yes, of course it's j, I don't have the repo on the machine I'm on right now so edited the file directly from github and missed that, I'm updating the pr.

cboulanger (approved):

derrell (approved):

oetiker (approved):

johnspackman (approved):

level420 (approved):

hkollmann (approved):

PR #9788, opened 2019-09-10
Minor grammatical fixes

derrell:


cboulanger (approved):
Great, thanks - I think the whole documentation probably could profit from a round of copyediting/proofreading....

hkollmann (approved):

PR #9792, opened 2019-09-16
Add `toUuid` to qx.core.Object to provide a UUIDv4 on demand

johnspackman:
This PR makes a change to how the hash returned from `qx.core.Object.toHashCode()` is generated so that it is a true UUID rather than just a sequential number based string.  This is useful for situations where the ID of an object needs to be shared between instances of Qooxdoo applications, for example client/server or postMessage between iframe windows.  Also, to support synchronization between Qooxdoo applications, this PR allows code to set the initial value for the hash code with `.setHashCode`.

There should be no backwards compatibility issues because the hash code is supposed to be an opaque string and has little or no relevance to user code.

This change can be reverted by setting the environment variable `qx.hashCode.classic` to `true`.

Most of the changes are in `qx.core.Object` and `qx.core.ObjectRegistry` - the other files are changed because there are a lot of references to `$$hash` throughout the code, and this has been replaced with `.toHashCode()`


johnspackman:
BTW I did a performance test on my MacBook (2.4Ghz i9) in Chrome - it took 2.2 seconds to create 100,000 UUIDs.  This is obviously slower than the old code which was incrementing a number and appending it to a string, but it can be turned off.

The default in this PR is that UUIDs are on

framework/source/class/qx/core/ObjectRegistry.js
```diff
@@ -162,15 +125,57 @@ qx.Bootstrap.define("qx.core.ObjectRegistry",
...
+              [1e7] +        // 10000000 +
+              -1e3 +         // -1000 +
+              -4e3 +         // -4000 +
+              -8e3 +         // -80000000 +
+              -1e11          // -100000000000,
+              ).replace(     // replacing
+                /[018]/g,    // zeroes, ones, and eights with
+                createUuidV4 // random hex digits
+              )
+        }
```
derrell: Please use the readable version of this. We don't care about eliminating a few bytes as that gist is attempting; we do very much care about readability. I had to read half way through that long reference gist discussion to figure out what the heck they were optimizing.

```
function()
{
  return (
    "00000000-0000-4000-8000-000000000000"
      .replace(
        /0/g,
        function()
          {
            return(0 | Math.random() * 16).toString(16)
          }))
}
```

derrell (changes requested):

oetiker (approved):

cboulanger:
How readable are log messages with this? Would it make sense to use an abbreviated hash code (like a shortened commit hash in git) [here](https://github.com/johnspackman/qooxdoo/blob/toHashCode-as-uuid/framework/source/class/qx/log/appender/Util.js#L46)?

level420 (approved):

framework/source/class/qx/util/Uuid.js
```diff
@@ -0,0 +1,151 @@
...
+
+   Copyright:
+     2004-2011 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Tristan Koch (tristankoch)
```
cboulanger: Update author here? 😃

cboulanger (approved):

cboulanger:
Why did you revert the short hash code?

johnspackman:
I've made changes to address both @derrell's question - the UUID generation is now a separate utility class, with a different implementation which is also 5 or 6 times faster.

But @cboulanger 's point was harder to address - at least, I can't find a solution that feels really right.  I don't think that there is an easy way to shorten a UUID into a psuedo-unique identifier for logging output, so what I ended up doing was appending the old-style hash code so you end up with something like `5db4a740-eb6d-4de0-b1ce-6a521a6215b9+15-0`.  I also added a `toShortHashCode` method to `qx.core.Object` to make it easy to get this `15-0` short code.

However that just didnt seem like the best way to go - I've reverted the commit I made with that solution.

My thoughts are that if we have a short and long version, then adding an API method (such as `toShortHashCode()`) is important because the identifier is opaque - we don't want code relying on the internal format of the string.  But if we're adding to the API, would it be better to just add `getQxUuid()` method (and `setQxUuid`) that exists in parallel to `toHashCode`?  But then we have the weird situation where we have *two* unique identifiers...!! 

On the other hand, is there any point in outputting the hash code into the logging output in the first place?  It is supposed to be an opaque identifier after all, and as there is no longer an object registry it is of limited use in the logging output.

My final thought is that it would be useful to be able to customise the formatting used by `qx.log.appender.*` (for example in multiple-window applications) so if someone really did want to log the hashCode/UUID they could add it back in at runtime.

So, what I'm proposing is that `toHashCode` returns a UUIDv4 like `5db4a740-eb6d-4de0-b1ce-6a521a6215b9` and that we remove hash codes from logging output.

johnspackman:
@cboulanger sorry, my post crossed!  I can't make my mind up which is the better way, so asking for opinions :)

cboulanger:
@johnspackman I agree. You couldn't do anything with the hash code to begin with, since there was no way to look it up somewhere or use the debugging console to identify it. For logging, I would instead use the `qxObjectId` if one has been set, and otherwise just output the class name.

cboulanger:
Maybe it would make sense to integrate a utility class that does automatic `qxObjectId` assignment on request. This would be useful in debugging situations to identify which object is concerned. See https://github.com/cboulanger/eventrecorder/blob/master/source/class/cboulanger/eventrecorder/ObjectIdGenerator.js

derrell (approved):

derrell:
> On the other hand, is there any point in outputting the hash code into the logging output in the first place? It is supposed to be an opaque identifier after all, and as there is no longer an object registry it is of limited use in the logging output.

Yes, it's very important to uniquely identify which object is generating the logging output. I use that to identify a specific instance of a class where there may be many generating similar output.

cboulanger:
@derrell How would you identify the instance? Is the hash in the DOM?

derrell:
And I would prefer just the very short ID in that output, like the old hash code. 

Maybe we need the uuid only sometimes, and it could be a second ID which is enabled when needed? I don't see a problem with a second ID which is universally unique, for this very different other purpose.

derrell:
I think sometimes the hash is in the dom, but that's not what I care about. I just need uniquely identified, which object is generating the output. (Early output might identify the arguments to the constructor, or something, so that I know which object is which should I need to know those details.)

cboulanger:
Would the uniqueness of the UUID suffer alot if, say, the last 5 digits were an hexadecimal value that would start with 00000 and would be incremented on each object creation? This way we would have a short unique local ID (last 5 digits of the UUID) and a globally unique one?

cboulanger:
Or. to make it more random, the last five digits could start with a random number. this would give us 1048575 objects, if that's not enough we can use the last 6 digits?

cboulanger:
On the other hand, a more pragmatic approach would be to just have two IDs - a local, incrementing one, and a global UUID.

derrell:
I don't see a big problem with two IDs. They are for very different purposes, and in one case they are intentionally and for good reason displayed; in the other, they'll be mostly used internally and across process boundaries.

cboulanger:
@johnspackman What if you left the `toHashCode()` as it was and add a new `toUuid()` method? This would seem like the most unobtrusive solution. Your object synchronization code would then would use the latter method instead of `toHashCode`? You could leave most of you changes intact since using a method is better than direct property access to `$$hash` (although I don't know if there is a performance cost).

johnspackman:
@cboulanger yes, having read the comments above I agree - I'll restore the original style of IDs in `toHashCode()` add a `toUuid()`;

cboulanger:
Do we really need the performance test bundle in the framework?

cboulanger (approved):

cboulanger:
Maybe you need to change the PR title to reflect the changed goal of the PR?

level420:
@johnspackman please have a look at the travis coverage run: it seems that a lot of tests are failing. Thank you!

framework/source/class/qx/log/Logger.js
```diff
@@ -578,9 +578,17 @@ qx.Bootstrap.define("qx.log.Logger",
       // Add relation fields
       if (object)
       {
+        function isQxCoreObject(clz) {
+          while (clz) {
+            if (clz.classname === "qx.core.Object")
+              return true;
```
cboulanger: You'll need to use braces to avoid linting errors: https://travis-ci.org/qooxdoo/qooxdoo/jobs/591902970#L542

framework/source/class/qx/log/Logger.js
```diff
@@ -578,7 +578,12 @@ qx.Bootstrap.define("qx.log.Logger",
       // Add relation fields
       if (object)
       {
-        function isQxCoreObject(clz) {
+        // Do not explicitly check for instanceof qx.core.Object, in order not
+        // to introduce an unwanted load-time dependency
+        function isQxCoreObject(object) {
+          if (object === object.constructor)
+            return false;
```
cboulanger: I am afraid this, again, won't be accepted by linting without braces.... :-(

johnspackman: :sigh: IMHO lint is 99% vanity and 1% sanity, this rule is part of the 99%...

cboulanger: We can change the rules - You could create an issue here (I think): https://github.com/qooxdoo/eslint-config-qx

derrell: I'm one of the ones who feels strongly that this lint rule is part of the 1% (and I' think I give it more of a 50/50 spread rather than 99/1 :-) ), because it prevents accidentally adding a new statement, e.g. a console.log inserted at line 584.5, and causing a complete change of operation (the return statement is then not part of the `if`). There are occasional times when such things are useful, though. I would be fine with enabling the lint option that allows a single line without braces, so this would end up looking like:

```
    if (object == object.constructor) return false;
```

johnspackman: I think that it depends a lot on the context, and for me it's all about readability and the flow of the code.  For example, enforcing indentation is a good solution for the problem of accidental new statements and preserves readability, so having a single line like that increases the chances of errors, eg

```
  if (object === object.constructor) return false;
    console.log("it's a constructor");
```

I find that I read the first token on the line plus the indentation to see when to skip ahead, and so one line statements are dangerous and time consuming because you have to read to the end.

Consistent indentation is really key, which to me means that this is troublesome:

```
    if (abc.getSomeReallyLong().methodName() == my.package.name.here.SomeClass.VALUE &&
        yetMoreCode.getBlahDeBlah() == my.package.name.here.SomeClass.OTHER_VALUE)
      return true;
```

That's not ideal IMHO, and should have brackets around `return true` - possibly on separate lines, a practice I usually detest because it introduces yet another extra bit of vertical cruft.

OTOH I find that the brackets rule is ugly for many simple statements, elongates the code and just gets in the way of reading the meaning of the code.

```
  if (a){
    doSomething(1);
  }
  if (b) {
    doSomething(2);
  }
```
as opposed to:
```
if (a)
  doSomething(1);
if (b)
  doSomething(2);
```

Ultimately though, these debates just go to show that that lint is highly subjective.  For me, I find lint occasionally useful, just not very often - it is particularly useless when it is only visible when I commit and travis kicks in, and pretty much the most it can possibly do at that stage is show up a code path which breaks and for which there is no unit test.

I'm not proposing changing the lint rules BTW, just having a moan 😆  It would be much better if (when) we could one day integrate lint on the fly with the continuous compilation of `qx compile` *and* present that somewhere front and centre (i.e somewhere other than in a terminal window that you never look at until it's too late...)

hkollmann (approved):

PR #9795, opened 2019-09-24
Reset user data of objects returned to the ObjectPool.

lucapivato:
If an app uses the ObjectPool to pool objects when getting an instance out of the pool the consumer should expect a clean object. A case in point would to use setUserData() on an event to handle custom information in a propagation chain. When the object is pooled and reused for the next event the user data should be empty.


lucapivato:
Please ignore.

PR #9796, opened 2019-09-25
add support for breaking within words in labels

johnspackman:
If you have a label with very little horizontal space (eg a table of some kind) and a user keys in a long bit of text, label will word wrap and calculate the height including the word wrap; however, words are space delimited, and users can enter values like "Netherlands/Belgium/Germany" and the  browser will not wrap because there are no spaces - instead the label is truncated at maximum width.

This PR add a property to label `breakWithinWords` which allows words to be broken; it defaults to `false` to avoid BC issues.

### DIFF
--- a/framework/source/class/qx/ui/basic/Label.js
+++ b/framework/source/class/qx/ui/basic/Label.js
@@ -123,6 +123,21 @@ qx.Class.define("qx.ui.basic.Label",
       init : true,
       apply : "_applyWrap"
     },
+    
+    
+    /**
+     * Controls whether line wrapping can occur in the middle of a word; this is
+     * typically only useful when there is a restricted amount of horizontal space
+     * and words would otherwise overflow beyond the width of the element.  Words
+     * are typically considered as separated by spaces, so "abc/def/ghi" is a 11 
+     * character word that would not be split without `breakWithWords` set to true. 
+     */
+    breakWithinWords :
+    {
+      check : "Boolean",
+      init : false,
+      apply : "_applyBreakWithinWords"
+    },
 
 
     /**
@@ -391,6 +406,9 @@ qx.Class.define("qx.ui.basic.Label",
       if (this.__webfontListenerId) {
         this.__fixEllipsis();
       }
+      if (rich && this.getBreakWithinWords()) {
+        styles.wordBreak = "break-all";
+      }
 
       return rich ?
         Label.getHtmlSize(content, styles, width) :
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/basic/Label.js
```diff
@@ -391,6 +406,9 @@ qx.Class.define("qx.ui.basic.Label",
       if (this.__webfontListenerId) {
         this.__fixEllipsis();
       }
+      if (rich && this.getBreakWithinWords()) {
+        styles.wordBreak = "break-all";
```
oetiker: is this supported on all browsers ?

johnspackman: yes, IE6 onwards https://caniuse.com/#search=word-break

hkollmann (approved):

cboulanger (approved):

derrell (changes requested):
Although it may be (far?) more difficult to implement, what would be far more useful than breaking at an arbitrary location, is breaking (when necessary to avoid wrapping) on matches of a regular expression. In your example case, that regular expression might contain a slash.

Before approving this, it's probably worthwhile to discuss this alternative.

johnspackman:
@derrell I agree, it is far more complicated!

To take that approach (unless you implement a rendering engine which understands the dimensions of each letter, kerning, etc) you'd have to split the text up and test the sizes of each combination, working until the boundaries of the container are reached.

The current version uses the browser to determine splits and to determine the best allowable dimensions; all this PR does is allow the calling code to fine tune how the browser does that.  The `break-all` functionality is a standard feature after all, even IE6 supports it so it seems reasonable to include it as-is.

Also, that does not solve the original problem - although the example is that users are filling in words that could be separated by spaces instead of slashes or commas, the real issue is that when you have a limited horizontal space, occasionally an end-user will key in something really long, truncating it is not desirable.

For example, there is a famous Welsh village called [Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch](https://en.wikivoyage.org/wiki/Llanfairpwllgwyngyll) (that genuinely is it's name), and no regex is going to be able to break that up :)

derrell (approved):

level420 (approved):

level420:
@oetiker do you still have objections?

oetiker:
> @oetiker do you still have objections?

oh sorry ... no all is well

oetiker (approved):

PR #9799, opened 2019-10-10
makes formatting of log output to be configurable and plugable

johnspackman:
The initial motivation of this PR was to allow the datetime formatting to be changed in log output - by default, the logging outputs the time as the number of milliseconds since the app started, which is concise and great for browser apps, but for long running node apps it's very inconvenient because what you really need to know is an actual date/time so that log entries can be related to real human events (like an end-user's bug report).  Changing the logging output (without this PR) is not possible because all appenders are baked in to use the static method `qx.log.appender.Util.toText()`

This PR changes the logging appenders so that they use an instance of the new `qx.log.appender.Formatter`; there is a default instance of Formatter (which can be replaced) and it has helper methods that can be used to put together the full logging output.

Also in this PR is a change to appenders - previously, every appender is uniquely identified by it's class name, but this prevents using the same class more than once; EG for node apps, it's quite reasonable to want to have a file-based appender, and to have multiple instances of a FileAppender, each one with a different set of filters and output filename.

This PR allows the appender to define a static `appenderName` property that will be used in preference to the `classname` property as a unique identifier.  This is obviously pretty ugly, but appenders are static classes at the moment and all appenders work by "convention" of having well known exposed static methods; this PR is deliberately not attempting to rewrite the entire Appender class implementation.


cboulanger (approved):

oetiker (approved):
Sorry, still early morning, but how does it help to set an appenderName if the appender is static? There will always only be one ...

johnspackman:
Because it's only the current appenders which are static - `qx.log.Logger` relies on being given a "thing" which supports a method called `process()` and a property called `classname` (and now also `appenderName`).  That "thing" could be a static class or an instance of a class derived from `qx.core.Object`, or probably just `{ process() {}, classname: "abc" }`.

This appenderName is useful if you have a proper object instance with a class derived from `qx.core.Object` - that instance will already have a `classname` via the prototype plus a non-static, overridable method called `process`.  You can pass this [non-static implementation of an] object as an appender, but the appenders need a unique ID.

IMHO the appenders need to be rewritten as a proper class, rather than this non-obvious kludge, but I want to keep that out of scope for now because there's quite a lot of code in the current appenders - plus it looks like there were some dependency issues that the original author was worried about.

oetiker:
ah ... it does not HAVE to be static ... in that case :) perfect

hkollmann (approved):

framework/source/class/qx/log/appender/Console.js
```diff
@@ -172,7 +172,8 @@ qx.Class.define("qx.log.appender.Console",
     process : function(entry)
     {
       // Append new content
-      this.__log.appendChild(qx.log.appender.Util.toHtml(entry));
+      var formatter = qx.log.appender.Formatter.getDefaultFormatter();
```
derrell: If the goal here is to be able to use different formatters, then why would we always retrieve the _default_ formatter? Should this instead be
```
var formatter = qx.log.appender.Formatter.getFormatter();
```
and the `formatter` property just has a default value, but it can be changed?

framework/source/class/qx/log/appender/Formatter.js
```diff
@@ -0,0 +1,308 @@
...
+      }
+      return qx.log.appender.Formatter.__defaultFormatter;
+    },
+    
+    /**
+     * Sets the default formatter
+     * 
+     * @param instance {qx.log.appender.Formatter}
+     */
+    setDefaultFormatter: function(instance) {
```
derrell: Yeah, this seems hokey to me. The default shouldn't be changeable; rather, there should be a default, I think, that `getFormatter` returns until `setFormatter` changes what is to be used.

derrell (changes requested):

level420 (approved):

framework/source/class/qx/log/appender/Formatter.js
johnspackman: done

derrell (approved):

framework/source/class/qx/log/appender/Formatter.js
hkollmann: No the test must be changed too.,.,.

PR #9800, opened 2019-10-12
fixes a bug with mixins when added via `qx.Class.patch`

johnspackman:
This PR fixes a tricky problem with `qx.Class.patch` where mixin constructors are not called; worse, it did not occur in `qx.debug` builds because qx.debug would enable a shortcut that accidentally avoided the bug (this shortcut has now been removed).  Another confusion was that it depends on dependency orders whether the problem occurs at all, and this could also lead to different behaviour between generator and compiler.

The test for this bug would be that after patching, this assert would fail (it should never fail):

```
qx.Class.patch(abc.MyClass, xyz.SomeMixin);

// *** This assert should never fail***
qx.core.Assert.assertTrue(abc.MyClass.constructor === abc.MyClass);
```

The problem comes down to the need for `qx.Class.patch` to sometimes have to wrap the class’ constructor if the mixin has a constructor and the class’ constructor has not previously been wrapped (this is another reason why the non-debug build can exhibit different behaviour).  The wrapping is necessary to automatically call constructors.  However, because the class instance is the `constructor` property (ie `qx.core.Object === qx.core.Object.constructor`) this means that wrapping constructor means changing the class instance as well, and this causes issues where the class instance has been held onto.

EG:
```
var tmp = abc.MyClass;
qx.Class.patch(tmp, xyz.SomeMixin);

// *** Do not use this assert, it will (legitimately) not always be true ***
qx.core.Assert.assertTrue(tmp === abc.MyClass);
```

The framework was holding onto class definitions, ie it was effectively falling foul of this `var tmp = …` example above, and this would cause `defer` to be given the old (wrong) class.

This is the correct way:
```
var tmp = abc.MyClass;
tmp = qx.Class.patch(tmp, xyz.SomeMixin);
// This assert is safe and will always be true
qx.core.Assert.assertTrue(tmp === abc.MyClass);
```

### DIFF
--- a/framework/source/class/qx/Class.js
+++ b/framework/source/class/qx/Class.js
@@ -372,27 +372,6 @@ qx.Bootstrap.define("qx.Class",
      * use, but as a formalized way (and a last resort) in order to patch
      * existing classes.
      * 
-     * <b>NOTE</b>: This changes the definition of the class, and can mean that the
-     * global clazz is not the same instance.  EG:
-     * 
-     * <pre>
-     *   var tmp = abc.MyClass;
-     *   qx.Class.patch(tmp, xyz.SomeMixin);
-     *   
-     *   // *** Do not use this assert, it will (legitimately) not always be true ***
-     *   qx.core.Assert.assertTrue(tmp === abc.MyClass);
-     * </pre>
-     *
-     * To be safe, if you have a temporary variable you must always reconnect it, and 
-     * the easiest way to achieve this is to get the return value from `patch`, eg 
-     * <pre>
-     *   var tmp = abc.MyClass;
-     *   tmp = qx.Class.patch(tmp, xyz.SomeMixin);
-     *   
-     *   // This assert is safe and will always be true
-     *   qx.core.Assert.assertTrue(tmp === abc.MyClass);
-     * </pre>
-     * 
      * <b>WARNING</b>: You may break working classes and features.
      *
      * @param clazz {Class} An existing class which should be modified by including a mixin.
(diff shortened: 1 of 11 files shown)


cboulanger (approved):

johnspackman:
I found that the fixes I had added improved things but did not entirely work, because they would not allow `obj = new abc.MyClass(); obj.constructor === abc.MyClass` - but only under some circumstances.

This problem is all caused by retrospectively adding a wrapper to a constructor, after the class and constructor have been defined and have been in use.

I don’t know whether this is something where I simply don’t understand exactly what’s going on under the hood and maybe, with a little more knowledge and work, I might be able to make retrospectively wrapping constructors work - but on second thoughts, I think that it is acceptable to always wrap a constructor and then completely avoid the problem in the first place.  

This was always happening when `qx.debug==true` anyway (which was a requirement for debug tests for instantiating abstract and singleton to work), and it is a very very tiny optimisation to avoid wrapping.

framework/source/class/qx/Class.js
```diff
@@ -1712,14 +1637,17 @@ qx.Bootstrap.define("qx.Class",
      *
      * @param base {Class} The base class.
      * @param mixins {Mixin[]} All mixins which should be included.
+     * @param type {String} the type of the class
      * @return {Boolean} true, if the constructor needs to be wrapped.
      */
-    __needsConstructorWrapper : function(base, mixins)
+    __needsConstructorWrapper : function(base, mixins, type)
```
derrell: Since you removed the call to `__needsConstructorWrapper` above, is it necessary to even still have this method here?

derrell (changes requested):

framework/source/class/qx/Class.js
johnspackman: ah, no...

oetiker (approved):

derrell (approved):

hkollmann (approved):

level420 (approved):

level420:
merging. thank you @johnspackman

PR #9803, opened 2019-10-28
build schema doc from schema file during building of site.

### LINKED ISSUE #583: "Unresolved use of symbol" warning, externalResources/script function
Like discussed on gitter, it could be an interesting feature to avoid  "Unresolved use of symbol" warning, for function declared in a plain javascript file defined in externalResource (or maybe elsewhere don't mind to give your opinion).

Just to expose the problem it solves, currently i have a plain javascript file with plenty "utility" function, i declared it in Manifest.json externalResource/script, i used these functions in many qooxdoo classes, it works without problem but the compiler throw a  "Unresolved use of symbol" warning for each use of function. The problem can be solved by adding on each class a @ignore for each function but it's quite time-consuming when you have like me many functions and many classes. The goal is so to automaticaly avoid these warning.

hkollmann:
to resolve https://github.com/qooxdoo/qooxdoo-compiler/issues/583 i added an auto generated markdown description of the compile-1-0-0.json Schema to the docs and the toolchain

### DIFF
--- a/.travis/build-site-v6
+++ b/.travis/build-site-v6
@@ -10,7 +10,7 @@ exe() { echo "\$ ${@/eval/}" ; "$@" ; }
 TARGET=deploy/$(cat ./version.txt)
 mkdir -p $TARGET
 
-npm i --no-save @qooxdoo/compiler
+exe npm i --no-save @adobe/jsonschema2md @qooxdoo/compiler
 
 # build apps  
 exe npx qx compile compile.json --target build  --feedback=0 --verbose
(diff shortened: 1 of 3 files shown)


level420:
@hkollmann travis fails in the markdown lint job complaining about a wrong self reference (I think).

oetiker (approved):

cboulanger (approved):
Thank you, that is very useful.!

level420 (approved):

derrell (approved):

johnspackman (approved):

PR #9805, opened 2019-10-29
check qx.promise setting

### LINKED ISSUE #9802: Setting environment variable "qx.promise" to false has no effect
**Describe the bug**
The environment setting "qx.promise" was introduced to be able to compile qooxdoo without promises (because that add about 100kb to the compiled code). Currently this behaviour is broken, qx.Promise is always added to the build, even with "qx.promise = false".

I guess this line causes the problem as it uses qx.Promise without checking the "qx.promise" environment setting: https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/event/handler/DragDrop.js#L321

**To Reproduce**
1. Add "qx.promise = false" to your compiler setting
2. Generate a source or build version
3. See that qx.Promise is part of the generated code

**Expected behavior**
qx.Promise should not be added to the generated code when "qx.promise" is set to false.

peuter:
to remove the hard-coded dependency to `qx.Promise`

closes #9802

### DIFF
--- a/framework/source/class/qx/event/handler/DragDrop.js
+++ b/framework/source/class/qx/event/handler/DragDrop.js
@@ -316,13 +316,19 @@ qx.Class.define("qx.event.handler.DragDrop",
      * @return {qx.Promise|String} One of <code>move</code>, <code>copy</code> or
      *    <code>alias</code>
      */
-    getCurrentActionAsync : function() {
+    getCurrentActionAsync : qx.core.Environment.select("qx.promise", {
+      "true": function() {
       var self = this;
+
       return qx.Promise.resolve(self.__detectAction())
         .then(function() {
           return self.__currentAction;
         });
-    },
+      },
+      "false": function () {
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

framework/source/class/qx/event/handler/DragDrop.js
```diff
@@ -316,13 +316,19 @@ qx.Class.define("qx.event.handler.DragDrop",
      * @return {qx.Promise|String} One of <code>move</code>, <code>copy</code> or
      *    <code>alias</code>
      */
-    getCurrentActionAsync : function() {
+    getCurrentActionAsync : qx.core.Environment.select("qx.promise", {
+      "true": function() {
```
derrell: Are these really supposed to be strings "true" and "false", not booleans?

derrell (approved):
Approved if those are really supposed to be strings. See inline comment.

PR #9807, opened 2019-11-05
Add a test on Number related locale

rommni:
Like discussed on qooxdoo/qooxdoo-compiler#587 add test to number related locale value.

Testing fr locale allow to test that the locale linked to defaultNumberingSystem are loaded and not just the first one declared in cldr files

### DIFF
--- a/framework/source/class/qx/test/locale/Number.js
+++ b/framework/source/class/qx/test/locale/Number.js
@@ -26,22 +26,22 @@ qx.Class.define("qx.test.locale.Number", {
         return;
       }
 
-      var Number = qx.locale.Number;
+      var number = qx.locale.Number;
       var useLocale = "C";
 
       var commonGroupSeparator = ",";
       var commonDecimalSeparator = ".";
 
-      this.assertEquals(commonGroupSeparator, Number.getGroupSeparator(useLocale));
-      this.assertEquals(commonDecimalSeparator, Number.getDecimalSeparator(useLocale));
+      this.assertEquals(commonGroupSeparator, number.getGroupSeparator(useLocale));
+      this.assertEquals(commonDecimalSeparator, number.getDecimalSeparator(useLocale));
 
       useLocale = "fr";
 
       var frenchLatinGroupSeparator = "\u202F"; // narrow no-break space
       var frenchLatinDecimalSeparator = ",";
 
-      this.assertEquals(frenchLatinGroupSeparator, Number.getGroupSeparator(useLocale));
-      this.assertEquals(frenchLatinDecimalSeparator, Number.getDecimalSeparator(useLocale));
+      this.assertEquals(frenchLatinGroupSeparator, number.getGroupSeparator(useLocale));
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/test/locale/Number.js
```diff
@@ -0,0 +1,67 @@
...
+  extend : qx.dev.unit.TestCase,
+
+  members: {
+    testNumberSeparators : function() {
+      // test not designed for generator build
+      if (!qx.core.Environment.get("qx.compilerVersion")) {
+        return;
+      }
+
+      var Number = qx.locale.Number;
```
cboulanger: Overwriting the native Number object is certainly legal, but is it a good idea? Maybe use a different symbol?

hkollmann:
I am missing the change of compile.json for the french locale?

framework/source/class/qx/test/locale/Number.js
rommni: Hum effectively i took the idea from the qx.test.locale.Date, i will change that.

rommni:
@hkollmann after more investigating the change wasn't needed, the fact is that for testing i was running the test with testtapper and this one was compiled using compile.json settings and so was lacking the fr locale, however the framework config settings contains the french locale.

cboulanger (approved):

hkollmann:
Can you add your testtapper compile.json as compileTest.json?

johnspackman (approved):

rommni:
@hkollmann of course i can, do you want me to push to my Manifest.json and qx-lock.json with testtapper add?  Maybe only the qx-lock? Currently i include in compile.json testtapper app only my qx.test.locale.Number class, should i let this or change it?

level420 (approved):

oetiker (approved):

hkollmann:
I suggest a project in testgui from where the testrunner can run. Your test is not working in the python toolchain. With this test project we are able to run the tests with the compiler tool chain.

cboulanger:
@hkollmann Do you mean the test *breaks* the python toolchain (because the tests seem to pass, or it simply doesn't run in the generator-based toolchain. Because if the latter is the case, it doesn't really matter, does it?

hkollmann:
It's not running. But i would like to archive the test project as a base for changing the test procedure

hkollmann (changes requested):
Please add the test project

rommni:
sorry a little busy currently, i'll looking that as soon as i have the time ;)

cboulanger:
Hi @rommni , any chance you can revisit and finalize this PR? Thank you.

rommni:
Hi @cboulanger sorry i totally forgotten this one, of course i will look at this (trying to do that in the week)

rommni:
Just a message to tell i haven't forget this but just not enough time these last days doing that as soon i can

hkollmann (approved):

hkollmann:
Testing is completely reworked

hkollmann:
repos is gone,

PR #9816, opened 2019-12-10
Blocker.forceUnblock() doesn't unblock in some cases

lucapivato:
If forceUnblock() is called before the blocker is shown it doesn't detach the "appear" handler and the blocker is shown anyway, unlike unblock().

If block() is called multiple times before the element is shown, the "appear" handler is re-registered losing the previous id, so if the app calls unblock() it cannot unregister the "appear" handler and the blocker is shown anyway.


derrell (approved):

hkollmann (approved):

johnspackman (approved):

level420 (approved):

cboulanger (approved):

cboulanger:
Full quorum, merging. Thank you, @lucapivato !

PR #9844, opened 2020-01-05
enable use native promise in event listeners

hkollmann:
With this change is it possible to just return an async function in an event listener. it's not longer necesarry to towrap into an qx.Promise


derrell (approved):
@johnspackman knows all of the details of this, so I defer to his comments

zaucker (approved):

cboulanger (approved):

oetiker (approved):

framework/source/class/qx/event/Utils.js
```diff
@@ -67,7 +68,7 @@ qx.Class.define("qx.event.Utils", {
      */
     track: qx.core.Environment.select("qx.promise", {
       "true": function(tracker, fn) {
-        if (typeof fn !== "function" && !(fn instanceof qx.Promise)) {
+        if (typeof fn !== "function" && !(fn instanceof qx.Promise) && !(fn instanceof Promise)) {
```
johnspackman: IMHO that kind of low level negotiation/detection of type is best suited to a common utility method, please can we move the implementation of this test into (for example) `qx.lang.Type.isPromise()`?  

Technically the Promises/A+ spec says that _“promise” is an object or function with a then method whose behavior conforms to this specification_, so in theory all an event has to do is `return { then() { console.log("hello"); } }`; the spec is at https://promisesaplus.com/

johnspackman (changes requested):

framework/source/class/qx/event/Utils.js
hkollmann: Good idea. 👍

cboulanger:
I tend to agree with @johnspackman. Instead of checking for prototypes ("is a"), why not checking for the least-common-denominator interface method ("has a") `then`? This would allow to use all kind of Promise implementations. Of course, this allows for false positives (objects that have a `then` method which are not Promises), but chances are very small that this would be a real issue.

cboulanger:
Or, to be on the safe side, we could explicitly check for the minimal API methods which the framework requires to function and therefore we require an implementation to provide. This would probably be `then`, `catch` and `all`, wouldn't it?

hkollmann:
Could be false positiv either

framework/source/class/qx/lang/Type.js
```diff
@@ -182,6 +183,24 @@ qx.Bootstrap.define("qx.lang.Type",
...
+     * Whether the value is a Promise.
+     *
+     * @param value {var} Value to check.
+     * @return {Boolean} Whether the value is a Promise.
+     *
+     * @ignore(Promise)
+     */
+isPromise : function(value)
+    {
+      var nat = (value instanceof (Promise || window.Promise));
```
derrell: I think this will crash if Promise is undefined. Try this in the console to see why I think so:
```
var nat = (value instanceof (xxx || window.Promise))
```

framework/source/class/qx/event/Utils.js
```diff
@@ -136,11 +137,11 @@ qx.Class.define("qx.event.Utils", {
           this.__addCatcher(tracker);
           return tracker.promise;
         }
-        if (fn instanceof qx.Promise) {
+        if ((fn instanceof qx.Promise) || (fn instanceof Promise)) {
```
cboulanger: `if (qx.lang.Type.isPromise(fn)) {`

framework/source/class/qx/event/Utils.js
```diff
@@ -136,11 +137,11 @@ qx.Class.define("qx.event.Utils", {
...
           return this.__thenPromise(tracker, fn);
         }
         var result = fn(tracker.result);
-        if (result instanceof qx.Promise) {
+        if ((result instanceof qx.Promise) || (result instanceof Promise)) {
```
cboulanger: `if (qx.lang.Type.isPromise(fn)) {`

framework/source/class/qx/lang/Type.js
```diff
@@ -182,6 +183,18 @@ qx.Bootstrap.define("qx.lang.Type",
...
       );
-    }
+    },
+
+
+    /**
+     * Whether the value is a Promise.
+     *
+     * @param value {var} Value to check.
+     * @return {Boolean} Whether the value is a Promise.
```
cboulanger: Maybe explain the type of checking that is done? "This will only check whether the argument is an object that has a `then` method, as proposed by https://promisesaplus.com/"

cboulanger (changes requested):

cboulanger (approved):

framework/source/class/qx/core/MEvent.js
```diff
@@ -230,7 +230,7 @@ qx.Mixin.define("qx.core.MEvent",
         if (oldData === undefined) {
           oldData = null;
         }
-        return this.__Registration.fireNonBubblingEvent(
+        return this.__Registration.fireEvent(
```
derrell: This looks like a major change that could affect lots of existing apps... Why did it change from non-bubbling to bubbling? (or am I misreading it?)

derrell (changes requested):

framework/source/class/qx/core/MEvent.js
hkollmann: fireEvent is using Registration.fireEvent.
fireDataEvent is using Registration.fireNonBubblingEvent.

2 similar functions with different return values is IMHO no good idea.

fireNonBubblingEvent could'nt return a promise. With reading the qooxdoo spec for bubbling to non-bubbling this should'nt break anything.

derrell (approved):

level420 (approved):

framework/source/class/qx/core/MEvent.js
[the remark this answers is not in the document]
hkollmann: fireEvent is using Registration.fireEvent.
fireDataEvent is using Registration.fireNonBubblingEvent.
2 similar functions with different return values is IMHO no good idea.
fireNonBubblingEvent could'nt return a promise. With reading the qooxdoo spec for bubbling to non-bubbling this should'nt break anything.

cboulanger:
Unfortunately, the tests fail: https://travis-ci.org/qooxdoo/qooxdoo/jobs/633809534#L794

@hkollmann, can you check?

hkollmann:
I know and i am working on this....

johnspackman (approved):

johnspackman:
thank you @hkollmann :+1:

Issue #9845, opened 2020-01-08
Setting Header Excluded on Table/TreeVirtual causes error on MouseOver


rad-pat:
**Describe the bug**
If you set the HeaderCellsVisible to false for a table/treevirtual, an error is generated when mouseover the control. This is because the getContentLocation() for the header is returning null in this instance at https://github.com/qooxdoo/qooxdoo/blob/c569049d5a1a2dfedd6af1e92b32acf7a8e24a0e/framework/source/class/qx/ui/table/pane/Scroller.js#L2075

Scroller.js:2075 Uncaught TypeError: Cannot read property 'left' of null
    at wrapper.qx.ui.table.pane.Scroller.prototype._getResizeColumnForPageX() [as _getResizeColumnForPageX] (Scroller.js:2075)
    at wrapper.qx.ui.table.pane.Scroller.prototype._onPointermovePane() [as _onPointermovePane] (Scroller.js:1115)

**To Reproduce**
Create tree as below and mouse over, see the errors in the console
`      var tree = new qx.ui.treevirtual.TreeVirtual("Tree");
      tree.set({
              width  : 400,
              rowHeight: 22
            });
      tree.setHeaderCellsVisible(false);
      this.getRoot().add(tree,
                        {
                          left : 10,
                          top : 30,
                          bottom : 30
                        });
      //
      // tree data model
      var dataModel = tree.getDataModel();
      var te = dataModel.addBranch(null, "Desktop", true);
      for (var i = 1; i < 3000; i++)
      {
        dataModel.addLeaf(te, "Some Item #" + i);
      }
      dataModel.setData();
`
**Expected behavior**
No errors

**Desktop (please complete the following information):**
 - OS - Ubuntu 18.04
 - Browser - Chrome
 - Version 79.0.3945.79

cboulanger:
@rad-pat Since this is probably trivial to fix, would you consider providing a pull request with a fix? Thank you.

rad-pat:
Sure, I'll create a PR, i think I have an idea how to fix

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo/pull/9846

PR #9846, opened 2020-01-08
Fix for #9845 - Use pane position if header is not visible

rad-pat:
Uses the pane left position in the case that the header is not visible. If both are excluded then returns -1 for ResizeColumn


derrell (approved):
Looks completely reasonable. Thanks!

cboulanger (approved):

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

PR #9855, opened 2020-01-23
ignore native objects to avoid warnings

derrell:
I am seeing these warnings in an application that doesn't (explicitly) use these classes. This should squelch the warnings... but should the compiler be ignoring these native types?


cboulanger (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

johnspackman (approved):

johnspackman:
The underlying issue is that there are a lot of classes being loaded due to legitimate dependencies, but which in reality do not need to be loaded for a given platform.  I've had to change one of the core classes because it uses a utility method that had the side effect of bringing the entire gesture/pointer events code into a nodejs app.

PR #9857, opened 2020-01-27
fix bug where you can't edit the first cell in a table until the focused cell has changed

johnspackman:
What happens is that the table.startEditing() is called by the tap handler, but the scroller still has it's focused row/col properties set to null even though the table's focused row/col is at `0`.  Because of coercion the scroller does not update its value from `null` to `0`, but later on refuses to allow editing because it has a `null` focus.


oetiker (approved):

derrell (approved):

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman

PR #9861, opened 2020-02-11
qx.ui.list.List: avoid flicker on model update

level420:
When having the property `variableRowHeight` set to true (which is the default), the `qx.ui.list.List` items flicker noticeably when manipulating the model of the list (e.g. be removing an item).
 
This is caused by unnecessarily resetting all item heights in method `__updateGroupRowHeight` (shrink all rows to default) and setting them again later on in `_setRowItemSize`  (set row height from item height) which is called deferred.

This PR avoids this by leaving `__updateGroupRowHeight` when  `variableRowHeight` is true.


oetiker (approved):

framework/source/class/qx/ui/list/List.js
```diff
@@ -678,10 +678,23 @@ qx.Class.define("qx.ui.list.List",
...
+       * In case of having variableItemHeigth set to true,
+       * the group item height has a variable height as well
+       * and will be set again in method _setRowItemSize 
+       * which is a deferred call, being run after all changes.
+       * Resetting the complete item sizes here and setting
+       * the height of the group items, only leads to an
+       * unnecessary flicker of the list items by shrinking and
+       * growing them again.
+       */ 
+      if(this.isVariableItemHeight()) {
```
derrell: For code consistency with qooxdoo style, there should be a space after `if`. Otherwise, this PR looks good.

derrell (changes requested):

cboulanger (approved):

hkollmann (approved):

framework/source/class/qx/ui/list/List.js
level420: done.

derrell (approved):

johnspackman (approved):

level420:
5 approvals. merging. thank you for approvals.

PR #9866, opened 2020-02-18
Fix double click on close button.

lucapivato:
Double clicking the close button causes the form to also maximize.

### DIFF
--- a/framework/source/class/qx/ui/window/Window.js
+++ b/framework/source/class/qx/ui/window/Window.js
@@ -1112,7 +1112,7 @@ qx.Class.define("qx.ui.window.Window",
      */
     _onCaptionPointerDblTap : function(e)
     {
-      if (this.getAllowMaximize()) {
+      if (this.getAllowMaximize() && e.getTarget() === this) {
         this.isMaximized() ? this.restore() : this.maximize();
       }
     },


johnspackman (approved):

oetiker (approved):

hkollmann (approved):

derrell (changes requested):
I'm thinking this is not expected behavior. The close button is part of the title bar. The close button doesn't handle a doubleclick event, so it's passed on to the title bar. One should normally not be double clicking on the close button, and if one does, I think the existing behavior is probably correct. If the user chooses to add other buttons to the title bar, I'm not sure we should predetermine how they, and the title bar, should interact.

In KDE on Linux, double clicking on the close button responds like close, i.e., the close button handles double click as equivalent to single click. "Fixing" this problem, if it needs fixing, is probably best done in the close button handling, not globally for the title bar.

Thoughts?

lucapivato:
I don't think the existing behavior is correct. Double clicking on the close button (or any other button on the caption bar) should not maximize the dialog. The issue is more evident in code that cancels the closing to ask the user to confirm closing. You get both events, close and maximize.

The pull request would work like KDE or Windows: double click on the close button closes the dialog, doesn't maximize it.

derrell (approved):
Ok, I think I'm convinced. Thanks.

oetiker:
The existing behavior is not correct, but 'ignoring' the double click might also not be the right thing ... because I guess when someone double clicks on these buttons they expect the action of the button to happen, so, if I understand @derrell correctly the proper solution would be to treat double click on the buttons the same as single click.

lucapivato:
But the patch is wrong :) Sorry, doesn't take in consideration the inner title widget inside the caption bar. My tests was without a title. How do I recall a pull?

derrell:
You can either close the PR and create a new one, or update your same branch and it will apply to this PR.

PR #9868, opened 2020-02-20
qx.ui.list.List: introduce a separate initialization method for the widget provider

level420:
In some situations it is desirable to override the the provider initialization in order to use a custom widget provider for `qx.ui.list.List`. Overriding the complete virtual list initialization method `_init` does not work because private methods are used therein.
 
This PR introduces a non breaking change by creating an overridable method `_initWidgetProvider` which only includes the instantiation of the widget provider.


cboulanger (approved):

hkollmann (approved):

derrell (approved):
We have this in a number of places in qooxdoo. It doesn't do much good to have a protected method that one is expected to override, that makes use of private variables, since it becomes impossible to override that method. @johnspackman I wonder if the compiler can warn about protected or public methods that make use of private members?

That said, the change requested here seems appropriate given the current situation.

johnspackman (approved):

johnspackman:
>  @johnspackman I wonder if the compiler can warn about protected or public methods that make use of private members?
@derrell you can certainly put it on the wish list :)  I wonder if it would be more helpful for protected than public members though, because IMHO public often deliberately wrap private members as part of encapsulation and documenting a public API; for protected members, there could be an argument that they are intended to have some kind of inner-workings knowledge.

level420:
Normally it's OK to have calls to private members if you have methods which are overridden in order to add code before or after calling `this.base(arguments,...)` without need to change the interna. 
But it fails in my situation, because if I create a new provider after the call to `this.base(arguments,...)` I'd have to destruct/dispose the already created instance.
In my classes/code wherever I'm in this situation I'm offering a override-able "instance creation methhod" which does a minimum of code, something like:
```
_createSomeInstance : function() {
  return new my.Some();
},

_init : function() {
  this.__some = this._createSomeInstance();
  ...
},
```

derrell:
@johnspackman I agree, it's more relevant for protected methods which are intended to be overridden, than for public methods which most frequently are not so intended.

oetiker (approved):

PR #9873, opened 2020-02-28
Replace misspelled method detatchPointerEvents in qx.ui.virtual.selection.Abstract

level420:
offer a correctly spelled detachPointerEvents method and deprecate the false spelled.

### DIFF
--- a/framework/source/class/qx/ui/virtual/selection/Abstract.js
+++ b/framework/source/class/qx/ui/virtual/selection/Abstract.js
@@ -110,8 +110,18 @@ qx.Class.define("qx.ui.virtual.selection.Abstract",
 
     /**
      * Detach pointer events from the managed pane.
+     *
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

hkollmann (approved):

derrell (approved):
Approving this as a good fix, but aren't there in-qooxdoo use cases of the now-deprecated method? If so, they should be corrected to use the properly spelled, new method name, too, shouldn't they?

level420:
@derrell I didn't find any other occurrence of detatchPointerEvents in the whole framework.

derrell:
Ok, great. Thanks for checking.

oetiker (approved):

johnspackman (approved):

level420:
merging. thank you for approving.

PR #9874, opened 2020-02-28
Make qx.ui.layout.Grid private member __colSpans and private method __getOuterSize protected

level420:
in order to allow method `_fixWidthsColSpan` to be override-able.

Without this PR you could not override `_fixWidthColSpan` completely as there is not possibility to access the private members `__colSpans` and `__getOuterSize`.

### DIFF
--- a/framework/source/class/qx/ui/layout/Grid.js
+++ b/framework/source/class/qx/ui/layout/Grid.js
@@ -1298,7 +1298,7 @@ qx.Class.define("qx.ui.layout.Grid",
      * have a protected getter which can be used other methods
      * to make them overridable
      *
-     * @return {Array} the  array
+     * @return {Array} the __colSpans array
      */
     _getColSpans : function() {
       return this.__colSpans;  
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -1191,7 +1192,7 @@ qx.Class.define("qx.ui.layout.Grid",
         };
       }
 
-      if (this.__colSpans.length > 0) {
+      if (this.getColSpans().length > 0) {
```
derrell: Missing an underscore? `_getColSpans` vs `getColSpans`?

derrell (changes requested):

framework/source/class/qx/ui/layout/Grid.js
level420: @derrell thank you! corrected!

cboulanger (approved):

derrell (approved):

oetiker (approved):

hkollmann (approved):

johnspackman (approved):

PR #9860, opened 2020-02-11
#9859 fixed wrong column width calculation with colSpan

Holdwyn:
I've adopted the existing solution for rowHeights with rowSpan.

PR #9875, opened 2020-03-02
Make qx.ui.layout.Grid spanned column width adaption configurable and per default BC

### LINKED ISSUE #9871: Breaking BC change introduced in qx.ui.layout.Grid with PR #9860 with colspan widths
It seems that with PR #9860 layout behaviour of `qx.ui.layout.Grid` changed when having non fixed width widgets like `qx.ui.basic.Atom` within a spanned grid cell.

With the new PR labels seem to grow out of bounds of the spanned columns now.

Here is a playground sample showing the behavior before the PR:
https://tinyurl.com/uqdkx6f

and here is a playground sample after the PR:
https://tinyurl.com/v756dgx

@Holdwyn could you please have a look into the playground samples and see if this is a behaviour which you would expect from your PR. Thank you.

level420:
This PR adds the property allowGrowSpannedCellWidth (default false) which allows to determine the behavior of spanning cells when the widget within a spanning cell needs more place than the width cumulated widths of the spanned columns.

Default false of the new property ensures backwards compatibility. True enables the new behavior introduced with #9860 

Fixes #9871

### DIFF
--- a/docs/desktop/layout/grid.md
+++ b/docs/desktop/layout/grid.md
@@ -37,7 +37,6 @@ Layout properties
 -   **column** *(Integer)*: The column of the cell the widget should occupy. Each cell can only contain one widget. This layout property is mandatory.
 -   **rowSpan** *(Integer)*: The number of rows, the widget should span, starting from the row specified in the `row` property. The cells in the spanned rows must be empty as well. (defaults to `1`)
 -   **colSpan** *(Integer)*: The number of columns, the widget should span, starting from the column specified in the `column` property. The cells in the spanned columns must be empty as well. (defaults to `1`)
--   **allowGrowSpannedCellWidth** *(Boolean)*: Allow growing of spanning cells' widths beyond the cumulated widths of the columns. The default behavior is that the width of the spanning cell is determined by the accumulated width of the columns (plus spacing). Setting this property to true lets the cell width grow as needed to show the widget in the spanning cell, which also enlarges the width of the spanned columns. (defaults to `true`)
 
 Alternative Names
 -----------------
--- a/docs/desktop/layout/grid.md
+++ b/docs/desktop/layout/grid.md
@@ -37,7 +37,7 @@ Layout properties
 -   **column** *(Integer)*: The column of the cell the widget should occupy. Each cell can only contain one widget. This layout property is mandatory.
 -   **rowSpan** *(Integer)*: The number of rows, the widget should span, starting from the row specified in the `row` property. The cells in the spanned rows must be empty as well. (defaults to `1`)
 -   **colSpan** *(Integer)*: The number of columns, the widget should span, starting from the column specified in the `column` property. The cells in the spanned columns must be empty as well. (defaults to `1`)
--   **allowGrowSpannedCellWidth** *(Boolean)*: Allow growing of spanning cells widths beyond the cumulated widths of the columns. The default behavior is that the width of the the spanning cell is determined by the cumulated width of the columns (plus spacing). Setting this property to true lets the cell width grow as needed to show the widget in the spanning cell, which also enlarges the width of the spanned columns. (defaults to `true`)
+-   **allowGrowSpannedCellWidth** *(Boolean)*: Allow growing of spanning cells' widths beyond the cumulated widths of the columns. The default behavior is that the width of the spanning cell is determined by the accumulated width of the columns (plus spacing). Setting this property to true lets the cell width grow as needed to show the widget in the spanning cell, which also enlarges the width of the spanned columns. (defaults to `true`)
 
 Alternative Names
 -----------------
--- a/docs/desktop/layout/grid.md
+++ b/docs/desktop/layout/grid.md
@@ -37,6 +37,7 @@ Layout properties
 -   **column** *(Integer)*: The column of the cell the widget should occupy. Each cell can only contain one widget. This layout property is mandatory.
 -   **rowSpan** *(Integer)*: The number of rows, the widget should span, starting from the row specified in the `row` property. The cells in the spanned rows must be empty as well. (defaults to `1`)
 -   **colSpan** *(Integer)*: The number of columns, the widget should span, starting from the column specified in the `column` property. The cells in the spanned columns must be empty as well. (defaults to `1`)
+-   **allowGrowSpannedCellWidth** *(Boolean)*: Allow growing of spanning cells widths beyond the cumulated widths of the columns. The default behavior is that the width of the the spanning cell is determined by the cumulated width of the columns (plus spacing). Setting this property to true lets the cell width grow as needed to show the widget in the spanning cell, which also enlarges the width of the spanned columns. (defaults to `true`)
 
 Alternative Names
 -----------------
--- a/framework/source/class/qx/ui/layout/Grid.js
+++ b/framework/source/class/qx/ui/layout/Grid.js
@@ -51,6 +51,12 @@
  *   widget should span, starting from the column specified in the <code>column</code>
  *   property. The cells in the spanned columns must be empty as well.
  * </li>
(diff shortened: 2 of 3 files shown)


framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -93,8 +99,10 @@ qx.Class.define("qx.ui.layout.Grid",
    *     Sets {@link #spacingX}.
    * @param spacingY {Integer?0} The vertical spacing between grid cells.
    *     Sets {@link #spacingY}.
+   * @param allowGrowSpannedCellWidth {Boolean?false} Allow growing of spanning cell widths beyond the cumulated widths of the columns.
+   *     Sets {@link #allowGrowSpannedCellWidth}.
    */
-  construct : function(spacingX, spacingY)
+  construct : function(spacingX, spacingY, allowGrowSpannedCellWidth)
```
oetiker: hmmm I am  bit ambivalent about adding extra positional arguments  ...

cboulanger (approved):

johnspackman (approved):

framework/source/class/qx/ui/layout/Grid.js
level420: I thought that this is suitable because it is a global acting layout property, just as spacingX and spacingY are.

hkollmann (approved):

docs/desktop/layout/grid.md
```diff
@@ -37,6 +37,7 @@ Layout properties
 -   **column** *(Integer)*: The column of the cell the widget should occupy. Each cell can only contain one widget. This layout property is mandatory.
 -   **rowSpan** *(Integer)*: The number of rows, the widget should span, starting from the row specified in the `row` property. The cells in the spanned rows must be empty as well. (defaults to `1`)
 -   **colSpan** *(Integer)*: The number of columns, the widget should span, starting from the column specified in the `column` property. The cells in the spanned columns must be empty as well. (defaults to `1`)
+-   **allowGrowSpannedCellWidth** *(Boolean)*: Allow growing of spanning cells widths beyond the cumulated widths of the columns. The default behavior is that the width of the the spanning cell is determined by the cumulated width of the columns (plus spacing). Setting this property to true lets the cell width grow as needed to show the widget in the spanning cell, which also enlarges the width of the spanned columns. (defaults to `true`)
```
derrell: add possessive apostrophe to "cells" so it becomes "Allow growing of spanning cells' widths..."

replace "cumulated" with "accumulated" throughout

double "the" in "is that the width of THE THE spanning cell"

framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -51,6 +51,12 @@
  *   widget should span, starting from the column specified in the <code>column</code>
  *   property. The cells in the spanned columns must be empty as well.
  * </li>
+ * <li><strong>allowGrowSpannedCellWidth</strong> <em>(Boolean)</em>: Allow growing 
+ * of spanning cells widths beyond the cumulated widths of the columns. The default behavior 
+ * is that the width of the the spanning cell is determined by the cumulated width of the
+ * columns (plus spacing). Setting this property to true lets the cell width grow as needed to
+ * show the widget in the spanning cell, which also enlarges the width of the spanned columns.
+ * </li>
```
derrell: Same comments as above

framework/source/class/qx/ui/layout/Grid.js
derrell: I think I agree with @oetiker that it's not necessary there, and it doesn't really match other classes which also have spacing parameters but not anything like this. It's easy enough to set the property if something other than the default is desired.

framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -108,6 +116,10 @@ qx.Class.define("qx.ui.layout.Grid",
     if (spacingY) {
       this.setSpacingY(spacingY);
     }
+    
+    if(allowGrowSpannedCellWidth === true || allowGrowSpannedCellWidth === false) {
```
derrell: If that parameter is removed, this block would be removed as well. If the parameter remains, then a more common paradigm might be `if (typeof allowGrowSpannedCellWidth == "boolean")`

framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -995,7 +1018,11 @@ qx.Class.define("qx.ui.layout.Grid",
         // increment the preferred column sizes.
         if (prefSpanWidth < hint.width)
         {
-          if (!qx.lang.Object.isEmpty(colFlexes)) {
+          // Do not adapt column widths to the width
+          // of the spanning cell if allowGrowSpannedCellWidth property
+          // is set to false
+          // See https://github.com/qooxdoo/qooxdoo/issues/9871
+          if (this.getAllowGrowSpannedCellWidth() === false || !qx.lang.Object.isEmpty(colFlexes)) {
```
derrell: What's here isn't wrong, but `if (! this.getAllowGrowSpannedCellWidth() || ...` vs `=== false` is more common

derrell (changes requested):

docs/desktop/layout/grid.md
level420: @derrell regarding your first comment: I meant here cells in plural. But you're the native speaker. should I change this then?

The other comments: will change that soon

derrell: Yes. In English, the general rule for possessive of a single thing is the singular form followed by `'s`, so if it belongs to a single cell, it would be `cell's`. If you are talking about belong to all of the cells, then the apostrophe comes after the plural form, so in this case, `cells'`.

level420: Ah OK. Changed that.

framework/source/class/qx/ui/layout/Grid.js
level420: @derrell removed the additional parameter

docs/desktop/layout/grid.md
level420: I mean changed it back again.

framework/source/class/qx/ui/layout/Grid.js
level420: done.

framework/source/class/qx/ui/layout/Grid.js
```diff
@@ -140,6 +140,21 @@ qx.Class.define("qx.ui.layout.Grid",
       check : "Integer",
       init : 0,
       apply : "_applyLayoutChange"
+    },
+    
+    
+    /**
+     * Allow growing of spanning cells widths beyond the accumulated widths of the columns.
```
derrell: I think you didn't understand me. This is pedantic, but if you want it correct, change this to:
`Allow growing of spanning cells' widths beyond...`
Note the apostrophe after "cells" to indicate that it is the widths of those cells that is being discussed.

oetiker (approved):

framework/source/class/qx/ui/layout/Grid.js
level420: OK. OK. Back again.

level420: Found the 11 apostrophe rules here : https://www.grammarbook.com/punctuation/apostro.asp
😁

derrell: Are you loving the English language and all of its inconsistencies, yet? :-)

derrell (approved):

framework/source/class/qx/ui/layout/Grid.js
level420: can't wait to see your first German text 😄

level420:
merging. thank you for the approvals and especially for @derrell and his English lessons 😁

Issue #9485, opened 2018-01-19
Add ${qx.iconset} environment variable and replace hardcoded icontheme


cboulanger:
@johnspackman has looked at the framework code and found that the iconthme "Tango" is often hardcoded. This needs to be replaced by `${qx.icontheme}`. Better yet, should add `${qx.iconset}` defined  as `/qx/icon/${qx.icontheme}` and then change references in the framework to just use @asset(${qx.iconset}/22/etc/etc.png).

cboulanger:
To take this a step further: It seems a bad idea to use hardcoded icon paths at all. To make theming possible that is independent of the (arbitrary) naming and ordering of the Tango and Oxygen icon sets that are shipped with the framework (side note: they look very 1990-ish), icon paths should always be aliases which are resolved by the theme used.

cboulanger:
Sorry for  just having opinions and not contributing any code ATM (just no time) - just jotting down some conceptual thoughts here: We should think of icons not in terms of their actual paths and what they they actually represent, but in terms of their _function_. For example, I use the alias "icon/16/apps/utilities-network-manager.png" from Oxygen in a tree to represent the root node of a access control list, even though it has nothing to do with a network. If I change the iconset, I might use a completely different icon from Tango (or whatever other iconset). That is why it is so important to establish a "best practice" in which users define icon aliases in the icon theme class and NOT assets in the classes. This might mean a bit more work upfront but prevent enormous headaches later. Qooxdoo already has all the tools for this approach and we should use it in the framework itself....

hkollmann:
You are right. On the other side this is a lot of work. Therefore we should fix the bug as a first step.

hkollmann:
The first step is done, all hardcoded themes are gone.
The rest is enhancement vor Version 7

PR #9878, opened 2020-03-03
remove hardcoded tango

hkollmann:
see https://github.com/qooxdoo/qooxdoo/issues/9485.

- removed the last hardcoded /Tango/ ref.
- fixed testing back to hardcoded assets,

### DIFF
--- a/framework/source/class/qx/test/bom/Attribute.js
+++ b/framework/source/class/qx/test/bom/Attribute.js
@@ -20,7 +20,7 @@
 ************************************************************************ */
 /**
  *
- * @asset(qx/icon/${qx.icontheme}/48/places/folder.png)
+ * @asset(qx/icon/Tango/48/places/folder.png)
  */
 
 qx.Class.define("qx.test.bom.Attribute",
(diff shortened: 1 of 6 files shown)


framework/source/class/qx/test/bom/Attribute.js
```diff
@@ -20,7 +20,7 @@
 ************************************************************************ */
 /**
  *
- * @asset(qx/icon/${qx.icontheme}/48/places/folder.png)
+ * @asset(qx/icon/Tango/48/places/folder.png)
```
derrell: This looks like a reverse diff. The stuff being added is Tango, with qx.icontheme being removed.

derrell (changes requested):

framework/source/class/qx/test/bom/Attribute.js
hkollmann: This is correct. I reverted the asserts for the unit tests.
Later in the unit test code  qx/icon/Tango/48/places/folder.png is used for testing.

cboulanger (approved):

oetiker (approved):

johnspackman (approved):

derrell (approved):

level420 (approved):

PR #9880, opened 2020-03-05
Promise docs

### LINKED ISSUE #9666: Write documentation on async events / qx.Promise
We need documentation on async events and qx.Promise and a short summary of it for blog/website

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/9666


docs/core/promises.md
```diff
@@ -1,14 +1,80 @@
...
+In that example, the console will print:
+```
+Connecting to server responded: Success
+```
+
+Promises are the code that makes `async` and `await` work in ES6 (which you can also use cross platform with Qooxdoo, provided you use the qooxdoo-compiler):
+
+```javascript
+async function connectToServer() {
+    await new qx.Promise((resolve, reject) => {
```
derrell: An async function returns a promise. I think this should be `return new qx.Promise(...` rather than `await new qx.Promise(...`

docs/core/promises.md
```diff
@@ -1,14 +1,80 @@
...
+main();
+```
+
+
+## Creating Asynchronous Properties
+When you write a Qooxdoo object, you typically create properties with `apply` methods and change events, and its quite possible that your apply method 
+could take some time to complete - for example, setting a property could require an asynchronous server roundtrip.  When this happens, it is necessary 
+to delay the firing the change event until the apply is complete, otherwise the event could be fired before the property value is fully applied.
+
+The return value for apply method can be a qx.Promise, in which case promise will wait for the promise to resolve successfully before firing the event; if
```
derrell: in which case *qooxdoo* will wait...

derrell (changes requested):
There's too much going on here to figure it all out, but I did notice a few things in documentation.

johnspackman:
ugh, sorry I screwed up my git again :(

johnspackman:
closing this and I'll open a new one shortly, incorporating @derrell 's comments

Issue #9876, opened 2020-03-02
qooxdoo's jsonrpc client needs some love


cboulanger:
- [x] Remove "experimental" status
- [x] "official" jsonrpc v2.0 support in docs/code
- [x] Make v2.0 default protocol
- [x] higher-level events (see https://github.com/teambition/jsonrpc-lite#jsonrpcparsemessage)
- [x] Promise support
- [x] Update http://www.qooxdoo.org/docs/#/communication/rpc, moving deprecated stuff to subpage or link to legacy docs
- [x] Update http://www.qooxdoo.org/docs/#/communication/rpc_server_writer_guide
- [x] Update http://www.qooxdoo.org/docs/#/about?id=communication

cboulanger:
Ongoing work is here: https://github.com/qooxdoo/qooxdoo/tree/cboulanger-modernize-jsonrpc

cboulanger:
Development of promise-based implementation has moved to https://github.com/qooxdoo/incubator.qx.io.jsonrpc . Closing

PR #9882, opened 2020-03-05
Modernize the qx.io.remote.Rpc API

cboulanger:
Addresses #9876. I have updated the documentation on communication generally and removed the deprecation of `qx.io.remote.*` which might be deficient architecturally, but is here to stay for the moment.


oetiker (approved):

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -1072,9 +1102,320 @@ qx.Class.define("qx.io.remote.Rpc",
...
+                  "Invalid jsonrpc data: Unknown request id",
+                  {
+                    request : data,
+                    response: response
+                  }
+                );
+                return reject(ex);
+              }
+              // remove the entry from the queue
+              this.__requestQueue.splice(this.__requestQueue.indexOf(originalRequest), 1);
```
derrell: This, too, is a slow operation that would be made infinitely faster if `__requestQueue` were a map instead of an array.

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -1072,9 +1102,320 @@ qx.Class.define("qx.io.remote.Rpc",
...
+                qx.io.remote.RpcError.type.INVALID_DATA,
+                "Invalid jsonrpc data for "+ method + ": result is missing.",
+                {
+                  request : originalRequest.data,
+                  response: response
+                }
+              );
+              return originalRequest.promise.reject(ex);
+            }
+            // response is a server call or notification
```
derrell: What is this? Is this a holdover from the qx1 server thing? I don't think it should be possible nor allowed here, right?

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -1072,9 +1102,320 @@ qx.Class.define("qx.io.remote.Rpc",
...
+    /**
+     * Sends a notification immediately without queueing it.
+     * See {@link qx.io.remote.Rpc#addNotification} and
+     * {@link qx.io.remote.Rpc#send}.
+     *
+     * @param method {String}
+     * @param params {Array|undefined}
+     */
+    sendNotification : function(method, params) {
+      this.addNotification(method, params);
```
derrell: This is inconsistent. `sendRequest` sends it without queuing it but `sendNotification` queues it. I think it should not be queued here.

derrell (changes requested):
@cboulanger Thanks for your work on this! The promise interface is a very nice addition, and upgrades to documentation are always welcome!

It is not appropriate to change the formatting of existing code, when you're modifying a file. Firstly, we use file-by-file code format consistency in qooxdoo, so you should follow the format already used in that file. There are differing opinions on the team about formatting, and our solution is to leave it file-by-file. Regardless of preference, it also makes it really difficult to see what the substance changes are during a code review, when 75% of the changes are formatting changes, as is the case in Rpc.js. I likely missed some things as a result. I'd appreciate if you'd revert the formatting changes, and I'll re-review to see what I missed.

That said, this is great. I've made a bunch of comments. Let me know if you disagree with any of them. If any of them don't make sense to you, and I'll try to explain.

docs/communication/README.md
```diff
@@ -0,0 +1,21 @@
+# Communication
+
+qooxdoo provides tree different communication APIs with a rich feature set:
```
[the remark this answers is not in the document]
cboulanger: Done, and rewritten as discussed on Gitter

docs/communication/jsonrpc.md
```diff
@@ -0,0 +1,204 @@
...
+The first parameter is the URL of the backend (in this example a backend on
+localhost). The second is the name of the service you'd like to call. 
+
+To work with the service, the Rpc class provides two different APIs, a newer
+one based on Promises, and a legacy one based on callbacks or events. Note that
+you cannot mix the two APIs because the JSONRPC ids are computed differently.
+
+## Promise-based API
+
+The Promise-based API fully conforms to the [JSON-RPC v2 specification](https://www.jsonrpc.org/specification),
```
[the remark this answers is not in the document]
cboulanger: changed to "The Promise-based API fully implements"

docs/communication/jsonrpc.md
```diff
@@ -0,0 +1,204 @@
...
+    var promise1 = client.addRequest("method1", ["foo"]);
+    var promise2 = client.addRequest("method2", ["bar"]);
+    var promise3 = client.addRequest("method3", ["baz"]);     
+    var promise4 = client.send();
+```
+
+Promises 1-3 will resolve with the individual results (or errors) and reject
+with jsonrpc errors for this specific call. Promise 4 will resolve with the
+jsonrpc data for all requests or reject with any transport-related error. This
+means you need to handle errors for both the individual requests and for `send()`.
```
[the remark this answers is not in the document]
cboulanger: The API offers to send a single request right away or to queue requests until they are sent in  a batch. That's as convenient as it gets :-) I'll try to explain it better.

cboulanger:
What I think I will do is to revert `qx.io.remote.Rpc` completely and put the new API into a new namespace and class `qx.io.remote.jsonrpc.Http`. This is a much cleaner solution, doesn't change any formatting and doesn't retrofit an existing API with a substantially different one. This will also allow to add other jsonrpc implementations such as one using websockets into this namespace.

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -1072,9 +1102,320 @@ qx.Class.define("qx.io.remote.Rpc",
...
+                  {
+                    request : data,
+                    response: response
+                  }
+                );
+                return reject(ex);
+              }
+              ids.push(id);
+              // find original request
+              var originalRequest = this.__requestQueue.find(function (item) {
```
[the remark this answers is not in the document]
cboulanger: good idea.

framework/source/class/qx/io/remote/Rpc.js
cboulanger: I don't understand - can you clarify what you mean?

framework/source/class/qx/io/remote/Rpc.js
cboulanger: It does the same thing - first queuing (1393/1409), then sending (1394/1410) the request/notification.

docs/communication/jsonrpc.md
```diff
@@ -0,0 +1,204 @@
...
+not return promises. 
+
+### Bi-directional calls
+
+Because of the client-server architecture of HTTP, which is used for
+transport, the bidirectional (peer-to-peer), real-time character of
+JSONRPC v2 cannot be fully implemented. Instead, any server-initiated
+remote procedure calls "piggy-back" on the responses to  client-initialed
+calls. If you want to continually listen to server calls, you must use
+polling or a different implementation.
```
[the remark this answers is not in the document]
cboulanger: In my opinion, it is totally up to the developer what s/he finds appropriate in this case, but I will remove the section.

docs/communication/request_io.md
```diff
@@ -1,119 +1,174 @@
...
-qooxdoo ships with two transport methods, interfaced by `qx.io.request.Xhr` and `qx.io.request.Jsonp`.
+## Choosing an appropriate transport
 
--   Choose `Xhr` whenever you can. `Xhr` offers true HTTP client functionality and exposes metadata associated with HTTP requests. It is agnostic of the data interchange format and does not make any specific demands on the backend.
--   If you are making cross-origin requests and need to support all popular browsers and/or the target server is not configured to accept cross-origin request (`Access-Control-Allow-Origin` header), you will need to use `Jsonp`. Only JSON is supported as data interchange format and the server needs to wrap responses in a JavaScript function call.
+qooxdoo ships with two transport methods, interfaced
+by `qx.io.request.Xhr` and `qx.io.request.Jsonp`.
 
-`Xhr` and `Jsonp` share a common interface. `AbstractRequest` defines the lowest common denominator of both transport methods.
+-   Choose `Xhr` whenever you can. `Xhr` offers true HTTP client functionality
```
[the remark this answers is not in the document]
cboulanger: I didn't -this was in the documentation.

docs/communication/request_io.md
```diff
@@ -1,119 +1,174 @@
...
-    req.bind("response", label, "value");
 
-Debugging
----------
+## Debugging
 
-If you encounter odd behavior, it might help to enable debugging of the IO classes. Debugging is controlled with the `qx.debug.io` setting. Provided you have allowed URL settings (`allowUrlSetting`), you can simply append `?qxenv:qx.debug.io:true` to the URL of your application.
+If you encounter odd behavior, it might help to enable debugging of
+the IO classes. Debugging is controlled with the `qx.debug.io` setting.
+Provided you have allowed URL settings (`allowUrlSetting`), you can
```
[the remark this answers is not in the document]
cboulanger: I don't know either - this needs to be clarified.

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -15,67 +15,50 @@
...
- * be expensive), so you as the user are responsible for avoiding them.
+ * converted automatically to and from their JSON representation.
+ *
+ * Server-initiated calls
+ *
+ * Because of the client-server architecture of HTTP, which we use here, the
+ * bidirectional (peer-to-peer), real-time character of JSONRPC v2 cannot
+ * be fully implemented. Instead, any server-initiated remote procedure
+ * calls "piggy-back" on the responses to  client-initialed calls. If
+ * you want to continually listen to server calls, you must use polling
```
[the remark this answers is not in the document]
cboulanger: It may be a "kludge" but I have use cases where websockets are not an option and a reverse transport via HTTP response is a rather elegant solution to that problem. But  I'll put it somewhere else.

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -84,48 +67,40 @@ qx.Class.define("qx.io.remote.Rpc",
...
-   *                                implementation, this is the fully
-   *                                qualified name of the class that offers
-   *                                the service methods
-   *                                (e.g. "my.pkg.MyService").
+   * @param serviceName {String}
+   *  identifies the service (dependent on the server implementation).
    */
-  construct : function(url, serviceName)
-  {
+  construct : function(url, serviceName) {
```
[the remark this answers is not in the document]
cboulanger: I will put my code in a different class and fully revert this class - I think that is the cleanest solution.

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -466,34 +523,42 @@ qx.Class.define("qx.io.remote.Rpc",
...
+        };
+        // parameters are optional according to the spec
+        if (parameters) {
+          if (qx.lang.Type.isObject(parameters) || qx.lang.Type.isArray(parameters)) {
+            requestObject["params"] = parameters
+          } else {
+            throw new Error("Parameters must be type Array (positional) or Object (named)");
+          }
+        }
+        // id: null is allowed but discouraged
```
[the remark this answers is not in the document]
cboulanger: The spec says this:
-  "A Notification is a Request object without an "id" member. "
- "The use of Null as a value for the id member in a Request object is discouraged, because this specification uses a value of Null for Responses with an unknown id. Also, because JSON-RPC 1.0 uses an id value of Null for Notifications this could cause confusion in handling."

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -545,33 +608,35 @@ qx.Class.define("qx.io.remote.Rpc",
 
       req.setCrossDomain(this.getCrossDomain());
 
-      if (this.getUsername())
-      {
+      // deprecated
```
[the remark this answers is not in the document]
cboulanger: Yes, I'll leave it alone.

framework/source/class/qx/io/remote/Rpc.js
derrell: My comment may have been one line too early. I think what I meant was, is it possible for `typeof method` to be undefined in 2.0 (line 1370)?

framework/source/class/qx/io/remote/Rpc.js
derrell: Yes, sorry. I misread it. I wonder, though, if this is the best implementation? If you're in the process of building up a batch, would it be useful to be able to send a request or a notification that is unrelated to that batch, e.g., addRequest, addRequest, sendRequest, addRequest, send... which would end up actually sending the single request of `sendRequest` before the three-request batch with sent with `send`...

cboulanger:
Closing. Reverting becomes to messy, I'll start anew.

framework/source/class/qx/io/remote/Rpc.js
cboulanger: Ok, you're sayint that a batch should be its own object, so building batches could be done separately? I like the idea conceptually, but can you think of a use case in which you would use it?

derrell: Say I wanted to send updates only when I've accumulated 20 of them, but I also have some periodic request I need to send. At present, those updates, each a call to `addRequest`, would be sent along with my periodic request (a call to `sendRequest`) even though I really want only the periodic request to be sent at that moment, and desire the updates to be sent later once I've accumulated enough of them. (Granted, I could create my own array of them, but in that case, I would prefer to pass an array to a (non-existent) `sendRequests` method or something of the sort.)

cboulanger: I am convinced, I will create a `qx.io.jsonrpc.Batch` object which can be used in such a case, plus the shorthand methods which send just one single call.

cboulanger: Maybe like this:
`qx.io.jsonrpc.Request` -> serializes to JSON
`qx.io.jsonrpc.Notification` -> serializes to JSON
`qx.io.jsonrpc.Batch.add()` -> adds one of the above
`qx.io.jsonrpc.Http` -> instance, has a `send()` method which takes one of the three objects above as an argument, i.e. either a request, a notification or a batch. 
What do you think?

cboulanger: Or should the client instance be generic and take a transport object as an injected dependency?

derrell: > Maybe like this:
> `qx.io.jsonrpc.Request` -> serializes to JSON
> `qx.io.jsonrpc.Notification` -> serializes to JSON
> `qx.io.jsonrpc.Batch.add()` -> instance, adds one of the above
> `qx.io.jsonrpc.Http` -> client instance, has a `send()` method which takes one of the three objects above as an argument, i.e. either a request, a notification or a batch.
> What do you think?

I like that general idea, and I like the injected transport concept to go along with it.  You could then do something like:
```
let batch1 = new qx.io.jsonrpc.Batch();
let batch2 = new qx.io.jsonrpc.Batch();
batch1.add(new qx.io.jsonrpc.Request(...)); // can intermix requests and notifications
batch2.add(new qx.io.jsonrpc.Request(...)); // to multiple batches
batch1.add(new qx.io.jsonrpc.Notification(...));
batch2.add(new qx.io.jsonrpc.Notification(...));
let transport = new qx.io.jsonrpc.transport.Http(configMap);
transport.send(batch1); // send a specified batch
transport.send(new qx.io.jsonrpc.Request(...); // directly send a specific request or notification
transport.send(batch2);
```

cboulanger: Cool. This and maybe some sugar object `const client = new qx.io.jsonrpc.Client("https://my.server.com/endpoint")` which auto-detects the transport type from the URL and has some convenience methods - but basically, you can work with the transports themselves.

Issue #9230, opened 2016-11-16
Method resetSelection of class qx.ui.list.List doesn't work


goldim:
The class `qx.ui.form.List` in comparison to class `qx.ui.list.List` has properly working method `resetSelection`. Instead of `mylist.resetSelection()` I use the following operation `mylist.setSelection(new qx.data.Array())`.

level420:
@goldim do you mean that ```qx.ui.formList.resetSelection``` does not work?

goldim:
@level420 no, I mean qx.ui.list.List.
Playground example: http://tinyurl.com/zleldu6

level420:
@goldim thank you for clarifying. Do you have some time investigating the cause of this bug? Best would be if you can offer a PR for this ;-)

goldim:
@level420 It's honor for me to take part in improving `qooxdoo` but I am not sure that I can cope with the problem because I don't have any knowledge about internal structure of the framework. But I'll try today, tomorrow or probably in the holidays. If I can not do anything I'll let you know.

level420:
@goldim thank you for helping the community!

stahl-riesa:
Instead of using the resetSelection method use this.getSelection().removeAll() to clear the selection.

The difference is how the selection is stored: qx.ui.form.list uses a private varibale, qx.ui.list.list a property exspecting an qooxdoo array like object. So if you change the selection property of qx.ui.list.list, you change the whole 'selection manager' array.

Maybe it could be helpful to extend qx.ui.virtual.selection.MModel's _applySelection method from
```
value.addListener("change", this._onChangeSelection, this);
if (old != null) {
  old.removeListener("change", this._onChangeSelection, this);
}
this._onChangeSelection();
```
to
```
value.addListener("change", this._onChangeSelection, this);
if (old != null) {
  old.removeAll(); <<< ADDED
  old.removeListener("change", this._onChangeSelection, this);
}
this._onChangeSelection();
```

What do you think?

goldim:
@stahl-riesa Thank you for reply. I tried to find a solution which will resolve this bug. But I can not do anything because of qooxdoo property system. How you said above it is needed to write own methods reset, init, set (etc) and private field to fix this bug. Now about using `this.getSelection().removeAll()` instead of `resetSelection` I've used `setSelection(new qx.data.Array())` this is the same thing and still doesn't resolve the problem with the method `resetSelection`.
[...]

PR #9883, opened 2020-03-06
fixes https://github.com/qooxdoo/qooxdoo/issues/9230

johnspackman:


framework/source/class/qx/ui/virtual/selection/MModel.js
```diff
@@ -46,14 +46,16 @@ qx.Mixin.define("qx.ui.virtual.selection.MModel",
...
+    this.__selection.addListener("change", this._onChangeSelection, this); 
+    this._applySelection(this.__selection, null);
   },
 
 
   properties :
   {
     /** Current selected items */
+    /* psuedo property - implemented manually in code below
     selection :
```
level420: Why is the property then needed at all?
Shouldn't this completely be removed as the automatic created getters and setters will else fight with the manual created ones?

level420 (changes requested):

framework/source/class/qx/ui/virtual/selection/MModel.js
johnspackman: no, that's why it's commented out - the selection property is needed, but the property mechanism isn't quite good enough so the property has to be implemented with hand crafted functions and events.  When I have to do this, I tend to put in a commented-out property and mark it as psuedo, just so that there is some kind of consistent documentation

cboulanger (approved):

oetiker (approved):

framework/source/class/qx/ui/virtual/selection/MModel.js
level420: Oh! I didn't notice that it was commented out.

level420 (approved):

level420:
@johnspackman I've hijacked the PR with this two commits because  the travis CI failed
https://github.com/qooxdoo/qooxdoo/pull/9883/commits/5428f646788555ea79cc513f6a3b033af68c4a4b
https://github.com/qooxdoo/qooxdoo/pull/9883/commits/00426ed179edfa8ceeea18685133d5895ca4c4bb

hkollmann (approved):
Great idea!

hkollmann (changes requested):
You should fire the changeSelection event somewhere.
This was done before through the property.

level420:
@hkollmann I think the `changeSelection` event is fired here: https://github.com/qooxdoo/qooxdoo/blob/90e7ee08e5551fbd196f75da1a52d6069b147124/framework/source/class/qx/ui/virtual/selection/MModel.js#L382

johnspackman:
> You should fire the changeSelection event somewhere.
> This was done before through the property.

There's no need because the value of the property never changes - the event is listed so that it appears to be a "real" property.  Because it is a qx.data.Array, the caller would have had to listen for `changeSelection` and when the value changes, remove & add `change` listeners on the value; the caller code does not have to be changed, but they will never need to remove & add `change` listeners

johnspackman:
@level420 thanks for making the changes to this PR, it's a very bad habit of mine to not run the tests etc before pushing :(

level420:
@johnspackman hmmm... but how do we know if `the selection` of the model and not the selection array changed? The semantics implies to me that I'm getting the event ... well ... when the selection changes.

johnspackman:
the value of `selection` never changes - the constructor sets an initial value, and then the `setSelection` just makes sure that the current value is updated rather than replaced, so therefore a `changeSelection` does not need to be fired

level420:
@johnspackman shouldn't then the event be removed? It is useless then. 

If someone doesn't know the internals of the implementation the first thought when reading the event name `changeSelection` is that the "selection" as per simple understanding (changed the selected list items by interaction or programmatically) of the list changed!

johnspackman:
The event has to stay because without it either binding will not work at all, or when binding you get a message about how "changeSelection" is not a supported event.

Ultimately, this is a rather awkward work around for limitations in the property mechanism

johnspackman:
> If someone doesn't know the internals of the implementation the first thought when reading the event name changeSelection is that the "selection" as per simple understanding (changed the selected list items by interaction or programmatically) of the list changed!

But that's ok, in that it will not cause a problem for them - just like existing code will not be broken.

cboulanger:
@level420 This is true, but I think it's something that needs to be (better?) documented  rather than changing the code to match users' expectations. You need to observe the selection array's events, not the selections object itself, i.e. `.getSelection().addlListener("change")` . Alternatively, if you bind the selection directly to something, you don't even have to think about it because it should simply work.

level420:
@cboulanger no I meant to remove the `selectionChange` event to not have false expectations what the event does. Now with the PR of @johnspackman it does nothing and is never fired.

johnspackman:
I think the confusion here is that all code must always `list.addListener("changeSelection"...` *and* must always `list.getSelection().addListener("change"...`.  There is no way in the property mechanism to specify that the `list.selection` property will never change it's value, and actually whether or not it changes its value is a private implementation detail.

It's entirely legitimate for someone to rewrite that list code, and make `list.setSelection` change the array instance - and as backwards compatibility goes, the previous implementation would do exactly that.

What I really would like to do is to make it so that Qooxdoo properties implement the same contract of behaviour as they do now, but allow the class to override every aspect about how the data is stored... but that's another PR on another day, so the comment is a kind of half way house

hkollmann (approved):

level420:
@johnspackman so in the end you argue that the `selectionChange` event should stay there even if it its never fired because someone may override the implementation and in this implementation it may be that a `selectionChange` event will be fired? 

But wouldn't then this implementation be also responsible to declare that event?

For me it doesn't feel right to have an event which is never fired just to have it handy for a scenarion which may never happen. Those using `qx.ui.list.List` will not understand why and for what purpose it exists.

johnspackman:
I'm saying that the `selection` is a property and that if you replace the Qooxdoo-manufactured implementation with a roll-your-own version, you need implement the property completely; and that if you don't, then other features (such as binding) will not work properly.

If you remove the event declaration, then not only is binding affected, but the `selection` property is not properly backwards compatible either.

level420:
OK understood. Then we may also need to implement `initSelection` to complete the property interface

derrell (approved):

PR #9888, opened 2020-03-10
Change jsonrpc default protocol to "2.0"

cboulanger:
This is BC because anyone who still uses the original "qx1" protocol can switch back by setting the "protocol" property accordingly.

### DIFF
--- a/framework/source/class/qx/io/remote/Rpc.js
+++ b/framework/source/class/qx/io/remote/Rpc.js
@@ -390,7 +390,7 @@ qx.Class.define("qx.io.remote.Rpc",
      */
     protocol :
     {
-      init : "qx1",
+      init : "2.0",
       check : function(val) { return val == "qx1" || val == "2.0"; }
     }
   },


johnspackman (approved):

level420 (approved):

framework/source/class/qx/io/remote/Rpc.js
```diff
@@ -390,7 +390,7 @@ qx.Class.define("qx.io.remote.Rpc",
      */
     protocol :
     {
-      init : "qx1",
+      init : "2.0",
```
level420: I'm not using it, but this is a BC breaking change. Or did I misunderstand something?

cboulanger: It's a breaking change but can be easily made BC by setting the property. I don't think many people are using the legacy mode. We can put it into the relase notes under "breaking changes".

oetiker: has there been any other mode than the qx1 mode ?

oetiker: all our apps use qx jsonrpc

hkollmann (commented):
Our's too. In qx1 mode.

cboulanger:
Actually, I don't care so much - we can leave it in "qx1" mode, given that I will add a wholly separate jsonrpc v2 client. I thought the idea was to deprecate "qx1" mode since it is non standard-compliant...

hkollmann:
Is this change really necessary for your new implementation? Otherwise i would prefer not to change the default

cboulanger:
@hkollmann No it is not necessary. I had the (wrong) impression that we agreed to deprecate "qx1". I can close this, no problem!

hkollmann:
Thanks, We can deprecate it with your new implementation. IMHO a lot of older programs use qx1 - they are programmed before rpc2 exists.

derrell:
We're not releasing the new code as part of the upcoming 6.0 release. We don't want new code using the qx1 variant of JSON RPC; we want them using 2.0. I feel pretty strongly that we should, in fact, make this change, and document it in the release notes. It's a trivial change to existing code, and gives people a JSON RPC 2.0 client they can use as 6.0 comes out the door.

hkollmann (approved):

cboulanger:
I probably need to update the tests for "qx1" mode.

derrell (approved):

oetiker (approved):

oetiker:
so with this change we are activating a protocol implementation that is not used and is going to be replaced with a new implementation anyway ? right ?

derrell:
@oetiker I'm not quite sure where you got that, but no, not right. JSON-RPC 2.0 is many years old and has been in extensive use. There are lots and lots of existing JSON-RPC 2.0 servers out there, and the 2.0 client code has been in use in qooxdoo for a number of years. I expect that the majority of users of qooxdoo JSON-RPC were setting protocol 2.0  (after painfully discovering that the client was not, by default, 2.0 compliant).  All this is doing is changing the default so that people who expect JSON-RPC to do what the standard says (2.0) get what they expect.

cboulanger:
Since the tests still all pass, there shouldn't be any backwards-compatibility issues if you rely on the custom extensions to the protocol. The main thing that changes is that all message objects have a "jsonrpc" property with value "2.0" which is required by standard-compliant servers, and which is probabply simply ignored by legacy ones.

cboulanger:
We should, however, insert deprecation warnings for anything non-standard in qooxdoo v6.1 or so, so that people have time to adapt their code, and remove all non-2.0 features in v7.0. The  client could then be moved to a qxl package if anyone needs it to support legacy apps.

PR #9896, opened 2020-03-13
Fix qx.test.core.Property test

cboulanger:
class config maps must be separately defined to not produce compiler errors

### DIFF
--- a/framework/source/class/qx/test/Part.js
+++ b/framework/source/class/qx/test/Part.js
@@ -177,7 +177,7 @@ qx.Class.define("qx.test.Part",
       partLoader.preload("juhu");
 
       var part = partLoader.getParts()["juhu"];
-
+      var self = this;
       part.getPackages()[0].loadClosure = function() {
         self.resume(function() {
           self.fail("load called twice!");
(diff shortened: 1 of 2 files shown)


hkollmann (approved):

level420 (approved):

johnspackman (approved):

derrell (changes requested):
I don't see anything wrong with what was there before. If it was causing a compiler error, doesn't that indicate a problem with the compiler, rather than with this code? We should allow what this code was originally doing. That paradigm is used extensively in demobrowser and playground apps.

cboulanger:
I'd say the test code is to assert a *run-time* behavior of the qooxdoo class system whereas the error thrown by the is a *compile-time* test, which correctly diagnoses a wrong class declaration in the code. I cannot think of a way how the compiler should know that the wrong declaration is purposely wrong. So in my view, it is not a bug of the compiler and declaring the config separately is the right response to the issue.

derrell (approved):
Understood. I see now what's going on there. (Is this a new test? If not, why did it not fail previously?)
Approving as is, but I am curious.

cboulanger:
So far, we've run the framework tests with the generator only. If you run them with the compiler and the TestTapper test runner, you get different errors such as this one.

derrell:
Ah, makes perfect sense. Thanks.

oetiker (approved):

Issue #9825, opened 2019-12-17
Optimize Navagtion bar for Docsify


oetiker:
* investigate foldable navigation bar
* less vertical spacing

cboulanger:
This might be useful: https://www.npmjs.com/package/docsify-sidebar-collapse

cboulanger:
Done with #9903, #9901  and #9900.

PR #9900, opened 2020-03-18
Add docsify sidebar collapse plugin

cboulanger:
- Adds https://github.com/iPeng6/docsify-sidebar-collapse
 - Part of #9825


derrell (approved):
This looks good. Approving. However, I don't know how I can determine if it actually works. Is there someplace to see the output?

hkollmann (approved):

johnspackman (approved):

cboulanger:
@derrell I tested it locally and it did work. It would be good if we had a preview on Documentation PRs but this would require some additional setup...

cboulanger:
@derrell It is live now: http://www.qooxdoo.org/docs/#/core/oo_feature_summary 

The behavior is this: If you click on a heading that contains sub-chapters, its will expand. If you click on the next heading, the previous one collapses its child nodes.

PR #9901, opened 2020-03-18
Fix sidebar

cboulanger:
This fixes broken links and a few layout problems in the sidebar. For example, if a heading is annotated with a URL, it is no longer formatted as a heading. This ultimately needs to be fixed in the CSS, of course, but I have no time for that and removing the link and moving it to an "Overview" sub-page seems like an adequate fix.

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -4,31 +4,33 @@
   - [Qooxdoo Apps](/apps.md)
   - [About](/about.md)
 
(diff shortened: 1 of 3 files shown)


cboulanger:
I also moved the quickrefs to the end of the respective chapter since I consider them "advanced" information which newcomers will not be using.

level420 (approved):

hkollmann (approved):

oetiker (approved):

Issue #9553, opened 2018-08-01
Flow layout does not re-render when no childs initially


mbgonicus:
In [this playground example](http://tinyurl.com/ybelbx9h) (see also below) I create a container with a flow layout and let it render. After a second, it adds a label in the then empty container.

Expectation: After the timeout, I can see the label.
Is: After the timeout, the label is invisible .

Remarks:
* In the dom, the elements are present, but the container has a height of 0.
* If you uncomment the line such that the container has a widget initially, both labels are visible as expected.
* The behaviour is identical in master and devel playground.

```js
var container = new qx.ui.container.Composite(new qx.ui.layout.Flow());
var label1 = new qx.ui.basic.Label("Label 1");
var label2 = new qx.ui.basic.Label("Label 2");

// container.add(label1);

qx.event.Timer.once(function() {
  container.add(label2);
}, null, 1000);

this.getRoot().add(container);
```

cboulanger:
@mbgonicus Can you create a new example in the new http://www.qooxdoo.org/qxl.playground/  and see if the bug is still there?

mbgonicus:
I don't understand, why you need a new playground. Just copy and paste the code above...

And yes, the bug still persists.

http://tinyurl.com/tsm2vdp

johnspackman:
The issue here is that the root element has a layout of type `qx.ui.layout.Canvas`, which means that you need to give explicit positioning and dimensions when you add a widget; for example, this code works as expected:

```
this.getRoot().add(container, { top: 0, right: 0, bottom: 0, left: 0 });
```

So what's confusing is that `qx.ui.layout.Canvas` is actually zooming it out to full size - and that this only happens when there is something like a browser resize event, but it will also later reset it to zero width & height later...

PR #9907, opened 2020-03-20
Fixes a problem with LayoutItem where dimensions are not always stored

johnspackman:
In https://github.com/qooxdoo/qooxdoo/issues/9553, the bug reported was that the container does not appear with non-zero sizes until an arbitrary event (eg browser window resize).

The size *was* given to a layout item at one point, but that the layout item choses to queue itself for resize so that the "height for width" calculation can be done.  The problem is that when that deferred resize happens, the item's container asks the item for it's current size - but because it has been deferred, the original sizes were not stored.

This fix just makes sure that the sizes are stored, and then continues with the deferred resize as before.


framework/source/class/qx/ui/core/LayoutItem.js
```diff
@@ -516,6 +496,29 @@ qx.Class.define("qx.ui.core.LayoutItem",
...
+       * 
+       * Note that it is important that this happens after the above first pass at calculating a 
+       * computed size because otherwise getBounds() will return null, and this will cause an
+       * issue where the existing size is expected to have already been applied by the layout.
+       * See https://github.com/qooxdoo/qooxdoo/issues/9553  
+       */
+      if (this.getHeight() == null && this._hasHeightForWidth()) {
+        var flowHeight = this._getHeightForWidth(width);
+        
+        if (flowHeight != null && flowHeight !== this.__computedHeightForWidth)
```
derrell: `!=` with null... Are you checking for falsy, or do you mean to check explicitly for null?

uitests/issue9553/readme.md
```diff
@@ -0,0 +1,5 @@
+# issue9553
+
+undefined
+
+undefined
```
derrell: Is this file supposed to be committed as it is?

derrell (changes requested):
see prior comments

framework/source/class/qx/ui/core/LayoutItem.js
johnspackman: That code is just moved down - I agree that the comparisons are suspect, but it dates back to before 2011, so I've just kept it as it is

uitests/issue9553/readme.md
johnspackman: oops, thats a bug in the skeleton generation

oetiker:
do you need the pngs in the unittest ?

oetiker (approved):

hkollmann (approved):

level420 (approved):

cboulanger (approved):

derrell (approved):

Issue #9645, opened 2019-03-22
'click' event not being fired for menu button in Firefox 66


milandamen:
**Description**
Since Firefox 66, the 'click' event is no longer being fired for qx.ui.menu.Button widgets. It still works in older versions of Firefox and newest Chrome.

**Reproduction**

    var menuBar = new qx.ui.menubar.MenuBar();
    
    var menuItem = new qx.ui.menubar.Button('Menu item');
    var submenu = new qx.ui.menu.Menu();
    var submenuItem = new qx.ui.menu.Button('Submenu item 1');
    
    submenu.add(submenuItem);
    menuItem.setMenu(submenu);
    menuBar.add(menuItem);
    
    this.getRoot().add(menuBar);
    
    submenuItem.addListener('click', function (event) {
      alert('submenu item 1 clicked');
    });
    
    
    
    
    var label = new qx.ui.basic.Label('Some label');
    this.getRoot().add(label, {top: 100});
    
    var menu2 = new qx.ui.menu.Menu();
    var item2 = new qx.ui.menu.Button('Submenu item 2');
    
    menu2.add(item2);
    label.setContextMenu(menu2);
    
    item2.addListener('click', function (event) {
      alert('submenu item 2 clicked');
    });

1. Open a menu. It does not matter if the menu was opened using a widget contextmenu (rightclick) or using a MenuBar Button.
2. Click on a menu item in the menu.

**Actual result**
No 'click' event gets fired by the browser, so Qooxdoo does not dispatch it to the rest of the application.

**Expected result**
The 'click' event gets fired by the browser and Qooxdoo will dispatch it.

**Possible cause**
This is speculation, but I think Qooxdoo might be stopping the browser from firing the 'click' event, because the browser fires 'pointerdown' and 'pointerup' correctly.
I have tried reproducing this with pure HTML without Qooxdoo and it fired the 'click' event correctly.

**Workaround**
Listen to 'execute' or 'tap' instead of 'click' on the menu button.

cboulanger:
Which version of qooxdoo are you using? We are just in the process of backporting fixes from v6 to v5 and this might already be fixed in version 6. @level420, can you comment?

milandamen:
I am using the tip of master at the moment (7ce1e45ba062a04871032847b6a3727a6c24fe2a). It's still broken on Firefox 66.0.1.

oetiker:
this is still present in that latest firefox and latest playground

[...]

PR #9908, opened 2020-03-20
Fixed #9906: The hideAll function of menu manager had the wrong context.

### LINKED ISSUE #9906: The menu does not always close
**Describe the bug**
The menu is not closed if a modal window opens. It also stays in front of the blocker. Seems to be broken since commit 8a382b3da40800e59bec78101d32ef2ac7ba1454.

**To Reproduce**
Here is a playground example: https://tinyurl.com/sq3wcpx

**Screenshots**
image

Tested on windows with Chrome and FF.

Holdwyn:


derrell (changes requested):
Although this probably is fixing a bug, it will have long-ranging effects on backward compatibility. Many apps may break. I'm not sure what the right solution is...

Also, the change from 100ms to 0ms may be wrong. Those delays are typically to allow the rendering queue to flush, and simply enquing it to run again immediately may not do the job. Those can be race conditions that may not show themselves immediately, but can be very hard to track down.

Holdwyn:
> Although this probably is fixing a bug, it will have long-ranging effects on backward compatibility. Many apps may break. I'm not sure what the right solution is...
> 
> Also, the change from 100ms to 0ms may be wrong. Those delays are typically to allow the rendering queue to flush, and simply enquing it to run again immediately may not do the job. Those can be race conditions that may not show themselves immediately, but can be very hard to track down.

I guess there is no need for a real timeout, it is enough to wait for the javascript tread to end. I tested to not raise the bug #9645 again. The click event still works with Firefox. With 100ms there is a noticeable flickering of the menu over the (colored) blocker.

derrell (approved):
The part I was most concerned about was a possible backward-compatibility issue, but re-reviewing, I see that it's not calling a user's callback function; it's calling a very specific function in the framework, which does in fact expect a particular context. I think this is a very appropriate change.

I'm still a bit concerned about the change in timeout value, but I suspect that your argument is valid, and that only a wait for the current JavaScript loop iteration context to end is adequate.

johnspackman (approved):

hkollmann (approved):

cboulanger (approved):

level420 (approved):

cboulanger:
Merging, thank you @Holdwyn !

PR #9912, opened 2020-03-24
Update Contents Overview

cboulanger:
And small fixes to qx server


hkollmann (approved):

oetiker (approved):

docs/contents.md
```diff
@@ -1,18 +1,44 @@
...
+detailed table of contents.
+
+## [Qooxdoo Core](core/) 
+
+A dedicated class system with custom class properties, data binding and a 
+runtime environment.
+
+## [Qooxdoo UI](gui/)
+
+The UI layer of Qooxdoo, providing a rich set of widget which can be used to
```
derrell: s/set of widget/set of widgets/

docs/contents.md
```diff
@@ -1,18 +1,44 @@
...
+create professional web-based applications.
+
+## [Communication](communication/)
+
+Qooxdoo provides tree different communication APIs with a rich feature set:
```
derrell: s/tree/three/

docs/contents.md
```diff
@@ -1,18 +1,44 @@
...
+
+Qooxdoo provides tree different communication APIs with a rich feature set: 
+HTTP Request APIs, a REST Interface which support databinding, and a JSONRPC 
+Client.
+
+## [Qooxdoo on the Server](server/)
+
+If you like to use the same development paradigm for the client and the server,
+Qooxdoo has you covered - Use Qooxdoo's class system and tooling with NodeJS or 
+Rhino.
```
derrell: Do we still need to mention Rhino? Hasn't it been pretty obsolete, with no maintenance, for many years?

docs/server/README.md
```diff
@@ -3,12 +3,12 @@
 This page is an overview of Qoodoo's server capabilities. It
 shows which parts of Qooxdoo can be used in a server environment
 or comparable scenario. It also serves as an introduction to all
```
derrell: s/to all/to those/

docs/server/README.md
```diff
@@ -3,12 +3,12 @@
...
-interested in using %{qooxdoo} on a %{JS} server environment.
+interested in using Qooxdoo in a server environment.
 
 ## Included Features
 
 This listing shows the core features of the *Qooxdoo Server* package. If
-you build your own package with the skeleton way of using %{qooxdoo},
+you build your own package with the skeleton way of using Qoxdoo,
```
derrell: s/Qoxdoo/Qooxdoo/

docs/server/requirements.md
```diff
@@ -1,10 +1,10 @@
 # Server Requirements
 
-qooxdoo server is a basic component that runs in many different contexts and
-environments, as it has very little dependencies to the underlying runtime. For
-use in command-line tools and programs you will need a corresponding JavaScript
-interpreter like Node.js or Mozilla Rhino. For use in HTML5 Web Workers you will
-need a browser that supports this technology.
+qooxdoo server is a basic component that runs in many different contexts
```
derrell: `qooxdoo` used to be always lower case. These days, we're starting to capitalize it in some places, as a proper noun. In this case, it's also at the beginning of a sentence. Let's figure out what we're doing with it, and then be consistent.

docs/server/requirements.md
```diff
@@ -1,10 +1,10 @@
...
+and environments, as it has very little dependencies to the underlying
```
derrell: s/very little/very few/

In English, if things are countable, you use "few", e.g., "There I saw a few clouds."
If they are not easily countable, you use "little", e.g., "There is very little water in that glass." or even "There is only a little sand in the box." (but if reworded to be countable... "There are only a few grains of sand in the box.")

derrell (changes requested):

docs/contents.md
cboulanger: It is still in the documentation. But I can remove it since probably no-one uses it anymore.

cboulanger:
@derrell Thanks! Changed as requested.

derrell (approved):

level420 (approved):

johnspackman (approved):

PR #9923, opened 2020-03-29
Update development docs - add howto

cboulanger:
forgot this section..


hkollmann (approved):

derrell (approved):
This is fine. For your consideration, though... I've seen recommendations to avoid "here" as a link, and to instead describe what it's linking to. In this case, you'd do something like,

```
See [the HOWTO](howto/) for ...
```

cboulanger:
@derrell Thanks, that makes total sense - I changed it.

PR #9928, opened 2020-04-08
fixes deployment

johnspackman:


cboulanger:
I am confused `-f` seems to be the parameter name for "feedback" (https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Compile.js#L115-L119), I thought `-c`/`--configFile` was the parameter name, but I don't see it in there ...

hkollmann (approved):

johnspackman:
@cboulanger you're right, hang on

cboulanger:
Where is the parameter name for the config file specified? `-c` is for clean, so it's not that. The Yargs docs still contain the old form (https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Compile.js#L158)

johnspackman:
it's in https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/Cli.js#L91-L94

johnspackman:
i'll fix the conflict with `-c` in the compiler tomorrow, but this should fix the deployment

cboulanger (approved):

PR #9934, opened 2020-04-17
Do not touch Content-Type when submitting FormData

oetiker:
When using XMLHttpRequest with FormData the Content-Type gets set automatically to `multipart/form-data` with an appropriate boundary.  But this will only work when the Content-Type has not been set already. Prevent qooxdoo from breaking this :)

### DIFF
--- a/framework/source/class/qx/io/request/Xhr.js
+++ b/framework/source/class/qx/io/request/Xhr.js
@@ -241,7 +241,8 @@ qx.Class.define("qx.io.request.Xhr",
     // overridden
     _getConfiguredRequestHeaders: function() {
       var headers = {},
-          isAllowsBody = qx.util.Request.methodAllowsRequestBody(this.getMethod());
+          isAllowsBody = qx.util.Request.methodAllowsRequestBody(this.getMethod()),
+          isFormData = (qx.Bootstrap.getClass(this.getRequestData()) == "FormData");
 
       // Follow convention to include X-Requested-With header when same origin
       if (!qx.util.Request.isCrossDomain(this.getUrl())) {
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/io/request/Xhr.js
```diff
@@ -254,7 +255,7 @@ qx.Class.define("qx.io.request.Xhr",
       }
 
       // By default, set content-type urlencoded for requests with body
-      if (this.getRequestData() && isAllowsBody) {
+      if (this.getRequestData() && isAllowsBody && !isFormData) {
```
derrell, line 258: Is there necessarily a one-to-one relationship between the requestData class being FormData, and `headers["Content-Type"] being set already? Would it make more sense, here, to test whether `! ("Content-Type" in headers)` ?

derrell (changes requested):

framework/source/class/qx/io/request/Xhr.js
oetiker: For FormData to work as a payload, Content-Type must not be set. The way the code presently works, qx will always set the Content-Type, it can be overwritten by th user, but there is no way to NOT set it, and unfortunately this is the only way FormData Posts will work.

cboulanger (approved):
I didn't even know about FormData  (https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects) until now.

johnspackman (approved):

level420 (approved):

hkollmann (approved):

framework/source/class/qx/io/request/Xhr.js
derrell: Ii think you misunderstood me. I'm not saying that a change isn't required. I agree, one is needed. What I'm suggesting is that this fix might not be as general as we'd like. If someone manually added that header, i.e., if `headers["Content-Type"] already has a value at this point for any reason, isn't that a better test than just whether the content type is FormData?

derrell: Or maybe I misunderstand the issue. "Content-Type must not be set" at all if the content type if FormData? That sounds odd...

derrell: Disregard my comments. I understand what's happening now. All good.

derrell (approved):

PR #9945, opened 2020-04-30
framework Manifest.json should not require anything

level420:
Both version and required version should be the same.

### DIFF
--- a/Manifest.json
+++ b/Manifest.json
@@ -87,9 +87,5 @@
     "class": "source/class",
     "resource": "source/resource",
     "translation": "source/translation"
-  },
-  "requires": {
-    "@qooxdoo/compiler": "^1.0.0-beta",
-    "@qooxdoo/framework": "^6.0.0-beta"
   }
 }
--- a/Manifest.json
+++ b/Manifest.json
@@ -90,6 +90,6 @@
   },
   "requires": {
     "@qooxdoo/compiler": "^1.0.0-beta",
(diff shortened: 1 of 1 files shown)


cboulanger:
I think the require entry needs to be removed completely. It seems to me that a framework requiring itself doesn't make sense. The  dependency order is compiler -> framework . The framework itself shouldn't have a dependency on the compiler.

hkollmann (approved):

hkollmann (changes requested):
IMHO @cboulanger is right!

level420:
OK, then the complete `required` attribute should be removed?
https://github.com/qooxdoo/qooxdoo/blob/master/Manifest.json#L91-L94

cboulanger:
I think it should be, but I have just seen that we made `requires` mandatory in the schema. This needs to be changed first, otherwise removing it breaks schema validation: 
https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/resource/qx/tool/schema/Manifest-1-0-0.json

hkollmann (approved):

level420:
as now with https://github.com/qooxdoo/qooxdoo-compiler/issues/668 required isn’t mandatory anymore we could proceed with this PR

cboulanger (approved):

johnspackman (approved):

oetiker (approved):

derrell (approved):

PR #9948, opened 2020-05-02
Scale default delay in async test methods

cboulanger:
the Unit test runner [calls `wait()` with no arguments for async functions](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/dev/unit/TestFunction.js#L126-L138)  which means that [the default timeout of 10000ms](https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/dev/unit/AsyncWrapper.js#L82) is applied, without the possibility of changing this timeout programmatically. This PR allows to scale the default timeout using the `qx.test.delay.scale` environment setting in the same way a given `delay` value would be scaled.

### DIFF
--- a/framework/source/class/qx/dev/unit/AsyncWrapper.js
+++ b/framework/source/class/qx/dev/unit/AsyncWrapper.js
@@ -41,14 +41,21 @@ qx.Class.define("qx.dev.unit.AsyncWrapper",
    */
   construct : function(delay, deferredFunction, context)
   {
-    for (var i=0; i<2; i++) {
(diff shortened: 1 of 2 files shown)


oetiker (approved):

hkollmann (approved):

johnspackman (approved):

derrell (approved):

level420 (approved):

level420:
5 approvals. thank you @cboulanger

Issue #9909, opened 2020-03-24
Breaking changes / migration paths


cboulanger:
This issue is to collect any breaking changes of v6.0.0 and possible migration paths, so that we have a list ready for our release notes.

hkollmann:
Default of rpc has changed from "qx1" to "2.0"

schnepe2:
`qx.ui.splitpane.Splitter` :
- New Property `knobVisible` toggles visibility of the splitter's button.
- Property `visible` toggles visibility of the whole splitter widget (no change from v5.0.x).

schnepe2:
- `qx.bom.client.Css`
Removed obsolete IE-Feature check methods `getFilterGradient` and `getFilterTextShadow`. If used in production code, consider them as being a `false` value.
- `qx.core.Environment`
Removed obsolete IE-Feature keys `"css.gradient.filter"` and `"css.textShadow.filter"`. If used in production code, consider them as being a `false` value.

hkollmann:
framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js: change vertical alignment from top to middle.

cboulanger:
Breaking changes are documented here: https://github.com/qooxdoo/qooxdoo/blob/master/CHANGELOG.md

PR #9950, opened 2020-05-05
Remove Firfox warning...

schnepe2:
`Expected ‘none’, URL, or filter function but found ‘progid’.  Error in parsing value for ‘filter’.  Declaration dropped.` during startup.

### DIFF
--- a/framework/source/class/qx/bom/client/Css.js
+++ b/framework/source/class/qx/bom/client/Css.js
@@ -426,18 +426,20 @@ qx.Bootstrap.define("qx.bom.client.Css",
     __isFilterSupported : function(filterClass, initParams)
     {
       var supported = false;
-      var value = "progid:" + filterClass + "(" + initParams + ");";
-      var el = document.createElement("div");
-      document.body.appendChild(el);
-      el.style.filter = value;
-
-      if (el.filters && el.filters.length > 0 &&
-        el.filters.item(filterClass).enabled == true)
+      if (qx.core.Environment.get("browser.name") != "firefox")
       {
-        supported = true;
-      }
-      document.body.removeChild(el);
+        var value = "progid:" + filterClass + "(" + initParams + ");";
+        var el = document.createElement("div");
+        document.body.appendChild(el);
+        el.style.filter = value;
 
+        if (el.filters && el.filters.length > 0 &&
+          el.filters.item(filterClass).enabled == true)
+        {
+          supported = true;
+        }
+        document.body.removeChild(el);
+      }
       return supported;
     },
 
--- a/framework/source/class/qx/bom/client/Css.js
+++ b/framework/source/class/qx/bom/client/Css.js
@@ -66,7 +66,7 @@ qx.Bootstrap.define("qx.bom.client.Css",
     getPlaceholder : function() {
       if (qx.core.Environment.get("engine.name") === "mshtml") {
         return false;
-      } 
+      }
       var i = document.createElement("input");
       return "placeholder" in i;
     },
@@ -264,22 +264,6 @@ qx.Bootstrap.define("qx.bom.client.Css",
 
 
     /**
-     * Returns <code>true</code> if the browser supports setting gradients
-     * using the filter style. This usually only applies for IE browsers
-     * starting from IE5.5.
-     * http://msdn.microsoft.com/en-us/library/ms532997(v=vs.85).aspx
-     *
-     * @return {Boolean} <code>true</code> if supported.
-     * @internal
-     */
-    getFilterGradient : function() {
-      return qx.bom.client.Css.__isFilterSupported(
-        "DXImageTransform.Microsoft.Gradient",
-        "startColorStr=#550000FF, endColorStr=#55FFFF00");
-    },
-
-
-    /**
      * Returns the (possibly vendor-prefixed) name this client uses for
      * <code>radial-gradient</code>.
      *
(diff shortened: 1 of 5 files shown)


johnspackman (approved):

derrell (approved):

cboulanger (approved):

level420 (approved):

hkollmann (approved):

hkollmann (approved):

hkollmann (changes requested):
filter is supported since firefox 35.

http://www.w3schools.com/jsref/prop_style_filter.asp
https://developer.mozilla.org/de/docs/Web/CSS/filter

which firefox version are you using?

schnepe2:
Currently 75.0 and 76.0 (windows)
That warning is showing up in FFs for quite a while though (at least since 57.0)

schnepe2:
> filter is supported since firefox 35.

filter is supported, but not the construct that qooxdoo is using to test it!
As the warning says!
'progid: ....'

hkollmann (commented):
Then the test is wrong. With your change the answer of __isFilterSupported is false for FF. Which is not correct.
Can we change the test instead not using progid?

schnepe2:
> Then the test is wrong. 
No it isn't! The method `__isFilterSupported` is only for checking some IE5.5 specific features.
See `getFilterGradient` and `getFilterTextShadow`
The name might not be perfect for that method however.
And documentation is also weak.

> With your change the answer of __isFilterSupported is false for FF. Which is not correct.
That result **is** correct for FF (see above)

> Can we change the test instead not using progid?
The method is only for testing special IE cases, so it's O.K. this way.

framework/source/class/qx/bom/client/Css.js
```diff
@@ -426,18 +426,20 @@ qx.Bootstrap.define("qx.bom.client.Css",
...
+        document.body.appendChild(el);
+        el.style.filter = value;
 
+        if (el.filters && el.filters.length > 0 &&
+          el.filters.item(filterClass).enabled == true)
+        {
+          supported = true;
+        }
+        document.body.removeChild(el);
+      }
```
oetiker, line 442: I would suggest if we touch this, then to deprecate the functions and always return `false` 
see also https://developer.mozilla.org/en-US/docs/Web/CSS/-ms-filter

we are not talking about genuine filter here but a M$ hack not required since IE8!

doing the whole sing and dance just takes valuable time ... at execution time.

better still remove the entire function and add deprecation notices to the getFilterGradient and getFilterTextShadow  calls.

framework/source/class/qx/bom/client/Css.js
```diff
@@ -426,18 +426,20 @@ qx.Bootstrap.define("qx.bom.client.Css",
...
     {
       var supported = false;
-      var value = "progid:" + filterClass + "(" + initParams + ");";
-      var el = document.createElement("div");
-      document.body.appendChild(el);
-      el.style.filter = value;
-
-      if (el.filters && el.filters.length > 0 &&
-        el.filters.item(filterClass).enabled == true)
+      if (qx.core.Environment.get("browser.name") != "firefox")
```
oetiker, line 429:
```suggestion
```

framework/source/class/qx/bom/client/Css.js
oetiker, line 442:
```suggestion
```

oetiker (changes requested):

hkollmann:
I would agree to @oetiker. IE 5.5 is dead. Return false and add an deprecated comment.

hkollmann:
@oetiker :  Removing the check and return always false leeds to a breaking change in 
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/decoration/MLinearBackgroundGradient.js#L155.

css.gradient.filter is set through
 qx.core.Environment.add("css.gradient.filter", statics.getFilterGradient);  

So i suggest to accept  the PR

oetiker:
@hkollmann the change as it stands makes no sense ... why test for obscrure things present only in browsers we do not support and then go ahead and exclude firefox just because it complains ... 

I can't see why this would break anything ...

``` javascript
__isFilterSupported : function(filterClass, initParams) {
  return false;
},
```

hkollmann:
Then we can remove this too:
from https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/ui/decoration/MLinearBackgroundGradient.js#L155:

      var styleImpl = this.__styleLinearBackgroundGradientAccordingToSpec;
      if (qx.core.Environment.get("css.gradient.legacywebkit")) {
        styleImpl = this.__styleLinearBackgroundGradientForLegacyWebkit;
      } else if (qx.core.Environment.get("css.gradient.filter") &&
        !qx.core.Environment.get("css.gradient.linear") && qx.core.Environment.get("css.borderradius")) {
        styleImpl = this.__styleLinearBackgroundGradientWithCanvas;
      } else if (qx.core.Environment.get("css.gradient.filter") &&
        !qx.core.Environment.get("css.gradient.linear")) {
        styleImpl = this.__styleLinearBackgroundGradientWithMSFilter;
      }

css.gradient.filter depends on __isFilterSupported in statics.getFilterGradient)

This is the only place where the settings are used.

schnepe2:
@hkollmann : ...like this? If accepted I'll also add the changes to #9909

hkollmann (approved):
great!

oetiker (approved):
yessss :)

Issue #9719, opened 2019-06-27
Conflicting property method qx.ui.splitpane.Splitter.isVisible with qx.ui.core.Widget


schnepe2:
The "fix" for #9649  introduced a (runtime-)warning, as it is "overriding" a generated property metghod:
`008327 Conflicting property method qx.ui.splitpane.Splitter.isVisible with qx.ui.core.Widget`

**To Reproduce**
Just run an app (source build) which has a Splitter.

**Expected behavior**
No warnings / errors whatsoever.

hkollmann:
@johnspackman : Why do you closed this?

johnspackman:
Because it was solved by the PR https://github.com/qooxdoo/qooxdoo/pull/9649 which was just merged

johnspackman:
oh hang on, no it's not!!

hkollmann:
It's another issue...

schnepe2:
Should I change the title so it doesn't contain the responsible PR?
(this is what triggered @johnspackman, right?)

schnepe2:
BTW: Last time such an issue came up (#9313), renaming of the method was proposed!
`showKnob` might be an alternative name for it.

schnepe2:
Took a look at the things PR #9649 introduced and I am not sure what to think of it.

1. The documentation of `qx.ui.splitpane.Splitter#visible` property is plain wrong, as it only applies the visibility of the Splitters **knob**!
2. The changes in `qx.ui.splitpane.Pane` seem also to just work with **one** special case
(Try a theme with a different sized knob and you'll probably see things go crazy).
3. Shouldn't all the public properties, members and methods of Splitter (`getChildControl("knob"),` `setVisibility("visible")`, `setVisibility("excluded")`, `getVisibility()`, etc. pp.) be enough for any application programmer to derive a class from it? This request looks like a very special use-case to me.


For the "quick fix" a rename of property `qx.ui.splitpane.Splitter#visible` into `qx.ui.splitpane.Splitter#knobVisible` might be enough, but still I feel like the whole idea has to be tested a lot more...

hkollmann:
done with https://github.com/qooxdoo/qooxdoo/pull/9951

PR #9951, opened 2020-05-05
Fix for #9719...

schnepe2:
`Conflicting property method qx.ui.splitpane.Splitter.isVisible with qx.ui.core.Widget`.
Property `visible` is for the whole widget (_as like all other widgets_),
property `knobVisible` is for the knob.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,5 @@
 ## Breaking changes
 
 * [qx.bom.Cookie] Previous versions of qooxoo use `escape()` and `unescape()` functions. Since those functions are deprecated, then now qooxdoo use `encodeURIComponent()` and `decodeURIComponent()` functions. This may break some cookies. There are no issues with special characters like `~!@#$%^&*(){}[]=:/,;?+\'"\\` but some unicode characters like `äëíöü` (etc) are encoded different by `escape()` and `encodeURIComponent()`, so you must take care of this change if you use unicode characters.
+
+* [qx.ui.splitpane.Splitter] New Property `knobVisible` toggles visibility of the splitter's button. Property `visible` toggles visibility of the whole splitte widget (no change from v5.0.x).
--- a/framework/source/class/qx/ui/splitpane/Pane.js
+++ b/framework/source/class/qx/ui/splitpane/Pane.js
@@ -187,7 +187,7 @@ qx.Class.define("qx.ui.splitpane.Pane",
       // is removed.
       splitter.addListener("resize", function(e) {
         var bounds = e.getData();
-        if (this.getChildControl("splitter").getVisible() && (bounds.height == 0 || bounds.width == 0)) {
+        if (this.getChildControl("splitter").isKnobVisible() && (bounds.height == 0 || bounds.width == 0)) {
           this.__blocker.hide();
         } else {
           this.__blocker.show();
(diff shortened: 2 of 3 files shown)


derrell (changes requested):
Unfortunately, this will break existing apps. Any app that sets the `visible` property will be broken. Also, based on the comment removed from line 89 of Splitter.js, it looks like it's not just the knob that is suppsoed to be affected, but the entire slitter...

schnepe2:
The `visible` property will affect the whole splitter!.
The `knowVisible` will only affect the knob.
This would fit the (former) comment even more!

But how can it be that every application has to tolerate this inconsistency and **warning**, while the implementation breaks all the expected API?
Everyone expects `visible` to affect the whole widget (that's what the warning is telling everybody) and those applications that might have used it in "the wrong" sense can easily be changed.

We've discussed this @ #9719 but no one seems to be able to present a cleaner solution. So after almost a year I thought: it is best to trigger this again.
I handle warnings as errors and this one cannot  be silenced by any non-framework action so far!

oetiker:
yes, indeed the API is really not very satisfactory, I remember that we have accepted a PR for this thing already some time back so in case this made it worse, it would be fine to back it out in some way as it came in during development. Better to resolve the issue before we go final with 6.0

oetiker:
in the worst case, we could provide a 'consistent' splitter under a new class name and deprecate the inconsistent one

derrell (changes requested):
Although this change is not backward compatible, it does seem appropriate and now, as we're about to make a Major Release, is the right time to do this kind of thing. I approve this change, but mark it as Request changes so that we can discuss and ensure that we have the release documentation (upgrade instructions) written before merging.

oetiker (approved):

cboulanger:
I support @derrell 's suggestion. @schnepe2 , can you add an entry to https://github.com/qooxdoo/qooxdoo/blob/master/CHANGELOG.md which explains the breaking change introduced by this PR?

hkollmann:
And add a hint here: https://github.com/qooxdoo/qooxdoo/issues/9909?
Many thanks!

schnepe2:
I've added the infos (to #9909 and to the PR).

Although I find it kind of questionable as there is no **change** when comparing to last release (5.0.2), just an added property.

cboulanger (approved):

cboulanger:
@schnepe2 So maybe I misunderstood? I thought a breaking change was involved - @derrell, you said this was not backwards compatible? I have to admit I did not have a look myself.

derrell:
It looks to me like previously, one would call `setVisible` to set the visibility of the *knob*. With this PR implemented, existing user code that calls `setVisible` would set the entire *slider's* visibility, not just the *knob*. I believe that with this PR, users must change their code to use `setKnobVisible` if they want to affect only the knob visibility, and not the entire slider's visibility.

(Prior to this PR, my reading indicates that there was no way to change the slider visibility because the slider's `visible` property was overridden by the one that only affected the knob; thus it was a bug and should be fixed.)

Am I reading this incorrectly?

schnepe2:
Look at the latest release (5.0.3-beta) and you'll find no overridden `visible` property in `Splitter`.
So this whole mess was introduced in master, completely ignoring that warning.
Any application depending on that behaviour[1] is using the development (master) branch anyway.
That application can now choose! And that "ill-feature" was never a working one anyway, or how would one set the visibility of the splitter? It only allowed to change the visibility of the knob, leaving no option to choose.

Why is this such a long discussion anyway?


[1] expecting `foo.setVisible(false)` to only affect foo's Knob

level420 (approved):

derrell (approved):

derrell:
> So this whole mess was introduced in master, completely ignoring that warning.

If that had been mentioned previously, I hadn't seen it. I thought we had a backward compatibility issue with the 5.x version, and backward compatibility is a *big deal* for most releases, thus the long discussion. Anyway, we're all set now. Thank you very much for persisting in pressing forward with this!

hkollmann (approved):

PR #9952, opened 2020-05-05
Deploy docs to new website

johnspackman:


derrell (approved):

cboulanger (approved):

hkollmann (approved):
I would prefer a GitHub Action based new soloution.
This should be just a quick hack

cboulanger (changes requested):
Thinking about it, let's not use the travis workflow anymore and create a dedicated github action workflow instead . See #9953

.travis/deploy_website
```diff
@@ -15,7 +15,7 @@ git config user.name "Travis CI deploy"
 git config user.email "no-reply@qooxdoo.org"
 git config push.default simple
 
-GIT_IO_URL="https://$GH_TOKEN@github.com/qooxdoo/qooxdoo.github.io.git"
+GIT_IO_URL="https://$GH_TOKEN@github.com/qooxdoo/website.io.git"
```
oetiker, line 18: that seems odd ... 

```suggestion
GIT_IO_URL="https://$GH_TOKEN@github.com/qooxdoo/website.git"
```

oetiker (changes requested):

cboulanger:
I think this PR is moot because we use a github workflow now. Can we close @johnspackman ?

PR #9953, opened 2020-05-05
Create deploy-docs.yml

cboulanger:
I hope this works.


hkollmann (approved):

hkollmann:
What's about the apps? 
In the moment they are build specially for the documentation with the current qooxdoo version.

Would'nt it not be easier to link to the current app Version e.g. https://qooxdoo.org/qxl.demobrowser?

johnspackman (approved):

PR #9959, opened 2020-05-08
Improve message bus

cboulanger:
- improves documentation and adds contributors
- adds support for regular expressions
- adds `subscribeOnce()` method

### DIFF
--- a/framework/source/class/qx/event/message/Bus.js
+++ b/framework/source/class/qx/event/message/Bus.js
@@ -24,8 +24,7 @@
  * The message bus registers subscriptions to topics and notifies subscribers when
  * a matching message is dispatched. A topic acts as a filter to select only those
  * messages which match the filter. It can be the name of a message, which can
- * be truncated with a `*` wildcard, or a regular expression which or a
- * string ending icontain a `*` as a wild
+ * be truncated with a `*` wildcard, or a regular expression.
  */
 qx.Class.define("qx.event.message.Bus",
 {
--- a/framework/source/class/qx/event/message/Bus.js
+++ b/framework/source/class/qx/event/message/Bus.js
@@ -199,8 +199,8 @@ qx.Class.define("qx.event.message.Bus",
     },
 
     /**
-     * Subscribes to a topic just for one dispach and automatically unsubscribe
-     * afterwards.
+     * Subscribes to a topic just for one dispatch and automatically unsubscribes
+     * after executing the message handler.
      *
      * @param topic {String|RegExp} Topic to subscribe to. see {@link qx.event.message.Bus#subscribe}
      * for details
--- a/framework/source/class/qx/event/message/Bus.js
+++ b/framework/source/class/qx/event/message/Bus.js
@@ -38,6 +38,7 @@ qx.Class.define("qx.event.message.Bus",
 
     /**
      * Shorthand method for {@link qx.event.message.Bus.getSubscription}
+     * @return {Object}
      */
     getSubscriptions : function() {
       return this.getInstance().getSubscriptions();
@@ -45,15 +46,32 @@ qx.Class.define("qx.event.message.Bus",
 
     /**
      * Shorthand method for {@link qx.event.message.Bus.subscribe}
+     * @param subscriber {Function} Message handler function
+     * @param context {Object} The execution context of the callback (i.e. "this")
+     * @return {Boolean} Success
      */
     subscribe : function(topic, subscriber, context)
     {
       return this.getInstance().subscribe(topic, subscriber, context);
+    },
 
+    /**
+     * Shorthand method for {@link qx.event.message.Bus.subscribeOnce}
+     * @param subscriber {Function} Message handler function
+     * @param context {Object} The execution context of the callback (i.e. "this")
+     * @return {Boolean} Success
+     */
+    subscribeOnce : function(topic, subscriber, context)
+    {
+      return this.getInstance().subscribeOnce(topic, subscriber, context);
     },
 
     /**
      * Shorthand method for {@link qx.event.message.Bus.checkSubscription}
+     * @param topic {String|RegExp} The topic that has been used when subscribing
+     * @param subscriber {Function} Message handler function
+     * @param context {Object} The execution context of the callback (i.e. "this")
+     * @return {Boolean} Whether a subscription was removed
      */
     checkSubscription : function(message, subscriber, context)
     {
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/event/message/Bus.js
```diff
@@ -191,30 +198,52 @@ qx.Class.define("qx.event.message.Bus",
...
+     * @return {Boolean} Success
+     */
+    subscribeOnce : function(topic, subscriber, context)
+    {
+      var that = this;
+      var modified_subscriber = function(msg) {
+        subscriber.call(context, msg);
+        that.unsubscribe(topic, modified_subscriber, context);
+      }
+      this.subscribe(topic, modified_subscriber, context);
```
derrell, line 218: This makes it impossible to unsubscribe because the user has no access to `modified_subscriber`. That may be enough of a downside to negate the usefulness of this new feature. Probably `subscribeOnce` should be left as an exercise for the user who needs to unsubscribe after a single use, by simply adding the `unsubscribe` call to their handler.

framework/source/class/qx/event/message/Bus.js
```diff
@@ -269,42 +306,45 @@ qx.Class.define("qx.event.message.Bus",
...
-        var pos = key.indexOf("*");
-
-        if (pos > -1)
+        var l = topic.length;
+        if (topic[l-1] === "*")
         {
-          // use of wildcard
-          if (pos === 0 || key.substr(0, pos) === msgName.substr(0, pos))
+          // use of wildcard, only allowed as "*" or at the end of the topic
+          if (l === 1 || topic.substr(0, l-2) === msgName.substr(0, l-2))
```
derrell, line 327: `l` looks very much like `1`. Suggest renaming the variable to `len`.

derrell (changes requested):

framework/source/class/qx/event/message/Bus.js
```diff
@@ -12,14 +12,19 @@
...
 ************************************************************************ */
 
 /**
  * A simple message bus singleton.
- * The message bus registers subscriptions and notifies subscribers when
- * a matching message is dispatched
+ * The message bus registers subscriptions to topics and notifies subscribers when
+ * a matching message is dispatched. A topic acts as a filter to select only those
+ * messages which match the filter. It can be the name of a message, which can
+ * be truncated with a `*` wildcard, or a regular expression.
```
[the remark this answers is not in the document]
cboulanger: Ok, thanks.

Yes, I added RegEx support as a more powerful companion to the "*" wildcard. It's just for convenience, though, one could also listen for "*" and then check the message name.

framework/source/class/qx/event/message/Bus.js
cboulanger: What would be a situation in which a self-unsubscribing subscriber would have to unsubscribed? The very point of this method is that you don't have to unsubscribe it - so the user doesn't need to have access to the outer function. I can certainly add a note to the documentation that it is not possible to unsubscribe, but as I said I wouldn't know why anyone would need to do that?

On the other hand, I added the method because writing self-unsubscribing message handlers is annoying ...

oetiker (approved):

framework/source/class/qx/event/message/Bus.js
derrell: An example would be that you set up something on the assumption that your app will initialize properly, but something times out and you now don't want that subscription to fire. I guess that sounds made up, but all of the times that you might unsubscribe from any other type of subscription would apply here as well.

Your solution of just documenting the fact that you can't unsubscribe seems very adequate.

derrell (approved):

hkollmann (approved):

framework/source/class/qx/event/message/Bus.js
cboulanger: Done.

framework/source/class/qx/event/message/Bus.js
cboulanger: Thanks for alerting me to this.

framework/source/class/qx/event/message/Bus.js
```diff
@@ -120,59 +116,69 @@ qx.Class.define("qx.event.message.Bus",
...
     },
 
 
     /**
-     * subscribes to a message
+     * Subscribes to a topic
      *
-     * @param message {String} name of message, can be truncated by *
-     * @param subscriber {Function} subscribing callback function
+     * @param topic {String|RegExp} Either a string, which can be truncated
```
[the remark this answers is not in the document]
cboulanger: Done

framework/source/class/qx/event/message/Bus.js
```diff
@@ -31,83 +36,74 @@ qx.Class.define("qx.event.message.Bus",
...
     },
 
-
     /**
-     * subscribes to a message
-     *
-     * @param message {String} name of message, can be truncated by *
-     * @param subscriber {Function} subscribing callback function
+     * Shorthand method for {@link qx.event.message.Bus.subscribe}
+     * @param subscriber {Function} Message handler function
```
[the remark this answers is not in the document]
cboulanger: Done.

cboulanger:
Just as a side note: I use this with JSON-RPC and a PHP backend which has to run a time-intensive task which might exceed the time limit that some servers might impose on PHP. The remote procedure runs the task and stops before the timeout and sends a "continue" message to the bus via a JSON-RPC notification, which then triggers a new call to the remote procedure, which then continues its task. When done, the notification dispatches a "done" message, which is handled by a "subscribeOnce" handler to continue frontend-stuff.

johnspackman (approved):

framework/source/class/qx/event/message/Bus.js
```diff
@@ -31,83 +36,76 @@ qx.Class.define("qx.event.message.Bus",
...
-     * checks if subscription is already present
-     * if you supply the callback function, match only the exact message monitor
-     * otherwise match all monitors that have the given message
-     *
-     * @param message {String} Name of message, can be truncated by *
-     * @param subscriber {Function} Callback Function
-     * @param context {Object} execution context
-     * @return {Boolean} Whether monitor is present or not
+     * Shorthand method for {@link qx.event.message.Bus.checkSubscription}
+     * @param topic {String|RegExp} The topic that has been used when subscribing
```
level420, line 72: The parameter is documented with name `topic` while it's still called `message` in the function parameter list below.

level420 (changes requested):

framework/source/class/qx/event/message/Bus.js
cboulanger: Updated the method signature

level420 (approved):

level420:
merging. thank you @cboulanger

PR #9963, opened 2020-05-09
remove twitter app because it is not working in the moment

hkollmann:


cboulanger (approved):

oetiker:
hmmm this is the only "complete" qooxdoo example app we have ... I guess we could replace the 'server' component with one which fakes twitter ...

hkollmann:
I removed it yet - saved it. Let's release v6. I hope i found some time to move it to the new toolchain and update the docs then.
Up to then better remove it.

hkollmann:
And we have a tutorial: http://www.qooxdoo.org/qxl.tutorial/

hkollmann:
twitter app can be used as tutorial with less work. 
The docs must be cleaned up - the idenica service is unusable. It's replaced by a local file service.js which delivers some hardcoded text.
This should be enough for a desktop tutorial.

PR #9967, opened 2020-05-14
Virtual DOM and JSX support

johnspackman:
This PR is about improving the Virtual DOM in `qx.html.*` and adding JSX support, primarily for use in server applications but also the browser.

The changes are:

- the `qx.html.Element` is split into `qx.html.Node` and `qx.html.Element`, and also the addition of `qx.html.Text`.  This change is entirely backwards compatible, and the addition of Text allows any arbitrary DOM structure to be represented (code in `qx.ui.*` does not have to use Text nodes because the UI classes always operate directly on the node to set/get text).
- Add JSX Factory implementation
- VDOM can be serialized into HTML text stream for delivery to the client browser
- An entire tree of the DOM on the browser (ie that which was built because the browser parsed the HTML text) can be reattached to a matching or sparse tree of `qx.html.Element` instances, based on the `qxObjectId`
- There are some additional properties and methods (eg `cssClass`, `addCssClass()`, `visible`, etc) which exist to simplify common tasks
- Document the new `qx.headless` property in the compiler which indicates that the code is compiled for the server
- There are a few warning fixes and other minor tweaks which have crept in

### DIFF
--- a/docs/desktop/gui/html.md
+++ b/docs/desktop/gui/html.md
@@ -87,7 +87,6 @@ qx.Class.define("mypackage.MySignupForm", {
   
   construct() {
     this.base(arguments, "div");
-    this.setCaption("Login");
     let body = this.getBody();
     body.add(<h1>Welcome to My Corporate Website</h1>);
     body.add(
--- a/docs/desktop/gui/html.md
+++ b/docs/desktop/gui/html.md
@@ -6,11 +6,11 @@ This document describes the ideas and concepts behind the classes in the `qx.htm
 Idea
 ----
 
-The classes in `qx.html` are wrappers around native DOM elements, which were created primarily to solve one major issue:
+The classes in `qx.html` form a Virtual DOM, which on the the server allow DOM-like manipulation and then serialization
+into HTML text, and on the browser the Virtual DOM is wrappers around native DOM elements which are created only when
+absolutely necessary.
 
-**Automatically keeping care of DOM manipulation and creation while dealing with large number of elements.**
-
-In details this means:
+This solves a number of problems:
 
 > -   **Automatic performance**: Programmatically constructing DOM hierarchies is hard to get fast because the order in which elements are nested can heavily influence the runtime performance. What `qx.html.Element` tries to do is keep the number of element instances to the minimum actually needed (DOM nodes are expensive, in terms of both performance and memory) and to insert the DOM nodes in an efficient manner. Further, all changes to the DOM are cached and applied in batch mode, which improves the performance even more.
 > -   **Normalized API**: Working with HTML DOM elements usually involves many browser differences, especially when it comes to reading and setting of attributes or styles. For each style, one has to remember whether a normalization method should be called or if the value can be set directly. `qx.html.Element` does this kind of normalization transparently. The browser normalization is based on the [existing low-level APIs](../../core/tech_website_apis.md).
@@ -21,8 +21,10 @@ Typical Use Cases
 
 > -   Building a widget system on top
 > -   Massively building DOM elements from data structures
+> -   Sharing DOM-specific code between client and server
 
-It may be used for smaller things as well, but incurs a large overhead. The size of the API, additional to a basic low-level package of qooxdoo is about 20 KB (5 KB gzipped). Also it consumes a bit more memory when all underlying DOM elements are created. Keep in mind that the instances are around all the time. Even when all jobs for a instance are done at the moment.
+It may be used for smaller things as well, but incurs a large overhead.  Also it consumes a bit more memory when all 
+underlying DOM elements are created. 
 
 Features
 --------
@@ -31,7 +33,8 @@ Features
 > -   Full cross-browser support through usage of low-level APIs e.g. `setStyle()`, `getAttribute()`, ...
 > -   Advanced children handling with a lot of convenience methods e.g. `addAfter()`, ...
 > -   Reuse existing markup as a base of any element via `useMarkup()`
-> -   Reuse an existing DOM node via `useElement()`
+> -   Reuse an existing DOM node via `useNode()`
+> -   Serialize into an HTML string
 > -   Powerful visibility handling to `include()` or `exclude()` specific sub trees
 > -   Support for scrolling and scroll into view (`scrollTo()`, `scrollIntoView()`, ...)
 > -   Integration of text selection APIs (`setSelection()`, `getSelection()`, ...)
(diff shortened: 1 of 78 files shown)


cboulanger (approved):
Looks very interesting and useful!

docs/desktop/gui/html.md
```diff
@@ -65,9 +68,107 @@ This element is used to create iframes to embed content from other sources to th
...
 
+
+Code Sharing
+------------
+When you create instances of `qx.html.Element` on the server, at some point you can call `serialize()` to convert it into
+an HTML string which the server can pass to the client; this is useful because it provides a structured mechanism for
+defining and manipulating the DOM before it reaches the browser.
+
+However, rather than simply creating instances of `qx.html.Element`, you can derive classes from it and this has a  
+particular advantage in that you code can be used on the client and the server, and will be able to connect to the
```
oetiker, line 79:
```suggestion
particular advantage in that your code can be used on the client and the server and will be able to connect to the
```

docs/desktop/gui/html.md
```diff
@@ -65,9 +68,107 @@ This element is used to create iframes to embed content from other sources to th
...
+        <p><a href="#">Click here if you have <b>forgotten your password</b></a></p>
+      );
+    this.setStyle("max-width", "600px");
+  },
+  
+  members: {
+    _createQxObjectImpl(id) {
+      switch(id) {
+      case "form":
+        let form = <form method="post" action="#"></form>;
```
oetiker, line 107: Does this require compiler support (jsx) ?

framework/source/class/qx/core/BaseInit.js
```diff
@@ -40,7 +40,7 @@ qx.Class.define("qx.core.BaseInit",
      * @return {qx.core.Object} The application instance.
      */
     getApplication : function() {
-      return this.__application || null;
+      return qx.core.BaseInit.__application || null;
```
oetiker, line 43: does `__application` survive the name mangling ... for the build version ... or does this still not happen in the compiler and only in the old infrastructure?

framework/source/class/qx/event/handler/Focus.js
```diff
@@ -1063,8 +1063,9 @@ qx.Class.define("qx.event.handler.Focus",
           return focusedElement;
         }
         // Check compound widgets
-        var widget = qx.ui.core.Widget.getWidgetByElement(focusedElement),
-          textField = widget && widget.getChildControl && widget.getChildControl("textfield", true);
+        var Widget = qx.Class.getByName("qx.ui.core.Widget");
```
oetiker, line 1066: I think Widget is a somewhat 'reserved' word ... https://www.w3.org/TR/widgets-apis/#widget ... also using that same word in upper and lowercase spelling does not feel right ...

framework/source/class/qx/event/handler/Mouse.js
```diff
@@ -177,16 +183,34 @@ qx.Class.define("qx.event.handler.Mouse",
...
+      
+      if (qx.core.Environment.get("qx.debug")) {
+        if (type == "mousewheel") {
+          var end = (window.performance && performance.now) ? (performance.now() + qx.bom.AnimationFrame.__start) : Date.now();
+          var diff = Math.round(end - start);
+          if (diff > 500) {
+            this.error("'mousewheel' event took " + diff + "ms - this may have a significant effect on UI experience");
+          } else if (diff > 100) {
+            this.warn("'mousewheel' event took " + diff + "ms - consider refactoring for smoother UI experience");
+          }
```
oetiker, line 211: cool :)

framework/source/class/qx/core/BaseInit.js
johnspackman: yes, although the compiler does not do private name mangling yet

framework/source/class/qx/event/handler/Focus.js
johnspackman: The `Widget` (capital W) is just a way to get the `qx.ui.core.Widget` class without triggering a compiler dependency, so as a class name it's correct

framework/source/class/qx/event/handler/Mouse.js
johnspackman: that code got pulled in through rebase :(

docs/desktop/gui/html.md
johnspackman: yes - i've updated the doc

framework/source/class/qx/bom/element/Attribute.js
```diff
@@ -333,6 +333,42 @@ qx.Bootstrap.define("qx.bom.element.Attribute",
     },
 
 
+    /**
+     * Serializes an HTML attribute into a writert
```
derrell, line 337: serializes into a *what*? I think that's supposed to say "writer" but I don't understand even that...

derrell (changes requested):
I've tagged this for a weekend review, as the changes are extensive. This is a very, very scary PR, though. It looks like it's mucking with low-level stuff that can affect all desktop apps. I'll spend time on it this weekend, but I suspect this would be best left until after the release of 6.0 because of the possibility of subtle breakage. That way, we'll have lots of time to test with it and not take the chance on messing up our first release in years.

framework/source/class/qx/bom/element/Attribute.js
johnspackman: typo, i'll fix the comment - the `writer` it refers to is the parameter that you call to pass strings to write the attribute definition

johnspackman:
> @derrell I've tagged this for a weekend review, as the changes are extensive. This is a very, very scary PR, though. It looks like it's mucking with low-level stuff that can affect all desktop apps. I'll spend time on it this weekend, but I suspect this would be best left until after the release of 6.0

That's a fair point, and I was a bit torn myself.  It has only had limited testing.

I'm happy to create a v7.x branch and make the PR in there?

framework/source/class/qx/core/BaseInit.js
oetiker: I think it would really help to get this name mangling in before we release as otherwise, we will never be able to put it in as it will break too much code of people who have ignored our conventions ...

framework/source/class/qx/event/handler/Focus.js
oetiker: your intention is clear to me but at least in github the word `Widget` with capital W gets syntax highlighting as if it had special meaning ... that might be confusing to people looking at the code ... maybe call it `WidgetClass` instead

derrell: I agree with @johnspackman. This is a qooxdoo convention, that a shorthand for a class is the final component of the class name with a leading capital letter as in that final component. The fact that it gets highlighted in some editor isn't, IMO, sufficient reason to break our standard convention.

derrell (changes requested):
Wow. There's a lot here! I spent about an hour on this, but could spend many more for a thorough review. Since this is supposed to be BC, I suggest we consider this for a 6.1 or 7.0 version, where it will get far more BC testing in existing apps before being officially released.

framework/source/class/qx/core/BaseInit.js
johnspackman: mangling is in now

docs/desktop/gui/html.md
```diff
@@ -65,9 +68,109 @@ This element is used to create iframes to embed content from other sources to th
...
+
+For example, consider this class:
+
+```
+qx.Class.define("mypackage.MySignupForm", {
+  extend: qx.html.Element,
+  
+  construct() {
+    this.base(arguments, "div");
+    this.setCaption("Login");
```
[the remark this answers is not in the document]
johnspackman: ah yes typo, fixed

framework/source/class/qx/bom/client/Html.js
```diff
@@ -353,7 +353,7 @@ qx.Bootstrap.define("qx.bom.client.Html",
      * @ignore(Image)
      */
     getDataUrl : function(callback) {
-      var data = new Image();
+      var data = new window.Image();
```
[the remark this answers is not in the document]
johnspackman: it suppresses the warnings because `window` is always an allowed/well known global but `Image` isn't when compiling for a NodeJS target.  Part of my goal was to improve NodeJS development, by adding VDOM and reducing the unnecessary dependencies (because this can cause code to be loaded which crashes) - connected with this was removing warnings which are issued on nodejs-only targets.

In any code which has to be cross platform, rather than adding a blanket `@ignore(window.*)` instead I tend to use `window.Image` because it's direct about what the code is referencing

johnspackman:
I'm marking this as do not move for now, until we confirm what we're going to do for 6.1 release PRs

PR #9969, opened 2020-05-14
Draft fix for NPM deployment - NOT FOR MERGING

cboulanger:
What this PR does

- renames compile config files
- compile.json just builds API Viewer
- handles recursive dependency regarding qx_server package

The crrent state of the PR is unfunctional because it needs to run locally without deploying anything to NPM. This is just for discussion! The PR mainly responds to today's broken NPM deploy pipeline that was caused by a bug in the qx_server package, which is a quasi-circular dependency between framework and compiler via the framework NPM package. This PR tries to prevent these kind of situations by compiling the qx_server package with both the NPM version AND itself with the hope that any break can be notices before the package is deployed to NPM. 

If you have a better idea, let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,8 @@ tmp/
 /deploy
 /build
 /qx_packages
+/qooxdoo-compiler
+/source
 
 # CSS compiler
 /framework/source/resource/qx/mobile/scss/.sass-cache/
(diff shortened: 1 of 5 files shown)


hkollmann:
IMHO we should do the release and build the deployment new after this. Based on GitHub Actions.

cboulanger:
Closing this, can be revisited later.

PR #9971, opened 2020-05-15
Add warning about result of this.tr("")

zaucker:
`this.("")` returns the header of the .po file (see specs in https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files).

This can be rather confusing to a developer who is not aware of this specification if she uses something like `this["tr"](textVar)` with textVar being the empty string.

### DIFF
--- a/docs/development/howto/internationalization.md
+++ b/docs/development/howto/internationalization.md
@@ -219,7 +219,7 @@ of the application may be specified in `compile.json`, in the
 
 This would add a German and a French translation to
 the project. For a more exhaustive list of available
-locales see [here](http://unicode.org/cldr/apps/survey).
+locales see [here](http:/cldr.unicode.org/index/survey-tool).
 
 A run of `npx qx compile --update-po-files` or its shorthand `npx qx
 compile -u` will generate a `.po` file for each configured locale,
--- a/docs/development/howto/internationalization.md
+++ b/docs/development/howto/internationalization.md
@@ -219,7 +219,7 @@ of the application may be specified in `compile.json`, in the
 
 This would add a German and a French translation to
 the project. For a more exhaustive list of available
-locales see [here](http:/cldr.unicode.org/index/survey-tool).
+locales see [here](http://cldr.unicode.org/index/survey-tool).
 
 A run of `npx qx compile --update-po-files` or its shorthand `npx qx
 compile -u` will generate a `.po` file for each configured locale,
--- a/docs/development/howto/internationalization.md
+++ b/docs/development/howto/internationalization.md
@@ -119,6 +119,10 @@ that setting if you plan to support locale switching during run time.
 If the string given to `tr` does not have a translation
 under the current locale, the string itself will be returned.
 
+If the string given to `tr` is the empty string, the header of the
+.po file is returned (which can be a bit confusing if done
+accidentally, but is correct according to the [PO specs](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files PO specs).
+
 There is one exception to the simple rule that all strings can just be
 replaced by wrapping them in an appropriate `this.tr()` function call:
 If init values of dynamic properties (core/understanding_properties)
--- a/framework/source/class/qx/locale/Manager.js
+++ b/framework/source/class/qx/locale/Manager.js
@@ -25,6 +25,9 @@
  * @require(qx.event.dispatch.Direct)
  * @require(qx.locale.LocalizedString)
  *
+ * Note: "translating" the empty string, e.g. tr("") will return the header
+ * of the respective .po file. See also https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#PO-Files
+ *
  * @cldr()
  */
 


zaucker:
One could argue, that although the behaviour if `tr()` is correct according to the PO specs, the behaviour is still unexpected to the average programmer and thus `tr()` (and friends) should just return the argument unchanged if it is `""` or `undefined` or `null`.
If there was a consensus on this, I'd make a PR for that.
It would be a breaking change, on the other hand one wonders, if really somebody is using `this.tr("")` to get the .po header in any real-world application.

cboulanger (changes requested):
Thank you. Can you add something to the documentation as well? The relevant page would be https://github.com/qooxdoo/qooxdoo/blob/master/docs/development/howto/internationalization.md

oetiker:
How about adding an additional method like this:

```
/**
         * Translate incoming data. Do NOT mark the string for translation
         * use the tr function for this. xtr is meant for backend strings.
         */
        xtr: function(messageId,varargs) {
            var nlsManager = qx.locale.Manager;
            if (messageId == ''){
                return '';
            }
            if (messageId instanceof Array) {
                return nlsManager.tr.apply(nlsManager, messageId);
            }
            if (messageId instanceof qx.data.Array) {
                return nlsManager.tr.apply(nlsManager, messageId.toArray());
            }
            return nlsManager.tr.apply(nlsManager, arguments);
        }
```

this would allow to pass strings from the backend without using 'tricks' to bypass the compiler tranlsation detection ... 

as an added benefit the backend can also send an array of strings as the messageId and thus get %x replacement ...

zaucker:
> Thank you. Can you add something to the documentation as well? The relevant page would be https://github.com/qooxdoo/qooxdoo/blob/master/docs/development/howto/internationalization.md

Sure. Let's see what other suggestions come along, I'll redo the PR to whatever the consensus is.

cboulanger:
Ok, if we're back to open discussion, on further reflection, my position would be to not follow the spec here, which is counter-intuitive, but return an empty string by default. No-one expects the qooxdoo API to follow the PO specs, and it is extremely unlikely that any application relies on this behavior. Even if that the case, v6 can contain breaking changes, and now is the time to do so, with adequate documentation. So my vote would be to simply add a check for a falsy value and  return the empty string.

cboulanger:
I support @oetiker's suggestion, but that should probably be dealt with in a different issue?

derrell:
I don't quite understand how @oetiker's suggestion would be used in practice, but I have a regular need for passing errors generated at the server back to the client where they are translated to the client browser language. It seems that's what this would allow? If so, I'm all for it! (I just need to understand how to use it.)

oetiker:
@derrell if you have an error from the backend and use `this.tr(error)` then the compiler is not happy (or at least the old one was not) as it tried to extract `error` as a string for translation ... so we had the habit of writing `this['tr'](error)` in our code to not get the compiler all riled up ...

so I propose to have an alternative method to `tr` where the compiler does NOT try to extract a string from the first argument ... 

furthermore my implementation supports backend 'error' strings where there is not a string but actually a string (messageId) as well as replacement aregument for %1 %2 are passed in along.

```javascript
var simple = 'server error';
var complex = ['complex %1 error %2', 'stuff','44'];
console.log(this.xtr(simple));
console.log(this.xtr(complex));
```

zaucker:
Apart from @oetiker's  suggestion for `xtr()` which sould cover @derrell's use case.
What is everybody's opinion for the default `tr()`behaviour. I'd favour @cboulanger's suggestion of adding a check for falsy and return an empty string in this case from a frontend programmer's point of view.
However, I wonder if this would actually break the compiler or if the compiler would then add an entry into the .po files with an empty string id (and thus overwriting the header).

johnspackman:
IMHO we should leave it as it is - we're talking about installing a workaround for a bug in the user's code?  And then adding a new function as a workaround because someone might need the data we're suppressing?

derrell:
I'm confused. Why would you call `this.tr()` with some non-constant string as an argument? That's not legal use of `this.tr()` and even doing it as `this["tr"]()` doesn't solve the problem. The string passed isn't going to be in any po file, and therefore won't ever be translated, and thus there's no need for `this.tr()`...

I would love to have a way to have errors generated at the server be translated by the client, but I think I'm inclined to agree with @johnspackman and leave this as is for now. It sounds like the use case is incorrect usage, so this should probably be thought through more.

zaucker:
> I'm confused. Why would you call `this.tr()` with some non-constant string as an argument? That's not legal use of `this.tr()` and even doing it as `this["tr"]()` doesn't solve the problem. The string passed isn't going to be in any po file, and therefore won't ever be translated, and thus there's no need for `this.tr()`...
> 
> I would love to have a way to have errors generated at the server be translated by the client, but I think I'm inclined to agree with @johnspackman and leave this as is for now. It sounds like the use case is incorrect usage, so this should probably be thought through more.

Well, if you supply translations somehow for the value in a variable passed to `tr()` it does the right thing. For example, if you'd define a select box with the item values wrapped in `this.tr()`, the .po files will get translations. When you then programmatically select one of those select box items, it will be displayed in the correct language. Or you can provide the translations as @oetiker suggested by loading them from the backend at program start (e.g. for your error messages).

But I admit, this is probably not how this was intended to be used. So I suggest that this PR be merged as is (just adding the hint about empty string returning the header), as this was very confusing to us. 
Or, if you don't like that either, we can also drop the PR.

@oetiker's `xtr()` method we can add in a separate PR which would also solve your error messages problem.

cboulanger:
I don't think its is  unusual to translate dynamic content. I can think of the following use cases, part of which I use in my app:
- using  `marktr()` to create a bunch of different messages that will be displayed based on some condition that is unknown at compile time, and to avoid code duplication
-using server-generated "po" files in the qooxdoo app to display  server-sent messages at runtime in the browser-language of choice without having to concern the server with locale issues. 
Maybe I am doing it all wrong, but these cases don't seem totally unreasonable?

cboulanger:
Yes, let's merge it with an additional entry in the docs, since the current behavior is unintuitive, even if it is correct from the PO specs point of view. I would still argue that from the users point of view, the correct behavior should be to translate an empty string with an empty string, but we can revisit the issue later should it come up again.

derrell (approved):
Approving these documentation changes as is. The remaining discussion topics regarding server-generated errors, etc.,  can be continued, but should probably be in a new issue (and postponed to post-6.0 release).

zaucker:
I added a sentence to the internationalization page in the docu as suggest by @cboulanger.
I'll talk to @oetiker about making a PR for his `xtr()` suggestion.

hkollmann (approved):

cboulanger (approved):

hkollmann:
Could you please check this   222:13-222:56  warning  Link to http://unicode.org/cldr/apps/survey is dead  no-dead-urls  remark-lint?

zaucker:
> Could you please check this   222:13-222:56  warning  Link to http://unicode.org/cldr/apps/survey is dead  no-dead-urls  remark-lint?

Is this the right issue for this question?

hkollmann:
Not really - but the linting goes wrong - so this must be fixed,

oetiker (approved):

hkollmann:
new url seems to be http://cldr.unicode.org/index/survey-tool

docs/development/howto/internationalization.md
```diff
@@ -219,7 +219,7 @@ of the application may be specified in `compile.json`, in the
 
 This would add a German and a French translation to
 the project. For a more exhaustive list of available
-locales see [here](http://unicode.org/cldr/apps/survey).
+locales see [here](http:/cldr.unicode.org/index/survey-tool).
```
cboulanger, line 222: there is a slash missing in `http://`

zaucker: Fixed

johnspackman (approved):

PR #9973, opened 2020-05-15
fix mangled private method names

johnspackman:
change the display names of functions so that if one is already provided, we use it rather than the key name - this is so that private mangling works, but does not show the mangled name.

(mangled private symbols are coming to the compiler)

### DIFF
--- a/framework/source/class/qx/Bootstrap.js
+++ b/framework/source/class/qx/Bootstrap.js
@@ -74,7 +74,7 @@ qx.Bootstrap = {
     {
       var value = functionMap[name];
       if (value instanceof Function) {
-        value.displayName = classname + "." + name + "()";
+        value.displayName = classname + "." + (value.name || name) + "()";
       }
     }
   },


derrell (approved):

derrell (changes requested):
Wait... Is this preserving the original name even though names are mangled? That defeats the purpose of making it difficult for someone to figure out what the code is doing. Once mangled, ONLY the mangled name should ever be available...

johnspackman:
Privates mangling is only intended to protect access to privates, not to obscure the code - if you want obfuscation, try https://obfuscator.io/ (I use that in production, it increases the size quite a bit but it raises the bar way *way* higher than private name mangling).

What happens is that the compiler translates this code:

```
members: {
  myFunc() {},
  __myPrivate() {}
}
```

into:

```
members: {
  myFunc: function myFunc() {},
  __myPrivate__PRIVATE_4: function __myPrivate() {}
}
```

This means that you cannot use another class to call "thatObject.__myPrivate" because it does not exist.  The function is declared as `function __myPrivate() {}` so that your call stack looks normal.

However, the framework will by default override the name of the function, and use the key from the `members` property, which we do not want because the compiler has already made sure that there is a good name.

I'll amend the PR I just did for compiler and strip the function naming in build targets, so the mangled name really is mangled and never appears.

However this PR *is* still required for source targets, and will have no effect on build targets - it should also be safe for use with the generator.

cboulanger (approved):

derrell:
> I'll amend the PR I just did for compiler and strip the function naming in build targets, so the mangled name really is mangled and never appears.

Thanks. I believe, with the generator, that the build version mangled names such that the original name was not available anyplace in the output. I'm not asking for anything other than what (I think/hope) we had before.... and it sounds like you're able to easily provide that.

derrell (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

PR #9977, opened 2020-05-18
fix npm deployment

johnspackman:
Minor amendments to @cboulanger 's update to add sanity-check to deployment - this builds the framework, then uses that to replace the compiler's version of the framework, and then rebuilds it.

This should ensure we don't have an issue where the cyclic dependencies between compiler and framework break the builds post deployment.

### DIFF
--- a/.travis/deploy_npm
+++ b/.travis/deploy_npm
@@ -40,7 +40,7 @@ echo ">>> Installing dependencies..."
 exe npm install
 
 echo ">>> Installing github.com/qooxdoo/qooxdoo-compiler"
-exe [ -d ./qooxdoo-compiler ] && exe rm -rf ./qooxdoo-compiler
+exe rm -rf ./qooxdoo-compiler
 exe git clone https://github.com/qooxdoo/qooxdoo-compiler.git --depth=1 --single-branch ./qooxdoo-compiler
 exe cd ./qooxdoo-compiler
 exe npm install
@@ -53,6 +53,11 @@ exe $QX compile --config-file=compile-server.json --target=build
 exe mkdir -p ./lib
 exe rm -rf ./lib/*
 exe cp -R compiled/build/qx_server ./lib
+
+echo ">>> Running Compiler Tests"
+exe cd ./qooxdoo-compiler
+exe test/test.linux.sh
+exe cd ..
 
 echo ">>> Compiling & deploying server library with compiler master using the newly built qx_server library..."
 exe rm -rf ./qooxdoo-compiler/node_modules/@qooxdoo/framework/lib
@@ -81,5 +86,6 @@ echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN:-}" > ~/.npmrc
 exe npm publish --access public
 
 
-
+echo ">>> Cleaning up"
+rm -rf qooxdoo-compiler source
 
--- a/.travis/deploy_npm
+++ b/.travis/deploy_npm
@@ -30,7 +30,7 @@ echo "-------------------------------------------------------------------------"
 cd $BASE_DIR
 if [ "{TRAVIS_TAG:-}" != "$FRAMEWORK_VERSION" ]; then
     echo "Bump version ..."
-    #exe tool/admin/bin/bumpqxversion.py $TAG_VERSION
+    exe tool/admin/bin/bumpqxversion.py $TAG_VERSION
 fi
 
 exe node -v
@@ -44,12 +44,12 @@ exe [ -d ./qooxdoo-compiler ] && exe rm -rf ./qooxdoo-compiler
 exe git clone https://github.com/qooxdoo/qooxdoo-compiler.git --depth=1 --single-branch ./qooxdoo-compiler
 exe cd ./qooxdoo-compiler
 exe npm install
-exe npm link
+QX=`pwd`/qx
 exe cd ..
-exe qx --version
+exe $QX --version
 
 echo ">>> Compiling & deploying server library with the compiler master using current qx_server library from NPM..."
-exe qx compile --config-file=compile-server.json --target=build
+exe $QX compile --config-file=compile-server.json --target=build
 exe mkdir -p ./lib
 exe rm -rf ./lib/*
 exe cp -R compiled/build/qx_server ./lib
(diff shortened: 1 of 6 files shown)


cboulanger (approved):
Thanks, I am happy you found this useful!

cboulanger:
This will run the compiler with the framework server library (qx_server) compiled from current master and should thus should catch a bug that breaks any version of the compiler (NPM or check-out) **before** the framework with qx_server is deployed to NPM. It does only check if the compiler can still compile. Any other compiler command that is affected by this framework change will go unnoticed. Maybe it would make sense to run the compiler tests as a step before NPM deployment, using the freshly brewed qx_server package?

.travis/deploy_npm
```diff
@@ -36,22 +36,50 @@ fi
...
-#compile everything
-exe npx qx compile --config-file=compileServer.json --target=build 
-#exe npx qx compile --config-file=compile.json       --target=build 
-#remove unnecessary stuff
-#exe find ./apps/transpiled -type f -name '*.map' -delete
-#exe find ./apps/transpiled -type f -name '*.js'  -delete
-#copy source
+
+echo ">>> Installing github.com/qooxdoo/qooxdoo-compiler"
+exe [ -d ./qooxdoo-compiler ] && exe rm -rf ./qooxdoo-compiler
```
derrell, line 43: I don't think you want `exe [ -d ./qooxdoo-compiler ] &&`. You're already using `rm -rf` which won't complain if it doesn't exist, and furthermore, since you're test is `-d`, for a directory, if there happened to be a file there called `qooxdoo-compiler` it wouldn't remove it. (Granted, highly unlikely.) In any case, I think you can get rid of the initial test.

.travis/deploy_npm
```diff
@@ -36,22 +36,50 @@ fi
...
+exe npm install
+QX=`pwd`/qx
+exe cd ..
+exe $QX --version
+
+echo ">>> Compiling & deploying server library with the compiler master using current qx_server library from NPM..."
+exe $QX compile --config-file=compile-server.json --target=build
+exe mkdir -p ./lib
+exe rm -rf ./lib/*
+exe cp -R compiled/build/qx_server ./lib
```
derrell, line 55: This appears not to be doing what the echo said it would do. This is using the compiled qx_server, not the current one in NPM...???

derrell (changes requested):

.travis/deploy_npm
cboulanger: The compiler has an NPM dependeny on `@qooxdoo/framework`, which is of course only the qx_server package (not the other framework classes). So the initial build of the framwork from master is done with qx_server from NPM. In the second step, this qx_server package is swapped out with the newly brewed one, in order to test if the compiler still works and to actually build and deploy the framework.

cboulanger:
Answering myself here: I just realized that this is the wrong place for running the compiler tests. This test has to be done in the workflow testing the PR, not in the deploy to NPM workflow.

.travis/deploy_npm
cboulanger: As I said in the other comment, it's actually too late here since any change that will break the compiler is already in master at that point. We need to move this kind of test to the earlier PR testing workflow.

johnspackman: i think it's still worth doing, because as it's cyclic dependencies, there could be an incompatibility that breaks during deployment - eg if compiler/PR and a framework/PR are going out together, both could pass tests, and then compiler/PR is merged and the framework/PR will no longer compile.  

By running the compiler tests as part of framework release, we know that we're not going to break the released compiler or the new framework.

johnspackman: i'm just testing a version that does this now

compile-all-apps.json
```diff
@@ -0,0 +1,142 @@
+{
```
hkollmann, line 1: Changes of the compile files should'nt be part of this PR.

hkollmann (changes requested):

compile-server.json
```diff
@@ -66,5 +67,6 @@
         "qx.event.handler.PointerCore"
       ]
     }
-  ]
+  ],
+  "$schema": "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/source/resource/qx/tool/schema/compile-1-0-0.json"
```
hkollmann, line 71: Here we need the new url

hkollmann (changes requested):

compile.json
```diff
@@ -50,92 +51,10 @@
...
-        "qx.test.*"
-      ],
-      "exclude": [
-      ]
     }
   ],
   "sass": {
     "compiler": "legacy"
   },
   "$schema": "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/source/resource/qx/tool/schema/compile-1-0-0.json"
```
hkollmann, line 59: Here too,

hkollmann (changes requested):

compile-all-apps.json
```diff
@@ -0,0 +1,142 @@
...
+        "qx.test.*"
+      ],
+      "exclude": [
+      ]
+    }
+  ],
+  "sass": {
+    "compiler": "legacy"
+  },
+  "$schema": "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/source/resource/qx/tool/schema/compile-1-0-0.json"
```
hkollmann, line 141: Here we need the new url

hkollmann (changes requested):

.travis/deploy_npm
```diff
@@ -36,22 +36,56 @@ fi
...
+exe test/test.linux.sh
+exe cd ..
+
+echo ">>> Compiling & deploying server library with compiler master using the newly built qx_server library..."
+exe rm -rf ./qooxdoo-compiler/node_modules/@qooxdoo/framework/lib
+exe cp -fR lib ./qooxdoo-compiler/node_modules/@qooxdoo/framework
+exe cp -f package.json Manifest.json ./qooxdoo-compiler/node_modules/@qooxdoo/framework
+exe rm -rf compiled
+exe $QX compile --config-file=compile-server.json --target=build
+exe $QX deploy --config-file=compile-server.json
```
hkollmann, line 68: IMHO here is --out=?? missing

.travis/deploy_npm
```diff
@@ -36,22 +36,56 @@ fi
...
+exe cd ..
+exe $QX --version
+
+echo ">>> Compiling & deploying server library with the compiler master using current qx_server library from NPM..."
+exe $QX compile --config-file=compile-server.json --target=build
+exe mkdir -p ./lib
+exe rm -rf ./lib/*
+exe cp -R compiled/build/qx_server ./lib
+
+echo ">>> Running Compiler Tests"
```
hkollmann, line 57: This tests are runing with the npm Version library version. Should'nt the new library not be copied before the tests?

.travis/deploy_npm
cboulanger: No, that's in the `compile.json`'s `deployPath` directive. That's the beauty of it and that's why they also need to be part of this PR.

.travis/deploy_npm
cboulanger: True, the tests need to be moved down to after line 66, I think.

hkollmann (approved):

oetiker (approved):

level420 (approved):

derrell (approved):

PR #9978, opened 2020-05-19
Basic tree node label editing

rad-pat:
I've had the need to edit the nodes in a tree and wondered if this might be useful to have as an option in the main source. It seemed that the basic edit wasn't too much trouble so I wondered why it was not included in the first instance? Perhaps there wasn't the requirement. The node editor works ok as a cell editor, but could do with being aligned with the node in the tree and I wasn't sure of the best way to go about that. Any pointers would be really helpful and I'll get this fixed up.


cboulanger:
This would be a great addition to the core. We just have to make sure that it doesn't break anything. Unfortunately, ther are no tests in the framework that test the behavior of the TreeVirtual widget, but could you maybe add a test case to https://github.com/qooxdoo/qxl.demobrowser/tree/master/source/demo/treevirtual using your fork, which could then be added to the demobrowser once the PR has been accepted?

cboulanger:
I'll try to test the PR branch with my app in the next few days to see how it works.

rad-pat:
Thanks for looking, yep I'll add some demo code

framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -244,7 +244,7 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel",
       // The tree column is not editable
```
derrell, line 244: This comment is no longer accurate, it seems.

framework/source/class/qx/ui/treevirtual/celleditor/NodeEditor.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+
+
+    // interface implementation
+    createCellEditor : function(cellInfo)
+    {
+      var cellEditor = this._createEditor();
+
+      // The value in the case of a Tree is a Node and we want the label
+      if (cellInfo.value === null || cellInfo.value === undefined) {
```
derrell, line 52: If `cellInfo.value` is required to be an object, as indicated by the value set below, then this test might better be
```
if (cellInfo.value === null || typeof cellInfo.value != "object")
```

derrell (changes requested):
This looks like a good addition.

derrell (approved):

level420 (approved):

johnspackman (approved):

cboulanger (approved):
Couldn't test it yet, but it looks good to me.

hkollmann (approved):

cboulanger:
Thank you @rad-pat for this addition!

PR #9979, opened 2020-05-19
Implement liveUpdateOnRxMatch property

oetiker:
If `liveUpdate` is set on a textField then a changeValue event is fired every time a key is pressed.

This PR adds a sort of restricted version of `liveUpdate`. Set `liveUpdateOnRxMatch` to a regular expression *or* string (which will be transformed into a regular expression). This will cause the `changeValue` event to be fired only once the content of the field matches the regular expression. This allows combining validation with live update behavior without the field going 'red' while typing without a chance of the value being valid. For an email address the following might be handy: `'^\S+@\S+\.\S{2,}$'`

### DIFF
--- a/framework/source/class/qx/ui/form/AbstractField.js
+++ b/framework/source/class/qx/ui/form/AbstractField.js
@@ -213,7 +213,17 @@ qx.Class.define("qx.ui.form.AbstractField",
       check : "Boolean",
       init : false
     },
-
+    /**
+     * Fire a {@link #changeValue} event whenever the content of the
+     * field matches the given regular expression. Accepts both regular
+     * expression objects as well as strings for input.
+     */
+    liveUpdateOnRxMatch :
+    {
+      check : "RegExp",
+      transform : "_string2RegExp",
+      init : null
+    },
     /**
      * String value which will be shown as a hint if the field is all of:
      * unset, unfocused and enabled. Set to null to not show a placeholder
@@ -524,6 +534,13 @@ qx.Class.define("qx.ui.form.AbstractField",
       }
     },
 
+    // property transform
+    _string2RegExp: function(value, old) {
+        if (qx.lang.Type.isString(value)) {
+            value = RegExp(value);
+        }
+        return value;
+    },
 
     // overridden
     tabFocus : function() {
@@ -585,6 +602,11 @@ qx.Class.define("qx.ui.form.AbstractField",
         // check for the live change event
         if (this.getLiveUpdate()) {
           this.__fireChangeValueEvent(value);
+        }
+        // check for the fireChangeValueOnRxMatch change event
+        var fireRx = this.getLiveUpdateOnRxMatch();
+        if (fireRx && value.match(fireRx)) {
+            this.__fireChangeValueEvent(value);
         }
       }
     },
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/form/AbstractField.js
```diff
@@ -213,7 +213,17 @@ qx.Class.define("qx.ui.form.AbstractField",
...
     },
-
+    /**
+     * Fire a {@link #changeValue} event whenever the content of the
+     * field matches the given regular expression. Accepts both regular
+     * expression objects as well as strings for input.
+     */
+    liveUpdateOnRxMatch :
+    {
+      check : "RegExp",
```
cboulanger, line 223: If the property setter checks for "RegExp" and throws an error if the value isn't one, you wouldn't be able to store a string in the first place, would you? I think to be able to set both is good, since the value could be bound from a JSON-marshalled object.

oetiker: the transform is applied before the check

cboulanger: I see!

cboulanger (approved):

hkollmann (approved):

framework/source/class/qx/ui/form/AbstractField.js
derrell: You don't actually need a "check" and it's confusing to have it as "RegExp" when you're really expecting a string as input. Instead, the transform function can throw an error to also indicate that validation fails. That would be a cleaner way of doing this, I think.

derrell (changes requested):

derrell:
This looks to be a particularly useful feature, BTW!

framework/source/class/qx/ui/form/AbstractField.js
oetiker: Well that is the way we describe the use case for the transform function in the documentation ... I am not expecting a string as input ... you can use both, a string OR a regular expression.

derrell: My point is that it looks like magic, as is. If `check` says RegExp then that's what will show up in the api docs, and the user will only think it can be that type. Maybe change it to:

```
check : function(value)
{
  return value instanceof RegExp || typeof value == "string";
}
```
Granted, by the time the function is called, it will always be a RegExp if what was passed in was a string, but this shows what the range of inputs can be, and provided the required checking as well.

If you feel strongly about leaving it as is, I'll withdraw my objection, but I do have a preference for something that accurately documents the legal range of inputs.

oetiker: The issue here really is about the auto-generated documentation ... that is why I documented the special behavior explicitly in the text. And yes I do feel strongly about it since this is how the transform property is supposed to be used.

Maybe we can think of a way to fix the documentation maybe integrating some jsdoc supplied for the transform function?

johnspackman (approved):

level420 (approved):

derrell (approved):

hkollmann:
Thanks @oetiker

derrell:
As of 3 Aug 2020, months after this PR was merged, this new property is not showing up at http://qooxdoo.org/qxl.apiviewer/#qx.ui.form.AbstractField. Is something wrong with the API viewer, or did this code get lost somehow?

PR #9980, opened 2020-05-22
Update AbstractImage.js

slachtar:
When changing table row height, images are aligned to top which  results on an ugly look, aligning it to middle fixes it.

### DIFF
--- a/framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
+++ b/framework/source/class/qx/ui/table/cellrenderer/AbstractImage.js
@@ -242,7 +242,7 @@ qx.Class.define("qx.ui.table.cellrenderer.AbstractImage",
           width: this.__imageData.width + "px",
           height: this.__imageData.height + "px",
           display: qx.core.Environment.get("css.inlineblock"),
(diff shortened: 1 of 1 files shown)


oetiker (approved):

cboulanger (approved):
I am trying to think of cases where this might be a backward-compatibility issue. But probably there aren't such cases. It certainly is a better default behavior.

hkollmann:
@slachtar: Could you write a comment here https://github.com/qooxdoo/qooxdoo/issues/9909 because it's not backward-compatibility.
So we should mention it in the release news.

level420 (approved):

derrell (approved):
This looks like it will be a visual BC difference for every application that that uses the image cellrenderer. This is the right time to make this change, as we come up on a major version bump, but it definitely needs to be in the release notes.

hkollmann (approved):

slachtar:
> @slachtar: Could you write a comment here #9909 because it's not backward-compatibility.
> So we should mention it in the release news.

Sorry, was in vacation and noticied you already added the comments, thank you for accepting the PR.

PR #9982, opened 2020-05-23
ColorUtil Enhancements

oetiker:
* handle hex4 and hex8 wherever hex3 and hex6 have been expected until now
* remove alpha if it is 1
* new functions `scale` and `adjust` modeled after the same functions in sass. They let you modify individual properties of an incoming color.


cboulanger (approved):

hkollmann (approved):

hkollmann:
Needs some linter fixes....

framework/source/class/qx/util/ColorUtil.js
```diff
@@ -164,83 +166,63 @@ qx.Bootstrap.define("qx.util.ColorUtil",
...
+      var color;
       if (this.isNamedColor(str))
       {
-        return this.NAMED[str];
+        color = this.NAMED[str].concat();
       }
       else if (this.isSystemColor(str))
       {
-        throw new Error("Could not convert system colors to RGB: " + str);
+          throw new Error("Could not convert system colors to RGB: " + str);
```
derrell, line 202: Why is it an error if it's a system color?

derrell (changes requested):
This looks like it will be a nice enhancement. I propose, however, that we treat the current master branch as feature-locked except for bug fixes (and the occasional late non-BC change) until release. Can this hold off for a 6.1 release?

framework/source/class/qx/util/ColorUtil.js
oetiker: I have no idea ... that has been there before

oetiker:
> This looks like it will be a nice enhancement. I propose, however, that we treat the current master branch as feature-locked except for bug fixes (and the occasional late non-BC change) until release. Can this hold off for a 6.1 release?

we can, but then we also have to hold off on the new theme ...

cboulanger:
@derrell You're right in principle, but I think we should have a new theme for the release and this appears necessary for the theme to work...

derrell:
I agree. The new theme has been in the planning for a long time, and I didn't realize that this was part of that work.

derrell (approved):

zaucker (approved):

johnspackman (approved):

PR #10000, opened 2020-06-02
Fix Boolean Renderer position

rad-pat:
With #9980, images are now vertically aligned to middle. The boolean renderer assumed alignment to top and thus added a padding of 4px which is now no longer required or the checkboxes look offset.


cboulanger (approved):

derrell (approved):

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

hkollmann:
Thanks @rad-pat

PR #10003, opened 2020-06-04
Add option to deselect on tap below rows in Table

rad-pat:
A requirement for our particular implementation so I thought I'd make it an option for the table. If you click in the control, but below the table rows, the current selection is removed.

### DIFF
--- a/framework/source/class/qx/ui/table/Table.js
+++ b/framework/source/class/qx/ui/table/Table.js
@@ -556,6 +556,18 @@ qx.Class.define("qx.ui.table.Table",
     },
 
 
+    /**
+     * Whether to reset the selection when the unpopulated table area is tapped.
+     * The default is false which keeps the behaviour as before
+     */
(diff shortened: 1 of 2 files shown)


cboulanger:
Is this a bevior known from other table implementations? It seems to be very special to me, but I am open to be convinced. Otherwise, as you know, simply create a subclass of Table and Scroller and add you behavior there.

rad-pat:
It seems like this behaviour reflects the sort of behaviour that you see in  e.g. Windows Explorer, or Files browser in Ubuntu that our users are familiar with. I can of course just implement this in my own class, but I thought the behaviour might be useful for everyone,

derrell (approved):
I have wanted, but never gotten around to implementing, something like this. Thanks.

hkollmann (approved):

oetiker (approved):

cboulanger (approved):

johnspackman (approved):

PR #10016, opened 2020-07-24
Allow to replace the default contains function.

adrelino:
This is just a tiny non-breaking addition to the API, but I consider it very useful.

With the freedom to supply my own contains function to the qx.ui.popup.Manager, I was finally able to fix some issues with too early closing of a parent popup when clicking inside a nested menu.

This allows e.g. to write a custom contains function that not only considers the layout parents, but also logical parents. 

- I needed it to display a table inside a popup withouth the popup closing when (de)-selecting columns from the table-column-menu.
- for my custom select boxes with search capabilities, closing using the _onBlur event of the selectBox was really hard to get right, since a click into the search filter would close the selectBox. I tried lots of workarounds, but with the notion of logical parents this got a whole lot easier.

### DIFF
--- a/framework/source/class/qx/ui/popup/Manager.js
+++ b/framework/source/class/qx/ui/popup/Manager.js
@@ -62,6 +62,16 @@ qx.Class.define("qx.ui.popup.Manager",
   {
     __objects : null,
 
+    __containsFunction : qx.ui.core.Widget.contains,
+
+    /**
+     * Replace the default contains function.
+     * 
+     * @param func {Function|qx.ui.core.Widget.contains}
+     */
+    setContainsFunction: function(func){
+      this.__containsFunction = func;
+    },
 
     /**
      * Registers a visible popup.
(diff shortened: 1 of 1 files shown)


oetiker (approved):
that searchable selectbox also sounds very nice :) care to contribute? I am about to write one too ...

derrell (approved):

cboulanger (approved):

hkollmann (changes requested):
What's about using a property instead of the setMethod?
With your solution you can not revert to the old contains function clearly.

adrelino:
> that searchable selectbox also sounds very nice :) care to contribute? I am about to write one too ...

They are here: https://github.com/adrelino/qooxdoo-extensions. Look for FilterSelectBox, FilterMultiSelectBox, FilterTreeSelectBox. 

The demo is currently hosted at http://www.adrian-haarbach.de/qooxdoo-extensions/build/. Once I know how to fix https://github.com/adrelino/qooxdoo-extensions/issues/1 it should also show up in https://qooxdoo.org/qxl.packagebrowser.

adrelino:
> What's about using a property instead of the setMethod?
> With your solution you can not revert to the old contains function clearly.

sure, I now changed it to use a property instead.

level420 (approved):

hkollmann (approved):
Great!

Issue #10018, opened 2020-07-31
qx.ui.tree.VirtualTree: method isShowTopLevelOpenCloseIcons always returns true and ignores property showTopLevelOpenCloseIcons


ofilippov:
**Describe the bug**
Class `qx.ui.tree.VirtualTree` has property:
```
showTopLevelOpenCloseIcons :
{
  check : "Boolean",
  init : false,
  apply : "_applyShowTopLevelOpenCloseIcons"
}
```

And it also has a method:
```
// Interface implementation
isShowTopLevelOpenCloseIcons : function() {
  return true;
}
```

This method should return correct property state.

**Expected behavior**
Method `isShowTopLevelOpenCloseIcons` of class `qx.ui.tree.VirtualTree` should return the same value as a method `getShowTopLevelOpenCloseIcons`

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo/pull/10019

PR #10019, opened 2020-07-31
https://github.com/qooxdoo/qooxdoo/issues/10018 qx.ui.tree.VirtualTre…

ofilippov:
…e: method isShowTopLevelOpenCloseIcons always returns true and ignores property showTopLevelOpenCloseIcons

*Solution*

Method isShowTopLevelOpenCloseIcons of class qx.ui.tree.VirtualTree should return the same value as a method getShowTopLevelOpenCloseIcons

### DIFF
--- a/framework/source/class/qx/ui/tree/VirtualTree.js
+++ b/framework/source/class/qx/ui/tree/VirtualTree.js
@@ -608,7 +608,7 @@ qx.Class.define("qx.ui.tree.VirtualTree",
     
     // Interface implementation
     isShowTopLevelOpenCloseIcons : function() {
-      return true;
+      return this.getShowTopLevelOpenCloseIcons();
     },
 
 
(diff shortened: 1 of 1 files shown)


framework/source/class/qx/ui/tree/VirtualTree.js
```diff
@@ -608,7 +608,7 @@ qx.Class.define("qx.ui.tree.VirtualTree",
     
     // Interface implementation
     isShowTopLevelOpenCloseIcons : function() {
-      return true;
+      return this.getShowTopLevelOpenCloseIcons();
```
derrell, line 611: I believe this whole function is unnecessary. I believe it is overriding the function of the same name which is automatically-generated for boolean properties. I believe we should probably remove this method entirely, if my recollection of the automatically-created methods is accurate...

derrell (changes requested):

framework/source/class/qx/ui/tree/VirtualTree.js
ofilippov: Yes, I agree. I've removed the whole method. Thanks!

derrell (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

PR #10023, opened 2020-08-10
Prevent dbl-click open/close button starting node edit

rad-pat:
With the tree node editing available, it was possible to click on the open/close button twice and invoke the edit, which is not desirable after this has happened a few times. 
The purpose of the PR is to capture the dblclick if it occurs on the open/close button and prevent the edit. There was some code for identifying the position of the open/close button in the Selection Manger. The calculation of this has been pushed to the tree renderer that was responsible for placing the button - seemed like sense. The position is then accessible via a method call on the main tree.

### DIFF
--- a/framework/source/class/qx/ui/treevirtual/TreeVirtual.js
+++ b/framework/source/class/qx/ui/treevirtual/TreeVirtual.js
@@ -95,6 +95,17 @@ qx.Class.define("qx.ui.treevirtual.TreeVirtual",
    *           return new qx.ui.table.columnmodel.Resize(obj);
    *         }
    *       </pre></dd>
+   *     <dt>tablePaneScroller</dt>
+   *       <dd>
+   *         Instance of {@link qx.ui.treevirtual.pane.Scroller}.
+   *         Custom table pane scroller for the tree
+   *         <pre class='javascript'>
+   *         function(obj)
+   *         {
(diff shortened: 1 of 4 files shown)


johnspackman (approved):

level420 (approved):

cboulanger (approved):

framework/source/class/qx/ui/treevirtual/TreeVirtual.js
```diff
@@ -155,6 +155,12 @@ qx.Class.define("qx.ui.treevirtual.TreeVirtual",
         };
     }
 
+    if (!custom.tablePaneScroller) {
```
derrell, line 158: Would you please add this to the documentation for `custom` at the top of the constructor.

derrell (changes requested):

framework/source/class/qx/ui/treevirtual/TreeVirtual.js
rad-pat: No probs, added

rad-pat: Added with latest update

derrell: Thanks.

derrell (approved):

hkollmann (approved):

PR #10030, opened 2020-08-28
Allow to debug invalid ids in qx.core.Id.MObjectId

cboulanger:
This adds three things to the API of `qx.core.Id.MObjectId`:

1) A second boolean parameter `onlyIfExists` to `getQxObject()` which, if true, makes the method return the object only if it has already been created (i.e. will not create it on demand),  otherwise, it returns undefined.
2) A second optional {qx.core.Object} owner` to `_createQxObject()` which defaults to `this` and allows to specify a different owner than the current object instance. 
3) An alias `createQxObject()` to the `_createQxObjectImpl()`method which as a method name is much nicer-looking and easier to remember - since we (at least @johnspackman and I) would like qooxdoo users to adopt using `_createQxObjectImpl()` as an architectural design pattern for qooxdoo apps, we need a better method name than the current cryptic one.


cboulanger:
and I forgot 
4) better debugging when a non-existing id is encountered.

framework/source/class/qx/core/MObjectId.js
```diff
@@ -157,59 +166,85 @@ qx.Mixin.define("qx.core.MObjectId", {
...
+      } else if (!(owner instanceof qx.core.Object)) {
+        throw new Error("Owner must be instance of qx.core.Object or undefined.");
+      }
+      var result;
+      if (typeof this.createQxObject == "function") {
+        // more user-friendly method name
+        result = this.createQxObject(id);
+      } else {
+        // method name of the original implementation
+        result = this._createQxObjectImpl(id);
```
oetiker, line 211: how about complaining a little in debug mode about the use of the old name ?

framework/source/class/qx/core/MObjectId.js
```diff
@@ -157,59 +166,85 @@ qx.Mixin.define("qx.core.MObjectId", {
...
-    _createQxObject: function(id) {
-      var result = this._createQxObjectImpl(id);
+    _createQxObject: function(id, owner) {
+      if (owner === undefined) {
+        owner = this;
+      } else if (!(owner instanceof qx.core.Object)) {
+        throw new Error("Owner must be instance of qx.core.Object or undefined.");
+      }
+      var result;
+      if (typeof this.createQxObject == "function") {
```
johnspackman, line 206: IMHO this is not a good way to change the API method name because it's non-obvious - the reader would normally expect the list of methods in `members` to define the API, and it is documented that the `_createQxObjectImpl` will be called, yet this change means that if the class *or any of it's descendant classes* implement `createQxObject` then the documented method is ignored.

I do think that in retrospect it would be good to change the naming here, because as @cboulanger says the object ID mechanism supports a coding pattern which is really useful in creating clear and concise widget UIs - more on that later, we've talked about a medium.com blog posting as well as docs to explain our thoughts.

IMHO the "create" method should stay protected though, because part of the API specification is that "create" will only be called once for each ID (and it could easily cause issues if that was not followed).  By making it protected and using the public `getQxObject` to control access, the `getQxObject` is a gatekeeper which reliably enforces those API rules.

I think I would prefer that `_createQxObjectImpl` and `_createQxObject` are simply swapped around, i.e. that the method which people override is called `_createQxObject` and the internal method which `getQxObject` defers to is `_createQxObjectImpl`.

This would be a breaking API change, but as we're talking about pre-release there is room to decide to do this; IMHO the benefits are better clarity because the design pattern that it promotes (or: that we intend to promote) is really very effective.

johnspackman: In line with @oetiker's comment below, we could detect the implementation of `_createQxObjectImpl` in debug mode and give a warning - this detection/warning would be scheduled to be removed in 7.x

framework/source/class/qx/core/MObjectId.js
```diff
@@ -157,59 +166,85 @@ qx.Mixin.define("qx.core.MObjectId", {
...
+        // more user-friendly method name
+        result = this.createQxObject(id);
+      } else {
+        // method name of the original implementation
+        result = this._createQxObjectImpl(id);
+      }
       if (result !== undefined) {
-        this.addOwnedQxObject(result, id);
+        owner.addOwnedQxObject(result, id);
```
johnspackman, line 214: Sorry but as it stands I'm not so keen on this either - although we have built it so that it is possible to add objects by ID, that's really intended for situations where the object does not implement `_createQxObjectImpl`, ie it's a work around for existing code.  The workaround means that objects can be located via a global ID string (eg "dlgLoginForm/edtUsername"), the initial use case was for automated testing.

The coding pattern which is so useful is centered around an object creating it's own objects, and they are exposed kind of like an API.  For example, a login form dialog box might expose objects with the IDs "edtUsername", "edtPassword", and "btnOk" and document them as two text fields and a button; there will be a lot more widgets used to make up that form but these are not necessarily exposed (or possibly the IDs could be prefixed with "_" or "__").  

Preserving the Owner/Ownee relationship is critical because it allows name conflicts to be avoided - but when we promote adding objects to another objects, we're breaking that protection because we are adding the new object directly into the namespace of the `owner`.  

This could have some use, in that you would be able to override the creation of a widget (eg provide the `edtUsername` textfield of a login dialog) **but** this will only work if the object has not yet been created by the dialog.  If the dialog is later modified in such a way that the constructor calls `getQxObject("edtUsername")`, then the calling code will immediately break.  The correct, and the only reliable, way to change the implementation of a field would be to derive a class from the login dialog and provide your own implementation of `_createQxObjectImpl`.

Also, if we consider the IDs an API, then the object has the right to define new objects and change or remove protected/private objects as needed; but this solution will add additional, owned objects into the namespace of an object on the fly.  Even putting name conflicts to one side for a moment, I'm not convinced that this is the right way either - it feels quite unstructured, and the equivalent of adding methods to an object on the fly.  I feel that the expected way to add functionality would be by deriving a class, and that would mean that the additions form part of the code structure (as opposed to runtime) and are much better understood and documented - you could for example easily detect all classes which derive and then check their `_createQxObjectImpl` and then know *for a fact* what API object IDs are valid in that object.

AIUI the intention of this change is to support compositing of objects, ie assist building the object graph - but I think that the correct way to do that is to either use the owner/ownee ID mechanism, where you can have "grand-owner/owner/ownee/etc" or to have a class which is specifically intended for compositing in that way.  That special "compositing" class would be the place to support a lot of other features which are very relevant to compositing (such as recording modified state, persistence, etc).

johnspackman (changes requested):
In this review, I've mentioned a really useful "coding pattern" that object ID mechanism enables - here's some work that @cboulanger has done which illustrate the clarity

[Before, without QxObject ID](https://github.com/cboulanger/bibliograph/blob/91602806be0902fc6e4fad7eba894b8577544a41/src/client/bibliograph/source/class/bibliograph/ui/main/Toolbar.js) 

[After rewrite based on QxObjectID](https://github.com/cboulanger/bibliograph/blob/develop/src/client/bibliograph/source/class/bibliograph/ui/main/Toolbar.js)

Note how in the "before" the code has to hand craft the dependency order of objects and has to expose methods for each object, whereas the "after" focuses on declaring the widgets - and in this way the exposed widgets become part of an API.

Also, much like `@childcontrol` it's our intention to add JSDoc comment support for this, eg:

```
/**
  * @objectId(“toolbar”, qx.ui.toolbar.ToolBar) the main toolbar for the window
  */
_createObjectImpl(id) {
  switch(id) {
    /** @objectId(qx.ui.toolbar.Button) the OK button
    case “btnOk”:
```

framework/source/class/qx/core/MObjectId.js
cboulanger: Thanks for the comments and I see your points. I'll change the PR (or what's left of it) later. Good to get the discussion going!

cboulanger:
As you say, the method names of this API are not ideal, in particular not very good in expressing the philosophy behind the system. If the priority is, as you say, that the hierarchy of owner/ownee is best not to be fiddled with (as in my attempt to attach an "ownee" to a different owner than the one producing it), `getQxObject` or `_createQxObject` doesn't really convey that idea. (`addOwnedQxObject` is better, but from the perspective of the design, a workaround, not a central part of the API). No immediate alternatives come to my mind, but maybe others have an idea. I wouldn't switch `_createQxObject` and `_createQxObjectImpl`, but rather find better semantics to replace them.

cboulanger:
The use case for a different "owner" is this: 

- I have a Plugin object which needs to attach some menu buttons created in the plugin object's `_createQxObjectImpl()` method to a menu in the toolbar of the main application
- These dynamically attached buttons (the plugin may or may not be activated) become part of the id hierarchy of the main application, so it can be called with `qx.core.Id.getQxObject("toolbar/foo-menu/plugin-created-button")`
- It's no problem to do this without the proposed change, just more verbose: 
```javascript
let menu = qx.core.Id.getQxObject("toolbar/foo-menu");
let button=this.getQxObject("plugin-created-button");
menu.add(button);
menu.addOwnedQxObject(button);
```
The button is now attached to the menu and no longer to the Plugin object, so in theory, if I would call `pluginObject.getQxObject("plugin-created-button")`, it would create a new one. 

My question is if my use case could or should be solved differently.

cboulanger:
I wonder if the idea behind `qx.core.Id` could be better expressed with "idChild"/"idParent" or "logicalChild"/"logicalParent" rather than with the owner/ownee or "qxObject" vocabulary. "parent" and "child" are well understood concepts in tree structures and the specific nature of the qx.core.Id API is that it is not about a layout hierarchy but a logical hierarchy.

cboulanger:
I reversed the controversial changes and left only the debugging part.

cboulanger:
I moved the discussion over to here so that we can continue it after this PR is closed: https://github.com/qooxdoo/qooxdoo/issues/10033

PR #10032, opened 2020-08-28
Allow to set a MinCellEditHeight on table

rad-pat:
When using a table with very tightly packed rows, using the in-line editor does not allow to see the full text within the editor. This PR allows a user to set a `minCellEditHeight` against a Table which adjusts the size of the FocusIndicator that hosts the cell editor itself, for the period of the edit.  This seems to work well as far as I can tell and only makes the adjustments if the property has been set so should have no effect on current usages.

### DIFF
--- a/framework/source/class/qx/ui/table/Table.js
+++ b/framework/source/class/qx/ui/table/Table.js
@@ -567,6 +567,16 @@ qx.Class.define("qx.ui.table.Table",
       apply : "_applyResetSelectionOnTapBelowRows"
     },
 
+    /**
+     * If set then defines the minimum height of the focus indicator when editing
+     */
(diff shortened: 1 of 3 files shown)


derrell (approved):

cboulanger (approved):

level420 (approved):

hkollmann:
Could you have alook at the failed test please?

hkollmann (approved):

rad-pat:
Test hopefully repaired, I shouldn't have touched the null comparison but my IDE highlights them and I can't help myself 😃

oetiker (approved):

hkollmann:
Could you merge the master? Thanks

PR #10039, opened 2020-09-21
Allow popups to be moved immediately to avoid the lag shown when MPlacement.placeToWidget() is used.

lucapivato:
This is a minor change that benefits popups that use placeToWidget(). Now setDomMove(true) is used only in qx.ui.core.DragDropCursor but it will benefit popups and remove the lag that is sometimes visible when the opener is moved or a flush executed between show() and placeToWidget().


oetiker:
hmm it seems this patch includes part of a previous one ...

hkollmann:
Tests are not working any longer?

lucapivato:
I really hate git. It showed only 1 file changed when I did the PR and it seems like a nightmare to do a PR for just one change, this thing takes all the changes that are already in my branch and there is no way to split them...

oetiker:
@lucapivato an easy way to fix PRs it to use interactive rebase `git fetch upstream;git rebase -i upstream/master` just squash everything and then push with --force this normally results in a nice and clean pr

PR #10040, opened 2020-09-21
Fixed Canvas right/bottom layout and added align support.

lucapivato:


cboulanger (approved):
Looks good to me although i don't know if this will break apps that work around the exisiting bug. However, we can be breaking in v6 if it is for a good cause.

oetiker (approved):
lgtm

level420 (approved):

hkollmann (approved):

derrell (approved):

johnspackman (approved):

PR #10041, opened 2020-09-21
qx.ui.layout.HBox: Local variable gaps is in the wrong scope.

lucapivato:


cboulanger (approved):

framework/source/class/qx/ui/layout/HBox.js
```diff
@@ -314,12 +314,12 @@ qx.Class.define("qx.ui.layout.HBox",
 
 
       // Compute gaps
-      var spacing = this.getSpacing();
+      var spacing = this.getSpacing(), gaps;
```
oetiker, line 317:
```suggestion
      var spacing = this.getSpacing();
      var gaps;
```

oetiker (approved):
also note as far as `var` is concerned, there is no scope there ... so this change does not change anything afaik

level420 (changes requested):
Even if var has no local scope, and both versions are equivalent, this code is clearer.
But please change it to what @oetiker suggested and put `var gaps;` in a separate line. Thank you!

hkollmann (approved):
After changes @oetiker suggested

level420:
@lucapivato sorry. I've highjacked your PR and changed the declaration of gaps to a separate line.

level420 (approved):

derrell (approved):

Issue #10029, opened 2020-08-28
404 errors for MaterialIcons


rad-pat:
Having compiled from latest source and compiler, since the tangible theme was merged, we see 404 errors for MaterialIcons. We haven't changed to use the new themes ourselves

**To Reproduce**
Create new Qooxdoo Application and run, observe 404 errors in browser console.
`qx create myapp --type desktop --noninteractive --out=.`

oetiker:
I could not reproduce that particular problem, still the MaterialIcons and the Tangible themes can not be used, because in the npm @qooxdoo/framework package the file https://github.com/qooxdoo/qooxdoo/blob/master/framework/Manifest.json is missing or rather instead of this file, it seems that the file https://github.com/qooxdoo/qooxdoo/blob/master/Manifest.json which is unfortunate as the master/Manifest is missing all the interesting bits to get the IconFonts working.

oetiker:
@johnspackman or @hkollmann can you help

hkollmann:
@rad-pat : Could you test this issue again? Thanks a lot.

rad-pat:
@hkollmann I can still observe the 404s with both my application and with a newly generated application using `@qooxdoo/compiler    v1.0.0-beta-20200928-1855` and Qooxdoo at `c446bb56ccdd41c3c25887f7b89ff3250e4717c8`

rad-pat:
I no longer get 404s for the MaterialIcons.

PR #10043, opened 2020-09-22
https://github.com/qooxdoo/qooxdoo/issues/10029: 404 errors for Mater…

hkollmann:
…ialIcons

add icons to Manifest.json re issue #10029

### DIFF
--- a/Manifest.json
+++ b/Manifest.json
@@ -86,6 +86,63 @@
     "encoding": "utf-8",
     "class": "source/class",
     "resource": "source/resource",
-    "translation": "source/translation"
+    "translation": "source/translation",
+    "webfonts":[
+      {
+        "name": "MaterialIcons",
+        "defaultSize": 32,
+        "comparisonString": "\ue1fe\uef39",
+        "resources": [
+          "qx/iconfont/MaterialIcons/materialicons-v50.eot",
+          "qx/iconfont/MaterialIcons/materialicons-v50.woff2",
+          "qx/iconfont/MaterialIcons/materialicons-v50.woff",
+          "qx/iconfont/MaterialIcons/materialicons-v50.ttf"
+        ]
+      },
+      {
+        "name": "MaterialIconsOutlined",
+        "defaultSize": 32,
+        "comparisonString": "\ue1fe\uef39",
+        "resources": [
+          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.eot",
+          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.woff2",
+          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.woff",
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

oetiker:
note that this is only 'half' the rent ... actually it will make things worse at first ... since this will cause the compiler to generate code that tries to load these fonts, even when they do not exist ... see https://github.com/qooxdoo/qooxdoo-compiler/issues/760

hkollmann:
MaterialIcons are part of qooxdoo resources - so they should be there

derrell (approved):

oetiker:
@cboulanger when working with a checked out version of qdooxoo, then `framework/Manifest.json` is used while the packaged version uses `Manifest.json` this is also at the root of this problem ... maybe it would make sense to not have virtually the same file twice in the source treen?

cboulanger:
@oetiker You probably meant to mention @hkollmann - I agree, but wonder if we should just keep it for now (except when there is a really simple and BC fix). Maybe we should move the whole source tree up one level in v7 and get rid of the "framework" folder entirely. We need to clean up the source tree anyways - so much stuff we can get rid of in v7.

hkollmann:
The Manifest.json in framework contains stuff use by the python tolchain. With v7 we should cleanup

oetiker (approved):

oetiker:
but to stress it again, merging this PR will make matters worse unless there is some sort of fix to the compiler, since now the code in the browser will try to load the material icon fonts despite the font files not being present as they would have to be copied using appropriate @asset lines ... which will only happen when you put these into your code or load the tangible theme ... 

without them the resulting code produced by the compiler will cause the error message this is said to prevent ... 

the error at the moment only shows in the git version of qooxdoo since the git version uses the Manifest.json in framework which already contains these lines ... while the packaged version of qooxdoo places the toplevel Manifest.json into framework and the toplevel Manifest.json does curently NOT contain the webfont line ... this will get fixed by the patch ... causeing the problem it is said to solve to appear for everyone

oetiker:
Instead of this patch, I would suggest to package the manifest sitting in framework with the npm release as this will make the npm release work the same a s the git version AND fix the compiler such that the resulting code will only try to load webfonts that have assets present or something ...

derrell:
It seems inappropriate to approve this with a question still outstanding in [@oetiker's last response](https://github.com/qooxdoo/qooxdoo/pull/10043#issuecomment-697992839). He is suggesting an alternative to this PR...

hkollmann:
The framework version contains stuff for the python tool chain - the one in the root not. That's the reason for the differentz versions.
Both versions contains the laoding of the webfonts.
Change of the compiler is another issue.

derrell (approved):

johnspackman (approved):

level420 (approved):

PR #10044, opened 2020-09-25
Add provision for updating a single table row

rad-pat:
Prompted by performance of refreshing a cell which was incrementing each second, allow to send an update for just the affected row. The method is constructed from parts taken from the full update, or an update due to scroll.

As part of this, I also spotted that the performance of getIds() in ResourceManager was not great. The change here halved the time taken in that method for me so hopefully that is also beneficial to everyone else.

### DIFF
--- a/framework/source/class/qx/ui/table/pane/Pane.js
+++ b/framework/source/class/qx/ui/table/pane/Pane.js
@@ -663,7 +663,7 @@ qx.Class.define("qx.ui.table.pane.Pane",
 
     _updateSingleRow: function(row) {
       var elem = this.getContentElement().getDomElement();
-      if (!elem) {
+      if (!elem || !elem.firstChild) {
         // pane has not yet been rendered, just exit
         return;
       }
@@ -679,9 +679,10 @@ qx.Class.define("qx.ui.table.pane.Pane",
 
       var tableBody = elem.firstChild;
       var tableChildNodes = tableBody.childNodes;
-      var rowElem = tableChildNodes[row];
+      var offset = row - firstRow;
+      var rowElem = tableChildNodes[offset];
 
-      if (row > modelRowCount || rowElem == null) {
+      if (row > modelRowCount || rowElem === undefined) {
         this._updateAllRows();
         return;
       }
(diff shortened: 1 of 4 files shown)


framework/source/class/qx/ui/table/pane/Pane.js
```diff
@@ -379,6 +383,8 @@ qx.Class.define("qx.ui.table.pane.Pane",
         this._scrollContent(scrollOffset);
       } else if (onlySelectionOrFocusChanged && !this.getTable().getAlwaysUpdateCells()) {
         this._updateRowStyles(onlyRow);
+      } else if (onlyRow) {
```
derrell, line 386: I think this is zero-relative, so a value of 0 means do it, whereas a value of undefined or false means don't do it. Maybe this test should if `else if (typeof onlyRow == "number" && onlyRow >= 0)`?

derrell (changes requested):

cboulanger (commented):
This looks like a very useful improvement. Since the table is such a central yet very complex widget, it might be good if we check out the PR branch and build our apps with, just it to make sure there are no side effects that might not be apparent and which do not show up in the tests.

framework/source/class/qx/ui/table/pane/Pane.js
rad-pat: Yes, good catch, thanks

framework/source/class/qx/ui/table/pane/Pane.js
```diff
@@ -283,7 +283,11 @@ qx.Class.define("qx.ui.table.pane.Pane",
       if (lastRow == -1 || lastRow >= paneFirstRow && firstRow < paneFirstRow + rowCount)
       {
         // The change intersects this pane
-        this.updateContent();
+        if (firstRow === lastRow) {
```
rad-pat, line 286: Perhaps this should also test the number of rows in the model, if there is only one then we may as well do the full update.

derrell (approved):
LGTM.

For future reference: This should really be two separate PRs: one to do what the title says; the other pertaining to ResourceManager and the Boolean renderer.

johnspackman (approved):

rad-pat:
> `rowElem == null` will also catch `rowElem == 0`. If you really mean null only, it must be `rowElem === null`

I think I actually probably mean `rowElem === undefined` ultimately - will fix it

oetiker (approved):

level420 (approved):

rad-pat:
This isn't quite right yet, just seen a problem when inserting a row. I'm going to split the PR into 2 as suggested.

PR #10045, opened 2020-09-28
Fix a problem on format with named parameters when there is no spaces…

rommni:
This commit fixes a bug when you had no spaces between two parameters in string format function (used by exemple in translation)

Exemple: 
`It's %{currentNumber}/%{totalNumber} done.`
The current code will detect only one argument named `currentNumber}/%{totalNumber`, where it should find `currentNumber` and `totalNumber`. It was because we were looking in arguments for non whitespace characters including `}` , i now excluded the `}` from the valid parameters characters.

### DIFF
--- a/framework/source/class/qx/lang/String.js
+++ b/framework/source/class/qx/lang/String.js
@@ -299,7 +299,7 @@ qx.Bootstrap.define("qx.lang.String",
     format : function(pattern, args)
     {
       var str = pattern;
-      var regexp = /%(\d+)|%{(\S*)}/g;
+      var regexp = /%(\d+)|%{(\S[^{]*)}/g;
       if (!Array.isArray(args)) {
         args = [args];
       }
--- a/framework/source/class/qx/lang/String.js
+++ b/framework/source/class/qx/lang/String.js
@@ -299,7 +299,7 @@ qx.Bootstrap.define("qx.lang.String",
     format : function(pattern, args)
     {
       var str = pattern;
-      var regexp = /%(\d+)|%{(\S[^{]*)}/g;
+      var regexp = /%(\d+)|%{(\S[^}]*)}/g;
       if (!Array.isArray(args)) {
         args = [args];
       }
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

hkollmann (approved):

framework/source/class/qx/lang/String.js
```diff
@@ -299,7 +299,7 @@ qx.Bootstrap.define("qx.lang.String",
     format : function(pattern, args)
     {
       var str = pattern;
-      var regexp = /%(\d+)|%{(\S*)}/g;
+      var regexp = /%(\d+)|%{(\S[^{]*)}/g;
```
derrell, line 302: I think the added portion should be `[^}]` instead of `[^{]` shouldn't it? We want to match everything up to the closing brace. Regardless of the use case (consecutive replacements as described in this PR or just a single replacement), a closing brace is not a legal character in the replacement variable name.

derrell (changes requested):

framework/source/class/qx/lang/String.js
derrell: An interesting question is whether a space is a legal character in the replacement variable name. Previously it was not allowed. With this change, it would be. I suspect that's incorrect; spaces should not be allowed in the variable name.

rommni: You're right it's clearly an error, i made differents tests and finally took the bad version of the regex. Thank you for notice

rommni: clearly a good question. I agree with you however I don't know how to do the regex to solve the problem and avoir the whitespace

derrell:
Sorry, I answered after seeing only your first comment, before you'd commented on the space question. Anyway, I think the regexp I provided should do what you're looking for. Give it some testing to be sure, though.

rommni:
Thank you ;) I made a little test and all seems good with your regex.

derrell (approved):

oetiker (approved):

level420 (approved):

PR #10046, opened 2020-09-29
Add row & cell attributes to cell markup

rad-pat:
As requested by @cboulanger, this adds initial row and column attributes to the markup to identify cells in the table.

### DIFF
--- a/framework/source/class/qx/ui/table/cellrenderer/Abstract.js
+++ b/framework/source/class/qx/ui/table/cellrenderer/Abstract.js
@@ -223,6 +223,8 @@ qx.Class.define("qx.ui.table.cellrenderer.Abstract",
         'left:', cellInfo.styleLeft, 'px;',
         this._getCellSizeStyle(cellInfo.styleWidth, cellInfo.styleHeight, this._insetX, this._insetY),
         this._getCellStyle(cellInfo), '" ',
+        'data-qx-table-cell-row="', cellInfo.row, '" ',
+        'data-qx-table-cell-col="', cellInfo.col, '" ',
         this._getCellAttributes(cellInfo),
         '>' +
         this._getContentHtml(cellInfo),


cboulanger (approved):
Very much in favor, obviously 😄 . Thank you. This will allow to locate table cell divs in the DOM via CSS selectors or get row/colum information from click events without having to know the internals of a qooxdoo app (for example, when using a web driver)

oetiker (approved):

framework/source/class/qx/ui/table/cellrenderer/Abstract.js
```diff
@@ -223,6 +223,8 @@ qx.Class.define("qx.ui.table.cellrenderer.Abstract",
         'left:', cellInfo.styleLeft, 'px;',
         this._getCellSizeStyle(cellInfo.styleWidth, cellInfo.styleHeight, this._insetX, this._insetY),
         this._getCellStyle(cellInfo), '" ',
+        'data-qx-row="', cellInfo.row, '" ',
+        'data-qx-col="', cellInfo.col, '" ',
```
derrell, line 227: This will be useful. Thanks! I think, though, since these are qx-generated attributes, the attribute names should begin with `qx` and should be descriptive and unambiguous. I'd suggest `qx-table-cell-row` and `qx-table-cell-col`.

derrell (changes requested):

framework/source/class/qx/ui/table/cellrenderer/Abstract.js
cboulanger: Good idea, I support that. This will make clear that it is part of a table.

cboulanger: Sorry, didn't read close enough - the attributes should really begin with "data"  to follow the specs and to support the dataset API:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*
https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset

cboulanger: So the unambiguous attribute name would be `data-qx-table-cell-(row|col)`.

derrell: I wasn't aware of that standard. Based on that, I agree with @cboulanger

rad-pat: Ok, I've updated the attribute names accordingly

hkollmann (approved):

derrell (approved):

level420 (approved):

PR #10051, opened 2020-09-30
Increase performance of Boolean cellrenderer

rad-pat:
We noticed performance issues with the Boolean renderer within a table. When analysed, it seemed that the method `getIds()` in `ResourceManager` was not optimal, changing to filter the keys seems to be a reasonable improvement there for the volume of resources we have.  Further analysis showed that only the Boolean renderer was using the `getIds()` method, and even then it would  take the first id given. So, changing the cell renderer to lookup the ID directly in the resource map improves things dramatically. I have left the original code there as a fallback in case there is not an exact match which seemed like it's purpose.

### DIFF
--- a/framework/source/class/qx/ui/table/cellrenderer/Boolean.js
+++ b/framework/source/class/qx/ui/table/cellrenderer/Boolean.js
@@ -131,13 +131,18 @@ qx.Class.define("qx.ui.table.cellrenderer.Boolean",
 
       // Retrieve the ID
       rm = qx.util.ResourceManager.getInstance();
-      ids = rm.getIds(this.__iconUrlTrue);
 
-      // If ID was found, we'll use its first (likely only) element here.
-      if (ids)
-      {
-        id = ids[0];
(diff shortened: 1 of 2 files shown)


cboulanger (approved):

zaucker (approved):

derrell (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
@rad-pat : could you update you branch? Thanks!

oetiker (approved):

hkollmann:
@rad-pat : Could you have a look at the travis test? e.-g. restart it? Thanks.

rad-pat:
@hkollmann I can't see any button to trigger a rebuild, do I have permissions?

hkollmann:
You could use an dummy commit. I can't restart it either.

level420:
I've restarted the travis job

level420:
thank you @rad-pat 
merging.

PR #10052, opened 2020-09-30
Provision for updating a single table row

rad-pat:
Prompted by performance of refreshing a cell which was incrementing each second, allow to send an update for just the affected row set by using the setRows() method of the Simple table model. The method is constructed from parts taken from the full update, or an update due to scroll.

### DIFF
--- a/framework/source/class/qx/ui/table/pane/Pane.js
+++ b/framework/source/class/qx/ui/table/pane/Pane.js
@@ -282,8 +282,12 @@ qx.Class.define("qx.ui.table.pane.Pane",
 
       if (lastRow == -1 || lastRow >= paneFirstRow && firstRow < paneFirstRow + rowCount)
       {
-        // The change intersects this pane
-        this.updateContent();
+        // The change intersects this pane, check if a full or partial update is required
+        if (firstRow === lastRow && this.getTable().getTableModel().getRowCount() > 1) {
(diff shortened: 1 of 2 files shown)


framework/source/class/qx/ui/table/pane/Pane.js
```diff
@@ -655,6 +661,46 @@ qx.Class.define("qx.ui.table.pane.Pane",
...
+      }
+
+      var modelRowCount = this.getTable().getTableModel().getRowCount();
+
+      var tableBody = elem.firstChild;
+      var tableChildNodes = tableBody.childNodes;
+      var offset = row - firstRow;
+      var rowElem = tableChildNodes[offset];
+
+      if (row > modelRowCount || rowElem === undefined) {
```
derrell, line 685: Since we're still supporting old browser versions where `undefined` could be overwritten (e.g., `var undefined = 23;`), I suggest replacing `rowElem === undefined` with `typeof rowElem == "undefined"`

derrell (changes requested):

cboulanger (approved):

level420 (approved):

oetiker (approved):

hkollmann (approved):

level420:
@rad-pat could you please update your branch? thank you!

derrell (approved):

PR #10058, opened 2020-11-03
Simple table model sort methods fixed to handle null, undefined.

### LINKED ISSUE #10014: Sorting a table containing empty fields (null, undefined) does not work and produces holes
If I try to sort a table contining empty fields (null, undefined) by clicking on the column header, the empty fields appear as holes scattered around and not all in the end or in the beginning.

image

see playground example:
https://tinyurl.com/y68bob5f

The reason is that the comparison of null,undefined with other values in the default comparator:

https://github.com/qooxdoo/qooxdoo/blob/97418ba9a62fed5eb00efbbac5190fdde66ed24b/framework/source/class/qx/ui/table/model/Simple.js#L74

is not symmetric, so the native sorting algorithm produces garbage:

```js
"abc" > undefined == false     —>   return -1
undefined > "abc" == false     —>   return -1
```

This is similar to #3378 but for non-numeric columns

adrelino:
This fixes #10014.

### DIFF
--- a/framework/source/class/qx/ui/table/model/Simple.js
+++ b/framework/source/class/qx/ui/table/model/Simple.js
@@ -71,8 +71,11 @@ qx.Class.define("qx.ui.table.model.Simple",
           return result;
         }
       }
-      if(obj1==null && obj2!==null) return -1;
-      if(obj2==null && obj1!==null) return 1;
+      if ((obj1 == null) && (obj2 !== null)) {
+        return -1;
+      } else if ((obj2 == null) && (obj1 !== null)) {
+        return 1;
+      }
       return (obj1 > obj2) ? 1 : ((obj1 == obj2) ? 0 : -1);
     },
 
@@ -98,8 +101,11 @@ qx.Class.define("qx.ui.table.model.Simple",
           return result;
         }
       }
-      if(obj1==null && obj2!==null) return -1;
-      if(obj2==null && obj1!==null) return 1;
+      if ((obj1 == null) && (obj2 !== null)) {
+        return -1;
+      } else if ((obj2 == null) && (obj1 !== null)) {
+        return 1;
+      }
       return (obj1 > obj2) ? 1 : ((obj1 == obj2) ? 0 : -1);
     },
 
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

framework/source/class/qx/ui/table/model/Simple.js
```diff
@@ -71,6 +71,8 @@ qx.Class.define("qx.ui.table.model.Simple",
           return result;
         }
       }
+      if(obj1==null && obj2!==null) return -1;
+      if(obj2==null && obj1!==null) return 1;
```
derrell, line 75: These will match any falsy value of obj1 and obj2, e.g., `false` or `0`,  with `==`. The part before `&&` on lines 74 and 75 must use `===` instead of `==`.

derrell: Similarly, in all cases below.

derrell (changes requested):
I'm not convinced we should make this change, even with the requested changes. This has the potential to affect many existing apps. OTOH, this can be a potentially breaking change for 6.0, if well documented in the release notes...

framework/source/class/qx/ui/table/model/Simple.js
adrelino: > These will match any falsy value of obj1 and obj2, e.g., `false` or `0`, with `==`. The part before `&&` on lines 74 and 75 must use `===` instead of `==`.

No, thats wrong.

obj1==null will only return true if obj1 is null or undefined, which is exactly the check I intended to use.

adrelino:
```js
console.log(null==0) // false
console.log(null==false) // false
console.log(null==null) // true
console.log(null==undefined) // true
```

johnspackman: @derrell is this really a breaking change?  it sounds like the table sort works fine except when nulls are involved, in which case the sort order is random(ish) and undesirable in all cases

derrell: @adrelino I stand corrected. Using `===` still seems prudent with falsy things, but I see that my statement was incorrect.

@johnspackman It's breaking only in that null was treated accidentally previously that is now explicit and possibly different.

I don't think it need be considered a breaking change.

derrell (approved):

hkollmann (approved):

level420 (approved):

oetiker (approved):

level420:
meging. Thank you @adrelino

PR #10059, opened 2020-11-17
Escape strings using codePointAt(), rather than charCodeAt()

rad-pat:
To fix an issue where emojis were not showing correctly in a table because the cell renderer was escaping the string value.
Quick demo here of the issue here https://tinyurl.com/yy5k255z

### DIFF
--- a/framework/source/class/qx/util/StringEscape.js
+++ b/framework/source/class/qx/util/StringEscape.js
@@ -40,7 +40,9 @@ qx.Bootstrap.define("qx.util.StringEscape",
       for (var i=0, l=str.length; i<l; i++)
       {
         var chr = str.charAt(i);
-        var code = chr.charCodeAt(0);
+        var code = str.codePointAt(i);
+
+        i += String.fromCodePoint(code).length-1;
 
         if (charCodeToEntities[code]) {
           entity = "&" + charCodeToEntities[code] + ";";
@@ -89,7 +91,7 @@ qx.Bootstrap.define("qx.util.StringEscape",
 
               // match hex number
               if (code.match(/^[0-9A-Fa-f]+$/gi)) {
-                chr = String.fromCharCode(parseInt(code, 16));
+                chr = String.fromCodePoint(parseInt(code, 16));
               }
             }
             else
@@ -98,7 +100,7 @@ qx.Bootstrap.define("qx.util.StringEscape",
 
               // match integer
               if (code.match(/^\d+$/gi)) {
-                chr = String.fromCharCode(parseInt(code, 10));
+                chr = String.fromCodePoint(parseInt(code, 10));
               }
             }
           }
--- a/framework/source/class/qx/test/lang/String.js
+++ b/framework/source/class/qx/test/lang/String.js
@@ -122,6 +122,8 @@ qx.Class.define("qx.test.lang.String",
 
       this.assertEquals("&lt;div id='1'&gt;&amp;nbsp; &euro;&lt;/div&gt;", qx.bom.String.escape("<div id='1'>&nbsp; €</div>"));
 
+      this.assertEquals("&#127774; 1", qx.bom.String.escape("🌞 1"));
+
       // textToHtml
       this.assertEquals("&lt;div id='1'&gt;<br> &nbsp;&amp;nbsp; &euro;&lt;/div&gt;", qx.bom.String.fromText("<div id='1'>\n  &nbsp; €</div>"));
 
@@ -134,6 +136,8 @@ qx.Class.define("qx.test.lang.String",
       this.assertEquals("juhu <>", qx.bom.String.unescape("juhu &lt;&gt;"));
 
       this.assertEquals("<div id='1'>&nbsp; €</div>", qx.bom.String.unescape("&lt;div id='1'&gt;&amp;nbsp; &euro;&lt;/div&gt;"));
+
+      this.assertEquals("🌞 1", qx.bom.String.unescape("&#127774; 1"));
 
       this.assertEquals(">&zzzz;x", qx.bom.String.unescape("&gt;&zzzz;x"));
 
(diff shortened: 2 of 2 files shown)


rad-pat:
Actually, I should probably add a test and also look at the unescape too. Any comments welcome though!

framework/source/class/qx/util/StringEscape.js
```diff
@@ -37,26 +37,21 @@ qx.Bootstrap.define("qx.util.StringEscape",
     {
       var entity, result = "";
 
-      for (var i=0, l=str.length; i<l; i++)
-      {
-        var chr = str.charAt(i);
-        var code = chr.charCodeAt(0);
+      for (var codePoint of str) {
```
cboulanger, line 40: Unfortunately, you need to use the old syntax for the time being, because we said we would keep supporting the Python generator for v6, and it doesn't understand ECMAScript 2015 features such as `for ... of`

framework/source/class/qx/util/StringEscape.js
```diff
@@ -37,26 +37,21 @@ qx.Bootstrap.define("qx.util.StringEscape",
...
+        var code = codePoint.codePointAt(0);
 
         if (charCodeToEntities[code]) {
           entity = "&" + charCodeToEntities[code] + ";";
-        }
-        else
-        {
+        } else {
```
cboulanger, line 45: Uh oh, @derrell will not like this :-)

derrell: lol. You're right... but this file is already a mess as far as mixed styles. I would prefer that it not be further mixed. In any case, though, formatting changes for the sake of formatting changes should not be included in a PR intending to fix a bug, but rather in its own PR, and in that latter PR, the format should be made consistent throughout the file.

framework/source/class/qx/util/StringEscape.js
```diff
@@ -37,26 +37,21 @@ qx.Bootstrap.define("qx.util.StringEscape",
     {
       var entity, result = "";
 
-      for (var i=0, l=str.length; i<l; i++)
-      {
-        var chr = str.charAt(i);
-        var code = chr.charCodeAt(0);
+      for (var codePoint of str) {
+        var code = codePoint.codePointAt(0);
```
derrell, line 41: I'm not very familiar with `codePointAt`. Is `codePointAt` guaranteed to yield the same result as `charAt` for characters other than the characters for which this PR is intended to correct? My concern is that this has an effect that is not backward compatible, on other, traditional characters that people might have in their programs.

derrell (changes requested):

rad-pat:
Revised to iterate through the string, adjusting the loop variable if we tackle a double-length Unicode char. Adjusts `unescape` to use fromCodePoint and adds tests for the offending string I was using.

framework/source/class/qx/util/StringEscape.js
rad-pat: I removed any style changes, just concentrating on fixing the issue

cboulanger (approved):

framework/source/class/qx/util/StringEscape.js
derrell: Thanks.

framework/source/class/qx/util/StringEscape.js
derrell: I found a polyfill for it at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt

cboulanger: But do we really need this polifyll? Since this is an API, not a  syntax thing? The generator doesnÄt have a problem, and the bowsers all support the API.

derrell: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt, IE does not support it, so there's a likelihood of BC problems. Am I mistaken?

cboulanger: I don't remember - do we still support IE for v6? But anyways, it's probably a quick fix to include the polyfill following the pattern in
https://github.com/qooxdoo/qooxdoo/blob/master/framework/source/class/qx/bom/client/EcmaScript.js and  https://github.com/qooxdoo/qooxdoo/tree/master/framework/source/class/qx/lang/normalize - although we don't have a dedicated polyfill class for the global object, do we? If we were to add it, it would have to be added as `qx.lang.normalize.GlobalObject` and here: https://github.com/qooxdoo/qooxdoo/blob/97418ba9a62fed5eb00efbbac5190fdde66ed24b/framework/source/class/qx/module/Polyfill.js . Is this correct, @derrell ?

cboulanger: Forget my last comment about the global object, we are talking about a method of the `String` object, so no new class is needed. Sorry for the confusion.

hkollmann: We should polyfill it.

level420:
@rad-pat will create a PR soon with the codeAtPoint polyfill.

framework/source/class/qx/util/StringEscape.js
johnspackman: > I don't remember - do we still support IE for v6?

Yes!  There are loads of users with IE11, and I think that officially it's not EOL for a while yet

oetiker (approved):

derrell (changes requested):
Approved pending merge of #10060

hkollmann (approved):

derrell (approved):
#10060 is now merged. Making my approval official.

level420 (approved):

rad-pat:
Rebased against master

level420:
merging. thank you @rad-pat

PR #10060, opened 2020-11-18
polyfill for String.prototype.codePointAt and String.fromCodePoint

level420:
according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint

Includes tests. Note that currently travis CI does not run any Internet Explorer tests, so the this polyfill is not tested there.

But I've run the testrunner manually for IE 11, Firefox, Chrome and current Edge on Win 10, Safari on Big Sur, both for the source and build version built with the python generator.

This PR is needed for #10059.


rad-pat:
@level420 , #10059 also requires the String.fromCodePoint() polyfill please.

oetiker (approved):

level420:
@rad-pat oops! yes. I've missed that. I'll also add that polyfill.

derrell (approved):

cboulanger (approved):

johnspackman (approved):

hkollmann (approved):

PR #10063, opened 2020-11-24
TreeVirtual - fixes bug where selection would not extend with shift+Up/Down

rad-pat:
Bug can be seen in old 5.02 Demo Browser
https://archive.qooxdoo.org/current/demobrowser/#treevirtual~TreeVirtual_Selections.html

Holding down shift and trying to selection multiple rows with Up/Down keys would only result in at most two rows selected. The selection could be obtained using the mouse but this appears to fix the issue which was that the anchorSelectionIndex of the selection model was modified by calling the private method `_addSelectionInterval()` .  The private method appears to be called to prevent the changeSelection event.


oetiker (approved):
neat

hkollmann (approved):
Is the same in current demo browser: http://qooxdoo.org/qxl.demobrowser/#treevirtual~TreeVirtual_Selections.html

cboulanger (approved):

johnspackman (approved):

level420 (approved):

level420:
Thank you @rad-pat

Issue #10064, opened 2020-11-24
TreeVirtual resets selection upon setData()


rad-pat:
There's likely a couple of things here, but they're related so we'll see how the discussion goes. 

In the SimpleTreeDataModel, in the `setValue()` method, there is a call to [`setData()`](https://github.com/qooxdoo/qooxdoo/blob/9cd0b1d89f14f3060b07d0192ed177abc78ae86f/framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js#L393) which basically rebuilds the data from the ground up and once complete then attempts to re-select what was previously selected but does not succeed. The failure to reselect can be seen in this [Demo](https://tinyurl.com/y5ntk3pf) if you select multiple rows and hit the refresh button. 

Removing the `setData()` and adding an update to the private `_rowArr` in `setValue()` seems like it should be enough to satisfy that the value has changed.  
I wondered if you knew why the `setData()` call might have been necessary in this position? 

The reason for asking the question is that we have the need to select a range of rows, edit a cell and then propagate the changed value across the selected rows, something we already do for tables. The presence of the `setData()` is resetting the selection when we handle the `dataEdited` event.  
In the meanwhile I have a custom dataModel that omits the `setData()` which seems to be working OK at the moment.

derrell:
That's probably my code (from ancient history). I can try to figure out why it does it the way it does. It might be easiest for me to review if you were to PR your changed version. Then I can try to see exactly what `setData` does that your code does not, and ensure that nothing else is likely broken by it. Are you able to PR that change?

rad-pat:
Yeah, I can make a PR definitely, I don't currently have a fix for the selections being reset upon `setData`,  but I do for making the `setValue` not call `setData`.

rad-pat:
Fixed by #10065

PR #10065, opened 2020-11-24
TreeVirtual - Prevent call to setData upon setValue()

rad-pat:
* Populate the private _rowArr

As mentioned in https://github.com/qooxdoo/qooxdoo/issues/10064, a call to set a value in the model currently causes a full rebuild of the tree which seems a bit heavy handed. This PR attempts to remove that full refresh and just populate  the required internals.

### DIFF
--- a/framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
+++ b/framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
@@ -389,8 +389,8 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel",
           return;
         }
         node.columnData[columnIndex] = value;
+        this._rowArr[rowIndex][columnIndex] = value;
       }
-      this.setData();
       // Inform the listeners
       if (this.hasListener("dataChanged"))
       {


framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -389,8 +389,8 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel",
           return;
         }
         node.columnData[columnIndex] = value;
+        this._rowArr[rowIndex][columnIndex] = value;
       }
-      this.setData();
```
derrell, line 393: So I think this change is valid except for one thing. The call to `setData` would also cause this code to be run:

```
        if (columnIndex == this.__sortColumnIndex) {
          this.clearSorting();
        }
```
which generates a `metaDataChanged` event which causes all content to be updated. I believe that that includes sorting. So with this change, it appears that if the column that the data being changed is in is the sorted column of the table, the table won't be resorted. Can you confirm?

rad-pat: I can't see how `setData()` would cause that code to run - the code you've highlighted is in the `setValue` of `qx.ui.table.model.Simple` I think, but we are overriding with the `setValue` in `qx.ui.treevirtual.SimpleTreeDataModel`?
Also, probably sorting of columns should not be allowed in a TreeVirtual with columns, the tree would not look very pretty 😄

derrell: Doh! Yeah, of course you're right about sorting a tree based on other columns.

The way it does what I said was, as I traced through it, because `setData` via `clearSorting` generated that `metaDataChanged` which called into Scroller and then each Pane. 

But it's all moot since sorting a tree is bogus. I *think* this change is safe. It's risky, in that it can affect a great number of applications. Since the release isn't imminent, I think we can probably go with it, and see if anyone encounters problems. I'm a bit scared, though...

derrell (approved):

cboulanger (approved):

framework/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
rad-pat: I still don't see it. The `setData` method is also overridden by the SimpleTreeDataModel wherein it doesn't use the `clearSorting` flag.  I'm happy to wait a while to merge this whilst I perform more testing. We have heavy use of TreeVirtualTable so hopefully would pick up any problems.

oetiker (approved):

level420 (approved):

level420:
@rad-pat please note that your branch is out of date. please re-integrate the changes in master

hkollmann (approved):

level420:
thank you @rad-pat

PR #10069, opened 2020-12-03
adds support for declaring environment checks

johnspackman:
adds support for declaring environment checks, to allow the compiler to know which class implements the checks identified with short codes - eg calling `qx.core.Environment.get("browser.name")` means that the class `qx.bom.client.Browser` has to be first, so that the implementation is in place.

There is a corresponding PR https://github.com/qooxdoo/qooxdoo-compiler/pull/775 in the compiler which needs to go out at the same time

### DIFF
--- a/framework/Manifest.json
+++ b/framework/Manifest.json
@@ -128,7 +128,35 @@
           "qx/iconfont/MaterialIcons/materialiconstwotone-v17.otf"
         ]
       }
-    ]
+    ],
+    "environmentChecks": {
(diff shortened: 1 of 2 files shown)


cboulanger (approved):

level420 (approved):

oetiker (approved):

derrell (approved):

hkollmann (approved):

johnspackman:
this should be safe to merge now that the compiler's latest release is live

PR #10073, opened 2021-01-05
Fix mixin this base support

johnspackman:
Adds a method which allows the compiler to emit code that handles `this.base` when outputting methods which are in a mixin.

This should have no effect in apps written for the generator.

It depends on a new environment setting `qx.compiler` which will be true if the compiler created your application (this will not be output until the next release of the compiler).

### DIFF
--- a/framework/source/class/qx/Class.js
+++ b/framework/source/class/qx/Class.js
@@ -1450,12 +1450,12 @@ qx.Bootstrap.define("qx.Class",
         // classes which are function like as well.
         if (base !== false && member instanceof Function && member.$$type == null)
         {
-          if (wrap == true)
+          if (!qx.core.Environment.get("qx.compiler") && wrap == true)
           {
             // wrap "patched" mixin member
             member = this.__mixinMemberWrapper(member, proto[key]);
           }
-          else
+          else if (wrap != true)
           {
             // Configure extend (named base here)
             // Hint: proto[key] is not yet overwritten here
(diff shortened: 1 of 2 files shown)


derrell (changes requested):
What problem is this solving? Does it change behavior of `this.base` calls in existing code? Will the effect be different in code created by the generator than code created by the compiler? (Just trying to get my head around the impetus for this PR.)

cboulanger:
@derrell The reason for the PR is given here: https://github.com/qooxdoo/qooxdoo/pull/10073/files#diff-ef036b269f9b963ea21c749b554258c32ff5856d4f51190e51db9455956c8f3cR308

derrell:
Yes I read that. That explains what it's doing, but not the answers to my other questions.

oetiker (approved):

johnspackman:
Without this patch you cannot call `this.base` in Mixin methods (although I think you may have been able to do with the generator in source targets but not build targets ... but don't quote me on that); the difference is that the compiler will always output code to call an explicit member, and will *never* use the implementation of `this.base` because `this.base` depends on using `arguments.caller/callee` which is disallowed by the browser in ES6 and onwards.

IE the compiler *must* know at compile time what method to call in exchange for `this.base`.

The problem is that when generating the code for a mixin at compile time, it is not possible to determine what method should be called for `this.base` because it depends entirely on what class that mixin has been included in.  Calls to `this.base` in mixin methods will crash.

This should not affect backwards compatibility, or the generator (hence the `qx.core.Environment` detection); this either fixes a bug, or adds a feature ... but if it adds a feature, it only adds the feature to code compiled with the compiler.

level420 (approved):

cboulanger (approved):

cboulanger:
This reminds me of this issue: https://github.com/qooxdoo/qooxdoo-compiler/issues/653 ...

derrell (approved):
Thanks for the response. I thought that in older code, `this.base` calls already worked and was used.

hkollmann (approved):

johnspackman:
@derrell 
> Thanks for the response. I thought that in older code, `this.base` calls already worked and was used.

I think I tried it once many years ago and couldnt get it to work with the generator; cam across it the other day with the compiler and decided to fix it instead of come up with a hacky solution in my code 😁 

@cboulanger 
> This reminds me of this issue: qooxdoo/qooxdoo-compiler#653 ...
Ah yes, still disappoointed I havn't done that yet :)  It has slightly different semantics to `this.base` but would be a much nicer solution...

PR #10074, opened 2021-01-08
escape form values to avoid code injection 

peuter:
by entering content like ``. to a form field which is using e.g. an email validator. That validator creates an invalid message that is usually shown by a tooltip. If you do not escape HTML code from form values they are interpreted as HTML code in the tooltip.


johnspackman (approved):

oetiker (approved):

cboulanger (approved):

hkollmann (approved):

hkollmann:
Please fix the tests too. Many thanks.

level420 (approved):

level420:
thank you @peuter

PR #10076, opened 2021-01-08
adds a `readOnly` property to List

johnspackman:
this is useful when the list contains selection that must not change, but which has to be scrollable (similar to `TextArea.readOnly`)


derrell (approved):
Big change, but looks fully BC.

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

level420:
Thank you @johnspackman

PR #10077, opened 2021-01-08
Add incremental search to VirtualSelectBox

zaucker:
- Enable incremental search on the list items by `selectBox.setIncrementalSearch(true)`.
- Highlighting can be customized by setting property `highlightStyle` to a style map (defaults to red and bold).
Screenshot 2021-01-08 at 17 32 49

### DIFF
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -480,7 +480,6 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
     __lastMatch : '',
     __filterUpdateRunning : 0,
     __filterInput : null,
-    __highlightStyle : null,
     __lastRich : false,
 
 
@@ -519,26 +518,40 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
       }, this);
     },
 
-    __applyHighlightStyle : function(value, old) {
-      var styles = [];
-      if (value != null) {
-        var keys = Object.keys(value);
-        var k;
-        for (k=0; k< keys.length; k++) {
+    __getHighlightStyle : function() {
+      var highlightAppearance =
+        qx.theme.manager.Appearance.getInstance().styleFrom("list-search-highlight");
+          
+      var highlightStyle = '', styles = [];
+      if (highlightAppearance != null) {
+        var keys = Object.keys(highLightAppearance);
+        for (var k=0; k< keys.length; k++) {
           var key = keys[k].replace(/[A-Z]/g, m => "-" + m.toLowerCase());
           styles.push(key + ':' + value[keys[k]]);
         }
-        this.__highlightStyle = styles.join(';');
+        highlightStyle = styles.join(';');
       }
-      else {
-        this.__highlightStyle = '';
-      }
+      return highlightStyle;
     },
 
     _highlightFilterValue : function (item)
     {
-      return this.__highlightStyle ? item.replace(item, '<span style="' + this.__highlightStyle + '">' + qx.module.util.String.escapeHtml(this.__filterValue) + '</span>')
-                                   : item;
+      var highlightStyle = this.__getHighlightStyle();
+      return (
+        highlightStyle
+          ? item.replace(
+            item,
+            [
+              '<span style="',
+              highlightStyle,
+              '">',
+              this.__lastRich
+                ? qx.module.util.String.escapeHtml(this.__filterValue)
+                : this.__filterValue,
+              '</span>'
+            ].join(""))
+          : item
+      );
     },
 
     __updateDelegate : function(lastFilterValue)
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -472,7 +472,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
 
     /*
     ---------------------------------------------------------------------------
-      FILTERING
+      INCREMENTAL SEARCH
     ---------------------------------------------------------------------------
     */
 
@@ -608,8 +608,8 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
         this.__initIncrementalSearch();
       }
       else {
-        this.__searchTimer.setEnabled(false);
         this.getChildControl('atom').setRich(this.__lastRich);
+        this.__searchTimer.setEnabled(true);
       }
     }
   },
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -81,16 +81,6 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
       check : "Boolean"
     },
 
-    highlightStyle :
-    {
-        nullable : true,
-        apply : '__applyHighlightStyle',
-        init : {
-            color      : '#FF0000',
-            fontWeight : 'bold'
-        }
-    },
-
     /** Current selected items. */
     selection :
     {
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -83,7 +83,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
      */
     incrementalSearch :
     {
-      apply : '__applyIncrementalSearch',
+      apply : '_applyIncrementalSearch',
       init : false,
       check : "Boolean"
     },
@@ -610,7 +610,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
       this.setDelegate(delegate);
     },
 
-    __applyIncrementalSearch : function(value, old)
+    _applyIncrementalSearch : function(value, old)
     {
       if (value) {
         this.__searchTimer.stop();
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -537,7 +537,8 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
 
     _highlightFilterValue : function (item)
     {
-      return this.__highlightStyle ? item.replace(item, '<span style="' + this.__highlightStyle + '">' + this.__filterValue+'</span>') : item;
+      return this.__highlightStyle ? item.replace(item, '<span style="' + this.__highlightStyle + '">' + qx.module.util.String.escapeHtml(this.__filterValue) + '</span>')
+                                   : item;
     },
 
     __updateDelegate : function(lastFilterValue)
--- a/framework/source/class/qx/ui/form/VirtualSelectBox.js
+++ b/framework/source/class/qx/ui/form/VirtualSelectBox.js
@@ -482,8 +482,7 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
     __filterInput : null,
     __lastRich : false,
 
-
-     __addFilterInput : function()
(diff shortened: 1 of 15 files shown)


cboulanger:
Cool! Would you add a demo for the widget browser? 

https://github.com/qooxdoo/qxl.demobrowser/blob/master/source/class/qxl/demobrowser/demo/virtual/SelectBox.js

cboulanger (approved):

zaucker:
> Cool! Would you add a demo for the widget browser?
> 
> https://github.com/qooxdoo/qxl.demobrowser/blob/master/source/class/qxl/demobrowser/demo/virtual/SelectBox.js

Will do.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -445,6 +467,150 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
+              e.preventDefault();
+              break;
+          }
+      }, this);
+    },
+
+    __applyMarkHighlightStyle : function(value, old) {
+      var styles = [];
+      if (value != null) {
+        var keys = Object.keys(value);
```
derrell, line 525: What do you think about adding the option for highlight style to be a string that maps to an appearance in the theme, since this is a fully themeable widget?

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -445,6 +467,150 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
+      this.__filterUpdateRunning++;
+      var filterValue = (lastFilterValue !== undefined) ? lastFilterValue : this.__filterInput.getValue();
+      this.__filterValue = filterValue;
+      // create and apply new filter
+      var delegate = {
+          filter : function(item) {
+              return item.match(filterValue);
+          },
+          configureItem : function(item) {
+              item.setRich(true);
```
derrell, line 554: Does this potentially allow for unanticipated injection of evil code? I understand that `rich` is a prerequisite to attaining this behavior, but wondering if we need a warning someplace that it will be automatically made `rich`? Since the search is always from the beginning of the list item value, maybe highlighting should be an option since the same early text will be highlighted in every row so is pretty but not absolutely necessary?

derrell (changes requested):
Very cool feature! I just added a potential security issue question that should probably be considered before merging.

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: That would make sense. Could you point me to an example?

oetiker: since the highlight is created using a span in html with the 'rich' property set on the value ... this does not mesh well with qooxdoo theming

framework/source/class/qx/ui/form/VirtualSelectBox.js
oetiker: true, the input would have to be escaped html escaped ...

zaucker: > Does this potentially allow for unanticipated injection of evil code? I understand that `rich` is a prerequisite to attaining this behavior, but wondering if we need a warning someplace that it will be automatically made `rich`? 

> Since the search is always from the beginning of the list item value, maybe highlighting should be an option since the same early text will be highlighted in every row so is pretty but not absolutely necessary? And if we don't highlight, then `rich` isn't necessary.
The search is intentionally **not** required to be from the beginning of the string. My use case is a select box containing addresses as items and always searching from the beginning is not very useful, especially of the item starts with a zip code or a town, because at least in our customer's application the street is much more likely to be "relatively" unique.
And thus the highlighting is not just pretty but necessary for the user to recognize why an item matches.
Screenshot 2021-01-08 at 20 27 25

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: Ah. yes, we are only highlighting part of the label string of the list item.

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: > true, the input would have to be escaped html escaped ...

Well, the only HTML that could be entered is what's already there in the item labels. Is this really an issue?

zaucker: But I added `escapeHtml()` around the filter value.

hkollmann (approved):

johnspackman (approved):

framework/source/class/qx/ui/form/VirtualSelectBox.js
derrell: The way to pull this from the theme is as so:
```
qx.theme.manager.Appearance.getInstance().styleFrom("list-search-highlight")
```

derrell: (And then we should add it to the few internal themes we ship with)

derrell: As stated in the "conversation", you can do it like this:

```
qx.theme.manager.Appearance.getInstance().styleFrom("list-search-highlight")
```
and then add "list-search-highlight" to the few shipped themes.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -445,6 +467,151 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
+        this.__highlightStyle = styles.join(';');
+      }
+      else {
+        this.__highlightStyle = '';
+      }
+    },
+
+    _highlightFilterValue : function (item)
+    {
+      return this.__highlightStyle ? item.replace(item, '<span style="' + this.__highlightStyle + '">' + qx.module.util.String.escapeHtml(this.__filterValue) + '</span>')
```
derrell, line 540: This is good. It solves the problem... except in the case where the user had turned on `rich` intentionally, in which case we shouldn't escape it. So something like this, maybe:

```
  return (
    this.__highlightStyle
      ? item.replace(
          item,
          [
            '<span style="',
            this.__highlightStyle,
            '">',
            this.__lastRich
              ? qx.module.util.String.escapeHtml(this.__filterValue)
              : this.__filterValue,
            '</span>'
          ].join(""))
      : item);
```

derrell (changes requested):

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: Done.

zaucker: Get highlight style via theming appearance and added the style to all the themes.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -519,26 +517,40 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
-    __applyHighlightStyle : function(value, old) {
-      var styles = [];
-      if (value != null) {
-        var keys = Object.keys(value);
-        var k;
-        for (k=0; k< keys.length; k++) {
+    __getHighlightStyle : function() {
+      var highlightAppearance =
+        qx.theme.manager.Appearance.getInstance().styleFrom("list-search-highlight");
+      console.log('__getHighlightStyle=', highlightAppearance);
```
cboulanger, line 523: Is this still there? Should be removed before merging.

derrell (approved):
Very cool feature! Thanks!

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: Thanks for the catch. Removed.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -70,6 +74,22 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
+
+    highlightStyle :
+    {
+        nullable : true,
+        apply : '__applyHighlightStyle',
+        init : {
+            color      : '#FF0000',
+            fontWeight : 'bold'
+        }
+    },
```
oetiker, line 92: So this does not seem to get used anymore ?

zaucker: Indeed, thanks. Removed.

framework/source/class/qx/ui/form/VirtualSelectBox.js
oetiker: the problem is that the list labels are normally not 'rich' ... escapeHtml should take care of the problem

framework/source/class/qx/ui/form/VirtualSelectBox.js
oetiker: hmm if there is 'evil' html outside the highlighted part, it will not get escaped I think

zaucker:
> > Cool! Would you add a demo for the widget browser?
> > https://github.com/qooxdoo/qxl.demobrowser/blob/master/source/class/qxl/demobrowser/demo/virtual/SelectBox.js
> 
> Will do.

https://github.com/qooxdoo/qxl.demobrowser/pull/12

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: > the problem is that the list labels are normally not 'rich' ... escapeHtml should take care of the problem

That would break the use of HTML formatted item labels together with incremental search.

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: > hmm if there is 'evil' html outside the highlighted part, it will not get escaped I think

Yes, as it wouldn't if the programmer has set them to rich himself (which is done exactly like here in one of the demobrowser examples) and doesn't take care what content he has in the model.

zaucker:
> Shouldn't an `apply` method be protected rather than private? Since an derived class might want to modify it...

Done.

hkollmann:
some open linting issues:

Warning: qx.ui.form.VirtualSelectBox (594, 7): Using an undeclared private class feature: '__highlightStyle'
  - Warning: qx.ui.form.VirtualSelectBox (535, 28): Using an undeclared private class feature: '__highlightStyle'
  - Warning: qx.ui.form.VirtualSelectBox (500, 49): Loop or condition statement without a block as body
  - Warning: qx.ui.form.VirtualSelectBox (536, 29): Loop or condition statement without a block as body

zaucker:
> some open linting issues:
> 
> Warning: qx.ui.form.VirtualSelectBox (594, 7): Using an undeclared private class feature: '__highlightStyle'
> 
> * Warning: qx.ui.form.VirtualSelectBox (535, 28): Using an undeclared private class feature: '__highlightStyle'
> * Warning: qx.ui.form.VirtualSelectBox (500, 49): Loop or condition statement without a block as body
> * Warning: qx.ui.form.VirtualSelectBox (536, 29): Loop or condition statement without a block as body

Fixed.

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: Fixed, I think.

level420 (approved):
Nice feature! Thank you @zaucker !

zaucker:
> Nice feature! Thank you @zaucker !

Thanks!

zaucker:
Just for documentation the latest theming:
Screenshot 2021-01-11 at 15 28 24

PR #10079, opened 2021-01-09
another needed require for deployment

hkollmann:


cboulanger (approved):

derrell (approved):

johnspackman:
I understand that the hotfix label allowed it to be merged with only 2 approvals, but IMHO this should not have been merged, it is the wrong solution.  AFAICT the issue is that the `qx.core.Environment` detector is not being loaded, but the requirement should be auto-detected by the compiler.  

As an aside, there are a number of classes which have `@use` or @require` which (Im guessing) were used to trick the generator into having a suitable side effect with regards to load order; as a rule, explicit load order manipulation should not be required and are problematic long term.

hkollmann:
Problem is that qx.io.remote.RequestQueue uses 

 maxConcurrentRequests :
    {
      check : "Integer",
      init : qx.core.Environment.get("io.maxrequests")
    },

the io.maxrequests entry is added to the enviroment in qx.bom.client.Transport - this is not detected by the compiler.

Another solution could be that the compiler loads all classes first wich defines enviroment entries.

But this would be your turn.

johnspackman:
Yes, that's my point - it is a bug in the compiler, not the framework (or at least, not in this class).  

This PR means that the bug  is hidden and will reoccur in other places, and as the original error message is `Error: Invalid key "init" of property "locale" in class "qx.locale.Manager"! The value is undefined: undefined` there is no obvious correlation between the error message and the solution or work around.

This in turn means that anyone can hit a blocker which can only be discovered by in-depth debugging, and they would probably have to be able to draw on quite a bit of Qooxdoo knowledge to successfully debug it and identify the work around.

The other problem is that once these manual overrides are added they tend to never get removed - when writing the dependency analysis part of the compiler, I discovered that there are lots of classes which have explicit dependencies added which are not necessary.  My guess is that they were introduced to work around similar (but different) problems with the generator, and AFAICR at the time I thought that the generator *could* handle it automatically, suggesting that it was to work around an early version of the generator.

hkollmann:
OK, what to do? 
Can you fix the compiler and i will revert this and https://github.com/qooxdoo/qooxdoo/pull/10078/commits/c57fcc989cfba7564a2cc41fbb8034eadbaadc80 ?

johnspackman:
I've been trying to reproduce it, but can't get the API viewer to compile - is there something missing from the repo?  Lets chat in gitter?

johnspackman:
My understanding is that this was reproducible with the API Viewer?  If that was correct, then this problem is not reproducible using a virgin user account on a linux VM.  I've used a fresh clone of the deployment repo in a brand new user account, and been able to edit the checked out `qooxdoo` repo to undo the change in this PR and rebuild and serve API viewer.

However, it wasn't quite as simple as that because I have also had to go through more dependency hell type problems to get to the point where I have a working, consistent compilation environment.  The issue is that, again, the framework and compiler versions get out of sync, partly because we're relying on npm to join the dots but npm isn't as automatic as we assume.

I'll make some pushes/PRs with my changes to the deployment and compiler repos this morning, but I think that this PR should be reverted as soon as that's done and a successful deployment has run (this PR won't do any harm in the mean time).

PR #10080, opened 2021-01-12
Make incremental search work with complex model

zaucker:
- Make models like `[ { key : 0, title : 'Elm street' } ]` work (will add example to demo browser)
- Make search case insensitive by using `matcH(RegExp)` instead of `split()`
- Handle edge cases properly


oetiker (approved):

cboulanger (approved):

derrell:
I am still bothered by a nagging concern about `rich` being set. If a legacy app using this class does not request `rich` for its list items, and does not turn on the new search capability, do its list items suddenly become `rich`?

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -591,6 +599,15 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
       this.__filterUpdateRunning--;
     },
 
+    _applyDelegate : function(value, old) {
+      if (old && old.configureItem && !value.configureItem) {
+        value.configureItem = function(item) {
+          item.setRich(true);
+        };
+      }
```
oetiker, line 607: this should only happen if incremental search is active. Is it actually necessary? does setDelegate override existing properties ?

oetiker: I think this is not neccesary in this way, as delegates get applied when they are configured ... configuring a new/different delegate does not override the existing ones ... the code for the handling of delegates is in the controller, which seems a bit odd as this means that `setDelegate` could act differently depending on where it is implemented ... but at least in the listController which I think is in us here, multiple invocations with different properties do not seem to interact ... 

https://github.com/qooxdoo/qooxdoo/blob/9e375001ea22a9f4a1ea79c9df353c8f9d5df177/framework/source/class/qx/data/controller/List.js#L281

rad-pat:
On updating Qooxdoo, I'm getting the following missing theme definition error which seems related to the previous PR for incremental search. Any chance you could include a fix for it within this PR?   `qx.theme.tangible.Appearance: [2661,2 to 2668,3] error: Unexpected property list-search-highlight in theme definition`

zaucker:
> On updating Qooxdoo, I'm getting the following missing theme definition error which seems related to the previous PR for incremental search. Any chance you could include a fix for it within this PR? `qx.theme.tangible.Appearance: [2661,2 to 2668,3] error: Unexpected property list-search-highlight in theme definition`

I'll look into it. Something is rather strange right now.

level420 (approved):

johnspackman (approved):

zaucker:
Converted to DRAFT so that it is not accidentally being merged. I hope I can finish it this weekend.

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: I think `delegate.configureItem` must handle `rich` explicitly when changing `delegate.filter` as items are dynamically generated.

zaucker:
> > On updating Qooxdoo, I'm getting the following missing theme definition error which seems related to the previous PR for incremental search. Any chance you could include a fix for it within this PR? `qx.theme.tangible.Appearance: [2661,2 to 2668,3] error: Unexpected property list-search-highlight in theme definition`
> 
> I'll look into it. Something is rather strange right now.

Fixed with previous PR.

zaucker:
> I am still bothered by a nagging concern about `rich` being set. If a legacy app using this class does not request `rich` for its list items, and does not turn on the new search capability, do its list items suddenly become `rich`?

- I added a `rich` property to the widget itself. Upon setting it to `true` both the widgets `atom` childcontrol and the delegate are set to `rich=true`. Thus this is completely under control of the user.
- There is now a none-rich highlighting as well:
Screenshot 2021-01-23 at 21 11 37

zaucker:
I added a default rich highlighting style in case the user is using a theme that does not set the highlighting style.

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -593,26 +623,49 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
       this.setLabelOptions(labelOptions);
-        var delegate = {
-          configureItem : function(item) {
-            item.setRich(true);
-          }
-      };
-      this.setDelegate(delegate);
+      this.setDelegate({
+        configureItem : function(item) {
+          item.setRich(true);
```
derrell, line 643: The item is set to rich below, in `_applyDelegate` and `_applyRich` as expected. I suspect this is old code that just didn't get deleted, since it sets the item's `rich` property to true regardless of whether the widtget's `rich` property is true.

derrell (changes requested):

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: You are right, this must be conditional. It has to be set here in case `setRich()` is called before `setIncrementalSearch()`.
Fixed.

derrell (changes requested):
Nice!

derrell (approved):

framework/source/class/qx/ui/form/VirtualSelectBox.js
```diff
@@ -521,53 +529,75 @@ qx.Class.define("qx.ui.form.VirtualSelectBox",
...
-      var parts = item.split(filterValue);
-      if (!this.__lastRich) {
-        parts[0] = qx.module.util.String.escapeHtml(parts[0]);
-        parts[1] = qx.module.util.String.escapeHtml(parts[1]);
+        var parts = item.match(new RegExp('(.*)' +  qx.module.util.String.escapeRegexpChars(filterValue) + '(.*)', 'i'));
+      parts[1] = parts[1] ? parts[1] : '';
+      parts[2] = parts[2] ? parts[2] : '';
+      var label;
+      if (highlightStyle) {
+        filterValue = qx.module.util.String.escapeHtml(this.__filterValue);
```
oetiker, line 561: If you asked to have the widget set to 'rich' above, then why to escape the string? Seems sort of the wrong way round. If rich is set, you have to assume that there is html in the labels, so you would want to refrain from escaping and using html of your own. in that case the `|` marker would make sense. I would suggest to use `*` though as this is generally used for emphasis im markdown.

If rich is not set though, you could use it for your own purposes, set it and then use html for highlightning ...

oetiker (changes requested):

framework/source/class/qx/ui/form/VirtualSelectBox.js
zaucker: - `__filterValue` is user input, thus I think it must be escaped if `rich=true` as it is going to be part of the output to the browser. I guess we could skip it if `rich=false`, but I think it doesn't matter in this case.
- The current implementation assumes that the visible label string itself doesn't contain HTML (it can be inside some HTML tags however). I guess one could try to strip the HTML tags of in both `__filterValue` and the label. I wonder if it is worthwhile (if at all reliable to implement).
I guess we can discuss this in person tomorrow.

cboulanger (approved):

hkollmann (approved):

oetiker:
@zaucker this is not in the final form yet, right  ?

zaucker:
> @zaucker this is not in the final form yet, right ?

Just trying to make it work.

oetiker (approved):
and we got a winner!!!!

hkollmann:
Lint failed. let is not allowed in v6.!

hkollmann (changes requested):
please make it v6 compliant! Only old syntax is allowed.

zaucker:
> please make it v6 compliant! Only old syntax is allowed.

Fixed.

hkollmann:
Linter is unhappy:

- Warning: qx.ui.form.VirtualSelectBox (770, 7): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (778, 38): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (774, 9): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (778, 9): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (627, 14): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (629, 14): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (756, 9): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (760, 9): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (639, 14): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (641, 14): Using an undeclared private class feature: '__highlightMarkers'
  - Warning: qx.ui.form.VirtualSelectBox (277, 11): Declared but unused variable or parameter: 'listIsOpen'

zaucker:
> Linter is unhappy:

Issues fixed.

How can I run the linter on my qooxdoo checkout?

hkollmann (approved):

hkollmann:
Coverage travis job  is broken. Therefore i just merge it.

hkollmann:
Thanks @zaucker !

PR #10085, opened 2021-01-16
Make cookies use SameSite=Strict by default

### LINKED ISSUE #10050: Qooxdoo's cookie handling needs to be updated to use modern standards
Firefox emits "Cookie 'cookie name' will be soon treated as cross-site cookie against 'file' because the scheme does not match" in the console. Other browser vendors might follow.

We need to update [`qx.bom.Cookie`](https://qooxdoo.org/qxl.apiviewer/#qx.bom.Cookie) in order to avoid to break our apps in Firefox (and elsewhere) as soon as this policy is implemented.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/10050


cboulanger (approved):

derrell (changes requested):
LGTM, but let's also add this to the non-BC list for release notes, per recent discussion, as part of this PR

derrell (approved):
Thanks!

hkollmann (approved):

level420 (approved):

oetiker (approved):

cboulanger:
Thank you @johnspackman !

PR #10086, opened 2021-01-17
Reorganize test directory

cboulanger:
As proposed on Gitter, this PR renames the "uitests" directory to "tests" and moves the individual test apps into an "apps" subdirectory. In addition, it adds a "_framework" app, which is a codeless library that runs the framework tests. 

The aim is to have a script which will run all applications in sequence with `qx test`, either with playwright if they are browser-based, or any other test mechanism defined in the application. 

None of this really works yet, this is just the first step towards having a comprehensive test suite for the framework which can be run manually by developers or automatically by a GitHub action.


derrell (approved):

hkollmann (approved):

johnspackman:
is there a reason why the `_framework` directory has a leading underscore?

cboulanger:
The only reason is that it shows up at the top of the list - when browsing GitHub or in the IDE, and for programmatic directory listings when running the meta-test script, so that the framework tests will be executed first. Of course, the order also could be also hard-coded in the script, but the underscore serves both visibility and ordering. But I am open for other solutions...

johnspackman:
I'd rather stay away from those kind of things because they run out of steam; I tend to use the command line a lot, and from that perspective underscores are more of a hindrance than a help.  When it comes to ordering, I could put together a script that prioritises `framework` but takes the rest of the directory (ie means that it is not necessary to hard code the entire list)?

cboulanger:
@johnspackman I was thinking to solve the visibility and priority issue by elevating the "framework" test app one level app and place it next to the "apps" dir. The test script will then run it seperately before iterating through the "apps" dirs. Better?

johnspackman:
👍

cboulanger:
@johnspackman Done. Can you have a look at the test apps after this has been merged (or take over the PR)? I wasn't able to get "drapndrop" to compile.

oetiker (approved):

johnspackman (approved):

level420 (approved):

PR #10088, opened 2021-01-18
Editing the value and pressing up/down keys updates the old value

lucapivato:
If you edit the value in the spinner and use the keyboard up/down keys it reverts to the previous value and increases or decreases the old value.

https://tinyurl.com/y5quor5f

Type 15 and press up, it shows 1.


lucapivato:
A better fix would be to remove he value property from spinnner.js and instead add getValue()/setValue() methods so that calling getValue() can read the value from "textfield" and always operate on the value in the input field. Thoughts?

lucapivato:
Or change _createChildControlImpl

        case "textfield":
          control.addListener("changeValue", this._onTextChange, this);
          to: control.addListener("input", this._onTextChange, this);

Maybe this is a cleaner fix.

cboulanger:
I cannot reproduce this on Firefox 85 / Windows 10 nor on Safari on Mac (Big Sur) - I am always getting "16" - what browsers do you see this in?

hkollmann:
It's working with edge too.

lucapivato:
@cboulanger @hkollmann did you try the playground sample?

It's not browser related since the code to update the value when editing is not there.

See video:

https://drive.google.com/file/d/1hqqvk9nCj-O9DecO8KwfjZTrP3BJwpLK/view

Tried with Firefox and Chrome.

Incidentally I just noticed that qooxdoo 6.0.0 doesn't load in IE:

image

hkollmann:
Just used the link above with my edge.

lucapivato:
@hkollmann Just did it again with Edge. Type 12 press arrow up without leaving the focus and the result is 2.

hkollmann:
OK, can reproduce it. I used the mouse not keyboard

derrell:
The key to replicating this issue is to type digits into the text field, and then without additional typing and without clicking anyplace else before hand, click on the up arrow. The newly-typed value has not yet been updated in the `value` yet, so presumably the up arrow takes the existing `value` and adds one to it, replacing the existing text.

hkollmann (approved):

derrell (commented):
@lucapivato Is your edit in https://github.com/qooxdoo/qooxdoo/pull/10088#issuecomment-762409142 saying that the change to use "input" instead of "changeValue" caused a mess; or that the original proposal in this PR also isn't working correctly?

lucapivato:
@derrell Yes the "input" change made a mess. The correct fix is to change keydown (insted of keyup) like this:

https://github.com/qooxdoo/qooxdoo/pull/10088/files

Originally I added the "changeValue" event to keyup but it didn't work when pressing key up on keeping it pressed to repeat. So the changeValue event has to be fired on keydown to update the value and let the kayboard work.

derrell (approved):

cboulanger (approved):

johnspackman (approved):

oetiker (approved):

hkollmann:
Thanks @lucapivato

PR #10089, opened 2021-01-18
Allows to skip the time consuming redundant travis tests by adding “[skip-travis]" to the PR title

cboulanger:
This makes sense especially with PRs which need to merge master multiple times when other PRs are merged, given that Travis is painfully slow at the moment. The idea is to change the PR title AFTER Travis has completed successfully at least once.


oetiker (approved):

johnspackman (approved):

derrell (approved):
I'll take your word for it that this works. I don't understand how it is intended to work.

cboulanger:
@derrell We're simply `curl`ing the website of the issue and grab its `<title>` tag [like so](https://github.com/qooxdoo/qooxdoo/pull/10089/files?file-filters%5B%5D=No+extension&hide-deleted-files=true#diff-dd65e64ec7368f861701f5cf9c1320570cd60810c07b1e227b86919bbc385473R2) to get the PR's title. It works with this very PR already - the individual Travis tests pass after one minute because the tests are never called. What this cannot do is shorten the time it takes for travis to run the workflow in the first place- which is long enough. But we'll get rid of it soon and GithHub Actions has a much richer syntax for this kind of conditional workflow.

hkollmann (approved):

level420 (approved):

PR #10095, opened 2021-01-22
add `flexShrink` to `qx.ui.layout.HBox` and `qx.ui.layout.VBox`

johnspackman:
The `flex` setting allows the HBox to shrink children as well as grow them, but there are situations where a child will insist that it has a minimum width requirements, and (prior to this PR) the HBox will not shrink that child to less than it's minimum width.

When the parent has a maximum width and there is not enough room for all of the children (partly because of one child's insistence that it has a large `minWidth`), this means that child will have a destructive effect on the layout, because even if you specify `flex` it will push all subsequent children (ie it's peers) off the screen.

This PR adds a new layout property `flexShrink` which only applies if the `flex` property is also set and there is not enough room for all children; it allows the container to nominate a child to have its `minWidth` ignored if there is not enough space for all children.


cboulanger (approved):

zaucker:
This sounds cool. Would it make sense to have the same for VBox? That would keep the two consistent in their behaviour and features.

framework/source/class/qx/ui/layout/HBox.js
```diff
@@ -371,6 +383,12 @@ qx.Class.define("qx.ui.layout.HBox",
               max : hint.maxWidth,
               flex : flex
             };
+            if (notEnoughSpace) {
+              var props = children[i].getLayoutProperties();
+              if (props && props.flexShrink) {
+                flexibles[i].min = 0;
```
derrell, line 389: Very, very cooll feature! It looks, though, like this eliminates the minWidth of all widgets in the layout, when there's not enough space, so the user has no control. What do you think about flexShrink being something other than a boolean, i.e., giving it a way to specify which child or children should be considered for further shrinking beyond their minWidth?

derrell (changes requested):

framework/source/class/qx/ui/layout/HBox.js
johnspackman: `flexShrink` is set on the child's layout properties, so it only applies to that one child :)

derrell (approved):
Thanks!

oetiker (approved):

level420 (approved):

hkollmann (approved):

PR #10096, opened 2021-01-22
Fix sporading Layout flush null errors.

lucapivato:
It's rare but sometimes we get a null error when a widget flushes Layout queue.

### DIFF
--- a/framework/source/class/qx/ui/core/queue/Layout.js
+++ b/framework/source/class/qx/ui/core/queue/Layout.js
@@ -106,7 +106,8 @@ qx.Class.define("qx.ui.core.queue.Layout",
           // This is an inner item of layout changes. Do a relayout of its
           // children without changing its position and size.
           var bounds = widget.getBounds();
-          widget.renderLayout(bounds.left, bounds.top, bounds.width, bounds.height);
+          if (bounds)
+            widget.renderLayout(bounds.left, bounds.top, bounds.width, bounds.height);
         }
       }
     },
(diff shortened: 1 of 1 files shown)


derrell (approved):
This sounds like a race condition (bounds isn't ready yet, or more likely the widget has already been disposed). This probably band-aids the problem, but is worth doing anyway. (Of course, it'd be nice to find the real cause of the problem...)

hkollmann (approved):

level420 (approved):

framework/source/class/qx/ui/core/queue/Layout.js
```diff
@@ -106,7 +106,8 @@ qx.Class.define("qx.ui.core.queue.Layout",
           // This is an inner item of layout changes. Do a relayout of its
           // children without changing its position and size.
           var bounds = widget.getBounds();
-          widget.renderLayout(bounds.left, bounds.top, bounds.width, bounds.height);
+          if (bounds)
+            widget.renderLayout(bounds.left, bounds.top, bounds.width, bounds.height);
```
cboulanger, line 110: Please add curly brackets as per the linting rules. I am surprise this passes...

hkollmann: it did´nt ...

cboulanger (approved):

johnspackman (approved):

johnspackman:
thank you @lucapivato

PR #10104, opened 2021-01-27
Fixes for unit tests

johnspackman:
fix: fixes a bug with classes with `null` extend not having its constructor assigned
fix: fixes a bug where anonymous inner classes could cause the outer class to be erased if the inner class is undefined
fix: fixes a bug in reading stack traces on chrome
fix: fixes unit tests


cboulanger (approved):
Wow, this must have been a lot of work. Thank you.

derrell (approved):

johnspackman:
FWIW the theme based tests would benefit from review - they depend on specific values in the theme, but this is very fragile because the theme can be legitimately edited to something different, which unexpectedly breaks the tests.  Also, some of the tests are actually useful to run on each of the themes (eg Form tests need to test that the invalid appearances are applied).  I've started to make it theme-agnostic, but there's bits in there which will take more work to do properly.  So long as we stick with Simple theme for the moment it should be fine

derrell:
How about including a complete theme in the test framework so that it is independent of the supplied themes that could legitimately be altered?

johnspackman:
> How about including a complete theme in the test framework so that it is independent of the supplied themes that could legitimately be altered? 

Yes, that would be ideal - although actually you need two for some of the tests so that we can test for things like colour changes if you swap out just the decorators.  It's just that what we have not should be sufficient to pass tests, it's just an FYI that it needs some love in the future

hkollmann (approved):

oetiker (approved):

hkollmann (approved):

hkollmann:
Now the generator tests do not work any longer?

cboulanger:
I fear we cannot have both. We can give the travis tests the boot, since we know they worked before - and run compiler-based test from now on.

cboulanger:
A different option would be to duplicate the code under `qx.test` namespace (with the changes in this PR) and place it as a Library (with a different namespace) in the `test` directory , and leave the Generator-based `qx.test` namespace as it is. This would allow to keep the Travis/Saucelab tests as they are.

level420 (approved):

PR #10106, opened 2021-01-27
Fix a bug in webfont loading [skip travis]

johnspackman:
fix: fix bug where loading multiple different webfonts from  the same family (one italic, one normal, etc) would trigger on load events on the first font only, meaning that font size calculations are slightly out


cboulanger (approved):

framework/source/class/qx/bom/webfonts/Validator.js
```diff
@@ -33,17 +33,25 @@ qx.Class.define("qx.bom.webfonts.Validator", {
 
   /**
    * @param fontFamily {String} The name of the font to be verified
+   * @param fontWeight {String} the weight of the font to be verified
+   * @param fontStyle {String} the style of the font to be verified
    * @param comparisonString {String?} String to be used to detect whether a font was loaded or not
    * whether the font has loaded properly
    */
-  construct : function(fontFamily, comparisonString)
+  construct : function(fontFamily, fontWeight, fontStyle, comparisonString)
```
derrell, line 41: BC issue here, if people were calling this expecting the comparison string to be the second argument?

derrell (changes requested):

derrell (approved):

framework/source/class/qx/bom/webfonts/Validator.js
```diff
@@ -34,16 +34,24 @@ qx.Class.define("qx.bom.webfonts.Validator", {
   /**
    * @param fontFamily {String} The name of the font to be verified
    * @param comparisonString {String?} String to be used to detect whether a font was loaded or not
+   * @param fontWeight {String} the weight of the font to be verified
+   * @param fontStyle {String} the style of the font to be verified
```
oetiker, line 38: since these two are optional too
```suggestion
   * @param fontWeight {String?} the weight of the font to be verified
   * @param fontStyle {String?} the style of the font to be verified
```

oetiker (approved):

hkollmann (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman

PR #10107, opened 2021-01-27
feat: adds imaging scaling to `qx.ui.basic.Image`[skip travis]

johnspackman:


cboulanger (approved):

framework/source/class/qx/ui/basic/Image.js
```diff
@@ -1046,16 +1066,90 @@ qx.Class.define("qx.ui.basic.Image",
...
      */
-    __updateContentHint : function(width, height)
-    {
-      // Compare with old sizes and relayout if necessary
-      if (width !== this.__width || height !== this.__height)
-      {
+    __updateContentHint : function(width, height) {
+      if (width === undefined || height === undefined) {
+        width = this.__width;
+        height = this.__height;
```
derrell, line 1072: Should these be independent tests?
```js
      if (width === undefined ) {
        width = this.__width;
      }
      if height === undefined) {
        height = this.__height;
      }
```

framework/source/class/qx/ui/basic/Image.js
```diff
@@ -108,6 +108,24 @@ qx.Class.define("qx.ui.basic.Image",
       apply : "_applyScale"
     },
 
+    /**
+     * Whether to preserve the image ratio (ie prevent distortion), and which dimension
+     * to prioritise
+     */
+    forceRatio : {
+      init : 'auto',
```
derrell, line 116: There was some semblance of scaling already. Does 'auto' here match the prior behavior? If not, we should start right now keeping track of what's not BC for our release notes.

derrell (changes requested):

framework/source/class/qx/ui/basic/Image.js
johnspackman: the previous version of "scaling" was just stretching the image; I've had this code in use for a while, I think that the "auto" setting was one that I chose because IMHO a stretched image is almost never useful, so the stretching was not useful.  You are right that this is a BC issue, I've added an entry in the changelog

derrell (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

cboulanger:
@johnspackman Thank you! Will now disable Travis.

PR #10110, opened 2021-01-28
Add russian language support

goldim:
There are also _qooxdoo_ feedreader and showcase where I noticed usage of different locales. I have to change them too, don't I? If answer is no then there it is.


framework/source/translation/ru.po
```diff
@@ -0,0 +1,294 @@
...
+msgstr "Сбросить длины столбцов"
+
+msgid "This field is required"
+msgstr "Заполните это поле"
+
+msgid "Visual"
+msgstr "Визуально"
+
+msgid "key_full_Alt"
+msgstr ""
```
derrell, line 102: Is there no translation for these that were left blank? If that's the case, then maybe translate them to English rather than making them blank?

derrell (changes requested):

framework/source/translation/ru.po
goldim: I thought it assigns English by default. All blanks I left on purpose. Thank you, I will change it and commit it.

derrell (approved):
Thanks!

johnspackman (approved):

hkollmann (approved):

level420 (approved):

oetiker (approved):

cboulanger (approved):

cboulanger:
@goldim Can you push a commit (can be anything) so that the checks workflow is updated? We have a problem with Travis.

cboulanger:
thank you @goldim - no pushing necessary, I figured it out.

PR #10122, opened 2021-02-09
correct behavior of vertical form slider (fix #10121)

### LINKED ISSUE #10121: Slider - wrong PageDown/PageUp behaviour
**Describe the bug**
In class qx.ui.form.Slider PageDown keyboard key increases slider value and PageUp keyboard key decreases slider value

**To Reproduce**
1. just use keyboard keys PageDown/PageUp and look at slider behaviour

**Expected behavior**
PageDown keyboard key decreases slider value and PageUp keyboard key increases slider value

**Additional context**
To fix bug modify function _onKeyPress in file \node_modules\@qooxdoo\compiler\node_modules\@qooxdoo\framework\source\class\qx\ui\form\Slider.js  
traded places 'case "PageDown":'  and  'case PageUp":'

rommni:
Currently our vertical slider have maximum value at bottom and minimum value at top.
In all other slider (w3 example: https://html.spec.whatwg.org/multipage/input.html#range-state-(type%3Drange)) the maximum value is at top and it seems logic on a form that if you want a greater value you push the slider to the top.

The Pr also fix bug #10121 which was related to the problem.

Be careful I think this is a Breaking Change, we now have  completely a different behavior for our vertical slider even if it is more logic so it can break some existing code.


johnspackman (approved):

cboulanger:
Please add a section to `changelog.md` under the heading "# Upcoming" > "## Breaking Changes" detailing the break with backward compatibility.

derrell (approved):
This is not backward compatible, but I feel that it's a bug fix and that it's been broken all of this time.

rommni:
> Please add a section to `changelog.md` under the heading "# Upcoming" > "## Breaking Changes" detailing the break with backward compatibility.

done

cboulanger (changes requested):
Sorry to be pedantic - the breaking change info will not be part of v6.0.x, but will got into the next release, so it needs to go into its own section on top (create both \#Upcoming and \##Breaking Changes  headline). We'll switch to a new release model shortly which will generate a changelog automatically, but we're not there yet.

rommni:
Oh yes sorry my bad i correct this

hkollmann (approved):

cboulanger (approved):

level420 (approved):

johnspackman:
Although I approved, it I think that this PR is wrong and should be reverted; when I tested your playground example, I could have sworn that page down was making the horizontal slider do the wrong thing, ie that Page Down made the value reduce (slider moved to the left) and Page Up made the value increase (slider moved to the right).  Testing it in the current playground does not do that, it works as expected.

Here is the playground: https://tinyurl.com/y46c7uav

I don't know whether I'm just imagining it or what, but I cannot fault the behaviour in that playground.  The offset of (0,0) should mean that the slider is at the top of the scroll bar and the top of the contents is being shown.  That is what happens with `window` and with `textarea`, and in a Latin-based language, we expect to read from top left to bottom right

What happens now that this PR has been merged is that (0,0) is the **bottom** left and the slider appears at the bottom.  This has a big effect on the user presentation, because (for example) lists etc are starting with the slider at the bottom and that's definitely not what the user expects.

Here's some screenshots showing the same code as in that playground, but with the PR applied:

Screenshot 2021-02-19 at 12 47 56

Screenshot 2021-02-19 at 12 48 13

johnspackman:
sorry, please ignore my previous comment - my git repo was screwed up; the current upstream/master is reporting the expected results

PR #10123, opened 2021-02-10
Merge compiler

johnspackman:
This PR does the main work of merging the compiler into framework; while this is suitable for testing and development (please see docs/contribute.md for getting started instructions) it is not ready for prime time.  Issues are:

* No deployment to npmjs of `qx` - this is by design really, until all the other issues are resolved.  
* API Viewer does not compile (lots of ENOENT errors for .json files) so has been removed from compile.json
* There is a dependency on `@qooxdoo/framework` (ie the released framework) which is getting picked up by the compiler and tests - this is not wanted, and work will be required to fix this in the short term - in the meantime, use `qx config set qx.library /path/to/qooxoo/repo`
* You will probably find that there are a number of tests which fail, for minor reasons. 
* Lint is slow and the compiler's ruleset have been relaxed because it generates 51,000+ "errors" about beautification.  Please **do not be tempted** to use reinstate checks and run `lint --fix`, lets get this working before allowing a second rate idiot computer program to pollute the codebase with an incomprehensibly large number of mass edits (assuming we would even want that).


hkollmann:
My thoughts:

- I would move the compiler testing stuff in a compiler subdir
- Remove apiviewer from packages
- upgrade the testnode package
- we should think about having compiler tests in an own project like the framework tests in test/framework

cboulanger:
Thank you for this. It's completely ok for this PR to mess things up horribly, it is to expected given the scale of this change. We'll need some time to decide how to fix all the ramifications of the integration. It is good that nothing is deployed to NPM at this point, I think that this can only happen once we have thought the semantic release stuff through thoroughly. I think we should merge this and then start rebuilding on the smouldering ruins. 

As written on Gitter, in order to fix the CLI we have to decide on how to handle the dependency of projects on the framework library, I propose that from 7.0.0 on, the dependence is declared by simply stating the "qx" library as dependency, in `requires` key of `Manifest.json` replacing `@qooxdoo/framework` and `@qooxdoo/compiler`. I could add this to the next PR after merging this one.

cboulanger (approved):

oetiker (approved):
I can't really judge ... the only thing to maybe think about would be if it would make sense to establish a compiler branch and then PR to this until it is in a good state to merge the whole thing with master and thus keep master in a working state ...

cboulanger:
@oetiker I think technically master IS in a working state if you only look at the qooxdoo framework code. Nothing has changed there, so you can still use it with an external compiler. What is really broken is the whole build pipeline and that's independent of whether the compiler is merged or not. So I think we should work on getting it back to work in master and work from there.

derrell (approved):

level420 (approved):
I’m trusting @johnspackman 
Too large to really review

hkollmann (approved):

PR #10125, opened 2021-02-11
Fix compiler bootstrapping

cboulanger:


johnspackman (approved):

README.md
```diff
@@ -29,6 +29,28 @@ For more information please see http://qooxdoo.org/LICENCSE .
...
+[NPM release](https://www.npmjs.com/package/@qooxdoo/qx) and a
+compiler called with `npx qx ...`. 
+
+In contrast, when hacking qooxdoo itself, you
+need to first execute the following steps:
+
+```bash
+git clone https://github.com/qooxdoo/qooxdoo.git
+cd qooxdoo
+npm install
```
hkollmann, line 44: npm ci - then package-lock is not changed

hkollmann (approved):

oetiker (approved):
if it was in a branch you could just happily merge way :)

cboulanger:
@oetiker That's true, but I needed the feedback! Much better PR now ... don't worry won't make a PR for each little change.

.gitignore
```diff
@@ -28,10 +28,10 @@ tmp/
 /apps
 /deploy
 /build
-/compiled
```
derrell, line 31: This seems wrong. Why would you want to commit the `comiled` directory?

derrell (changes requested):

.gitignore
cboulanger: This is a duplicate entry. There is already a `compiled` in line 27

derrell (approved):

PR #10128, opened 2021-02-12
Refactoring and modularization of user code migration process

cboulanger:
still unfinished & unfunctional, just showing the current thinking.


source/class/qx/tool/cli/Utils.js
```diff
@@ -0,0 +1,105 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2017-202 The authors
```
derrell, line 8: year invalid

source/class/qx/tool/cli/Utils.js
```diff
@@ -0,0 +1,105 @@
...
+        });
+        exe.stderr.on("data", data => {
+          if (this.argv.verbose) {
+            qx.tool.compiler.Console.error(data.toString());
+          }
+        });
+        exe.on("close", code => {
+          if (code !== 0) {
+            let message = `Error executing '${cmd} ${args.join(" ")}'. Use --verbose to see what went wrong.`;
+            throw new qx.tool.utils.Utils.UserError(message);
```
derrell, line 51: Should it throw here, or better to call `reject`?

source/class/qx/tool/migration/IMigration.js
```diff
@@ -0,0 +1,38 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2017-202 The authors
```
derrell, line 8: year invalid

source/class/qx/tool/migration/M6_0_0.js
```diff
@@ -0,0 +1,183 @@
...
+     * @inheritDoc
+     * @see {qx.tool.migration.IMigration#migrate()}
+     */
+    migrate: async function (options = {}) {
+      let {announceOnly} = options;
+      let needFix = false;
+      let pkg = qx.tool.cli.commands.Package;
+      let cwd = process.cwd();
+      let migrateFiles = [[path.join(cwd, pkg.lockfile.filename), path.join(cwd, pkg.lockfile.legacy_filename)], [path.join(cwd, pkg.cache_dir), path.join(cwd, pkg.legacy_cache_dir)], [path.join(qx.tool.cli.ConfigDb.getDirectory(), pkg.package_cache_name), path.join(qx.tool.cli.ConfigDb.getDirectory(), pkg.legacy_package_cache_name)]];
+      if (qx.tool.migration.Utils.checkFilesToRename(migrateFiles).length) {
```
derrell, line 41: Some comments in here saying what this code is trying to accomplish would be really useful. It's not obvious.

derrell (commented):
This is WIP, but I took a look through it anyway. Lots going on here.

hkollmann:
What's about adding some unit tests in qx.test.tool.cli namespace?

cboulanger:
@cboulanger Oh yes certainly. This PR is only 1/3-finished. Just wanted to let you guys know the direction it intends to take.

cboulanger:
Ready for review.

johnspackman (approved):

level420 (approved):

hkollmann (commented):
is  test/cli/compiler/test-migrations/v6.0.0/unmigrated/contrib/.gitkeep needed?

bin/tools/utils.js
```diff
@@ -216,6 +218,379 @@ fs.copyFileSync("bin/build/qx.cmd", "bootstrap/qx.cmd");
...
+
+  /**
+   * Rotates files so that this file does not exist, by renaming the existing file to have a ".1"
+   * appended, and the ".1" to be renamed to ".2" etc, up to `length` versions
+   *
+   * @param filename {String} filename to rotate
+   * @param length {Integer} maximum number of files
+   * @async
+   */
+  rotateUnique: async function(filename, length) {
```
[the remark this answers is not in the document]
cboulanger: Oh, I just copy & pasted utility methods from  `qx.tool.utils.*` to make them available to the test methods, which do not have qx available. The state of utility methods in `qx.tool` is a real mess, which has to do with its ad-hoc and decentral development. There is a lot of redundancy and overlapping methods, plus duplicate functionality targeting earlier node versions. It needs a thorough clean-up, but that means a lot of work. In particular, we need some classes which can be accessed both by qx and the nodeJS runtime - which would probably work with the "qx.server" package.

The method you point to is probably not needed by the test pipeline and can be removed.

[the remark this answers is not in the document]
cboulanger: I think `.gitkeep` is needed because git doesn't store empty directories. I need the directoy because it must be renamed by the migration.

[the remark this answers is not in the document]
johnspackman: @hkollmann if it's not used anywhere then it can be deleted; I recognise it as one of mine, I use it for rotating log files, but it's not very well defined (or sophisticated) so does not need to be in the framework unless its used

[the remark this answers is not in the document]
hkollmann: @cboulanger : Could you delete it? Inclusive the then not needed test?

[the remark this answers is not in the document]
cboulanger: I will delete it.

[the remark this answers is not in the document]
cboulanger: Done.

source/class/qx/tool/cli/Utils.js
cboulanger: done.

source/class/qx/tool/migration/M6_0_0.js
cboulanger: The migration code has been modularized, with each step in one separate method- the name should tell you what it does: https://github.com/cboulanger/qooxdoo/blob/refactor_migration/source/class/qx/tool/migration/M6_0_0.js

hkollmann (approved):

cboulanger:
Copying this over from Gitter: What this PR does is to allow granular migrations, i.e. for every version change (of course this mostly applies to breaking ones), a migration file can be added (e.g. `qx.tool.migration.M7_1_3`). Ideally, our workflow would be configured to require that any breaking change MUST contain such a file, so it is guaranteed to provide a migration path. If the user updates the framework, the CLI will warn that migrations exist that have to be applied and tell the user to qx migrate.

The migration classes should extend from `qx.tool.migration.BaseMigration` which contains useful utility methods. Any method starting with "migration" will be run, which makes it easy to organize migration steps easily. Here is an example for the 6.0.0-beta to 6.0.0 migration: https://github.com/cboulanger/qooxdoo/blob/refactor_migration/source/class/qx/tool/migration/M6_0_0.js

The PR also includes:
- the removal of `qx.library` config setting
- change of the way the qooxdoo library is looked up - now the canonical methods are 
   - `qx.tool.config.Utils.getQxPath()` to get the qooxdoo version in the environment (i.e. usually the NPM module, but it will autodetect the repo also) of the current working dir, 
   - `qx.tool.config.Utils.getAppQxPath()` to get the qooxdoo version used by the package/library/application in the current working dir; this can be different from the `getQxPath()` value. It is calculated either by looking up the qooxdo library referenced in the `libraries` key of `compile.json`, which takes precedence, or by looking at the `@qooxdoo/framework` dependency information in the Manifest. 
- allows to use the `this.(debug|info|warn|error)` methods from within server code.

source/class/qx/tool/cli/commands/Command.js
```diff
@@ -60,353 +57,129 @@ qx.Class.define("qx.tool.cli.commands.Command", {
...
-      let {applications} = await this.getProjectData();
-      if (applications instanceof Array && applications.length) {
-        return path.resolve(process.cwd(), applications[0].path);
+      const semaphore = path.join(process.cwd(), ".qxmigrationcheck");
+      try {
+        await fsp.stat(semaphore);
+        this.debug(`Not checking migration because check is already in progress.`);
+      } catch (e) {
+        // run migration in dry-run mode
+        await fsp.writeFile(semaphore,"");
```
derrell, line 105: We've got a race condition here. Since we're using a semaphore, we're clearly concerned with two processes doing migrations at the same time... but it's possible that the `stat` runs in one of the processes and finds no error, an operating system task switch occurs and `stat` runs in the other process finding no error, and as each process continues, they each enter `writeFile`. In general, it is better not to do a test to see if you can do something and then do it (which causes the race condition), but rather to just "try to do it" and let it fail if it can't.

In this case, that would be done by getting rid of the `stat` and passing options to `writeFile` to tell it to error if the file exists. This sample code shows the general idea: if /tmp/xxx already exists, this will fail.

```
fsp.writeFile("
  /tmp/xxx", 
  "hello world",
  {
    flag : fs.constants.O_WRONLY | fs.constants.O_CREAT | fs.constants.O_EXCL,
    mode: 0o644
  })
 .then(() => console.log("worked"))
 .catch((e) => console.log("failed: " + e));
```

source/class/qx/tool/config/Utils.js
```diff
@@ -0,0 +1,308 @@
...
+      dir = path.resolve(dir || process.cwd());
+      let root = path.parse(dir).root;
+      while (dir !== root) {
+        // 1. Manifest.json files
+        if (await this.isQxLibrary(dir)) {
+          return dir;
+        }
+        // 2. node_modules folders
+        let npmdir = path.join(dir, "node_modules", "@qooxdoo", "framework")
+        if (fs.existsSync(path.join(npmdir, "Manifest.json"))) {
```
derrell, line 143: consider replacing sync call with async one?

source/class/qx/tool/migration/BaseMigration.js
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Given an array of [newPath,oldPath], filter by those which exist
+     * at oldPath and not at newPath
+     * @param fileList {[]}
+     * @return []
+     */
+    checkFilesToRename(fileList) {
+      let filesToRename = [];
+      for (let [newPath, oldPath] of fileList) {
+        if (!fs.existsSync(newPath) && fs.existsSync(oldPath)) {
```
derrell, line 128: sync -> async ?

source/class/qx/tool/migration/BaseMigration.js
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Checks if the given file or array of files contains a given text
+     * @param {String|String[]} files
+     * @param {String} text
+     * @return {Promise<Boolean>}
+     */
+    async checkFilesContain(files, text) {
+      files = Array.isArray(files) ? files : [files];
+      for (let file of files) {
+        if ((await fsp.stat(file)).isFile() && (await fsp.readFile(file, "utf8")).includes(text)) {
```
derrell, line 144: This looks to me like if `fsp.stat().isFile()` returns false, we'll be doing `if false.includes(text)` which will throw. Is that what's intended?

source/class/qx/tool/migration/M6_0_0.js
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Check for legacy compile.js - needs manual intervention
+     * @return {Promise<{applied: number, pending: number}>}
+     */
+    async migrateCompileJs() {
+      let migrationInfo = this.getRunner().createMigrationInfo();
+      let compileJsFilename = path.join(process.cwd(), "compile.js");
+      if (await fs.existsAsync(compileJsFilename)) {
+        let data = await fs.readFileAsync(compileJsFilename, "utf8");
```
derrell, line 39: `existsAsync` looks unnecessary. Is there a reason not to simply try to read the file, and ignore the ENOEXIST error? (This one is likely not a race condition; just seems superfluous.)

derrell (changes requested):

bin/tools/utils.js
```diff
@@ -5,19 +5,21 @@ const child_process = require("child_process");
 //var fsPromises = require("fs").promises;
 // node 8 compatibility
 const {promisify} = require('util');
+const stat = promisify(fs.stat);
```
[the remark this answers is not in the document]
cboulanger: The whole file is a basket full of random utility functions with different dependencies, so this is just as it is toto make it work. 

It would be a waste of time to clean up this file because it needs to be replaced by a fully rewritten static class that is included in the "qooxdoo" (qx.server) npm package and this way can be accessed by nodeJS. As I have written elsewhere, the whole utility method business requires a rewrite, but this is outside the scope of this PR.

bin/tools/utils.js
```diff
@@ -216,6 +218,355 @@ fs.copyFileSync("bin/build/qx.cmd", "bootstrap/qx.cmd");
...
+    function copy(statFrom, statTo) {
+      if (statFrom.isDirectory()) {
+        var p;
+        if (statTo === null) {
+          p = fsPromises.mkdir(to);
+        } else {
+          p = Promise.resolve();
+        }
+        return p.then(() => fsPromises.readdir(from)
+          .then(files => Promise.all(files.map(file => t.sync(path.join(from, file), path.join(to, file), filter)))));
```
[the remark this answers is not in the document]
cboulanger: This function is just copy & pasted from another `qx.tool.utils` utility function which is not accessible to the NodeJS process. Same problem as before, so please just ignore it for the moment.

source/class/qx/tool/cli/commands/Command.js
cboulanger: You are right in principle, but the use case are only for the case that one `qx` command triggers another one before it has finished, and the chance of a race condition here is almost zero because of the time it takes before a `qx` CLI command actually executes. If this was mission-critical, I'd rewrite it, but it is probably not even real problem - I was already thinking to remove it altogether.

source/class/qx/tool/config/Utils.js
cboulanger: Done. Thanks.

source/class/qx/tool/migration/BaseMigration.js
cboulanger: As per the Docs, the method always returns a <Stats> object. https://nodejs.org/api/fs.html#fs_class_fs_stats even if there is no file.

derrell: That's ugly. A function that begins "is" generally returns a boolean. Anyway, not our problem.

derrell (approved):

source/class/qx/tool/migration/M6_0_0.js
cboulanger: Sure, try/catch would work here just the same. This is just habit. But there is a drawback - if you just `try` the readFileAsync in order not to catch errors from the other code inside the `if` clause, you have to set some flag insde the try:
```javascript
let flag = false;
try { 
  let data = await fs.readFileAsync(compileJsFilename, "utf8");
  flag = true;
} catch(e){}
if (flag) { 
// do the other code that was inside the if clause
} 
 ```

I am not sure this is really superior.

Comment (derrell) [reply source/class/qx/tool/migration/M6_0_0.js:39]: Wouldn't `finally` solve that?

Not important. Was just a thought.

Comment (cboulanger) [reply source/class/qx/tool/migration/M6_0_0.js:39]: You are right! That's exactly what `finally` is for. I forgot about it.

Comment (cboulanger) [reply source/class/qx/tool/migration/M6_0_0.js:39]: (I think when I first learned javascript, it didn't have finally)

**Review by oetiker — APPROVED** [review]

File: test/cli/compiler/test-migrations.js [hunk 13]
```diff
@@ -0,0 +1,29 @@
...
+
+test("Test migrations", async assert => {
+  try {
+    const baseDir = path.join(__dirname, "test-migrations", "v6.0.0");
+    const unmigratedDir = path.join(baseDir, "unmigrated");
+    const migratedDir = path.join(baseDir, "migrated");
+    await testUtils.deleteRecursive(migratedDir);
+    await testUtils.sync(unmigratedDir, migratedDir);
+    console.log("\n# upgrade alert\n");
+    await testUtils.runCommand(migratedDir, qx, "clean");
```
hkollmann, line 17: Why do not use runcompiler here?

hkollmann (changes requested):

test/cli/compiler/test-migrations.js
cboulanger: What would be the equivalent commands?

hkollmann:
Sorry runCompiler only runs compile process.
But there is a helper in utils to fetch qx - so there is no need to do it yourself

Von meinem iPad gesendet

test/cli/compiler/test-migrations.js
```diff
@@ -0,0 +1,29 @@
+const test = require("tape");
+const path = require("path");
+const process = require("process");
+const testUtils = require("../../../bin/tools/utils");
+const fsp = require("fs").promises;
+const qx = path.resolve(__dirname, "..", "..", "..", "bin", "source", "qx");
```
hkollmann, line 6: const qx = testUtils.getCompiler()

hkollmann (changes requested):

test/cli/compiler/test-migrations.js
cboulanger: Strangely, that breaks the tests, because getCompiler() returns the built version of the compiler instead of the source version. However, that means that there's something wrong, since both versions should produce the same output. Let me check.

cboulanger:
@hkollmann The problem was that I hadn't recompiled the "build" compiler. Now everything works in source or build mode.

I have changed the API for updating runtime migration information. Instead of passing around an object with that info, I added two methods `markAsApplied()` and `markAsPending()` to the base class. Much simpler and cleaner and less prone to breakage.

Now everything is ready. @hkollmann, once you have approved, you can merge it!

hkollmann (approved):

PR #10131, opened 2021-02-14
Enable testing

hkollmann:
- fix all compiler unit tests
- fix all compiler integration test
- add compiler test project
- add tests for compiler and framework in project root so that qx test in root will lint, test compiler and test framework
- add some tap output so that ```qx test 2>nul | tap-summary``` looks nice
- add test call to package json


cboulanger (approved):

package.json
```diff
@@ -23,8 +23,8 @@
   "scripts": {
     "devtools": "node source/resource/qx/tool/bin/build-devtools",
     "website": "node source/resource/qx/tool/bin/build-website",
-    "pretest": "node bootstrap-compiler",
-    "test": "node ./bin/build/qx lint --warnAsError && node ./bin/build/qx --block-global-framework test"
+    "pretest": "node bootstrap-compiler --target=build",
+    "test": "node ./bin/build/qx test --fail-fast --block-global-framework 2>nul | npx tap-summary"
```
derrell, line 27: `2>nul` looks like a Windowsism. On Linux, that would be `2>/dev/null` (unless you wanted to write stderr to a file named `nul`). Does this code need to run operating system independent?

derrell (changes requested):

package.json
hkollmann: And how to redirect stderr to stout?
Yes it should be  operating system independent.

hkollmann: found it. Same as in windows. 2>&1

oetiker (approved):

derrell (approved):

level420 (approved):

johnspackman (approved):

johnspackman:
thanks @hkollmann

PR #10132, opened 2021-02-15
fix: Allow space in html with contenteditable field

### LINKED ISSUE #10130: Not able to add space on contenteditable
**Describe the bug**
qooxdoo somehow disable the event for space. Modern js editor for example quill, editorjs make use of contenteditable
in order for it to work. But since qooxdoo currently seems to be blocking user from key in space character. This editor
not able to work properly

**To Reproduce**
Here is a minimal playground code to reproduce this problem https://tinyurl.com/y3xbpckt

**Expected behavior**
Type "Hello", space then "World" should produce "Hello World" but the result is "HelloWorld"

 - OS: Catalina 10.15.7
 - Browser : firefox, chrome & safari
 - Version latest version

rommni:
Add a check before intercept the space event to handle case of contenteditable inside qx.ui.embed.Html()

fix #10130


derrell (approved):
This is a little bit scary since it's in an abstract class and could have far-reaching implications, but this looks safe AFAICT...

rommni:
> This is a little bit scary since it's in an abstract class and could have far-reaching implications, but this looks safe AFAICT...

To be honnest i was a little faired to do changes on this file. However it seems that the edit is quite small and shouldn't have unexpected effect. If someone see a possible problem please don't mind to yell it ;)

hkollmann (approved):

oetiker (approved):

oetiker:
how about adding a "Backport" label to things that should seem worthy of adding to 6 as well

cboulanger (approved):

level420 (changes requested):
I'm not sure that this PR is really needed in order to fix what was the cause for opening issue #10130 as when quilljs or other editors are used like demonstrated in https://tinyurl.com/yyda8ptu entering spaces works as expected.
So I'd like to wait for @rommni @paragasu to check if we only need this PR to solve the editor integration problem or if it is needed to solve some general flaw.

oetiker:
I think we do need the patch, but it is probably a bit too early in the file ... look here https://github.com/qooxdoo/qooxdoo/blob/3f74b97fab98b9e483b21c811f4b48692c79f8c1/source/class/qx/ui/root/Abstract.js#L262 that is the place where the target is handled. So the editContent case should be handled here

oetiker (changes requested):
move change a bit down

johnspackman:
I'm curious to know why other editors don't have this problem, but OTOH the example that @rommni has given is pretty distilled down to showing that a contentEditable `div` is not behaving as you'd expect; in the code, `el` is the DOM for the `qx.ui.embed.Html` and presumably `originalTarget` is the actual embedded contentEditable `<div>`.

However there is an issue in that `originalTarget` is not always the contentEditable `div`, for example if the cursor is in a nested child element; I changed the code to this:

```
var button1 = new qx.ui.embed.Html('<div contenteditable=true><p><b>Nospace!</b></p></div>');
```

and the originalTarget is the `<b>` and it has `contentEditable` set to `inherit`

It seems that the original implementation only works when the embedded element is contentEditable, because the DOM event will bubble up to the Qooxdoo widget.

Also, given that the nature of the test whether `this` widget is editable, it seems slightly wrong that a global handler like this is working that out by sniffing the encapsulated DOM - after all, in theory at least AIUI an editor could be implemented with a `canvas` tag, ie the implementation of the widget is private.

So, how about this: `qx.ui.core.Widget` should expose a new API method called something like `isAcceptsKeyboardInput` (which returns false by default) and that is what this global handler checks; that method would be refined, so `qx.ui.embed.Html` knows that it has to check every DOM element from `originalTarget` up to `getContentElement` looking for `contentEditable==="true"`, whereas `qx.ui.form.AbstractField` sniffs depending on DOM name

johnspackman (changes requested):
please see the comment in the main thread for my "request changes" review

level420:
@johnspackman quilljs does not have the problem "unable to enter space" either. See https://tinyurl.com/yyda8ptu

cboulanger:
It seems we have the following choices:

1) We leave everything as it is, but add a page in the documentation that instructs how to correctly embed a HTML editor so that the space-eating behavior does not occur. 
2) We include the fix after making sure that this doesn't break anything
3) We research the root problem  and create a solution that goes at the heart of the problem.

Plus, a qooxdoo package that exposes the editor as a qooxdoo widget (and works around the problem) is always welcome!

oetiker:
> It seems we have the following choices:
> 
>     1. We leave everything as it is, but add a page in the documentation that instructs how to correctly embed a HTML editor so that the space-eating behavior does not occur.
> 
>     2. We include the fix after making sure that this doesn't break anything
> 
>     3. We do some more research and create a solution that fixes the root of the problem.
> 
> 
> Plus, a qooxdoo package that exposes the editor as a qooxdoo widget (and works around the problem) is always welcome!

as pointed out in the example, there definitely IS a problem ... and even the original code is trying to handle it:

https://github.com/qooxdoo/qooxdoo/blob/3f74b97fab98b9e483b21c811f4b48692c79f8c1/source/class/qx/ui/root/Abstract.js#L264

but it leaves out the `contentEditable` check ... this fix at least would be appropriate in any case ... even without going all the way (like @johnspackman  suggests).

rommni:
Sorry guys not much time these days. Your comments seems interesting we will need to investigate more. I will try to do that as soon as I have some time.

cboulanger:
@rommni Any progress on the issue? We do want to release a first beta version of v7.0.0 quite soon and it would be great to have the fix included (maybe also in a v6.0.1 maintenance release)

oetiker (approved):

hkollmann:
Should we close this?

johnspackman:
no, leave it open for now - @rommni if it's OK with you I'd like to take over this PR?  Im doing some work in that area so it might be convenient for both of us?

hkollmann:
@johnspackman Could we close this?

johnspackman:
oh wow!  I'd forgotten about this.  I think this is an issue which annoys me at least once a week with CodeMirror.  We could close it, because while it's on my radar I'm obviously not getting round to looking at it

PR #10134, opened 2021-02-17
Fix tests

hkollmann:
With this fixes to the tests i came down to 7 not working tests. 
Details see commit logs.

For this last 7 i need some help because i do not understand what's happening.
TAP version 13
not ok 837 - qx.test.ui.toolbar.OverflowHandling:testShowItemRemoved - [16.97 ms] - Event (showItem) not fired.
not ok 1209 - qx.test.Mixin:testPatchOverwritten - [0.60 ms] - Cannot find base class method called sayJuhu for mixin qx.MPatch, when viewed from qx.Patch1
not ok 1694 - qx.test.bom.rest.Resource:test: invoke action ignores invalid check in production - [14.46 ms] - Expected value to be a regular expression but found !
not ok 1722 - qx.test.bom.rest.Resource:test: dispose request on loadEnd - [32.86 ms] - resume() called before wait()
not ok 1730 - qx.test.bom.rest.Resource:test: fire started - [0.81 ms] - Timeout of 500 ms reached before resume() was called.
not ok 1952 - qx.test.io.rest.Resource:test: invoke action ignores invalid check in production - [18.60 ms] - Expected value to be a regular expression but found !
not ok 2773 - qx.test.lang.normalize.Date:test parse() - [12.79 ms] - Expected '-377736727708000' (identical) but found '-377736728100000'!
DONE testing 3226 ok, 7 not ok, 172 skipped - [324119 ms]


source/class/qx/test/data/singlevalue/Array.js
```diff
@@ -100,8 +100,6 @@ qx.Class.define("qx.test.data.singlevalue.Array",
       });
       this.__label = new qx.test.data.singlevalue.TextFieldDummy();
 
-      // remove all bindings
```
cboulanger, line 103: Is it safe to simply delete this? Can we do it because the tests don't run in one single execution context any more?

hkollmann: removing off all bindings here is not necessary - test works without. Removing of all bindings is deprecated - so tests should'nt depent on it.

cboulanger: Ok thanks!

cboulanger (approved):

johnspackman (approved):

level420 (approved):

oetiker (approved):

derrell (approved):
I'll take others' word for it.

PR #10138, opened 2021-02-22
Update docs for v7

cboulanger:


hkollmann:
Assumed that the new build chain is setup and running?
Should'nt  we merge this after we set it up?

cboulanger:
This is the documentation that is visible only at https://qooxdoo.org/qooxdoo, not the one at  https://qooxdoo.org/docs . It can and _should_ reflect current master. This change updates it to what is already in the code. In fact, in the future code changes should _always_ be accompanied by documentation changes.

hkollmann (approved):

derrell (approved):

oetiker (approved):

Issue #10117, opened 2021-02-05
TODO Post-v6-release


cboulanger:
More complicated items or those requiring a discussion should have their individual issue, to which we link here:

- [x] Make sure that all badges in https://qooxdoo.org/status/ and https://github.com/qooxdoo/qooxdoo don't display any broken state for master, disable until everything works again
- [x] remove all generator-related directories (application, component, tool)
- [x]  merge https://github.com/qooxdoo/qooxdoo/pull/10104
- [x] move `framework/source` to `source` 
- [x] fix tests
- [x] fix qx lint issues
- [x] merge compiler into framework 
    - [x] update config & compiler `getQxPath()`
    - [x] move open issues from qooxdoo/qooxdoo-compiler/issues to qooxdo/qooxdoo/issues 
    - [x] update `qx package migrate` to update `"$schema": "https://github.com/qooxdoo/qooxdoo-compiler/XXX/YYY.json",` `"$schema": "https://qooxdoo.org/schema/YYY.json",` in the configuration files
- [ ] archive the https://github.com/qooxdoo/deployment script
- [x] update https://github.com/qooxdoo/qooxdoo-compiler to reflect new base in framework repo
- [x] adapt the existing deployment script to run testing of merged framework & compiler
- [x] re-integrate documentation, so that it can be versioned and PR can contain changes to both code and documentation at the same time
- [ ] Create infrastructure for automating tasks:
    - [ ] Create bot user which we can use to automate tasks which don't need to be attributed to one of the team members, see https://stackoverflow.com/questions/29177623/what-is-a-bot-account-on-github
    - [ ] have a look whether https://github.com/probot/probot can be useful 
- [x] https://github.com/qooxdoo/documentation/issues/24 : create workflow that publishes documentation from qooxdoo/qooxdoo in qooxdoo/documentation in a versioned form (i.e. qooxdoo.org/documentation/6.0/ ... qooxdoo.org/documentation/7.1/, etc.
- [ ]  https://github.com/qooxdoo/qooxdoo/issues/10109 : find a consensus on the exact form of semantic-release (see and long gitter discussion) by providing an implementation PR that can be discussed:
    - [ ] add semantic-release script to the new deployment script
    - [x] edit deployment script to release the version in `package.json` (ie do not automatically update version numbers)
    - [ ] adapt PR workflow to only accept semantic-release-conformant commit messages 
[...]

PR #10139, opened 2021-02-22
fix some more tests. see https://github.com/qooxdoo/qooxdoo/issues/10117#issuecomment-782139616

hkollmann:
fix tests where qx.debug has to been false
fix error with wrong missing self

Only 2 left:

not ok 837 - qx.test.ui.toolbar.OverflowHandling:testShowItemRemoved - [7.41 ms] - Event (showItem) not fired.
not ok 1209 - qx.test.Mixin:testPatchOverwritten - [0.48 ms] - Cannot read property 'call' of undefined
DONE testing 3209 ok, 2 not ok, 214 skipped - [273779 ms]

### DIFF
--- a/test/framework/compile.json
+++ b/test/framework/compile.json
@@ -41,8 +41,7 @@
       "title": "Qooxdoo TestTAPper",
       "environment": {
         "qx.icontheme": "Tango",
-        "qxl.testtapper.testNameSpace": "qx.test",
-        "qx.debug": false
+        "qxl.testtapper.testNameSpace": "qx.test"
       },
       "include": [
         "qx.test.*"
--- a/test/framework/compile.json
+++ b/test/framework/compile.json
@@ -41,7 +41,8 @@
       "title": "Qooxdoo TestTAPper",
       "environment": {
         "qx.icontheme": "Tango",
-        "qxl.testtapper.testNameSpace": "qx.test"
+        "qxl.testtapper.testNameSpace": "qx.test",
+        "qx.debug": false
       },
       "include": [
         "qx.test.*"
--- a/docs/README.md
+++ b/docs/README.md
@@ -39,11 +39,11 @@ or you can install it globally.
 
 ### Global installation
 
-The following command line installs the Qooxdoo compiler so that it becomes
-available anywhere via your path settings.
+The following command line installs the Qooxdoo framework globally, which includes
+the compiler and makes it available anywhere via your path settings.
 
 ```bash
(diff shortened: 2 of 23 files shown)


test/framework/compile.json
```diff
@@ -41,7 +41,8 @@
       "title": "Qooxdoo TestTAPper",
       "environment": {
         "qx.icontheme": "Tango",
-        "qxl.testtapper.testNameSpace": "qx.test"
+        "qxl.testtapper.testNameSpace": "qx.test",
+        "qx.debug": false
```
derrell, line 45: As discussed on gitter, this looks like it will cause tests that require `qx.debug=true` to fail or not be tested properly, e.g., tests of `check` in properties. Do we really want to make this change?

derrell (changes requested):

test/framework/compile.json
hkollmann: I suggest this steps

1. Accept this and setup a test pipline
2. Change compiler to optionally check settings at runtime
3. Revert this and test again

cboulanger: Wouldn't it maybe be better to just skip this test instead until 2) is implemented, rather than changing qx.debug für all tests?

hkollmann: We can do so. But qx.debug seems to have no effect on testing. After switching off no other test fails

hkollmann: And there are some more tests which should be reenabled. I mean the php based remote test where the backend should be changed to an express based node server in compile.js

cboulanger: I would see it this way: this test is testing something that is currently impossible because the compiler works differently from the generator. So the test itself makes no sense. The conclusion, to me, is that the test should be disabled rather than changing the environment to make the test think it succeeds when it really doesn't 😄

oetiker (approved):
lets get testing working again :-) thank you @hkollmann for all the work!

level420 (approved):

test/framework/compile.json
derrell: > We can do so. But qx.debug seems to have no effect on testing. After switching off no other test fails

That seems to imply that we don't test that `check` works correctly, if nothing else. Ok, in the interest of getting tests working, I agree, we just go with this.

derrell (approved):

cboulanger (approved):

johnspackman (approved):

PR #10140, opened 2021-02-23
fixes bug in handling mixins, which caused qx.test.Mixin:testPatchOverwritten to fail in the compiler

johnspackman:
see also https://github.com/qooxdoo/qooxdoo/issues/10117#issuecomment-782914590

### DIFF
--- a/source/class/qx/Class.js
+++ b/source/class/qx/Class.js
@@ -1451,21 +1451,31 @@ qx.Bootstrap.define("qx.Class",
         // Hint: Could not use typeof function because RegExp objects are functions, too
         // Protect to apply base property and aspect support on special attributes e.g.
         // classes which are function like as well.
-        if (base !== false && member instanceof Function && member.$$type == null)
-        {
-          if (!qx.core.Environment.get("qx.compiler") && wrap == true)
-          {
-            // wrap "patched" mixin member
-            member = this.__mixinMemberWrapper(member, proto[key]);
-          }
-          else if (wrap != true)
-          {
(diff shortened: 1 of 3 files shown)


cboulanger (approved):

hkollmann (approved):

source/class/qx/Class.js
```diff
@@ -1451,21 +1451,31 @@ qx.Bootstrap.define("qx.Class",
...
-          if (!qx.core.Environment.get("qx.compiler") && wrap == true)
-          {
-            // wrap "patched" mixin member
-            member = this.__mixinMemberWrapper(member, proto[key]);
-          }
-          else if (wrap != true)
-          {
-            // Configure extend (named base here)
-            // Hint: proto[key] is not yet overwritten here
+        if (base !== false && member instanceof Function && member.$$type == null) {
```
derrell, line 1454: Is there any other possible falsy value of `$$type`? If so, `==` should be `===`

derrell (changes requested):

source/class/qx/Class.js
johnspackman: possibly, but this is legacy code that I've preserved unchanged (including from https://github.com/qooxdoo/qooxdoo/pull/10073) ; quite a bit of this code has only limited documentation and is quite intricate, and because of how it is spread across several classes it is difficult to be sure what something is doing.

level420 (approved):

derrell (approved):

oetiker (approved):

PR #10141, opened 2021-02-23
Add migration for v7.0.0, tooling updates, test fixes

cboulanger:
This also moves CLI logging to qx.log.appender.NodeConsole and adds colorization


cboulanger:
Oh, and it does also uprade the Manifest.json JSON-Schema

source/class/qx/log/appender/NodeConsole.js
```diff
@@ -37,6 +37,32 @@ qx.Class.define("qx.log.appender.NodeConsole", {
      */
     __ERR : null,
 
+    /**
+     * Whether to use color codes
+     */
+    __useColors: false,
```
hkollmann, line 43: What's about a property for this purpose?

source/class/qx/tool/cli/Cli.js
```diff
@@ -360,7 +354,7 @@ Versions: @qooxdoo/compiler    v${qx.tool.compiler.Version.VERSION}
             let backup = filepath + ".old";
             await fs.copyFileAsync(filepath, backup);
             if (!this.argv.quiet) {
-              qx.tool.compiler.Console.warn(`*** A backup of ${lockfile} has been saved to ${backup}, in case you need to revert to it. ***`);
+              this.warn(`*** A backup of ${lockfile} has been saved to ${backup}, in case you need to revert to it. ***`);
```
johnspackman, line 357: This isn't quite right - the point of `this.warn`, `this.debug`, `this.info`, etc is to output messages for developers which can be redirected into different log appenders based on the runtime configuration for logging.  That means that (for example) if you want to debug code inside `qx.tool.cli.commands.Compile` you can increase the logging level for just that class - or you could increase it for `qx.tool.cli.commands.*`.  Or, as you did above on line 134, set the default logging to "debug" and get maximum of output for debugging.

However, these methods are not suitable for interacting with the user, who simply wants to get feedback to better understand what is happening, or have everything quiet unless there is a problem.

As a developer, you might instruct a user to add a logging appender configuration which takes output from `this.info` & `this.warn` etc and direct it to a file.  

When a user asks for output to be "verbose" they are not asking for more debugging info, they're just asking for more feedback.

It is important to be clear about these differences, otherwise code cannot add periodic logging statements with `this.info` and `this.warn` without polluting the console output for the end user - which pushes the developer into using `this.debug` everywhere, which in turn means that the developer's debug logging output is polluted with everything (because `this.debug` is usually reserved for more detailed and code-specific output).

The way to look at it is that console output to the user is "curated" and carefully thought through, but as a developer you want to be focused on how you write code which can be debugged without a debugger (eg because it's running on an user's computer)

This has definitely got muddled over the past few years and you're right that it needs to be cleaned up; there should be a central, easily accessible method to output to the user - eg `qx.tool.compiler.Cli.printVerbose` that only outputs if `--verbose` is set, or maybe `this.printVerbose` for `qx.tool.cli.commands.Command` instances

Each use of `this.warn/debug/info` etc as well as `qx.tool.compiler.Console.warn/info/etc` also needs to be reviewed

source/class/qx/tool/cli/Cli.js
```diff
@@ -110,14 +111,11 @@ qx.Class.define("qx.tool.cli.Cli", {
...
-      //  to load until we have parse the command line args
-
-      title +=
-`
-Versions: @qooxdoo/compiler    v${qx.tool.compiler.Version.VERSION}
+      title += `
+Version: v${await qx.tool.config.Utils.getQxVersion()}
 `;
       title += "\n";
+      // noinspection HtmlDeprecatedTag
```
hkollmann, line 118: Could you explain this comment?

test/cli/compiler/test-migrations.js
```diff
@@ -1,35 +1,51 @@
...
+const colorize = require('tap-colorize');
 const path = require("path");
-const process = require("process");
 const testUtils = require("../../../bin/tools/utils");
 const fsp = require("fs").promises;
 const digest = require("util").promisify(require("dirsum").digest);
-//const qx = path.resolve(__dirname, "..", "..", "..", "bin", "source", "qx");
-const qx = testUtils.getCompiler();
-const debug = false;
+//const qxCmdPath = path.resolve(__dirname, "..", "..", "..", "bin", "source", "qx");
```
hkollmann, line 7: Remove line?

hkollmann (changes requested):

johnspackman (changes requested):

docs/development/compiler/configuration/Manifest.md
```diff
@@ -79,9 +79,9 @@ This is a sample file:
     the following fields:
     - **name** : Author name.
     - **email** : Author email address.
+    - **maintainer**: Boolean, can be used to mark those authors who are currently
+      maintaining the code, as opposed to previous authors who are no longer involved.
   - **version**: Version of the library.
-  - **Qooxdoo-versions**: List of versions of the Qooxdoo framework this library
```
oetiker, line 83: is there anyway to encode that the library only works starting with a particular qx release ?

oetiker (approved):

source/class/qx/tool/cli/Cli.js
```diff
@@ -34,6 +33,8 @@ qx.Class.define("qx.tool.cli.Cli", {
       throw new Error("qx.tool.cli.Cli has already been initialized!");
     }
     qx.tool.cli.Cli.__instance = this;
+    // use node console log appender with colors
+    qx.log.appender.NodeConsole.setUseColors(true);
```
derrell, line 37: This looks to be hard-coding that colors are to be used with any instance of Cli. It should be using a value obtained from the user. Defaulting to `true` is fine as long as the user has control to turn it off.

source/class/qx/tool/migration/BaseMigration.js
```diff
@@ -93,23 +93,46 @@ qx.Class.define("qx.tool.migration.BaseMigration",{
...
+        throw new TypeError("Argument must be string or number");
+      }
+      this.setApplied(this.getApplied()+numberOfMigrations);
     },
 
     /**
-     * Marks a migration step as pending
+     * Marks one or more migration steps as pending
+     * @param {Number|String} param Optional. If number, number of migrations to mark
+     * as applied, defaults to 1; if String, message to be `announce()`ed
```
derrell, line 117: copy/paste error? `as applied` should be `as pending` here I think

source/class/qx/tool/migration/M6_0_0.js
```diff
@@ -57,10 +57,10 @@ qx.Class.define("qx.tool.migration.M6_0_0", {
       ];
       // change names in .gitignore
       if ((await this.checkFilesToRename(migrateFiles)).length) {
-        let result = await this.renameFiles(migrateFiles);
+        await this.renameFiles(migrateFiles);
```
derrell, line 60: This looks potentially dangerous. It calls `renameFiles` and subsequently looks at `dryRun`. Does `renameFiles` actually rename files? If so, maybe it should be inside the `else` block?

derrell (changes requested):

cboulanger:
Thank you all for your reviews! Much appreciated.

source/class/qx/log/appender/NodeConsole.js
cboulanger: It is a static class so the whole property boilerplate would have to be manually set up. Can we do without, given that there doesn't really seem to be a need for a getter here?

source/class/qx/tool/cli/Cli.js
cboulanger: That's my IntelliJ IDE complaining about the use of `<commands>` in the help text, which it thinks is a HTML tag - this is to switch the warning off. But I can remove it.

test/cli/compiler/test-migrations.js
cboulanger: I can do that. However, we should somehow make it possible to run the source version too. but you are right, this isn't the way to do it.

source/class/qx/tool/cli/Cli.js
cboulanger: I fully agree. User feedback has nothing to do with logging. A good indication is if the message in question would make sense to be logged to the development console in the browser, which is clearly not the case here. 

On the contrary, it seems to make sense to have a user interaction API that is agnostic as to the medium by which the message is shown to the user and her feedback is requested - it could be a console or a dialog widget. Maybe we need a new abstract API `qx.io.interaction` with its own  `qx.io.interaction.appender`s? 

But can we do that in a second step? I feel this is outside of the scope of this PR.

docs/development/compiler/configuration/Manifest.md
cboulanger: `qooxdoo-versions` has been deprecated in v6 and removed in v7 in favour of the `requires.@qooxdoo/framework` key, which is a semver range that indicates the exact range of qx versions the library  is supposed to work with.

source/class/qx/tool/cli/Cli.js
cboulanger: You are right. I will add a CLI argument to turn it on explicitly.

source/class/qx/tool/migration/BaseMigration.js
cboulanger: Yes 🙄 . Humanity will one day perish because of a copy & paste error in the nuclear launch application.

source/class/qx/tool/migration/M6_0_0.js
cboulanger: `this.renameFiles` does its own check whether `dryRun` is enabled and renames only in this case. But you are right, this is not obvious from the method name. Not sure if I should rename it or revert this magic, so that one has to write a bit more code but it's more obvious what's going on.

source/class/qx/tool/cli/Cli.js
johnspackman: I think that a new namespace with appenders would be overkill, because there is only likely to be one appender and that's `console.out`.  

The `qx.tool.compiler.Console.warn/info/debug` methods centralise the methods that are called to output to the user, and while they could benefit from detecting whether `--verbose` or `--quiet` are set and suppressing console output automatically, they are adequate.

In our gitter chats, I think I may have guided you down the route of making the changes that you have - apologies if I did, but please can you revert the calls to `qx.tool.compiler.Console.info/warn/etc`

source/class/qx/tool/cli/Cli.js
cboulanger: Done.

test/cli/compiler/test-migrations.js
cboulanger: Done.

source/class/qx/tool/cli/Cli.js
cboulanger: I disagree, I think there would be use cases for more than one appender; I also think that a central user-facing message system should be more general than `qx.tool.compiler.Console`. But this is a different question beyond the scope of this PR, and for the purpose of the PR, we should use what we have. I reverted all this.(info|warn|error) and left the debug ones. We will have to look at them carefully whether they are user interaction or log messages later. If this bugs me enough, I'll create an incubator on the basis of which we can continue the discussion.

source/class/qx/tool/migration/M6_0_0.js
cboulanger: I renamed the method to make more obvious what they do.

hkollmann (approved):

source/class/qx/tool/cli/Cli.js
cboulanger: Done.

source/class/qx/tool/migration/BaseMigration.js
cboulanger: Fixed.

derrell (approved):

test/cli/compiler/test-qx-package.js
```diff
@@ -2,37 +2,37 @@ const test = require("tape");
 const fs = require("fs");
 const testUtils = require("../../../bin/tools/utils");
 const fsPromises = testUtils.fsPromises;
-
-test("Create app", async assert => {
+require("process").chdir(__dirname);
+test("Create app", async t => {
```
johnspackman, line 6: renaming of `assert`

test/cli/compiler/test-migrations.js
```diff
@@ -1,35 +1,50 @@
...
+test.createStream().pipe(colorize()).pipe(process.stdout);
+
+// debugging
+const debug = true;
+const debugArg = debug ? "--debug --colorize" : "--colorize";
+
+function testMigration(maxVersion, numMigrationsExpected, checksumExpected) {
+  return async t => {
+    try {
+      t.comment(`>>>>>>>>>>>>>>>>>> TESTING v${maxVersion} migration <<<<<<<<<<<<<<<<<<<`);
```
johnspackman, line 19: what is `t` here?  looks like it's `tape`

johnspackman (changes requested):

test/cli/compiler/test-externals.js
```diff
@@ -1,29 +1,31 @@
 const test = require("tape");
 const testUtils = require("../../../bin/tools/utils");
 const fsPromises = testUtils.fsPromises;
+require("process").chdir(__dirname);
 
-test("test externals", async assert => {
+test("test externals", async t => {
   try {
```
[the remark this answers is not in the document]
cboulanger: It's the tape convention, which is used throughout the tape documentation.  The object passed into the test function contains assert fuctions, but many other ones to, which are nonsensical if called `assert` - for example, `assert.end()` doesn't assert that the test ends, it signals that it ends.

test/cli/compiler/test-migrations.js
cboulanger: exactly - that's why it is not good to name the object `assert`

test/cli/compiler/test-qx-package.js
cboulanger: As before, I think we should switch to the tape convention so that others aren't confused.

test/cli/compiler/test-migrations.js
cboulanger: and the function returns a function that can be fed into the tape test function, that's why it looks like it.

source/class/qx/tool/compiler/resources/ScssConverter.js
```diff
@@ -101,19 +101,19 @@ qx.Class.define("qx.tool.compiler.resources.ScssConverter", {
...
     /** @type {Boolean} Whether to copy .scss files */
     COPY_ORIGINAL_FILES: false,
-    
+
     isNewCompiler() {
       if (qx.tool.compiler.resources.ScssConverter.USE_V6_COMPILER === null) {
         console.warn("DEPRECATED: Using the Qooxdoo v5 style of SASS Compilation; this is backwards compatible " +
             "but the default will change in v7 to use the new style (see https://git.io/JfTPV for details, and how " +
-            "to disable this warning).");
+            "to disable qx.tool.compiler.Console.warning).");
```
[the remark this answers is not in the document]
cboulanger: Oups. Thanks for spotting this!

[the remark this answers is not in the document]
cboulanger: this was a sloppy RegEx Search and Replace without escaping the dot character! My bad.

test/cli/compiler/test-externals.js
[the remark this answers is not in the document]
johnspackman: hmm, I thought that I'd picked up the `assert` from tape when i wrote this; but I think that the issue remains, which is that `t` is undescriptive, confusing because it can be used (and has been used) for other things beginning with "t" (like `this`).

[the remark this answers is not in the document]
cboulanger: See https://github.com/substack/tape/#example - I assume that most users of tape follow this convention and thus I don't think it gets confused, as opposed if you call it `assert`. Not convinced?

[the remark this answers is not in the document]
cboulanger: Among other methods in the passed object t.plan(n), t.end(err), t.fail(msg), t.pass(msg), t.comment() and t.timeoutAfter(ms) all do not assert anything. We could also use a shadowed `test` or `tst` if you prefer, but I would move away from `assert`.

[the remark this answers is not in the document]
johnspackman: i'd already checked that webpage before i commented.  

I'm not convinced because I'm really concerned about other tape users, especially as we're planning on ditching tape in exchange for everything being part of qooxdoo test framework, and IMHO just because `tape` authors would like to suggest that `t` should be reserved for `tape`, that's not something which is established - quite the opposite, there are 62 uses of `var t=this` in framework and google shows that it's a common idiom.

also, imho mass changes to code without a clear reason isn't a good idea - this applies to reformatting as well as these kind of changes.

test/cli/compiler/test-migrations.js
johnspackman: renaming it to `tape` would be better than `t`, because `t.comment` looks like `this.comment` ... the reason I asked what `t` is, is because I had to go through almost the entire file to follow what was happening, and that costs time.  I could  guess that `t` was probably not `this`, but I couldn't tell if it was the outer scope `t` either.

test/cli/compiler/test-externals.js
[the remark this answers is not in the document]
cboulanger: I have a different opinion, but it is your code, and you should like it if you work on it!  I reverted.

johnspackman (approved):

level420 (approved):

cboulanger:
Thanks everybody for your thorough review!

PR #10143, opened 2021-02-23
add qxWeb support

hkollmann:


cboulanger (approved):

johnspackman (approved):

level420 (approved):

oetiker (approved):

oetiker:
can we backport this to the 6 branch please ?

hkollmann:
@oetiker : After merge?

derrell (commented):
My only concern with this is that none of us on the team use qxWeb, so turn around on finding regressions may be long...

oetiker:
@derrell I am using it :)

hkollmann:
@derrell: I agree - but discussion on gitter shows that there are people who use it. 
IMHO we will not support it. Just compile it.

oetiker:
not the website bits but the event and dom selection machinery ... which is also used by qx desktop btw, just hidden away

derrell (approved):

Issue #10147, opened 2021-02-25
call to this.base do not work any longer


hkollmann:
**Describe the bug**
this.base(arguments) do not work any longer in not precompiled code e.g. in compile.js

**To Reproduce**
checkout qxl.apiviewer
try to compile it with the new version 7.0 compiler

**Additional context**
In Object.base args.callee.base is undefined since  https://github.com/qooxdoo/qooxdoo/commit/cee7fac7ce71b38e863b736dab002c930ccf5bf6

This seems to be an effect of 
```
// If the class has it's own implementation, we need to remember that method in the
            //  mixed-in method's `.base`; wrap the method with a closure so that it can have a
            //  `.base` set, if we were to set `member.base` it would mean that the mixin can
            //  only be added into one class
            if (proto[key]) {
               member = qx.lang.Function.create(member, {always: true});
               member.base = proto[key];	              
```

hkollmann:
done with https://github.com/qooxdoo/qooxdoo/pull/10148

PR #10148, opened 2021-02-25
fix for #10147: call to this.base do not work any longer 

hkollmann:
with https://github.com/qooxdoo/qooxdoo/commit/cee7fac7ce71b38e863b736dab002c930ccf5bf6 the base function is copied to this[args.callee.name].base

### DIFF
--- a/source/class/qx/core/Object.js
+++ b/source/class/qx/core/Object.js
@@ -155,6 +155,11 @@ qx.Class.define("qx.core.Object",
 
     /**
      * Call the same method of the super class.
+     * 
+     * Either the compiler translate all calls to this.base
+     * into mypkg.MyBaseClass.prototype.myMethod.call(this, 123);
+     * this method is still needed for use in compile.js or playground
+     * which are not precompiled
      *
      * @param args {IArguments} the arguments variable of the calling method
      * @param varargs {var?} variable number of arguments passed to the overwritten function
--- a/source/class/qx/core/Object.js
+++ b/source/class/qx/core/Object.js
@@ -162,9 +162,14 @@ qx.Class.define("qx.core.Object",
      */
     base : function(args, varargs)
     {
+      let func = args.callee.base;
+      if (!func) {
+         func = this[args.callee.name].base;
+      }
+
       if (qx.core.Environment.get("qx.debug"))
       {
-        if (!qx.Bootstrap.isFunctionOrAsyncFunction(args.callee.base)) {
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

derrell (approved):

oetiker (approved):

johnspackman:
Is `qx.core.Object.base` actually called anywhere, except with the generator?  The compiler will translate `this.base` into an explicit function call, ie this:

```
  myMethod() {
    this.base(arguments, 123);
  }
```

becomes

```
  myMethod() {
    mypkg.MyBaseClass.prototype.myMethod.call(this, 123);
  }
```

johnspackman (approved):

hkollmann:
Yes it's called in all compile.js files.
They are not precompiled

source/class/qx/core/Object.js
```diff
@@ -162,9 +162,14 @@ qx.Class.define("qx.core.Object",
      */
```
johnspackman, line 162: Please can you add to the method comment, reminding that `this.base` is still required even with the compiler because there are occasions such as `compile.js` or playground where the compiler is not used

johnspackman (changes requested):

source/class/qx/core/Object.js
hkollmann: sure. Done

johnspackman (approved):

source/class/qx/core/Object.js
```diff
@@ -155,16 +155,26 @@ qx.Class.define("qx.core.Object",
...
+     * this method is still needed for use in compile.js or playground
+     * which are not precompiled
      *
      * @param args {IArguments} the arguments variable of the calling method
      * @param varargs {var?} variable number of arguments passed to the overwritten function
      * @return {var} the return value of the method of the base class.
      */
     base : function(args, varargs)
     {
+      let func = args.callee.base;
```
level420, line 170: I don't know if this is of any relevance from now on, but it doesn't make any difference here if to use `var` instead of `let` as the declaration is the first statement in the function.

level420 (changes requested):

source/class/qx/core/Object.js
hkollmann: Changed

level420 (approved):
thank you @hkollmann

PR #10155, opened 2021-02-27
extend qx search

hkollmann:
search in __dirname too. otherwise framework is not found if you ln qx and use it in an directory which is not a subdir of qooxdoo


cboulanger (approved):

cboulanger:
Just had the exact same problem!

oetiker (approved):

derrell (approved):

johnspackman (approved):

cboulanger:
thank you @hkollmann

PR #10158, opened 2021-02-28
Fix migration tests

cboulanger:
This fixes the migration tests and various bugs and inconsistencies in the `qx package` commands


level420 (approved):

oetiker (approved):

derrell (approved):

hkollmann (approved):

cboulanger:
Just added another fix: `qx package X` methods now use the application qooxdoo version now by default, not the one given by the environment. This is necessary in order to pick up a qooxdoo library path in `compile.js` (it would otherwise always use the NPM version).

johnspackman (approved):

Issue #10072, opened 2021-01-04
Errors generated from qx.ui.table.pane.Scroller


rad-pat:
I have a number of similar stack traces from our Beta system relating to `_onPointerMovePane` in `qx.ui.table.pane.Scroller`. I presume these are generated upon destruction of a table but have been unable to replicate the errors myself just yet. From the traces, I could create a PR to protect against one of the errors in `_getRowFromPagePos`, but wonder if it's perhaps just a case of disconnecting the events upon table destruction instead, or perhaps testing for if the table is being disposed in a few places. 
Once I have some feedback on best way to fix, I'm happy to put together the PR. If you have any tips to replicate the errors too, then I'd appreciate that help.

**Stack Traces**
```
TypeError: Cannot read property 'length' of null
   at i.qx.ui.table.model.Abstract.prototype.getColumnCount() [as getColumnCount] (https://plaidcloud.net/client/plaid/index.js:1:1526219)
   at i.qx.ui.table.Table.prototype.getTableColumnModel() [as getTableColumnModel] (https://plaidcloud.net/client/plaid/index.js:1:1444136)
   at i.qx.ui.table.pane.Scroller.prototype._getResizeColumnForPageX() [as _getResizeColumnForPageX] (https://plaidcloud.net/client/plaid/index.js:1:1586844)
   at i.qx.ui.table.pane.Scroller.prototype._onPointermovePane() [as _onPointermovePane] (https://plaidcloud.net/client/plaid/index.js:1:1577349)
```
```
TypeError: Cannot read property 'top' of null
   at i.qx.ui.table.pane.Scroller.prototype._getRowForPagePos() [as _getRowForPagePos] (https://plaidcloud.net/client/plaid/index.js:1:1587485)
   at i.qx.ui.table.pane.Scroller.prototype._onPointermovePane() [as _onPointermovePane] (https://plaidcloud.net/client/plaid/index.js:1:1577485)
   at https://plaidcloud.net/client/plaid/index.js:1:742127
```

rad-pat:
Fixed by #10169

PR #10159, opened 2021-03-02
Fix table errors

rad-pat:
Removes event listeners from the tablemodel when the table is destroyed to prevent the exception which is replicated in https://tinyurl.com/ybsxaq9x 
Also adds some null protection to pane and header position in the Scroller.


derrell (changes requested):
I thought it was already there, but I can't find it: I think that all event listeners should be removed by `qx.core.Object`'s superclass-of-all `dispose` method. I think maybe `qx.core.Object` should have the single added line, near the top of its `dispose`:
```
qx.event.Registration.removeAllListeners(this);
```

Thoughts?

johnspackman (approved):

oetiker (approved):

level420 (approved):

cboulanger (approved):

rad-pat:
@derrell, your suggested change seems like that would make sense but I guess everything would need a thorough test if that change was implemented.

derrell:
Agree. By visual inspection, it seems to be of no harm. I'm currently setting up to run the full test suite with that change in place. I'll let you know the results.

derrell:
@rad-pat Please see #10163. If you have a way to confirm that it fixes the problems that your PR was trying to solve, that'd be wonderful.

rad-pat:
@derrell, I gave it a quick go. It seems that I can replicate one of the errors for sure that is dealt with in this PR - the detach of event handlers from the TableModel. Easy to replicate by building a table and then setting a timeout to populate the data (as if being retrieved async from a backend), dispose the table in the meantime. Same method as https://tinyurl.com/ybsxaq9x

I had trouble replicating the mousemove events anyway but I suspect they would be fixed by your PR.

Perhaps we need the change to Table from this PR, maybe include the null protection for panePos & headerPos and then the event removal in your PR?

derrell:
@rad-pat I don't quite understand your note. Are you saying that the change provided in #10163 *does* solve (most of) the problems you were seeing; or *does not*?

If it does, then once #10163 is merged, I think modifying your PR to include only the aspects that are not related to removing listeners would be a good thing to do.

If it does not, I'm pretty interested in why it doesn't...

rad-pat:
@derrell, sorry for the unintelligible ramblings.  I think that I found a reasonable way to test the scenarios and initially found that neither PR appeared to be working if I was firing a mouse tap event on the supposedly destroyed paneScroller 2000ms after it was destroyed - I still got the event handler to fire. 

Looking further into it, the paneScrollers are destroyed via a call to `Widget.destroy()` which adds then to a dispose queue. If I change that call to a `dispose()` then your PR works well to disconnect the events immediately which means the removeListeners in my PR could be removed.

I'm not entirely sure of the implications of changing a destroy() to a dispose() here but it looks like it is required, or a call to flush the dispose queue??? https://github.com/qooxdoo/qooxdoo/blob/39aea126f77fd46a50dd699107a0d43ebb988bb0/source/class/qx/ui/table/Table.js#L1273

derrell:
@rad-pat Please pull in #10169 which should, I think, resolve that problem.

rad-pat:
@derrell, indeed that does fix that particular issue, thanks. 
I shall reduce this PR to just the necessary parts to prevent the error when a table is destroyed and the table model with which the table was built (which may not be destroyed) is subsequently updated.
Your PRs will resolve https://github.com/qooxdoo/qooxdoo/issues/10072 directly

rad-pat:
@derrell, are you happy with the selected changes I have left behind here?

derrell (approved):
LGTM! Thanks!

hkollmann (approved):

PR #10160, opened 2021-03-02
Remove the only(?) mention of a "require" class def key

cboulanger:
`qx.Class.defing("x",{ require: xxx}` seems to have been a thing is an earlier version of qooxdoo - it is no longer supported by the compiler and needs to be removed from the docs.


johnspackman (approved):

oetiker (approved):

hkollmann (approved):

hkollmann:
doc change only - 2 approvals.
Thanks @cboulanger !

PR #10163, opened 2021-03-02
Remove all listeners in qx.core.Object's destructor

derrell:
This was inspired by #10159 which added a bunch of listener removal to Table. It seemed that the proper, and more general fix, would be to do that in the superclass of all qooxdoo objects, `qx.core.Object`. This should negate the need for #10159, I believe.

The tests which fail with this change are the same ones that fail without it; this causes no new failures in the test suite. That said, it's still a dangerous change, as it affects all objects, so thorough reviews are welcome!

### DIFF
--- a/source/class/qx/core/Object.js
+++ b/source/class/qx/core/Object.js
@@ -345,6 +345,13 @@ qx.Class.define("qx.core.Object",
         }
       }
 
+      // Remove all listeners.
+      //
+      // This must be done early, since it calls
+      // qx.core.ObjectRegistry.toHashCode(target) which would add a
+      // hash code back in after code here has cleaned it up.
+      qx.event.Registration.removeAllListeners(this);
+
       // Deconstructor support for classes
       var clazz = this.constructor;
       var mixins;
@@ -372,8 +379,6 @@ qx.Class.define("qx.core.Object",
         // Jump up to next super class
         clazz = clazz.superclass;
       }
-
-      qx.event.Registration.removeAllListeners(this);
 
       this.$$disposing = false;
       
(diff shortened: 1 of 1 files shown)


hkollmann (approved):

cboulanger (approved):

level420 (approved):

oetiker (approved):
hmmm I seem to remember that there are many places where listeners are being removed explicitly (one by one) in class specific disposal code ...

this seems like a much more elegant solutions ...

but it begs the question:

* why was it not done like this in the first place ... 
* is there something problematic happening inside the removeAllListeners code ?

If this really is THE solution, then we should also get rid of all the 'specific' removal instructions ...

Since people are probably also removing listeners explicitly in their own disposal methods, how does an explicit removal act AFTER this removeAllListeners code has run ?

derrell:
I checked that, but another set of eyes to confirm would be good. The path to executing `removeListener` is

- `qx.core.MEvent`'s `removeListener()
- `qx.event.Registration.removeListener()`
- `qx.event.Manager`'s `removeListener()`
  - line 717 returns false if there is not an active listener

Agree?

oetiker:
Yes I agree, double removal seems fine ... 

I am wondering about code like this:

https://github.com/qooxdoo/qooxdoo/blob/0e800f18e7ddd320898bdb7e335abb0c7d00131a/source/class/qx/ui/mobile/form/TextField.js#L90

always getting confused about destruct and dispose ...

derrell:
Calling the `dispose` method of a class instance in turn calls the `destruct` defined in that class.

I think I'm not understanding you. It seems you're saying that multiple removals are fine but then pointing to multiple removals as an example of something that may not work...???

johnspackman (approved):

cboulanger:
thank you @derrell !

PR #10164, opened 2021-03-03
Update pullapprove.yml

cboulanger:
This mainly removes the travis check.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -2,10 +2,10 @@ version: 3
 
 # see https://docs.pullapprove.com/config/groups/
 groups:
-  # source code documentation change only: 2 approvals necessary
-  documentation:
+  documentation: documentation changes require 2 approvals
+    description:
     conditions:
-      - "'documentation' in labels"
+      - glob("docs/**/*.md") in files
     reviews:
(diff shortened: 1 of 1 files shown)


oetiker (approved):

.pullapprove.yml
```diff
@@ -2,20 +2,19 @@ version: 3
 
 # see https://docs.pullapprove.com/config/groups/
 groups:
-  # source code documentation change only: 2 approvals necessary
-  documentation:
+  documentation: documentation changes require 2 approvals
+    description:
     conditions:
-      - "'documentation' in labels"
+      - glob("docs/**/*.md") in files
```
hkollmann, line 8: would change to docs/**

hkollmann (changes requested):

.pullapprove.yml
cboulanger: I thought about this, but there is also code in `docs/` - that is, the Docsify code. But you're right, it should be possible to also change images etc. easily.

hkollmann (approved):

derrell (approved):

level420 (approved):

johnspackman (approved):

PR #10166, opened 2021-03-03
Using widget.setUserBounds() may break the layout.

lucapivato:
When a child widget is already added to a container using the Dock or Table layout managers that cache the layout children and you call setUserBounds() or resetUserBounds() the parent's layout manager is not updated and will ignore the user bounds breaking the layout.

### DIFF
--- a/source/class/qx/ui/core/LayoutItem.js
+++ b/source/class/qx/ui/core/LayoutItem.js
@@ -792,6 +792,14 @@ qx.Class.define("qx.ui.core.LayoutItem",
      */
     setUserBounds : function(left, top, width, height)
     {
+      
+      if (!this.__userBounds) {
+        var parent = this.$$parent;
+        if (parent) {
+          parent.updateLayoutProperties();
+        }
+      }
+      
       this.__userBounds = {
         left: left,
         top: top,
@@ -810,8 +818,16 @@ qx.Class.define("qx.ui.core.LayoutItem",
      */
     resetUserBounds : function()
     {
-      delete this.__userBounds;
-      qx.ui.core.queue.Layout.add(this);
+      if (this.__userBounds) {
+        delete this.__userBounds;
+
+        var parent = this.$$parent;
+        if (parent) {
+            parent.updateLayoutProperties();
+        }
+
+        qx.ui.core.queue.Layout.add(this);
+      }
     },
 
 


source/class/qx/ui/core/LayoutItem.js
```diff
@@ -792,6 +792,14 @@ qx.Class.define("qx.ui.core.LayoutItem",
      */
     setUserBounds : function(left, top, width, height)
     {
+      
+      if (!this.__userBounds) {
+        var parent = this.$$parent;
+        if (parent) {
+          parent.updateLayoutProperties();
+        }
```
derrell, line 800: `__userBounds` has not yet been set (it happens later, at line 803). Will there be any effect of calling `parent.updateLayoutProperties()` when there were no user bounds and there are still no user bounds? Won't that leave the layout exactly as it was? Should this be called after setting `__userBounds`?

derrell (changes requested):

source/class/qx/ui/core/LayoutItem.js
oetiker: I have a similar question ... if userBounds are already set, and get changed, then the updateLayoutParent would not be called ?

oetiker:
Ping @lucapivato

source/class/qx/ui/core/LayoutItem.js
lucapivato: The issue is different:
- create a widget without user bounds and add it to a container with Dock or Table layouts (the two layouts that cache the children list using getLayoutChildre(), which will include the widget just added)
- call setUserBounds() on the widget, it will move the widget but as soon as the parent performs a layout again it will break the layout and treat the widget as a docking child since the layout children list has not been updated.
- same thing happens the other way around, resetting the user bounds will not update the layout children list in the parent, unless something else causes the _invalidChildrenCache flag to be set.

Calling LayoutItem.updateLayoutProperties() is just to keep the same coding style, otherwise the call should be parent.invalidateLayoutChildren() which end up in the same place: just setting layout._invalidChildrenCache to true. So at the next layout round it will discard the cached layout children list and rebuild it with or without the user bounds children.

oetiker (approved):
now I get it :)

lucapivato:
> now I get it :)

Thanks! :) It's a border-line issue, I know, but we use the qooxdoo layout system heavily and I'm trying to get some of the enhancements/fixes in the wisej-qx fork into the main when they make sense or it's easy for me...

derrell (approved):

cboulanger (approved):

hkollmann (approved):

johnspackman (approved):

PR #10167, opened 2021-03-04
Fix description placed in wrong key

cboulanger:
Sorry I put the description in the wrong line.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -3,7 +3,7 @@ version: 3
 # see https://docs.pullapprove.com/config/groups/
 groups:
   documentation:
-    description: documentation changes require 2 approvalss
+    description: documentation changes require 2 approvals
     conditions:
       - '"docs/*" in files'
     reviews:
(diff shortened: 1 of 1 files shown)


hkollmann (approved):

oetiker (approved):

.pullapprove.yml
```diff
@@ -2,10 +2,10 @@ version: 3
 
 # see https://docs.pullapprove.com/config/groups/
 groups:
-  documentation: documentation changes require 2 approvals
-    description:
+  documentation:
+    description: documentation changes require 2 approvalss
```
derrell, line 6: One too many 's'es at the end of "approvalss"

derrell (changes requested):

.pullapprove.yml
cboulanger: Thanks, I fixed it.

derrell (approved):

level420 (approved):

johnspackman (approved):

PR #10169, opened 2021-03-05
Remove all listeners when destroying a widget (deferring disposal)

derrell:


level420 (approved):
My only concern is if this has a performance impact on „volatile“ objects like events. But lets try

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):
Oh! this PR is for qx.ui.core.Widget! then I've no problem with it!

johnspackman (approved):

PR #10170, opened 2021-03-05
enable framework tests

hkollmann:
- add test action for framework with ubuntu, windows against node 10,12,14
- enable unhandledRejection in qx.Promise
- enable parameter handling in compile.js
- fix output format with own formatters
- fix unittests for firefox and ubuntu
- use own log function in qx.tool.util.Utils.print

result of testing could be seen here: https://github.com/hkollmann/qooxdoo/actions/runs/624874104.


.github/workflows/test-framework.yml
```diff
@@ -0,0 +1,42 @@
+name: run framework tests
+
+on:
+  push:
+    branches:
+      - '*'
+      - '!master'
```
derrell, line 7: If we have tests run with master, wouldn't we be more likely to catch problems earlier?

source/class/qx/Promise.js
```diff
@@ -903,7 +902,12 @@ qx.Class.define("qx.Promise", {
         return;
       }
       qx.Promise.__initialized = true;
-      qx.bom.Event.addNativeListener(window, "unhandledrejection", qx.Promise.__onUnhandledRejection.bind(this));
+      var isNode = typeof process !== "undefined";
+      if (isNode) {
+        process.on("unhandledRejection", qx.Promise.__onUnhandledRejection.bind(this));
```
derrell, line 907: Maybe check whether the user has already registered an unhandled rejection handler? Or remove this one when they do register one?

derrell (changes requested):

oetiker (approved):
cool ...

.github/workflows/test-framework.yml
oetiker: I think tests will run on the PR but on on master after the PR has been merged

source/class/qx/Promise.js
hkollmann: This is old framework code. Only adopted to node.

derrell: Still an issue, but ok.

derrell (approved):

.github/workflows/test-framework.yml
hkollmann: Test will run on all branches except master. Therefore the !

derrell: Yes I understand that. My question was, why shouldn't the tests run on master?

cboulanger (approved):
Thank you very much for the work involved!!

.github/workflows/test-framework.yml
hkollmann: ok. 2 reasons:
  - it's tested through the pull request before therefore not neccsessary
  - it's time consuming and prevents us for mass merges.
  -

derrell: Thanks

cboulanger:
Very clever new way of parsing the test command console output!

hkollmann:
Based on idea of @oetiker

johnspackman (approved):

level420 (approved):

cboulanger:
Thank you very much @hkollmann . Tests are BACK!!

PR #10172, opened 2021-03-09
Update test workflow

cboulanger:
This hopefully allows the tests to run in a MacOS VM on GitHub. 

In general, it seems a bad idea to hardcode OS versions like this, making it necessary to manually update the qooxdoo code after each OS release.


source/class/qx/bom/client/OperatingSystem.js
```diff
@@ -126,6 +126,9 @@ qx.Bootstrap.define("qx.bom.client.OperatingSystem",
       "Win95" : "95",
 
       // OS X
+      "Mac OS X 10_15" : "10.15",
+      "Mac OS X 10_14" : "10.14",
+      "Mac OS X 10_13" : "10.13",
       "Mac OS X 10_13" : "10.13",
```
derrell, line 132: This looks like line 131 and 132 are the same.

derrell (changes requested):

derrell (commented):
@cboulanger Did you intend to check in a new file (.github/workflows/test-framework-macos.yml)? The commit message didn't indicate so.

derrell (approved):
Disregard @cboulanger. I didn't see the other commit messages.

level420 (approved):

level420 (changes requested):
@cboulanger It may be necessary to also add the dotted notation: `Mac OS X 10_14` and `Mac OS X 10.14`. same with `Mac OS X 10_15` and `Mac OS X 10.15`. one notation comes from safari the other from firefox or/and chorme afair. I may not be needed by the tests as they presumably run on webkit/safari, but for the framework in general.

cboulanger:
@level420 Can you open a separate issue for this? This PR is just about fixing the tests. I think the whole OS version sniffing stuff is broken, it shouldn't be doing string comparisons which have to be manually updated. But that requires more thought and is beyond the scope of this PR.

level420 (approved):

.github/workflows/test-framework.yml
```diff
@@ -18,25 +18,41 @@ on:
...
+    name: ${{ matrix.os }}, node ${{ matrix.node-version }}, ${{ matrix.browser }}
     runs-on: ${{ matrix.os }}
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         node-version: [10.x, 12.x, 14.x]
+        browser: [no-browser, chromium, firefox, webkit]
```
hkollmann, line 29: if you change this to on the whole handling will be done in compile.js.
The later testing is not neccesary then.

.github/workflows/test-framework.yml
```diff
@@ -18,25 +18,41 @@ on:
...
+
+      - name: install and run browser tests with ${{ matrix.browser }}
+        if: "matrix.browser != 'no-browser' && !(matrix.os == 'windows-latest' && matrix.browser == 'webkit')"
+        run:  |
+              npm test -- --browsers=${{ matrix.browser }} --terse --set-env qx.test.delay.scale=5
+
+      - name: install and run node tests
+        if: "matrix.browser == 'no-browser'"
+        run:  |
+              cd test/tool
```
hkollmann, line 57: npm test will only work in the root.

level420:
@cboulanger I've seen that one test target is macos latest which would imply that if this is really the latest then it must be macos 11.2 aka Big Sur. for this OS the `navigator.userAgent` strings are:

safari 14.0.3: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Safari/605.1.15" 
chrome 89: "Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.82 Safari/537.36"
firefox 86: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.16; rv:86.0) Gecko/20100101 Firefox/86.0"

Only chrome still reports the real current version of the OS. All others seem to be now freezing the navigator.userAgent to a descent version and in the end deprecate the userAgent property of navigator. In the future there will not be any reliable possibility to detect the OS version in a browser so we should be prepared for that.

See https://groups.google.com/a/chromium.org/g/blink-dev/c/-2JIRNMWJ7s/m/yHe4tQNLCgAJ

cboulanger:
@level420 Big Sur Safari reports "10.5.6" - I tested it on my machines and it seems to be intentional: https://bugs.webkit.org/show_bug.cgi?id=213622 . Should this ever change then we can update it. As I said, we should rather deprecate the whole API and remove any check that relies on it from the framework. But that's another issue.

.github/workflows/test-framework.yml
cboulanger: grr, this is a leftover from the previous attempt. thanks.

hkollmann (approved):

level420:
I'm just curios: did you check the output of `window.navigator.userAgent` an Big Sur and safari?

> @level420 Big Sur Safari reports "10.5.6" - I tested it on my machines and it seems to be intentional: https://bugs.webkit.org/show_bug.cgi?id=213622 . Should this ever change then we can update it. As I said, we should rather deprecate the whole API and remove any check that relies on it from the framework. But that's another issue.

source/class/qx/bom/client/OperatingSystem.js
```diff
@@ -126,6 +126,8 @@ qx.Bootstrap.define("qx.bom.client.OperatingSystem",
       "Win95" : "95",
 
       // OS X
+      "Mac OS X 10_15" : "10.15",
```
oetiker, line 129: I wonder what string Big Sur (MacOS 11.4) will output

oetiker (approved):

cboulanger:
@oetiker @level420 As said before, on MacOS 11 Safari, `window.navigator.userAgent` => `"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Safari/605.1.15" ` . It seems to be a feature, not a bug: https://bugs.webkit.org/show_bug.cgi?id=213622 .

johnspackman (approved):

PR #10174, opened 2021-03-10
Command activenes matches that of the group

voger:
When CommandGroup is set inactive and then commands are added to it, the new commands don't follow the active status of the group until that status changes. One would expect that since the group is inactive, all the commands should be inactive. Here is an example

```javascript
      const commandGroup = new qx.ui.command.Group();
      commandGroup.setActive(false);

      const command = new qx.ui.command.Command("Enter");
      command.addListener("execute", function () {
        alert("Executed");
        }, this
      );

      commandGroup.add("enter", command);
```
Working playground example https://tinyurl.com/y6vfad54

This PR fixes that by aplying CommandGroup's active status to the command, before adding it to it's list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ at the bottom. Before that, the 2 values were reversed
 
 - `qx.ui.command.Group` fixed a bug where new `qx.ui.command.Command` added
   was not set the `active` status of the group, thus staying active even if the 
-  group was inactive.
+  group was inactive. This may affect code that relied in such behavior.
 
 ## Deprecations:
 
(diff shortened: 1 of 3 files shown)


derrell (approved):

level420 (approved):

oetiker (approved):

cboulanger (approved):
This is a bug fix which breaks existing behavior (rightly so), can you please also add an entry to the release notes? Thank you!

hkollmann (approved):

voger:
> This is a bug fix which breaks existing behavior (rightly so), can you please also add an entry to the release notes? Thank you!
Done

PR #10185, opened 2021-04-13
refactor: fromFontUriToCharCode

adrelino:
working code from Image.js for Icon font paths containing the size, e.g. `@MaterialIcons/home/16` now also called from Decoration.js, e.g. used in cellrenderer.Image.

### DIFF
--- a/source/class/qx/ui/basic/Image.js
+++ b/source/class/qx/ui/basic/Image.js
@@ -936,11 +936,6 @@ qx.Class.define("qx.ui.basic.Image",
 
       if (isFont) {
         var sparts = source.split("/");
-        var fontSource = source;
-        if (sparts.length > 2) {
-          fontSource = sparts[0] + "/" + sparts[1];
-        }
-
 
         var ResourceManager = qx.util.ResourceManager.getInstance();
         var font = qx.theme.manager.Font.getInstance().resolve(source.match(/@([^/]+)/)[1]);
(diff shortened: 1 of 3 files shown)


oetiker (approved):
nice

johnspackman (approved):

derrell (approved):
This is a blind approval to move the process along, on the assumption that others on the team understand what this code is doing far better than I do, to make the merge decision.

hkollmann (approved):

source/class/qx/ui/basic/Image.js
```diff
@@ -936,11 +936,6 @@ qx.Class.define("qx.ui.basic.Image",
 
       if (isFont) {
         var sparts = source.split("/");
```
level420, line 938: Is sparts still needed?

level420 (changes requested):

source/class/qx/ui/basic/Image.js
adrelino: Yes, in line  [L945](https://github.com/adrelino/qooxdoo/blob/73c9bf4879e00a7804a881f0a6c5c8362c7be1ff/source/class/qx/ui/basic/Image.js#L954):
```js
var size = parseInt(sparts[2] || qx.theme.manager.Font.getInstance().resolve(source.match(/@([^/]+)/)[1]).getSize());
```
for the sizing of the image. 

I know that this line is duplicated now with my new code in ResourceManager, but this was for the sake of a more minimal diff. Only Image.js considers the sizing hint from the material icon, Decorator.js doesn't.

level420: Ah! I see! OK. Thank you.

level420 (approved):

level420:
5 approvals. merging. thank you @adrelino

adrelino:
Thanks for blindly merging.

Now I had some more time to create a small playground example to show what went wrong before. 

[Here](https://tinyurl.com/ydqrtbw2) is a minmal playground example with the following content:
```js
// window
var win = new qx.ui.window.Window("Table").set({
  layout : new qx.ui.layout.Grow(),
  allowClose: false,
  allowMinimize: false,
  contentPadding: 0
});
this.getRoot().add(win);
win.moveTo(30, 40);
win.open();

// table model
var tableModel = new qx.ui.table.model.Simple();
tableModel.setColumns(["Image"]);
tableModel.setData([["@MaterialIcons/home"]]);

// table
var table = new qx.ui.table.Table(tableModel)
win.add(table);

// Image in column 4
var tcm = table.getTableColumnModel();
tcm.setDataCellRenderer(0, new qx.ui.table.cellrenderer.Image());

// Create a button
var button1 = new qx.ui.form.Button("Scaled", "@MaterialIcons/home/64");
this.getRoot().add(button1,{left : 150, top  : 50});
// Add an event listener
  tableModel.setData([["@MaterialIcons/home/64"]]);
});
```
Once the playground is redeployed, the click on the button should work and not throw an error anymore.

PR #10186, opened 2021-04-14
Fix TextArea's autoSize behaviour on android tablets.

peuter:
the scrollTop value used to calculate the TextArea's height might be fractional in some cases (e.g. on an
android tablet with opened virtual keyboard). That breaks assertions in dimension values some layout managers (they expect integers).
And in these cases the autoSize feature of the TextArea is broken.

Rounding fixes the problem.

Initial error reported here: https://github.com/CometVisu/CometVisu/pull/1057#issuecomment-819035567


derrell (approved):

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

level420:
5 approvals. merging. thank you @peuter

PR #10199, opened 2021-05-28
Fixed wrong local scope in splice()

lucapivato:
Wrong local scope. Works anyway but it's better not to redefine a local in a new scope and then use it outside of the scope.

### DIFF
--- a/source/class/qx/data/Array.js
+++ b/source/class/qx/data/Array.js
@@ -368,20 +368,22 @@ qx.Class.define("qx.data.Array",
         }
       }
       // fire an event for the change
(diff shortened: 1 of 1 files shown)


derrell (approved):
Agree. That was sloppy programming. Your change is far superior.

hkollmann (approved):

oetiker (approved):

cboulanger (approved):

level420 (approved):

level420:
5 approvals. Thank you @lucapivato

PR #10200, opened 2021-06-04
Extends the Virtual DOM implementation to support Text nodes

johnspackman:
Extends the Virtual DOM implementation to support Text nodes, which is for use in node.js applications and when using JSX

### DIFF
--- a/source/class/qx/html/Root.js
+++ b/source/class/qx/html/Root.js
@@ -48,7 +48,7 @@ qx.Class.define("qx.html.Root",
     this.base(arguments);
 
     if (elem != null) {
-      this.useElement(elem);
+      this.useNode(elem);
     }
   },
 
@@ -71,7 +71,7 @@ qx.Class.define("qx.html.Root",
      * @param elem {Element} the dom element to set
      * @throws {Error} if the element is assigned again
      */
-    useElement : function(elem)
+    useNode : function(elem)
     {
       // Base call
       this.base(arguments, elem);
--- a/source/class/qx/bom/element/Attribute.js
+++ b/source/class/qx/bom/element/Attribute.js
@@ -136,8 +136,8 @@ qx.Bootstrap.define("qx.bom.element.Attribute",
         $$elementObject : 1,
 
         // Used by qx.ui.core.Widget
-        $$widget        : 1,
-        $$widgetObject  : 1,
+        $$qxObjectHash  : 1,
+        $$qxObject      : 1,
 
         // Native properties
         checked     : 1,
@@ -156,8 +156,8 @@ qx.Bootstrap.define("qx.bom.element.Attribute",
 
       qxProperties :
       {
-        $$widget : 1,
-        $$widgetObject : 1,
+        $$qxObjectHash : 1,
+        $$qxObject : 1,
         $$element : 1,
         $$elementObject : 1
       },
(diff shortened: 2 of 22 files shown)


cboulanger (approved):

source/class/qx/html/Root.js
```diff
@@ -71,7 +71,7 @@ qx.Class.define("qx.html.Root",
      * @param elem {Element} the dom element to set
      * @throws {Error} if the element is assigned again
      */
-    useElement : function(elem)
```
oetiker, line 74: this seems to be a public method name ... why change it ?

oetiker (approved):
wow ... can't really judge if this will have no side effects ... but go for it, this if master and it odes look good.

please fix the tabs

source/class/qx/html/Root.js
johnspackman: Only to better match the meaning (it's defined in `qx.html.Node`, and can be used to "use" text nodes, so "element" is a bit of a weird name); `useElement` still exists but is deprecated, and it is rare in the existing code to need to call it so i expect the change to be low impact

hkollmann (approved):

johnspackman:
> wow ... can't really judge if this will have no side effects 

Fair comment, this has been in my main repo for 9-12 months now I think, and has been in use in production for quite a while too.

Tab fixes coming up

level420 (approved):

source/class/qx/html/Root.js
oetiker: ok

oetiker:
@derrell  ping ?

derrell (approved):
Blind acceptance based on others' reviews. Too much in here I don't know enough about.

johnspackman:
thanks for reviewing; i have a bug fix i'm working on, please hold off before merging

johnspackman:
should be ready to merge now

cboulanger (approved):

rad-pat:
Seems to have had an effect on Radio & Checkboxes for me (I just tried to bring in master today)
Oh, and its not applied the Group Box header style.
Before:
image

After:
image

Is this because we have overridden some of the styles ourselves?

johnspackman:
@rad-pat please can you create a reproducable test case and file an issue?

rad-pat:
Yep, I'll try to put one together

adrelino:
I have the feeling that this is the cause for worse performance when calling `setSource` repeatetly on `qx.ui.basic.Image` as used by `ListItems` in combination with material icons.
At least I just upgraded our portal to the new qooxdoo and it now hangs a lot when interacting with the GUI.
When I click on pause in the chrome debugger while the GUI hangs, I get the following stack trace:
qooxdoo_perf_stacktrace.

`qx.ui.basic.Image.js` and `qx.html.Element.js` that are at the top of the stack trace are all files changed in this commit. 

the `_willBeSeeable` function in Node.js is marked as performance hungry, which is why I guess I end up here when clicking pause
```
    /**
     * Walk up the internal children hierarchy and
     * look if one of the children is marked as root.
     *
     * This method is quite performance hungry as it
     * really walks up recursively.
     * @return {Boolean} <code>true</code> if the element will be seeable
     */
```

My question now is, why is this function called so often now after this commit?

adrelino:
Here is a performance profile of a GUI interaction that blocked for 38 seconds! Recorded with the Performance Tab in chrome dev tools: 
performance_list_blocking
The same function names keep appearing...

johnspackman:
Can you get this into a reproducible example? One which shows the huge hits your taking would be really useful, but if not then just a comparative example where 6.x vs master has a noticeable difference

adrelino:
> Can you get this into a reproducible example? One which shows the huge hits your taking would be really useful, but if not then just a comparative example where 6.x vs master has a noticeable difference

I finally found the time to put a minimal example together:
```js
function makeSelectBox(){
  var box = new qx.ui.form.SelectBox();
  for (var i=0; i<50; i++) {
    box.add(new qx.ui.form.ListItem("Free text", "@MaterialIcons/text_fields/12"));
  }
  return box;
}

// Create a button
var button1 = new qx.ui.form.Button("Click me", "pull10200/test.png");

var doc = this.getRoot();

doc.add(button1, {left: 100, top: 50});

// Add an event listener
button1.addListener("execute", function() {
  for(var i=0; i<3; i++){
    mainSelect.setSelection([mainSelect.getChildren()[i]]);
  }
});

var mainSelect = makeSelectBox();
doc.add(mainSelect, {left: 100, top: 100});

var vBox = new qx.ui.container.Composite(new qx.ui.layout.VBox());
doc.add(vBox, { left: 100, top: 130 });
// Add an event listener
mainSelect.addListener("changeSelection", ()=>{
  var startTime = performance.now()
  vBox.removeAll();
  /* eslint no-alert: "off" */
  for (let i = 0; i < 20; i++) {
    vBox.add(makeSelectBox());
  }
  var endTime = performance.now();
  alert(endTime - startTime);
});
```

The same code above shows a huge difference in performance between qx 6 and 7, I host both version here:
Demo qooxdoo v7: https://adrelino.github.io/pull10200
Demo qooxdoo v6: https://tinyurl.com/yc94hqv3

Just click the button or change the topmost selectbox' selection to start the measurements.

with qooxdoo v6 I get consistent timings of around 200 msec at most whenever I change the selection of the topmost selectbox.
with qooxdoo v7, the first time I get a timing of around 500msec, but which each subsequent selection change the time doubles. So in addition to beeing less performant, there seems to be a garbage collection problem or something like that.

PR #10201, opened 2021-06-04
Adds CheckedList, CheckedSelectbox and data controller

johnspackman:
CheckedList (and CheckedSelectBox) are easier to use than multiple selection, because checkboxes cannot be accidentally turned off so easily and have a more appropriate appearance.  This PR also includes a controller for binding


source/class/qx/data/controller/CheckedList.js
```diff
@@ -0,0 +1,344 @@
...
+   Authors:
+     * John Spackman (github.com/johnspackman)
+
+************************************************************************ */
+
+/**
+ * Extension of `qx.data.controller.List` which adds support for `qx.ui.form.CheckedList`
+ * and `qx.ui.form.CheckedSelectBox`.
+ *
+ * The principal is thart the underlying `List` controller implementation has a model which
```
derrell, line 23: typo: `thart` vs `that`

source/class/qx/data/controller/CheckedList.js
```diff
@@ -0,0 +1,344 @@
...
+ */
+qx.Class.define("qx.data.controller.CheckedList", {
+  extend: qx.data.controller.List,
+  
+  /**
+   * Constructor
+   *
+   * @param model {qx.data.Array?} the model array
+   * @param widget {qx.ui.core.Widget?} the widget target
+   * @param path {String} the path in the model for the caption
```
derrell, line 38: `path` is optional. Both `widget` and `path` default to null. I think these should be `qx.ui.core.Widget?null` and `String?null` respectively, as in the superclass:

```
   * @param target {qx.ui.core.Widget?null} The widget which should show the
   *   ListItems.
   *
   * @param labelPath {String?null} If the model contains objects, the labelPath
   *   is the path reference to the property in these objects which should be
   *   shown as label.
```

source/class/qx/data/controller/CheckedList.js
```diff
@@ -0,0 +1,344 @@
...
+      this.__inUpdateChecked = true;
+      try {
+        // Maps of the widget item, indexed by the hashcode of the model item      
+        let children = {};
+        let toUncheck = {};
+        
+        target.getChildren().forEach(item => {
+          let itemModel = item.getModel();
+          let hash = itemModel ? itemModel.toHashCode() : "null";
+          children[hash] = item;
```
derrell, line 317: It appears possible for multiple items to hash to "null" thus overwriting items in the children map. Is this confirmed to be acceptable here?

derrell (changes requested):
Looks useful!

cboulanger (approved):
Another great addition. I assume you'll add examples to the Demo Browser and Widget Browser (also important so that Theme authors can check that their theme works with the new widgets).

oetiker (approved):

hkollmann (approved):

level420 (approved):

derrell (approved):

PR #10202, opened 2021-06-04
Additons to `qx.ui.list` and `qx.ui.virtual`

johnspackman:
There are a lot of changes in this PR, to the list and to the underlying mechanism; this solves some bugs, but adds new features:

* Auto sizing rows and columns (varying sized rows allowed)
* Wrap rows onto multiple columns
* Support multiple columns (like a table), where each column has a specific type (eg a DateColumn etc)
* Fixes bugs with label rendering

There are BC issues here, although it should be noted that some of the bug fixes meant that BC was unavoidable, and also that there was code (eg some layers) which could not work.  There is more work to be done, but this is a good start.


hkollmann (changes requested):
Could you add a comment for the BC issues in CHANGELOG.md?
Thanks.

source/class/qx/ui/list/List.js
```diff
@@ -59,911 +59,980 @@
...
-     * qx.ui.virtual.layer.WidgetCell layer.
-     *
-     * Recalculates the item sizes in a deffered call,
-     * which only happens if we have variable item heights
-     */
-    _onLayerUpdated: function () {
-      if (this.isVariableItemHeight() === false) {
-        return;
+qx.Class.define("qx.ui.list.List", {
+	extend: qx.ui.virtual.core.Scroller,
```
derrell, line 71: Without reference to coding style, per se, can we avoid use of tabs for indentation? Qooxdoo has traditionally used 2 spaces of indentation. Tabs are "defined" as moving to multiples of 8 spaces, and in fact, that's how it appears, for example, in this github PR. That much indentation makes JavaScript lines really, really long. Do you think you can replace tabs with 2 spaces each in here?

derrell (changes requested):

johnspackman:
Yes, sorry it’s supposed to be 2 spaces, must be because I switched ide recently

cboulanger (approved):
Looking forward to seeing this in action - similarly you'll probably have some examples for the widgetbrowser and demobrowser ready for when this is merged?

oetiker (approved):

hkollmann (approved):

johnspackman:
oops, i missed the ui tests for this and checkboxlist, i'll add them shortly

derrell (approved):

oetiker:
@level420 ping ?

level420 (approved):

oetiker:
There still seeem to be lots of failing tests with chrome ... guess they have a common cause ?

hkollmann:
Due to es6ify this can not be merged any longer.
Can we close it @johnspackman ?

PR #10204, opened 2021-06-04
various minor warnings and references

johnspackman:


hkollmann (approved):

source/class/qx/bom/client/Html.js
```diff
@@ -353,7 +353,7 @@ qx.Bootstrap.define("qx.bom.client.Html",
      * @ignore(Image)
      */
     getDataUrl : function(callback) {
-      var data = new Image();
+      var data = new window.Image();
```
derrell, line 356: With the advent of DOM simulation in server-side JS (Node), maybe it would make sense to avoid use of `window` and use, instead, `globalThis` which is implemented in all of the modern JS engines, and presumably mapped as necessary when transpiled, as well.

derrell, line 356: With the advent of DOM simulation in server-side JS (Node), maybe it would make sense to avoid use of `window` and use, instead, `globalThis` which is implemented in all of the modern JS engines, and presumably mapped as necessary when transpiled, as well. I have found `globalThis` to be very useful when running the same code both client- and server-side.

source/class/qx/xml/Document.js
```diff
@@ -132,7 +132,7 @@ qx.Bootstrap.define("qx.xml.Document",
       }
 
       if (qx.core.Environment.get("xml.domparser")) {
-        var parser = new DOMParser();
+        var parser = new window.DOMParser();
```
derrell, line 135: Ditto

derrell (changes requested):

oetiker (approved):

cboulanger (approved):

level420 (approved):

derrell (approved):

PR #10205, opened 2021-06-04
minor fixes to `qx.data.Array`

johnspackman:


source/class/qx/data/Array.js
```diff
@@ -368,22 +368,20 @@ qx.Class.define("qx.data.Array",
...
-      var type = "";
       var removed = amount > 0;
       var added = arguments.length > 2;
       if (removed || added) {
         var addedItems = qx.lang.Array.fromArguments(arguments, 2);
 
         if (returnArray.length == 0) {
-          type = "add";
-          end = startIndex + addedItems.length;
+          var type = "add";
```
hkollmann, line 377: this should be let do avoid global var hints!

hkollmann (changes requested):

derrell (approved):

oetiker (approved):

level420 (approved):

cboulanger (approved):

hkollmann (approved):

PR #10210, opened 2021-06-04
fixes a bug where native placeholder styling wasn't actually working

johnspackman:


hkollmann (approved):

derrell (approved):
I don't quite understand, but looks reasonable...

oetiker (approved):
very nice. I gather it does work for the existing themes ? seems that someone took a bit of a shortcut here when writing the original version ... while implementing tangible I ran into a bunch of these

johnspackman:
not tested, but I think so because its taking its cues from the theme.

level420 (approved):

cboulanger (approved):

Furynka:
@johnspackman hello, does it really makes sense to delete lines 63 and 66? AFAIK now those styles are not applied unless you use mshtml/edge engine. Otherwise it's just dead code.

I would have expected 'qx.ui.style.Stylesheet.getInstance().addRule(selector, "color: " + color);' to be called at the end.

johnspackman:
@Furynka you're right, there's something very odd about this - the code could either be substantially reduced, or (more likely) there's something wrong with the commit.  I'll have to test it and find out

PR #10211, opened 2021-06-08
fixes memory leak found by @derrell

johnspackman:


cboulanger (approved):

derrell (approved):

level420 (approved):
followed your diskussion on gitter. I wonder if removing the initialization of `this.__bindings[hash]` here https://github.com/qooxdoo/qooxdoo/pull/10211/files#diff-d0286f2fb9d580e4024e28a80c16bd74b670d4b9a4fc465cfa7e43151d7ca807L1338 with an empty array (which, if needed, shouldn't be done here), will have any negative effects elsewhere.

oetiker (approved):
the removal of the [] initialization can't be an issue since calling this method is something which happens depending what happens to the binding, so you can not relie on anything being initialized in here anyway

level420:
@oetiker yes I know. I meant that it had maybe hidden a missing initialization somewhere else.

hkollmann (approved):

PR #10212, opened 2021-06-09
fix unit test

johnspackman:
This is needed to fix other PRs

### DIFF
--- a/source/class/qx/data/Array.js
+++ b/source/class/qx/data/Array.js
@@ -969,9 +969,7 @@ qx.Class.define("qx.data.Array",
      */
     forEach : function(callback, context)
     {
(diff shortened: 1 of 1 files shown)


cboulanger (approved):
Approved, but I don't see any difference (other than using the new syntax)?

hkollmann (approved):

oetiker (approved):

oetiker:
> Approved, but I don't see any difference (other than using the new syntax)?

the difference is that `this` has a different context when using the `(x)=>` syntax

derrell (approved):

level420 (approved):
ah! with the arrow function we do not have to save `this` into the function context as it does not have its own context

PR #10213, opened 2021-06-09
Remove firefox because it always times out

johnspackman:


cboulanger (approved):
Approving, but this should only be a temporary measure that should be fixed afterwards - FF should certainly be part of the tested browsers.

hkollmann (approved):

level420 (approved):

oetiker (approved):

derrell (approved):

PR #10214, opened 2021-06-11
fix missing typescript template

hkollmann:
in the moment it's not possible to create a typescript definition file

Should be backportet to current compiler


johnspackman (approved):

cboulanger (approved):
Including a static copy of the known-good compiler makes sense to me - even if it seems out of the scope of this PR :-)

oetiker (approved):

level420 (approved):

oetiker (approved):

derrell (approved):

Issue #10146, opened 2021-02-24
COMPILER: move to dart-sass


hkollmann:
node sass is not longer supported: https://www.npmjs.com/package/node-sass

hkollmann:
should be backported to 6.0

johnspackman:
I just ported something to it and have to go back - see https://github.com/dart-lang/sdk/issues/27979#issuecomment-704462646 which says that it gets worse with jsdom (which we use)

hkollmann:
I have the same issue as i tried to ported it..

PR #10215, opened 2021-06-11
Move to sass

hkollmann:
see https://github.com/qooxdoo/qooxdoo/issues/10146.
I found a way to trick out the sass loading problem.
This should solve all node-sass problems.

Should be backportet to compiler


cboulanger (approved):
This is excellent, thank you very much!

johnspackman:
Does this solve the compatibility problem with jsdom?

hkollmann:
> Does this solve the compatibility problem with jsdom?

Not in generally. It solves the problem loading sass in qooxdoo compiler. We do not use jsdom there - we use outr own generated windows problem.

You can use this approach to load sass with jsdom. I have'nt tried it it.
The idea is to trick out sass:

  process.versions.electron = "0.0.0";     Tells sass that we are in a node enviroment
  window.document.currentScript = {src: ""}; Prevents sass to access to the document.scripts variable

oetiker:
sass seems quite a bit slower ... but then again, I guess we do not have a performance problem there, or do we?
https://www.peterbe.com/plog/quick-comparison-between-sass-and-node-sass

oetiker (approved):

level420 (approved):

derrell (approved):

johnspackman (approved):

PR #10217, opened 2021-06-15
add a timeStamp macro to the compiler

level420:
`${timeStamp}` macro which is replaced by the seconds since epoch at compile/deploy time which could be used to prevent browser caching in URLs


cboulanger (approved):

derrell (approved):

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

level420:
5 approvals. merging

PR #10218, opened 2021-06-15
Re-enable NPM deployment

cboulanger:
Has been dry-run-tested in my fork. See big caveat in file. This is all I could do in an evening and not anything that can replace a real automated workflow.


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,53 @@
...
+# i.e. `npm install @qooxdoo/qooxdoo@beta` will get you the last beta version
+# published, anything else goes to the `latest` tag.
+#
+# This workflow rests on the highly dubious assumption that pushes to the master
+# branch will always contain a "beta" version and pushes to the "stable" branch
+# (currently 6.0), such as for bug fix releases or backports will not.
+#
+# Every push involves incrementing the version number by hand, otherwise no
+# package will be published - the workflow won't do it for you. A beta version
+# will need some kind of secondary versioning such as 7.0.0-beta.202106152115
```
derrell, line 16: Where would that incrementing of the version number be done? I'm not seeing where a version number comes from, unless it's related to `uses: martinbeentjes/npm-get-version-action@master`

cboulanger: The individual PR author needs to do it herself, in the source, in package.json. Unless the version is changed, it won't get published to NPM.

cboulanger: This is the least-common-denominator approach here because any automated versioning requires that we have a consensus on the versioning policy first, which we don't have.

cboulanger: This let's us decide when and with which version we release an NPM package - we can do backports and 6.x. fixes this way and we can periodically release a 7.0.0-beta.XXXXXXX package.

derrell: Ok, so if I'm understanding properly, the "version" line in `package.json` would get changed from
```
  "version": "7.0.0-beta",
```
to something like, in your example,
```
  "version": "7.0.0-beta.202106152115",
```
And if a new version isn't published if the version number doesn't change, then this means that it really is under developer control whether a new version is published. The developer who wants to publish a new version updates `package.json` and that change gets merged into master, and a new version gets published. This sounds perfect as a quick-to-implement solution. Did I describe it correctly?

cboulanger: Yes, exactly

derrell (approved):
Yay!

johnspackman (approved):

oetiker (approved):

level420 (approved):

hkollmann (approved):

PR #10223, opened 2021-06-22
Detects a corruption problem caused by duplicate target output directories

### LINKED ISSUE #10222: Private member in a mixin can overwrite a private member in class it's mixed into
In my app, I've found that in the build version, the `__Property` member of `qx.core.Object`  is found to be null in the destructor, so when it goes to dereference `$$store`, it crashes.

I have tracked down the cause of this problem to be that qx.core.Object uses mixin qx.core.MObjectId, and in the build version, both `qx.core.Object`'s `__Property` member and `qx.core.MObjectId`'s `__ownedQxObjects` member are assigned private variable name `__P_26_0`. It looks like the order of instantiation in Class would have `qx.core.Object`'s members list initialized first, and then mixins are loaded so their members lists are initialized. What appears to be happening is that `qx.core.Object`'s `__Property` (`this.__P_26_0`) is initially assigned `qx.core.Property` as expected, but when the `qx.core.MObjectId` mixin is loaded, its `__ownedQxObjects` (again, `this.__P_26_0`) is assigned `null`. It looks to me like it's not feasible/allowable to ever assign the same private variable name like `__P_26_0` to a mixin as to a class, since privates are expected to be unique even between mixins and the classes they're mixed in to.

I confirmed this is the cause of the crash problem by manually changing, in `compiled/build/index.js`, all occurrences of `__P_26_0` within the mixin definition to have a different name. The problem then went away.

johnspackman:
Detects a corruption problem caused by duplicate target output directories

fixes https://github.com/qooxdoo/qooxdoo/issues/10222

### DIFF
--- a/source/class/qx/tool/cli/commands/Compile.js
+++ b/source/class/qx/tool/cli/commands/Compile.js
@@ -733,6 +733,7 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
       /*
        * There is still only one target per maker, so convert our list of targetConfigs into an array of makers
        */
+      let targetOutputPaths = {};
       let makers = [];
       targetConfigs.forEach(targetConfig => {
         if (!targetConfig.appConfigs) {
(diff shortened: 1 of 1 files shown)


derrell (approved):

cboulanger (approved):

derrell (approved):
Thanks!

oetiker (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman

PR #10224, opened 2021-06-22
Fixes a bug where qooxdoo classes were not preserved

### LINKED ISSUE #10221: Initial render of RadioButton is incorrect
Since https://github.com/qooxdoo/qooxdoo/pull/10200, the initial render of a Radiobutton (or a checkbox) is not correct. Once clicked on by mouse it appears to render ok. 

I've attached a sample project that replicates the issue for me. 
[displaybug.zip](https://github.com/qooxdoo/qooxdoo/files/6676125/displaybug.zip)

Also seems broken by the same PR, a custom decoration to apply a css transform to make a Groupbox label uppercase. This is included in the project.

**Screenshots**
Before:
image

After:
image

 - OS: Ubuntu 18.04
 - Browser:  Chrome 90.0.4430.93

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/10221

### DIFF
--- a/source/class/qx/html/Node.js
+++ b/source/class/qx/html/Node.js
@@ -793,7 +793,12 @@ qx.Class.define("qx.html.Node",
       if (data)
       {
         // Import listeners
-        qx.event.Registration.getManager(elem).importListeners(elem, data);
+        let domEvents = {};
+        let manager = qx.event.Registration.getManager(elem);
+        for (let id in data)
+          if (manager.findHandler(elem, data[id].type))
+            domEvents[id] = data[id];
+        qx.event.Registration.getManager(elem).importListeners(elem, domEvents);
 
         // Cleanup event map
         // Events are directly attached through event manager
--- a/source/class/qx/html/Node.js
+++ b/source/class/qx/html/Node.js
@@ -795,9 +795,11 @@ qx.Class.define("qx.html.Node",
         // Import listeners
         let domEvents = {};
         let manager = qx.event.Registration.getManager(elem);
-        for (let id in data)
-          if (manager.findHandler(elem, data[id].type))
+        for (let id in data) {
+          if (manager.findHandler(elem, data[id].type)) {
             domEvents[id] = data[id];
+          }
+        }
         qx.event.Registration.getManager(elem).importListeners(elem, domEvents);
 
         // Cleanup event map
--- a/source/class/qx/html/Element.js
+++ b/source/class/qx/html/Element.js
@@ -764,8 +764,6 @@ qx.Class.define("qx.html.Element",
       this.base(arguments, fromMarkup, propertiesFromDom);
       var elem = this._domNode;
       
-      if (this.$$x)
-        debugger;
       // Copy attributes
       var data = this.__attribValues;
       if (data) {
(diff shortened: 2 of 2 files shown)


cboulanger (approved):

source/class/qx/html/Node.js
```diff
@@ -793,7 +793,12 @@ qx.Class.define("qx.html.Node",
       if (data)
       {
         // Import listeners
-        qx.event.Registration.getManager(elem).importListeners(elem, data);
+        let domEvents = {};
+        let manager = qx.event.Registration.getManager(elem);
+        for (let id in data)
+          if (manager.findHandler(elem, data[id].type))
+            domEvents[id] = data[id];
+        qx.event.Registration.getManager(elem).importListeners(elem, domEvents);
```
derrell, line 801: We've seen a recent bug caused by someone adding statements within an `if` or `for` but neglecting to add braces when braces were previously not technically necessary because there was only a single statement. To avoid potential future problems, would you mind adding braces to the `for` and `if` even though there is only a single statement in each at present?

derrell (changes requested):

source/class/qx/html/Node.js
johnspackman: I'm in two minds about this, firstly because my objection to unnecessary curly braces "as a rule" is that I think that it sounds like the error you're talking about might have been because of an assumption that there will always be curly braces made it easy to rely on indentation as an indicator, but my guess is that makes it possible to make the same error in reverse, ie have a break in indentation and wrongly intended braces mislead the eye in the same way.  Lint picks up inappropriate indentation (unless you use `lint fix`, IMHO `lint fix` should be avoided for this reason)

But apparently our lint configuration no longer requires curly braces, and this change has "crept in through the back door", and probably from changes that I made in the compiler repo to relax the linting rules; this has obviously come through into the framework without discussion and that's unintentional and regrettable, because lint changes should be a matter of discussion and agreement, regardless of my opinion.  I presume that it's a side effect of merging the repos that we either didn't discuss or I didn't pick up on when it was raised.

I'm not sure how we resolve this going forward, and would suggest some kind of discussion about it - for the moment, I'm going to add the braces here to preserve the status quo while we figure out what to do (which might mean reinstating the lint check and bulk updating the code to comply), but I would ask that we make a decision fairly soon so that PRs can be automatically reviewed for this kind of compliance

derrell (approved):
Thanks for making the change got now, and yes, let's discuss for long-term policy.

source/class/qx/html/Node.js
oetiker: I would also prefer to have braces everywhere ... for consistencies sake ...

oetiker (approved):

source/class/qx/html/Node.js
hkollmann: Me too!

hkollmann (approved):

source/class/qx/html/Node.js
johnspackman: I take it back, those lint rules are not turned off!  The reason that curly braces is not enforced any more is something to do with eslint config; the eslint plugin config still requires curly braces ([see here](https://github.com/qooxdoo/eslint-config-qx/blob/master/index.js#L80-L84)) but `qx lint` is not picking up that config.

See https://github.com/qooxdoo/qooxdoo/issues/10225

level420 (approved):

derrell:
Thanks, @johnspackman

PR #10233, opened 2021-07-16
Character detection in `qx.lang.String`

johnspackman:


hkollmann (approved):

source/class/qx/lang/String.js
```diff
@@ -146,6 +149,57 @@ qx.Bootstrap.define("qx.lang.String",
...
+     */
+     isUpperCase(str) {
+      if (str.length == 0)
+        return false;
+      for (let i = 0; i < str.length; i++) {
+        let type = qx.lang.String.__characterTypes[str[i]];
+        if (type !== "upper")
+          return false;
+      }
+      return true;
```
oetiker, line 165: how about this

```suggestion
      return qx.lang.String.__characterRx.upper.test(str);
```

source/class/qx/lang/String.js
```diff
@@ -396,5 +450,16 @@ qx.Bootstrap.define("qx.lang.String",
...
+
+  defer(statics) {
+    let tmp = {};
+    for (let str = "ABCDEFGHIJKLMNOPQRSTUVWXYZ", i = 0; i < str.length; i++)
+      tmp[str[i]] = "upper";
+    for (let str = "abcdefghijklmnopqrstuvwxyz", i = 0; i < str.length; i++)
+      tmp[str[i]] = "lower";
+    for (let str = "0123456789", i = 0; i < str.length; i++)
+      tmp[str[i]] = "digit";
+    statics.__characterTypes = tmp;
```
oetiker, line 463:
```suggestion
    statics.__characterRx = {
      upper: RegExp(/^\p{General_Category=Uppercase_Letter}+$/u),
      lower: RegExp(/^\p{General_Category=Lowercase_Letter}+$/u),
      digit: RegExp(/^\p{General_Category=Decimal_Number}+$/u)
    };
```

source/class/qx/lang/String.js
```diff
@@ -146,6 +149,57 @@ qx.Bootstrap.define("qx.lang.String",
...
+     */
+     isLowerCase(str) {
+      if (str.length == 0)
+        return false;
+      for (let i = 0; i < str.length; i++) {
+        let type = qx.lang.String.__characterTypes[str[i]];
+        if (type !== "lower")
+          return false;
+      }
+      return true;
```
oetiker, line 182:
```suggestion
      return qx.lang.String.__characterRx.lower.test(str);
```

source/class/qx/lang/String.js
```diff
@@ -146,6 +149,57 @@ qx.Bootstrap.define("qx.lang.String",
...
+     */
+     isDigits(str) {
+      if (str.length == 0)
+        return false;
+      for (let i = 0; i < str.length; i++) {
+        let type = qx.lang.String.__characterTypes[str[i]];
+        if (type !== "digit")
+          return false;
+      }
+      return true;
```
oetiker, line 199:
```suggestion
      return qx.lang.String.__characterRx.digit.test(str);
```

oetiker (changes requested):
this would probably be faster AND take care of things like Ä and Ç and so on as well

source/class/qx/test/lang/String.js
```diff
@@ -28,6 +28,30 @@ qx.Class.define("qx.test.lang.String",
...
+      this.assertTrue(!qx.lang.String.isLowerCase("A"));
+      this.assertTrue(!qx.lang.String.isLowerCase("S"));
+      this.assertTrue(!qx.lang.String.isLowerCase("X"));
+      this.assertTrue(!qx.lang.String.isLowerCase("4"));
+      this.assertTrue(!qx.lang.String.isUpperCase("b"));
+      this.assertTrue(!qx.lang.String.isUpperCase("t"));
+      this.assertTrue(!qx.lang.String.isUpperCase("z"));
+      this.assertTrue(!qx.lang.String.isUpperCase("3"));
+      this.assertTrue(!qx.lang.String.isLowerCase("2"));
+      this.assertTrue(!qx.lang.String.isUpperCase("5"));
```
derrell, line 51: These tests would be easier to read as
```
this.assertFalse(qx.lang.String.isUpperCase("5"));
```

I couldn't figure out why it was checking for a number being uppercase, until I saw that hard-to-see `!`. I'll approve anyway, but I suggest making this change unless I'm missing something.

derrell (approved):

source/class/qx/lang/String.js
johnspackman: i assumed that regex would be massively slower, but didn't try it.  I'll work up some quick tests

oetiker: There is also the problem of international language support ... with the proposed implementation, the method would have to called `isEnglishUppercase`  :)

cboulanger (commented):
Yes please use RegEx to make this truely international - otherwise it wouldn't catch "Ä" or "Ô".

source/class/qx/lang/String.js
johnspackman: fair comment.  I've compared jsfiddle https://jsfiddle.net/srcgux9n/ and in Chrome there is between 35% and 50% overhead for regex, but it still does 1,000,000 regex in under 50ms - Firefox RegEx is faster than maps!

cboulanger (approved):

oetiker (approved):

level420 (approved):

PR #10235, opened 2021-07-16
adds an environment variable so that startup timings are optional

johnspackman:
for a cleaner startup (most visible in node apps)


hkollmann (approved):

oetiker (approved):
it would be nice to know about this somewhere in the documentation ...

derrell (approved):

cboulanger (approved):
As @oetiker said, it would be great to document this somewhere, but we don't have a comprehensive list (or even a non-comprehensive one) in our docs. That's probably something that should be auto-generated from the source, but that's another issue.

level420 (approved):

PR #10236, opened 2021-07-16
adds better warnings when fonts or font icons cannot be resolved

johnspackman:
the previous code would crash with an obscure error


hkollmann (approved):

oetiker (approved):

derrell (approved):

cboulanger (approved):
Thanks, the only thing being that our style normally requires curly braces everywhere - the ususal 😄

level420 (approved):

hkollmann:
Here we have a regression in tests: 
not ok 106 - qx.test.ui.basic.Image:testWebFontImage - [110.20 ms] - Cannot determine charCode from source: @FontAwesome/f004

PR #10237, opened 2021-07-17
all nodes >= 8.x should be ok

oetiker:


derrell (approved):

cboulanger (approved):

docs/README.md
```diff
@@ -24,7 +24,7 @@ your application that were referenced either directly or indirectly from your
 code.
 
 The Qooxdoo compiler is called `qx` and runs in node.js. Qooxdoo requires
-Node.js version 8.x or 10.x. to work.
+Node.js version  >= 8.x to work.
```
johnspackman, line 28: do we really have to support node v8, or even v10?  AIUI they are both way past end-of-life, node 12 is in "maintenance" mode and will end of life in April 2022 (only 9 months away).

As a general rule, the framework is likely to support older versions of node because there is very little in the framework (if anything) which is node-version specific, principally because we can transpile to support language features on older node versions.

But, transpilation only allows us to support language features, it does not include a backporting of the node runtime.  For example, `require("fs").promises` was (I think) added in Node v10, and therefore any Qooxdoo code which uses it will not run on Node v8.

The node runtime will only affect tools like the compiler (assuming that we do not start adding framework utility classes that are node-specific), but new versions of node will always bring new runtime features that we could start using.

IMHO this means that the compiler must run on current LTS node at a minimum (currently v12) but can target older versions; we could state that we "endeavour" to support end-of-life versions of node as a target, but that is not a guarantee.

Ideally, we would say that the compiler actually requires the current _active_ version of node (currently node v14) so that we can have the freedom to use the latest and greatest features.

https://nodejs.org/en/about/releases/

johnspackman (changes requested):

docs/README.md
cboulanger: you are right, we should not support end-of-life versions, at least not explicitly. If people have special requirements, we can still look into it, but we should not encourage reliance on legacy versions.

derrell: The reason for supporting Node 8 is that it's the last version that has i386 builds. That was critical to me in my prior job, and may be important for others... We don't need to claim support of it, but we should continue, I think, transpiling by default to code that will run in Node 8. I don't see any reason that we must support building with Node 8, however.

oetiker: So how about we just say `current version of Node.js` ?

level420 (approved):

docs/README.md
hkollmann: +1

cboulanger (changes requested):
As people with no-longer-supported versions of Node experience problems, we should explicitly discourage the use of these versions. We can always find solutions for special use cases that require older versions, but this should not be advertised. The language should thus be `Node.js version >= 12.x`.

cboulanger (approved):

hkollmann:
@oetiker : Could you change the text to "current version of Node.js"?

hkollmann (approved):

johnspackman (approved):

PR #10238, opened 2021-07-19
forces curly braces on all block statements and corrects framework code to comply

### LINKED ISSUE #10225: Linting is broken
`qx lint` is not working as expected - it does not enforce things like "always have curly braces", which is specified here: https://github.com/qooxdoo/eslint-config-qx/blob/master/index.js#L80-L84

Adding an eslint config to package.json does not work, eg this code:

```
  "eslintConfig": {
    "extends": [
      "@qooxdoo/eslint-config-qx"
    ],
    "rules": {
      "curly": "error"
    }
  },
```

should, in theory (and according to the eslint-config-qx README), configure eslint to take the config in `@qooxdoo/eslint-config-qx` and turn on the curly rule, but in fact it just causes all of the eslint rules to revert to default and *only* the `curly` rule to change.

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/10225

### DIFF
--- a/compile.json
+++ b/compile.json
@@ -247,7 +247,7 @@
        "no-unsafe-negation": "off",
        "new-parens": "off",
        "semi": "off",
-       "curly": "off",
+       "curly": "error",
        "no-mixed-spaces-and-tabs": "off",
        "no-extra-semi": "off",
       "indent": "off",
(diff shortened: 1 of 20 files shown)


cboulanger (approved):

derrell (approved):
Great! (This is actually fewer changes to conform to the new eslint requirement than I would have expected.)

hkollmann (approved):

oetiker (approved):

level420 (approved):

PR #10243, opened 2021-07-28
makes the format string available from `qx.util.format.DateFormat`

johnspackman:


cboulanger (approved):

source/class/qx/util/format/DateFormat.js
```diff
@@ -425,6 +425,15 @@ qx.Class.define("qx.util.format.DateFormat",
       return locale;
     },
 
+    /**
+     * Returns the original format string
+     * 
+     * @returns {String}
```
derrell, line 431: The annotation here should be `@return` (singular)

derrell (changes requested):

oetiker (approved):

hkollmann (approved):

level420 (approved):

derrell (approved):

PR #10247, opened 2021-07-30
modifies the URL which are output so that they include the timestamp …

johnspackman:
…of the file; this allows strong (even immutable) caching of source files without stale content

### DIFF
--- a/source/class/qx/tool/compiler/targets/meta/AbstractJavascriptMeta.js
+++ b/source/class/qx/tool/compiler/targets/meta/AbstractJavascriptMeta.js
@@ -1,3 +1,25 @@
+/* ************************************************************************
+ *
+ *    qooxdoo-compiler - node.js based replacement for the Qooxdoo python
+ *    toolchain
+ *
+ *    https://github.com/qooxdoo/qooxdoo-compiler
+ *
+ *    Copyright:
+ *      2011-2021 Zenesis Limited, http://www.zenesis.com
+ *
+ *    License:
+ *      MIT: https://opensource.org/licenses/MIT
+ *
+ *      This software is provided under the same licensing terms as Qooxdoo,
+ *      please see the LICENSE file in the Qooxdoo project's top-level directory
+ *      for details.
+ *
+ *    Authors:
+ *      * John Spackman (john.spackman@zenesis.com, @johnspackman)
+ *
+ * ************************************************************************/
+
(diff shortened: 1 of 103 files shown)


source/class/qx/tool/compiler/targets/meta/AbstractJavascriptMeta.js
```diff
@@ -1,3 +1,25 @@
+/* ************************************************************************
+ *
+ *    qooxdoo-compiler - node.js based replacement for the Qooxdoo python
+ *    toolchain
+ *
+ *    https://github.com/qooxdoo/qooxdoo-compiler
```
derrell, line 6: This isn't right any longer, is it? (Might be a global change that should be made in a separate PR.)

derrell (approved):

oetiker (approved):

level420 (approved):

hkollmann (approved):

source/class/qx/tool/compiler/targets/meta/AbstractJavascriptMeta.js
johnspackman: yes - rather than fix it a bit here and a bit elsewhere, I've done a fix-it-everywhere PR in https://github.com/qooxdoo/qooxdoo/pull/10249

johnspackman:
The last few commits include a fix where the webfont sources can be overridden by the application - eg the application includes the paid-for FontAwesome Pro, where the package/framework only use the open source FontAwesome Free, in which case the font name will be the same, but the font files are different.

cboulanger (approved):

hkollmann (changes requested):
Checks are not working any longer. Could you repair them?
Thanks!

cboulanger:
If this is a node10 problem, I suggest we remove node 10

hkollmann:
No it's a problem with the new parameter

johnspackman:
lint error, curse of the curly braces.  i'll fix it shortly.

hkollmann:
Hi John,
problem is this test in test-deps.js: 
src.match(/src="appone\/index.js"/)

the newly created index.html file contains 

src="appone/index.js${indexJsTimestamp}"

should i fix the test in your PR?

johnspackman:
@hkollmann np, just done it

hkollmann (approved):

PR #10250, opened 2021-08-11
bug in detecting whether a webfont has been loaded

johnspackman:
fixes a minor issue where a listener is attached to detect the webfont being loaded, but if the webfont is already loaded the event will never fire in practice

### DIFF
--- a/source/class/qx/bom/webfonts/WebFont.js
+++ b/source/class/qx/bom/webfonts/WebFont.js
@@ -57,7 +57,7 @@ qx.Class.define("qx.bom.webfonts.WebFont", {
     },
 
     /**
-     * Fired when the font has loaded successfully
+     * Indicates that the font has loaded successfully
      */
     valid: {
       init: false,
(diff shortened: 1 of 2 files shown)


oetiker (approved):

level420 (approved):

lucapivato:
And the same listener is added in AbstractFIeld.js, it should probably check isValid() there too.

source/class/qx/bom/webfonts/WebFont.js
```diff
@@ -93,6 +102,7 @@ qx.Class.define("qx.bom.webfonts.WebFont", {
     _onWebFontChangeStatus : function(ev)
     {
       var result = ev.getData();
+      this.setValid(!!result.valid);
```
derrell, line 105: Are you sure this accomplishes the goal? The `changeValid` event will not be fired if the property is set to the same value it already has. Is there any possibility of that occurring here?

derrell (changes requested):

derrell (approved):

hkollmann (approved):

cboulanger (approved):

PR #10252, opened 2021-08-24
Fixes CSP error on inline script

rad-pat:
When running our Qooxdoo app with a tighter security policy, the `onload` attribute of an iFrame is not permitted. 
This amendment appears to fix the issue, and should (hopefully) be backwards compatible if someone has specified an `onload` attribute directly.

### DIFF
--- a/source/class/qx/bom/Iframe.js
+++ b/source/class/qx/bom/Iframe.js
@@ -39,7 +39,6 @@ qx.Class.define("qx.bom.Iframe",
      */
     DEFAULT_ATTRIBUTES :
     {
-      onload : "qx.event.handler.Iframe.onevent(this)",
       frameBorder: 0,
       frameSpacing: 0,
       marginWidth: 0,
@@ -66,14 +65,20 @@ qx.Class.define("qx.bom.Iframe",
       var attributes = attributes ? qx.lang.Object.clone(attributes) : {};
       var initValues = qx.bom.Iframe.DEFAULT_ATTRIBUTES;
 
-      for (var key in initValues)
-      {
-        if (attributes[key] == null) {
+      for (var key in initValues) {
+        if (!(key in attributes)) {
           attributes[key] = initValues[key];
         }
       }
 
-      return qx.dom.Element.create("iframe", attributes, win);
+      var elem = qx.dom.Element.create("iframe", attributes, win);
+      if (!("onload" in attributes)) {
+        elem.onload = function() {
+          qx.event.handler.Iframe.onevent(elem);
+        };
+      }
+      return elem;
+
     },
 
 


oetiker (approved):

johnspackman (approved):

cboulanger (approved):

source/class/qx/bom/Iframe.js
```diff
@@ -73,7 +72,11 @@ qx.Class.define("qx.bom.Iframe",
         }
       }
 
-      return qx.dom.Element.create("iframe", attributes, win);
+      var elem = qx.dom.Element.create("iframe", attributes, win);
+      if (attributes["onload"] == null) {
```
derrell, line 76: This doesn't look right. Are you really checking for `null`? If so, it should use `===` instead of `==`. Are you instead checking for whether `onload` is not in the `attributes` map? In that case, I'd recommend either the specific  `if (! ("onload" in attributes))` or the more generic but oft-used, `if (! attributes["onload"])`. (I'd recommend the former.)

derrell (changes requested):

source/class/qx/bom/Iframe.js
rad-pat: Yep, I'll fix it. I was just following form from line 70, but I'll change that too

derrell (approved):

level420 (approved):

level420 (changes requested):
@rad-pat It seems that some iframe specific tests are failing now:
```
not ok 257 - qx.test.ui.embed.Iframe:testHiddenSetSource - [1.10 ms] - Timeout of 10000 ms reached before resume() was called.
not ok 303 - qx.test.ui.embed.Iframe:testGetWindow - [11.70 ms] - Timeout of 10000 ms reached before resume() was called.
not ok 2562 - qx.test.dom.Hierarchy:testIsRenderedIframe - [4.80 ms] - Timeout of 10000 ms reached before resume() was called.
not ok 2674 - qx.test.html.Iframe:test: set source to URL with same origin - [4.00 ms] - Timeout of 10000 ms reached before resume() was called.
not ok 2675 - qx.test.html.Iframe:test: update source on navigate - [20.00 ms] - Expected value to be a string but found null!
not ok 2845 - qx.test.html.Iframe:test: skip setting source if frame is already on URL - [4.00 ms] - Timeout of 10000 ms reached before resume() was called.
not ok 2853 - qx.test.html.Iframe:test: set null source if frame is cross-origin - [4.20 ms] - Timeout of 10000 ms reached before resume() was called.
DONE testing 3230 ok, 7 not ok, 176 skipped - [380094 ms]
Test 'framework test' failed with exit code 7.
```

rad-pat:
Hmm, ok. I'll check it out.

rad-pat:
I need a clue please. I've tried `qx package install --save=0 qooxdoo/qxl.testtapper` but I either get 
```
Error: Error: Error while reading compile.js
Error: Cannot find module 'upath'
```
when I run that in Qooxdoo library path, or 
```
Error: 'qooxdoo/qxl.testtapper' has no stable release compatible with qooxdoo version 7.0.0-beta.
             To install anyways, use --release x.y.z.
             Please ask the library maintainer to release a compatible version
```
when I run it under my own project folder

hkollmann:
Testtapper is already installed in the framework.

Just try to run npm test

derrell (commented):
@rad-pat What was the error in the prior versions, causing the tests to fail, i.e., what error was actually generated that was hidden by the tests?

rad-pat:
Honestly, I'm not 100% sure, but I *think* by capturing the `load` event and then using `qx.event.handler.Iframe.onevent`, it was then firing a `load` event and I wonder if it caused some stack issue....

level420 (approved):
Now all tests pass. Thank you @rad-pat !

level420:
5 approvals. merging. thank you @rad-pat

PR #10255, opened 2021-10-05
Adds letter-spacing to Font Modification

ArturWesner:
See https://developer.mozilla.org/de/docs/Web/CSS/letter-spacing

### DIFF
--- a/source/class/qx/test/bom/Font.js
+++ b/source/class/qx/test/bom/Font.js
@@ -189,7 +189,7 @@ qx.Class.define("qx.test.bom.Font",
       var keys = Object.keys(styles);
 
       this.assertMap(styles, "Method 'getStyles' should return a map!");
-      this.assertEquals(9, qx.lang.Object.getLength(styles), "Map should contain 8 key!");
+      this.assertEquals(9, qx.lang.Object.getLength(styles), "Map should contain 9 keys!");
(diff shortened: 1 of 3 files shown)


source/class/qx/test/bom/Font.js
```diff
@@ -181,7 +189,7 @@ qx.Class.define("qx.test.bom.Font",
       var keys = Object.keys(styles);
 
       this.assertMap(styles, "Method 'getStyles' should return a map!");
-      this.assertEquals(8, qx.lang.Object.getLength(styles), "Map should contain 8 key!");
+      this.assertEquals(9, qx.lang.Object.getLength(styles), "Map should contain 8 key!");
```
derrell, line 192: Error message should say "Map should contain 9 keys!" (changes both the number and fixes the plural of "key")

derrell (changes requested):

source/class/qx/test/bom/Font.js
ArturWesner: Oops, missed that. Thanks!

derrell (approved):

johnspackman (approved):

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

Issue #9685, opened 2019-06-05
Better accessibility for qooxdoo apps (Implement ARIA)


cboulanger:
Currently, qooxdoo lacks any special support for screenreaders and other devices that help, for example, sight-impaired users to navigate and use qooxdoo applications. 
We need to implement [ARIA roles, properties and states](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basics) in the HTML generated by the qooxdoo runtime.

hkollmann:
Done

PR #10277, opened 2021-11-12
Adds more accessibility support by adding some ARIA attrs

ArturWesner:
Hello qooxdoo Team,

my company needs more accessibility support for our project and i got the task to accomplish this!
So i had a look into the code and found that qooxdoo does not implement the ARIA specifications yet.
[WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/) adds semantic information to the web content in order to allow assistive technologies to better help persons with disabilities.
I added ARIA attributes to several components and modified a few focus handling interactions. Sadly i didn't have the time to work on all qx components (yet). On monday i will add a new patch regarding menus, toolbars and tables. Regarding tables i will have to add handling of the aria-activedescendant.
The current patch adds aria-attributes to the following components: Button, Checkbox, RadioButton, ToggleButton, SelectBox, Combobox, Slider, Table, Tabview.
I tested the changes with the Screenreader [NVDA](https://www.nvaccess.org/)

I don't know if i can reply before monday, i added the patch today so you're able to have a look at it now already.
Discussion and criticism are welcome.


cboulanger:
This is phantastic, thank you very much!
Implements #9685

johnspackman (approved):

oetiker (approved):
Great! One nit pick: repeated calls to `this.getContentElement()` could replaced by a extra variable ... be a bit DRYer

hkollmann:
Please fix some lints:

Error: error: Unexpected lexical declaration in case block (no-case-declarations) at source/class/qx/ui/form/AbstractSelectBox.js:149:11:
  147 |             quickSelection: true
  148 |           });
> 149 |           const listId = "list-" + control.toHashCode();
      |           ^
  150 |           const childrenContainerEl = control.getChildrenContainer().getContentElement();
  151 |           childrenContainerEl.setAttribute("id", listId);
  152 |           childrenContainerEl.setAttribute("role", "listbox");


error: Unexpected lexical declaration in case block (no-case-declarations) at source/class/qx/ui/form/AbstractSelectBox.js:150:11:
  148 |           });
  149 |           const listId = "list-" + control.toHashCode();
> 150 |           const childrenContainerEl = control.getChildrenContainer().getContentElement();
      |           ^
  151 |           childrenContainerEl.setAttribute("id", listId);
  152 |           childrenContainerEl.setAttribute("role", "listbox");
  153 |           this.getContentElement().setAttribute("aria-owns", listId);

hkollmann (changes requested):
Fix lint

source/class/qx/ui/table/cellrenderer/Abstract.js
```diff
@@ -168,7 +168,7 @@ qx.Class.define("qx.ui.table.cellrenderer.Abstract",
      */
     _getCellAttributes : function(cellInfo)
     {
-      return "";
+      return `role=gridcell aria-readonly=${!cellInfo.editable}`;
```
cboulanger, line 171: If this is to be backported to v6, which supports the legacy python generator, the backtick syntax probably won't work. Can you use ES5 syntax here, just to be sure?

source/class/qx/ui/table/cellrenderer/AbstractImage.js
```diff
@@ -255,10 +255,11 @@ qx.Class.define("qx.ui.table.cellrenderer.AbstractImage",
     // overridden
     _getCellAttributes : function(cellInfo)
     {
-      var tooltip = this.__imageData.tooltip;
+      const baseAttrs = this.base(arguments, cellInfo);
 
+      var tooltip = this.__imageData.tooltip;
       if (tooltip) {
-        return "title='" + tooltip + "'";
+        return `${baseAttrs} title='${tooltip}'`;
```
cboulanger, line 262: Same thing here. This will change in v7, which no longer supports the generator, but it would be good if we can backport the ARIA feature

cboulanger (changes requested):
Please replace template string syntax with old ES5 syntax so that the PR can be backported to v6

source/class/qx/ui/table/cellrenderer/AbstractImage.js
hkollmann: Do we really need the backport to version 6? Should'nt we bring out v7 instead?
Or change the the not supported syntax during backport?

Backtick syntax reads much nicer...

cboulanger: ARIA support should also be available for legacy apps, I think. I like backticks better, too, but here it would be so easy to stay backwards-compatible. After v7 is out, we no longer need to worry about such things, but at the moment, we still support v6.

ArturWesner:
Thank you for your feedback!
Changes: 
- repeated calls to this.getContentElement() replaced by extra variable
- Lint Fixes
- ES6 String Concatenation -> ES5
- ARIA Support for Toolbar and Menu (+ Child Components)
- aria-activedescendant Handling in  qx.ui.Table
  Comment here: I realised this in a kinda "hacky" way. I needed to set a cell id to every cell and set it as attribute in the table when focus changes. The Problem here was getting the cell id after the HTML Element was created. One way would have been to compute the element by the position depending on the col and row. But i felt that it was easier to set the cellId depending on the table hash, row and col. Then it was easy to compute the id in the table method. Since the cells get rerendered on change this worked so far. Do you see any problems doing it this way?
  
 I tested all additions with the Screenreader NVDA again.
  
  Summary of Components which support ARIA now:
- Button
- Checkbox
- RadioButton
- ToggleButton
- SelectBox
- Combobox
- Slider
- Table
- Tabview
- Toolbar
- Menu
- ARIA orientation where possible

cboulanger (approved):

ArturWesner:
I am sorry for the multiple small commits. Forgot something everytime.
I added 2 aria attrs to the table and changed the focus handling a bit. It was possible that the moveFocusedCell calls from pressing keyboard keys didn't went through when there was no focused cell before. This was strange because the table showed a focus. Turns out the Scroller set the focus to the first cell, but the table didn't. I found that handling a little bit weird for keyboard only users. The cell didn't get announced when the table got focus because of that too.

I removed the first cell focus from the Scroller and changed the moveFocusedCell so that the first possible cell gets focused if there was no cell focus before

johnspackman (approved):

oetiker:
> I am sorry for the multiple small commits. Forgot something everytime. I added 2 aria attrs to the table and changed the focus handling a bit. It was possible that the moveFocusedCell calls from pressing keyboard keys didn't went through when there was no focused cell before. This was strange because the table showed a focus. Turns out the Scroller set the focus to the first cell, but the table didn't. I found that handling a little bit weird for keyboard only users. The cell didn't get announced when the table got focus because of that too.
> 
> I removed the first cell focus from the Scroller and changed the moveFocusedCell so that the first possible cell gets focused if there was no cell focus before

you could do an interactive rebase to squash everything into one commit, this would also make backporting simpler ...

hkollmann:
I always do the squash during merge into master!

hkollmann (approved):

level420 (approved):

source/class/qx/ui/form/AbstractSelectBox.js
```diff
@@ -293,7 +305,7 @@ qx.Class.define("qx.ui.form.AbstractSelectBox",
       }
 
       // hide the list always on escape
-      else if (!listPopup.isHidden() && identifier == "Escape")
+      else if (!listPopup.isHidden() && (identifier == "Escape" || identifier == "Tab"))
```
derrell, line 308: This is a change of general behavior, not just ARIA attribute-related. I wonder if this is appropriate???

source/class/qx/ui/form/MenuButton.js
```diff
@@ -142,6 +156,9 @@ qx.Class.define("qx.ui.form.MenuButton",
 
       if (menu)
       {
+        if (this.isFocusable() && !qx.ui.core.FocusHandler.getInstance().isFocused(this)) {
```
derrell, line 159: Also a change of behavior, but looks like it actually fixes a bug.

source/class/qx/ui/form/MenuButton.js
```diff
@@ -241,6 +261,7 @@ qx.Class.define("qx.ui.form.MenuButton",
     {
       switch(e.getKeyIdentifier())
       {
+        case "Space":
```
derrell, line 264: This too. Typically Space and Enter are treated equivalently. I think maybe these enhancements and changes of behavior should be in a separate PR, though.

source/class/qx/ui/form/RadioGroup.js
```diff
@@ -522,7 +522,26 @@ qx.Class.define("qx.ui.form.RadioGroup",
 
       if (value) {
         value.set(groupedProperty, true);
+
+        // If Group is focused, the selection was changed by keyboard. Switch focus to new value
```
derrell, line 526: Enhancement, it seems?

source/class/qx/ui/menu/Manager.js
```diff
@@ -597,6 +598,7 @@ qx.Class.define("qx.ui.menu.Manager",
             this._onKeyPressRight(menu);
             break;
 
+          case "Tab":
```
derrell, line 601: Enhancement

source/class/qx/ui/table/Table.js
```diff
@@ -1805,8 +1826,9 @@ qx.Class.define("qx.ui.table.Table",
       var col = this.__focusedCol;
       var row = this.__focusedRow;
 
-      // could also be undefined [BUG #4676]
+      // could also be undefined [BUG #4676]. In that case default to first cell focus
       if (col == null || row == null) {
+        this.setFocusedCell(0, 0, true);
```
derrell, line 1831: I think using `==` for these two comparisons with null rather than `===` makes this the equivalent of
```
if (! col || ! row)
```
which means that if col is 1 and row is 0, instead, it will use 0, 0 which isn't what's desired...

source/class/qx/ui/table/pane/Scroller.js
```diff
@@ -697,17 +697,13 @@ qx.Class.define("qx.ui.table.pane.Scroller",
...
         this.updateVerScrollBarMaximum();
-
-        if (this.getFocusedRow() === null && rowCount > 0 && colCount > 0)
-        {
-          this.setFocusedCell(this.getFocusedColumn()||0, 0);
-        }
-        else if (this.getFocusedRow() >= rowCount)
+        
+        const focusedRow = this.getFocusedRow();
+        if (focusedRow !== null && focusedRow >= rowCount)
```
derrell, line 706: I'm not quite following this change, but it looks to me not to be equivalent. What is the purpose of this change?

source/class/qx/ui/toolbar/Button.js
```diff
@@ -19,32 +19,12 @@
 
 /**
  * The normal toolbar button. Like a normal {@link qx.ui.form.Button}
- * but with a style matching the toolbar and without keyboard support.
+ * but with a style matching the toolbar.
```
derrell, line 22: Behavior change

derrell (changes requested):
This will be a nice addition. There are a number of seemingly unrelated changes, identified in my comments, that I think should be pulled into a different PR and discussed separately, unless there is some need for them to be included here.

source/class/qx/ui/form/AbstractSelectBox.js
ArturWesner: This is removable. I added it at my first tries because i made the list and its elements focusable too. In the current version this is not needed anymore, but i kept it because i liked it and saw it in another framework too.
Closing SelectMenus with Tab is not a W3C requirement.

source/class/qx/ui/form/MenuButton.js
ArturWesner: Agree, but needed. If open is called the Button didn't get focused

source/class/qx/ui/form/MenuButton.js
ArturWesner: Change of behavior

source/class/qx/ui/form/RadioGroup.js
ArturWesner: Kinda needed. Radio Selection could change without focus moving, resulting in no reading out to the screenreader

source/class/qx/ui/menu/Manager.js
ArturWesner: Kinda needed, but could be solved in another way. The menu did not disappear when the opener button blurred. This could have been handled by a blur listener, but this change had the side effect of the enhancement i added to the selectbox: Tab returns focus to the opener button

source/class/qx/ui/table/Table.js
ArturWesner: val == null checks if the val is null or undefined since null == undefined. Quick search: https://dorey.github.io/JavaScript-Equality-Table/

source/class/qx/ui/table/pane/Scroller.js
ArturWesner: It is not meant to be equivalent. I gave a short explanation [here:](https://github.com/qooxdoo/qooxdoo/pull/10277#issuecomment-970145759)

> I am sorry for the multiple small commits. Forgot something everytime. I added 2 aria attrs to the table and changed the focus handling a bit. It was possible that the moveFocusedCell calls from pressing keyboard keys didn't went through when there was no focused cell before. This was strange because the table showed a focus. Turns out the Scroller set the focus to the first cell, but the table didn't. I found that handling a little bit weird for keyboard only users. The cell didn't get announced when the table got focus because of that too.
> 
> I removed the first cell focus from the Scroller and changed the moveFocusedCell so that the first possible cell gets focused if there was no cell focus before

source/class/qx/ui/toolbar/Button.js
ArturWesner: Needed. The button is not reachable for keyboard users otherwise

source/class/qx/ui/toolbar/CheckBox.js
```diff
@@ -24,27 +24,6 @@ qx.Class.define("qx.ui.toolbar.CheckBox",
...
-  *****************************************************************************
-  */
-
-  construct : function(label, icon)
-  {
-    this.base(arguments, label, icon);
-
-    // Toolbar buttons should not support the keyboard events
-    this.removeListener("keydown", this._onKeyDown);
-    this.removeListener("keyup", this._onKeyUp);
```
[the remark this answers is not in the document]
ArturWesner: This is true, but needed for more accessiblity support. I agree that some enhancements which are not required could be discussed in another pr, but imo this one is required and meant to be discussed in this pr

source/class/qx/ui/table/Table.js
derrell: The specific behavior of `==` for 0, null, and undefined is difficult to remember. My preference here would be to break it out so it's clear what's meant:
```
if (col === null || typeof col == "undefined" || row === null || typeof row == "undefined")
```
That's a bit anal retentive on my part. Maybe just document in a comment above that the comparison for null also compares true for undefined but not for 0?

source/class/qx/ui/toolbar/CheckBox.js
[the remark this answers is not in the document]
derrell: Understood. Why don't you add a section to the qooxdoo.git/docs/release-notes.md that says that you added the accessibility support, and also documents the other enhancements included here. Since they're not backward compatible, they probably won't be back-ported into 6.x but rather be released with 7.0 where behavior changes are allowed (but should be documented in the release notes).

source/class/qx/ui/menubar/Button.js
```diff
@@ -24,25 +24,6 @@ qx.Class.define("qx.ui.menubar.Button",
...
-  *****************************************************************************
-  */
-
-  construct : function(label, icon, menu)
-  {
-    this.base(arguments, label, icon, menu);
-
-    this.removeListener("keydown", this._onKeyDown);
-    this.removeListener("keyup", this._onKeyUp);
-  },
```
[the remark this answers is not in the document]
ArturWesner: As you guessed in another issue: The Menubar Button could not be toggled by keyboard users

source/class/qx/ui/table/Table.js
ArturWesner: Will change as suggested!

source/class/qx/ui/toolbar/CheckBox.js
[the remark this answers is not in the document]
ArturWesner: Should i open another PR with the enhancements which are not required? In that PR i would document these then.
In this PR i will add an section to qooxdoo.git/docs/release-notes.md about (more) accessibility support to the mentioned components as suggested.
On a sidenote: Why are they not backward compatible?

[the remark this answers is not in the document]
hkollmann: They changed the expected behavior of old applications. You can'nt be shure that anyone relies on this.
So you should split the PR into a part which aria support and a part with better accessiblity support if this is possible

hkollmann:
btw: not ok 276 - qx.test.ui.form.ComboBox:testFocusTextOnClose - [45.80 ms] - Called fail().

Seems that changes in accessiblity support breaks one of the tests.

ArturWesner:
That test will be removed because it has no purpose anymore. I will mention the change causing this in the .md file. I removed the test locally and npm test runs through now without any problems.
Regarding splitting the PR: I will look into that next week problably. Can't say for sure if i have time this week.
Please don't merge in the meanwhile

hkollmann:
Why is the test obsolete?

ArturWesner:
I seperated all behaviour changes into another PR. 
Some of them were important and needed for better accessibility support, but they would cause this ARIA PR to not be backportable

> Why is the test obsolete?
I added a behaviour change (now in the other PR) in the combobox: The input text got selected when the menu closed. That was the reason keyboard events did not pass to the input field, which is the wanted behaviour. And the two tests did exactly test this.

derrell (approved):

source/class/qx/ui/menu/Menu.js
```diff
@@ -506,6 +519,17 @@ qx.Class.define("qx.ui.menu.Menu",
...
     },
 
+    // property apply
+    _applyOpener : function(value, old)
+    {
+      // ARIA attrs
+      if (value) {
+        this.getContentElement().setAttribute("aria-labelledby", value);
+      } else {
+        this.getContentElement().removeAttribute("aria-labelledby");
```
rad-pat, line 529: On latest master, I'm getting an error here on dispose because `this.getContentElement()` is null when the call to `setContextMenu(null)` is made from Widget https://github.com/qooxdoo/qooxdoo/blob/28066cf52fe240d009986e67fde9768d4b5ec714/source/class/qx/ui/core/Widget.js#L4004 I've tried to build a repl but can't seem to replicate. Does it just make sense to check the ContentElement here anyway?

hkollmann:
Could you prepare a pr?

Many thanks
Henner Kollmann

Von meinem iPad gesendet

PR #10284, opened 2021-11-19
Add behaviour changes for better keyboard accessibility

ArturWesner:
Additional and seperate PR for [#10277](https://github.com/qooxdoo/qooxdoo/pull/10277)
Probably not backportable


docs/release-notes.md
```diff
@@ -97,6 +97,24 @@ simply executing `qx test`, using pluggable test runners.
...
 [Documentation](https://qooxdoo.org/documentation/#/development/testing/)
 
+## Component behaviour changes
+
+In hindsight of better keyboard accessibility support some components received changes regarding their behaviour:
+
+| Component | Changes |
+| --- | --- |
+| qx.ui.form.AbstractSelectBox | `Tab` Key is now equally treated as `Escape` and closes the menu |
+| qx.ui.form.ComboBox | - `Down` Key now opens the menu without pressing `Alt` <br /> - Textfield text is not fully selected anymore when the menu closes <br /> - Pressing input specific keys while the menu is open will close the menu|
```
hkollmann, line 107: did'nt you mean "is fully selecteded"?

ArturWesner: No, all text is currently getting selected when the menu closes. This got changed in the PR. tabFocus is no longer called in handler _onPopupChangeVisibility

hkollmann (approved):

derrell (changes requested):
This looks good. My only concern here relates to use of the Tab key. It looks like Tab will now close a menu/list, but will not move to the next focusable field. (`e.stop()` is called). I don't recall ever using Tab to close a menu so I don't know what the generally expected behavior is, but my own expectation would be that if it closed the menu, it would also move to the next field. Maybe it does that even though my reading of the code indicates that it wouldn't? Do you have knowledge of the generally expected behavior when Tab is pressed on an open list?

ArturWesner:
Your assumptions about the behaviour with this change are correct, the Tab key closes the menu and does not move focus to the next target. I added it in this PR because Vuetify (Vue Framework) and the native HTML select component handle it that way and i got used to it.
I tried to find a source explaining the generally expected behaviour, but couldn't find anything. The w3c ARIA document doesn't say anything either.
Seems like its preference only. I have no problem reverting it. Only a few changes are needed then. Just tell me your decision

derrell:
If a native HTML Select works this way, that seems like the definitive answer (assuming it works the same in all popular browsers). 

I just ran a quick test, using Chrome. I found two distinct behaviors:
```
<html>
  <head>
  </head>

  <body>
    <label for="pet-select">Choose a pet:</label>

    <select name="pets" id="pet-select">
        <option value="">--Please choose an option--</option>
        <option value="dog">Dog</option>
        <option value="cat">Cat</option>
        <option value="hamster">Hamster</option>
        <option value="parrot">Parrot</option>
        <option value="spider">Spider</option>
        <option value="goldfish">Goldfish</option>
    </select>

    <select name="pets" id="pet-select">
        <option value="">--Please choose another option--</option>
        <option value="dog">Dog</option>
        <option value="cat">Cat</option>
        <option value="hamster">Hamster</option>
        <option value="parrot">Parrot</option>
        <option value="spider">Spider</option>
        <option value="goldfish">Goldfish</option>
    </select>


  </body>
</html>
```
If you tab into the first `select` and press Down a few times to select a different item without opening the list, and then press TAB, it moves to the next `select` item. If instead, you press ALT-Down to open up the list and press Down a few times to select a different item, then pressing TAB just closes the opened list, but leaves focus in the same `select`.

If we are emulating that behavior (or whatever portions of that behavior are relevant), then I think we're good.

ArturWesner:
Oh, i did not know about Alt+Up/Down in the selectbox. I have added it to the handling. Now it works the same in all scenarios you described

derrell (approved):
Thanks!

oetiker (approved):

level420 (approved):

cboulanger (approved):

PR #10290, opened 2021-12-07
Adds ARIA role to window and aria helper functions to widget

ArturWesner:
I recognized that aria-label, aria-labelledby and aria-describedby are often needed when updating the code for screenreaders. The added functions in qx.ui.core.Widget help to achieve this.
This should be backportable. My only concern is the use of rest parameters (to achieve variadic functions). I checked https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters and the supported Node Version is 6+. So this shouldn't be a problem, right?

### DIFF
--- a/source/class/qx/ui/core/Widget.js
+++ b/source/class/qx/ui/core/Widget.js
@@ -3827,7 +3827,74 @@ qx.Class.define("qx.ui.core.Widget",
 
 
 
+    /*
+    ---------------------------------------------------------------------------
+      ARIA attrs support
+    ---------------------------------------------------------------------------
+    */
+    
+    /**
+     * Sets the string which labels this widget. This will be read out by screenreaders. Needed if there is no
+     * readable text/label in this widget which would automatically act as aria-label.
+     * @param {String} label Labelling Text
+     */
+    setAriaLabel: function(label) {
+      this.getContentElement().setAttribute("aria-label", label);
+    },
 
+    /**
+     * Marks that this widget gets labelled by another widget. This will be read out by screenreaders as first
+     * information.
+     * Similiar to aria-label, difference being that the labelling widget is an different widget and multiple
+     * widgets can be added.
+     * @param {Array} labelWidgets Indefinite Number of labelling Widgets
+     */
+    addAriaLabelledBy: function(...labelWidgets) {
+      this.__addAriaXBy(labelWidgets, "aria-labelledby");
+    },
+
+    /**
+     * Marks that this widget gets described by another widget. This will be read out by screenreaders as last
+     * information. Multiple Widgets possible.
+     * @param {Array} describingWidgets Indefinite Number of describing Widgets
+     */
+    addAriaDescribedBy: function(...describingWidgets) {
+      this.__addAriaXBy(describingWidgets, "aria-describedby");
+    },
+
+    /**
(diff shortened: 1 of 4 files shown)


source/class/qx/ui/form/DateField.js
```diff
@@ -602,7 +602,22 @@ qx.Class.define("qx.ui.form.DateField",
...
-    }
+    },
+
+    // overridden
+    setAriaLabel: function(label) {
+      this.getChildControl("textfield").setAriaLabel(label);
+    },
+
+    // overridden
+    addAriaLabelledBy: function(...labelWidgets) {
```
oetiker, line 613: this will require some `arguments` and `apply` or `call` magic when backporting, but it would certainly be possible.

oetiker (approved):

source/class/qx/ui/core/Widget.js
```diff
@@ -3827,6 +3827,76 @@ qx.Class.define("qx.ui.core.Widget",
...
+    /**
+     * Sets either aria-labelledby or aria-describedby
+     * @param widgets {qx.ui.core.Widget[]} Indefinite Number of widgets
+     * @param ariaAttr {String} aria-labelledby | aria-describedby
+     */
+    __addAriaXBy: function(widgets, ariaAttr) {
+      if (!["aria-labelledby", "aria-describedby"].includes(ariaAttr)) {
+        throw new Error("Only aria-labelledby or aria-describedby allowed!");
+      }
+      let idArr = [];
```
derrell, line 3874: In addition to replacing the rest operator with use of `arguments` as previously mentioned, `let` and `const` ES6 features are not recognized by the old generator, so will have to be replaced if backported.

derrell (approved):

johnspackman:
@ArturWesner the problem with using ES6 isn't whether the client or node can support it, it's that Qooxdoo v6.x needs to be processed by the deprecated python toolchain, and it strictly only understands ES5.  This means that if you use the rest operator in 6.x branch, then no-one will be able to compile it (without switching from `./generate.py` to the new `qx` compiler).

Once we hit v8, this will not be a problem any more :)

johnspackman:
> Alright, got it. Thanks! I've tested an ES5 syntax solution and it seems to be working: `var labelWidgets = Array.prototype.slice.call(arguments);` 

FWIW there's also `qx.lang.Array.fromArguments(arguments)`

> I'd like to keep the ES6 way though. What is the best way to handle this then? Create a new cloned PR with the needed ES5 changes and keep this PR?

Yes, good idea

hkollmann:
This is obsolete, is'nt it?
Same as https://github.com/qooxdoo/qooxdoo/pull/10291?

ArturWesner:
Functionallity-wise it's the same, yes. But i would like to keep the ES6 way if possible. It is much cleaner and easier to read/understand. The idea was to merge the es5 version first, then backport it and finally merge the es6 version, overriding the es5 version.
If that's too much of a hassle, just keep the es5 version and delete this PR. That would be fine too

hkollmann:
We can not merge this until we haved moved to version 7. v6 should be work with phython toolchain

cboulanger (approved):

hkollmann (approved):

level420 (approved):

PR #10295, opened 2021-12-10
Fix babel

hkollmann:
- fix babel docs
- add doc fo new api bubble afterProcessFinished
- babel options are not handled as documented


cboulanger (approved):

johnspackman (approved):

oetiker (approved):

level420 (approved):

source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -571,10 +571,10 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
     createMakersFromConfig: async function(data) {
       const Console = qx.tool.compiler.Console.getInstance();
       var t = this;
-
       if (data.babelOptions) {
-        if (!data.babelConfig) {
-          data.babelConfig = { options: data.babelOptions };
+        if (!data?.babel?.options) {
+          data.babel = data.babel || {};
+          data.babel = { options: data.babelOptions };
```
derrell, line 577: This isn't just adding `options`; it's overwriting `data.babel`. If before line 575, `data` contained:
```
data : 
{
  babel : 
  {
    x : "hello"
  },
  babelOptions : 
  {
    z : "hi"
  }
}
```

then after line 577, data now contains:
```
data : 
{
  babel : 
  {
    options : 
    {
      z : "hi"
    }
  }
}
```

`x` has been lost.

Unless the intent was to lose the original value of `data.babel`, I think line 577 should instead be:
```
data.babel.options = data.babelOptions;
```

derrell (changes requested):

source/class/qx/tool/cli/commands/Compile.js
derrell: (followed possibly by `delete data.babelOptions;`)

derrell (approved):

PR #10303, opened 2021-12-13
Fix for UTC parsing not working correctly

rad-pat:
Per old bug https://github.com/qooxdoo/qooxdoo/issues/7946, UTC parsing was not working correctly. Also the timezone format identifiers 'Z' & 'z' were not being processed if non-literal in the format string.

 * Use date.UTC() to correctly determine the UTC date.
 * Adds isoDateTimeTz as a format to parse timezone details in the date string, i.e. offset from UTC

### DIFF
--- a/source/class/qx/util/format/DateFormat.js
+++ b/source/class/qx/util/format/DateFormat.js
@@ -763,7 +763,7 @@ qx.Class.define("qx.util.format.DateFormat",
         min         : 0,
         sec         : 0,
         ms          : 0,
-        tzOffsetMins: 0
+        tzOffsetMins: null
       };
 
       var currGroup = 1;
@@ -828,7 +828,7 @@ qx.Class.define("qx.util.format.DateFormat",
       }
 
       var date;
-      if (this.__UTC || dateValues.tzOffsetMins !== 0) {
+      if (this.__UTC || dateValues.tzOffsetMins !== null) {
         var utcMs = Date.UTC(dateValues.year, dateValues.month, dateValues.day, (dateValues.ispm) ? (dateValues.hour + 12) : dateValues.hour, dateValues.min, dateValues.sec, dateValues.ms);
         if (dateValues.tzOffsetMins !== 0) {
           utcMs += (dateValues.tzOffsetMins * 60000);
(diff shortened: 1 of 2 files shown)


source/class/qx/util/format/DateFormat.js
```diff
@@ -1207,10 +1216,28 @@ qx.Class.define("qx.util.format.DateFormat",
         dateValues.hour = parseInt(value, 10) % 12;
       };
 
-      var ignoreManipulator = function(dateValues, value) {
-        return;
+      var timezoneManipulator = function(dateValues, value) {
+        var regEx = new RegExp("([+-]*)(\\d\\d)[:]*(\\d\\d)*$");
```
derrell, line 1220: This regex doesn't look right. It allows 0 or more leading `+` or `-`, and zero or more colons. That means that this value is accepted (but `tzResults[1]` does not equal `-`):
```
"---23:::13"
```

I think you probably mean "options" (0 or 1) in both cases, so you'd use this:
```
var regEx = new RegExp("([+-]?)(\\d\\d)[:]?(\\d\\d)?$");
```

Even that, though, doesn't invalidate bad input. I think there are additional constraints necessary to detect bad input.

derrell (changes requested):

source/class/qx/util/format/DateFormat.js
oetiker: I think `var regEx = new RegExp("([+-]?)(\\d\\d)(?::?(\\d\\d))?$");` should do.

oetiker (changes requested):
see regexp comment

source/class/qx/util/format/DateFormat.js
rad-pat: Thanks, I've updated the regex. This regex is only used for extracting the parts, not for validation. Validation is already done via regexes on lines 1784 & 1791 for rules 'Z' and 'z' respectively.

cboulanger (approved):

level420 (approved):

derrell (approved):

derrell (approved):

hkollmann (approved):

oetiker (approved):

PR #10310, opened 2021-12-28
Check contentElement exists before modification

rad-pat:
Fixes a bug on destruction where opener is set to null and getContentElement returns null

Reference:
https://github.com/qooxdoo/qooxdoo/pull/10277#discussion_r775014305_

PR #10311, opened 2021-12-28
fix empty contentElement and add docs for accessibility

ArturWesner:
- fixes bug report in [#10277](https://github.com/qooxdoo/qooxdoo/pull/10277#discussion_r775014305) and #10310
- fixes wrong paths for `best practices` and `avoid antipatterns` in docs\development\howto\README.md
- adds accessibility document to howto


source/class/qx/ui/menu/Menu.js
```diff
@@ -522,11 +522,12 @@ qx.Class.define("qx.ui.menu.Menu",
...
     _applyOpener : function(value, old)
     {
+      const contentEl = this.getContentElement();
       // ARIA attrs
       if (value) {
-        this.getContentElement().setAttribute("aria-labelledby", value);
-      } else {
-        this.getContentElement().removeAttribute("aria-labelledby");
+        contentEl.setAttribute("aria-labelledby", value);
+      } else if (contentEl) {
```
hkollmann, line 529: Should'nt this be a general test?

const contentEl = this.getContentElement();
if  (contentEl) {

...
}

source/class/qx/ui/menu/AbstractButton.js
```diff
@@ -358,10 +358,12 @@ qx.Class.define("qx.ui.menu.AbstractButton",
       {
         this._excludeChildControl("arrow");
         
-         // ARIA attrs
-        contentEl.removeAttribute("aria-haspopup");
-        contentEl.removeAttribute("aria-expanded");
-        contentEl.removeAttribute("aria-controls");
+        // ARIA attrs
```
hkollmann, line 361: No test if value is set?

source/class/qx/ui/form/SelectBox.js
```diff
@@ -430,10 +430,11 @@ qx.Class.define("qx.ui.form.SelectBox",
       }
 
       // Set aria-activedescendant
+      const contentEl = this.getContentElement();
       if (current && current[0]) {
-        this.getContentElement().setAttribute("aria-activedescendant", current[0].getContentElement().getAttribute("id"));
-      } else {
-        this.getContentElement().removeAttribute("aria-activedescendant");
+        contentEl.setAttribute("aria-activedescendant", current[0].getContentElement().getAttribute("id"));
```
hkollmann, line 435: No check in this case?

source/class/qx/ui/form/ComboBox.js
```diff
@@ -337,10 +337,11 @@ qx.Class.define("qx.ui.form.ComboBox",
       }
 
       // Set aria-activedescendant
+      const textFieldContentEl =  this.getChildControl("textfield").getContentElement();
       if (current && current[0]) {
-        this.getChildControl("textfield").getContentElement().setAttribute("aria-activedescendant", current[0].getContentElement().getAttribute("id"));
-      } else {
-        this.getChildControl("textfield").getContentElement().removeAttribute("aria-activedescendant");
+        textFieldContentEl.setAttribute("aria-activedescendant", current[0].getContentElement().getAttribute("id"));
```
hkollmann, line 342: No check in this case?

hkollmann (changes requested):

source/class/qx/ui/menu/Menu.js
ArturWesner: I don't see a reason why the content element should be null or undefined in the first if case. That goes for all the places i added the check or did not.

docs/development/howto/accessibility.md
```diff
@@ -0,0 +1,42 @@
+# Accessibility
```
derrell, line 1: This is a useful new page. There are a bunch of grammatical and usage errors that are too difficult to make PR comments on. Maybe ping me after this gets merged, and I'll create a new PR to fix those problems. None are bad enough to prevent merging this, I think.

derrell (approved):
See my inline comment for post-merge additional changes.

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

source/class/qx/ui/menu/Menu.js
derrell: I would probably have added, at line 525.5, something like,

```
if (! contentEl) {
  return;
}

That handles the case where the content element doesn't exist right away (we won't do anything). In that case, line 529 could get rid of the `if` constraint.

I agree, as is it's correct but a bit hard to read.
```

ArturWesner:
I've updated the PR with better readable code.
Could you apply the newest commit 'fix error when applying aria-labelledby to menu opener' to latest branch too please? I would have to make a new PR for this, but its only 3 lines

hkollmann:
We need a new pr based on latest

hkollmann (approved):

johnspackman (approved):

PR #10316, opened 2021-12-31
Add `qx es6ify`

johnspackman:
To try this out on your own repo, just use `qx es6ify source/class`, or checkout my branch https://github.com/johnspackman/qooxdoo/tree/es6ify-test-branch

Once this is merged we can put a hold on new PRs, run es6ify and then merge again; when we do that, we can also add the prettier-on-commit git hook


source/class/qx/tool/cli/commands/Es6ify.js
```diff
@@ -0,0 +1,89 @@
...
+
+   Authors:
+     * John Spackman (john.spackman@zenesis.com, @johnspackman)
+
+************************************************************************ */
+
+const fs = require("fs");
+const process = require("process");
+const path = require("upath");
+const rimraf = require("rimraf");
```
hkollmann, line 22: Not used.

source/class/qx/tool/cli/commands/Es6ify.js
```diff
@@ -0,0 +1,89 @@
...
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * John Spackman (john.spackman@zenesis.com, @johnspackman)
+
+************************************************************************ */
+
+const fs = require("fs");
+const process = require("process");
```
hkollmann, line 20: Not used

source/class/qx/tool/compiler/Es6ify.js
```diff
@@ -0,0 +1,283 @@
...
+      let result = babelCore.transform(src, config);
+
+      let prettierConfig = await prettier.resolveConfig(this.__filename, { editorConfig: true })||{};
+      prettierConfig.parser = "babel";
+      let prettyCode = prettier.format(result.code, prettierConfig);
+      
+      await fs.promises.writeFile(this.__filename, prettyCode, "utf8");
+    },
+
+    __pluginFunctionExpressions() {
```
hkollmann, line 168: What is done here?
Could you provide an example?

hkollmann (changes requested):

derrell:
Looking at @johnspackman's test branch https://github.com/johnspackman/qooxdoo/tree/es6ify-test-branch, it appears that es6ify does not convert `var` to `let`. They would be interchangeable if hoisted to the top of their enclosing function. If we're gong to convert the code to es6, it'd be nice to do it completely, not half-heartedly. Is there an option to do var -> let conversion?

hkollmann (approved):

.prettierrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "trailingComma": "none",
```
oetiker, line 2: I find the trailing comma particularly useful in preventing accidental errors when extending a list ...

cboulanger (changes requested):
Trying this out on `qxl.dialog` , the following problem became apparent (thanks to @hkollmann for spotting it!): a checking function in a property definition becomes rewritten, which is not recognized by the compiler anymore.
See:
- https://github.com/qooxdoo/qxl.dialog/blob/v3.2.0/source/class/qxl/dialog/Progress.js#L27-L30 rewritten to https://github.com/qooxdoo/qxl.dialog/blob/b3dff40ce681beebd97d0d37bc3d6eeeb4ad01c3/source/class/qxl/dialog/Progress.js#L25-L28
- building fails: https://github.com/qooxdoo/qxl.dialog/runs/4685875028?check_suite_focus=true#step:4:102

johnspackman:
> Looking at @johnspackman's test branch https://github.com/johnspackman/qooxdoo/tree/es6ify-test-branch, it appears that es6ify does not convert `var` to `let`. They would be interchangeable if hoisted to the top of their enclosing function. If we're gong to convert the code to es6, it'd be nice to do it completely. Is there an option to do var -> let conversion?

That's a nice idea but the task of converting ES5 to ES6 in the best most complete way possible would be a huge engineering challenge and is beyond the scope of this PR; changing `var` to `let` is achievable and would be only one of those steps, but I don't really see the benefit of it apart from saying it's more ES6-ish.  Hoisting all variable declarations up to the top of the function can make the function less readable, not more.  

IMHO the point of this tool is to make enough of a difference to improve readability that cannot be improved by code formatting alone.  In a recent PR, I was asked to remove ES6 constructs because it clashed with the ES5 constructs next to it, and this fixes that exact problem.

In a recent chat on Gitter, the question of `super` was raised - I think that's a very good idea which can be implemented and `super` has been on the todo list for a while too, I'll try and take a look and add it in

.prettierrc.json
johnspackman: Funny, I think the exact opposite :)  I think it's just as likely to cause errors because you cant see the end of the list; the only only place i've seen it before is in Dart code where it seems like a way to get around their "constructor hell".  I turned it off here so that it follows what we've done before (didnt it used to be an error in Javascript to have trailing commas?)

cboulanger:
> Trying this out on `qxl.dialog` , the following problem became apparent (thanks to @hkollmann for spotting it!): a checking function in a property definition becomes rewritten, which is not recognized by the compiler anymore. See:

@johnspackman might this be the result of a v7 compiler without the PR merged trying to compile code that has ben `es6ified`?

.prettierrc.json
level420: I remember back to some problems with internet explorer not accepting trailing commas (I love them in perl!), especially with trailing commas of array elements (elisions).
I know that IE will be dead midth of this year, but I'm not sure about the implications allowing them.

level420: which means that I'm against allowing trailing commas :-)

level420 (approved):
In @johnspackman s code we trust :-)

hkollmann:
Problem is:

  properties: {
    /**
     * The percentage of the progress, 0-100
     */
    progress: {
      check: function(value) {
        return qx.lang.Type.isNumber(value) && value >= 0 && value <= 100;
      },
      init: 0,
      nullable: false,
      event: "changeProgress",
      apply: "_applyProgress",
    },


this syntax is correct.  This case should not be es6ified.

hkollmann:
BTW:

    progress: {
      check: value => qx.lang.Type.isNumber(value) && value >= 0 && value <= 100,
      init: 0,
      nullable: false,
      event: "changeProgress",
      apply: "_applyProgress",
    },
is working

derrell:
That example works because `this` is not used in the check method. Is there a difference of context between the example and the original code?

hkollmann:
It's not a question of runtime. It's a compiler error.
Babel is not translating the first

derrell:
I misunderstood you. I thought you were suggesting that the compiler transpile to the "fat arrow" function style. Disregard.

cboulanger:
> Trying this out on `qxl.dialog` , the following problem became apparent (thanks to @hkollmann for spotting it!): a checking function in a property definition becomes rewritten, which is not recognized by the compiler anymore. See:
> 
>     * https://github.com/qooxdoo/qxl.dialog/blob/v3.2.0/source/class/qxl/dialog/Progress.js#L27-L30 rewritten to https://github.com/qooxdoo/qxl.dialog/blob/b3dff40ce681beebd97d0d37bc3d6eeeb4ad01c3/source/class/qxl/dialog/Progress.js#L25-L28
> 
>     * building fails: https://github.com/qooxdoo/qxl.dialog/runs/4685875028?check_suite_focus=true#step:4:102

Turns out this was due to an unrelated bug now fixed by @hkollmann .

cboulanger (approved):

hkollmann:
@cboulanger: I changed Progress manually. Could you test if it's rewritten by es6ify?

johnspackman:
@hkollmann I've tried using es6ify on this:
```
  properties: {
    progress: {
      check: function(value) {
        return qx.lang.Type.isNumber(value) && value >= 0 && value <= 100;
      },
      init: 0,
      nullable: false,
      event: "changeProgress",
      apply: "_applyProgress",
    }
  }, 
```
and the output is this:
```
  properties: {
    progress: {
      check(value) {
        return qx.lang.Type.isNumber(value) && value >= 0 && value <= 100;
      },
      init: 0,
      nullable: false,
      event: "changeProgress",
      apply: "_applyProgress",
    },
  },
```

after which, `qx compile` works ok for me

johnspackman:
darn, hang on a minute this isnt ready...

johnspackman:
ok this is done now; the branch https://github.com/johnspackman/qooxdoo/tree/es6ify-test-branch is updated with the results of this es6ify and tests are passing

.prettierrc.json
oetiker: well, if you do a little google survey you will see, that as in may things perl has been way ahead of the curve ... in JS world, support for trailing (final) commas is being added in more and more places. So removing them or even disalowing them here seems to a rather anachronistic.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

https://blog.logrocket.com/best-practices-using-trailing-commas-javascript/

https://time2hack.com/are-you-using-trailing-commas-in-your-javascript/

johnspackman: but this isn't perl, and as the mozilla doc says, support has been added progressively, JSON does not support it at all, and it has inconsistencies between object literals and arrays.  The logrocket article shows how once is OK but you have to be careful not to add undefined entries in the array, IMHO meaning that allowing trailing commas means it's easier to add a bug.  The time2hack article finishes by saying that "We write code in Clean way for other developers to understand`, as if this somehow automatically qualifies as a higher grade of coding and the reader is stupid and inconsiderate and obfuscating to not just see it their way.

This is, of course, one of those arguments that something like prettier is supposed to resolve.  If we have to have commas at the end of lines then we can, but it just seems like another arbitrary change to something that has worked just fine for generations and which has just as many disadvantages as advantages.

oetiker: Not only Dart has trailing commas, 

Kotlin added them in 1.4 https://medium.com/@sanketmeghani/kotlin-1-4-trailing-commas-53c690eced32 

Go has them https://dave.cheney.net/2014/10/04/that-trailing-comma

Rust has them https://users.rust-lang.org/t/trailing-commas/13993

Trailing commas are a good thing -- They prevent a common bug when modifying existing code, and they make diffs shorter and more readable. 

So forcing them out of existance seems to be a rather odd move ...

johnspackman: But I don't see that comparing it to other languages makes any sense; I havn't tried to find any, but do C++, C, etc?  Java does not.  Would we care if they did?  

> They prevent a common bug when modifying existing code

They also allow a common bug though ??

> So forcing them out of existance seems to be a rather odd move ...

Which we've never allowed before, so requiring them now is an equally odd move ...

derrell: Just to add another vote, I agree with @johnspackman that trailing commas should be disallowed (removed by prettier).

oetiker: here is another one, but I guess your minds are made up ... 

https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

but I guess rational arguments don't matter, because as one commenter puts it:

> I understand the benefits — or, I should say, the arguments in favour — but I still hate trailing commas.

I guess I can still configure es6ify to pretty up my own code with dangling commas, right ?

oetiker (approved):

.prettierrc.json
johnspackman: I've had to write a lot of Dart/Flutter code over the last 9 months and having massively nested arrays and parameter lists is almost baked into the API; the default settings for Dart is to require a trailing comma and my experience is that always having commas makes it noticably harder to read (especially without an IDE to highlight the end of a statement, eg when browsing some code in Github).  Perhaps this is because a lot of Flutter devs are advocating the "style" which can lead to a single statement literally being hundreds of lines long, but I genuinely think that it does not have a net gain except a limited use case where there is a simple list and you want to mark & copy a horizontal line instead of a staggered one.

I understand the claimed benefits, I just think that they are not justified and don't outweigh the disadvantages.

> I guess I can still configure es6ify to pretty up my own code with dangling commas, right ?

Yes, this prettierrc.json only affects the framework; es6ify searches for the nearest prettierrc.json file, starting in the directory containing the .js file being prettier-ed and then going up through parent directories until it finds one.

derrell (approved):

PR #10319, opened 2022-01-02
Add popular NPM keywords [ci skip][skip ci]

cboulanger:
Adds some popular fitting NPM keywords so that qooxdoo shows up when people browse them

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "cli",
     "framework",
     "front-end",
+    "fontend",
     "mobile",
     "back-end",
     "widgets",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,18 @@
   "keywords": [
     "qooxdoo",
     "compiler",
-    "cli"
+    "cli",
+    "framework",
+    "front-end",
+    "mobile",
+    "back-end",
+    "widgets",
+    "gui",
+    "databinding",
+    "classes",
+    "interfaces",
+    "mixins",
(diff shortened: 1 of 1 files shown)


hkollmann (approved):

package.json
```diff
@@ -6,7 +6,18 @@
   "keywords": [
     "qooxdoo",
     "compiler",
-    "cli"
+    "cli",
+    "framework",
+    "front-end",
+    "mobile",
+    "back-end",
```
derrell, line 13: backend (vs back-end) and frontend are common spellings. Maybe add them without the dashes as well?

derrell (approved):

oetiker (approved):

package.json
```diff
@@ -6,7 +6,18 @@
   "keywords": [
     "qooxdoo",
     "compiler",
-    "cli"
+    "cli",
+    "framework",
+    "front-end",
```
oetiker, line 11:
```suggestion
    "front-end",
    "fontend",
```

package.json
oetiker, line 13:
```suggestion
    "back-end",
    "backend",
```

cboulanger: The main NPM keywords are actually with the dashes, but sure, we can add the other spelling as well.

johnspackman (approved):

level420 (approved):

PR #10320, opened 2022-01-03
Fix jsdoc parameter type of MProperty.prototype.set [ci skip ci]

cboulanger:
The the type of the first parameter of `MProperty.prototype.set` is not `Map | String`, but `Object | String` since `Map` is very specific type (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) not an `Object` *used* as a map or dictionary. We might have this documentation problem at other places, too, but it is specifically salient here because it leads to linter warnings in IDEs.


derrell:
I agree that with modern JS, this should change. However, just changing `Map` to `Object` reduces clarity. An `Object` can be pretty much anything, whereas `Map` typically indicates that it was a simple object with members, and could be JSONified. I believe `Map` is used in many places to indicate this.

I'd suggest that we allow a special type such `ObjectMap` or `JSONifyableObject` or some such thing, in place of `Map` which now has a different meaning than it did in the past.

I recommend against simply changing occurrences of `Map` to `Object`, as that loses meaning.

derrell (changes requested):
see my comment in Conversation

cboulanger:
@derrell You are right but that's the JSDoc standard that IDEs rely on, you cannot introduce symbols that don't make sense to standard-compliant tools. And `Object` is completely correct: you can feed in any `Object` instance to `MProperty.prototype.set` and it will apply it to the target instance - as long as the properties of the target receive the right type (as checked in the setter methods), it will be totally fine. It doesn't need to be a JSONifyable object - this wouldn't describe the method signature correctly.

cboulanger:
For example, this would be legal: `myQxObject.set({foo:"bar", errorFunc: err => console.error(err)})`

derrell:
Our usage of JSDoc isn't standards-compliant in general, anyway, because in many places, we swap the order of the type and the variable name, per the standard. (I think it was a PR in the past year that allowed either order.)

In this particular case, the documentation specifies that `data` "is a map of property values", which is adequate additional documentation. Maybe the general case requirement, if we want to replace `Map` with `Object` elsewhere, is to ensure that the documentation states that it is a key/value map.

derrell (approved):

derrell:
I've been researching JSDoc. There appears to be no constraints on the type in curly braces, so tools should handle whatever we decide to put there. We can use something meaningful.

I understand your point about it not necessarily needing to be JSONifyable. You mentioned Dictionary, and that's actually a common term that describes what is expected. I suggest we use Dictionary in lieu of Object, for clarity.

derrell (changes requested):
See further comments in Conversation. Suggest replacing Object with Dictionary

cboulanger:
@derrell In my opinion, we should strive to make our JSDoc comments digestable by modern IDEs, which try hard to parse source code and do post-hoc typing. #a `Dictionary` type doesn't exist, so WebStorm, for example, will show an error. IMHO, the JSDOC type hints should refer to Types that actually exist, not to semantics that we invent.

cboulanger:
We can use the free-style comments after the type declaration to inform the user what the type *should* contain, then a more generic type doesn't hurt...

derrell:
I think it's much more important that documentation be as meaningful as possible, than that some particular IDE barfs or doesn't barf on it. There is no requirement in JSDoc for the type to exist, so it's really an error of WebStorm to complain about it.

That said, we can probably create a real type, e.g., `qx.type.Dictionary` (which extends Object) that nicely documents its own purpose, so that the type actually exists...

cboulanger:
@derrell Sure  - let's discuss this in  #10321? In this particular case, `Object` is correct, you can pass any type of Object to `MProperty.prototype.set`, and `Map` is clearly wrong, so the proposed change is unrelated to any more general discussion.

cboulanger:
To illustrate the problem: this will be the case for all advanced real-time code parsers/linters
Bildschirmfoto 2022-01-03 um 18 30 36

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

hkollmann:
@derrell : Could you confirm this now and can we do the general change later?

derrell (approved):
For this particular issue, which documents that the `data` parameter is a map of property values, I'm fine with switching the type from Map to Object for now, and we'll do the more general change later.

level420 (approved):

Issue #10220, opened 2021-06-17
COMPILER: inconsistent behaviour for paths to js source files in map files between source and build target


level420:
Currently when using the compiler to compile the source version of an app `qx compile -t source` the created `.map` files contain absolute paths to the original js source files. 

A simple test on linux reveals:
```
npm i -g @qooxdoo/compiler
cd /tmp
qx create test --noninteractive
cd test
qx compile -t source
head -4 compiled/source/test/package-1.js.map
```
the resulting output is:
```
{
  "version": 3,
  "sources": [
   "/usr/lib/node_modules/@qooxdoo/compiler/node_modules/@qooxdoo/framework/source/class/qx/Bootstrap.js",
  ...
}
```
now if we compile the build target with
```
qx compile -t build
head -1 compiled/build/test/index.js.map
```
results in:
```
{
  "version":3,
  "sources":[
    "../../usr/lib/node_modules/@qooxdoo/compiler/node_modules/@qooxdoo/framework/source/class/qx/Bootstrap.js",
   ...
}
```
The different behavior complicates serving of the js source files via a web server for debugging purposes.

Both compile targets build and source should use absolute paths in map files.

In the luxury version there could be a configuration option in `compile.json` which allows to select between relative and absolute paths e.g. `sourceMapRelativePaths: true` or similar.

Issue #10145, opened 2021-02-24
COMPILER: move from uglify-es to terser


hkollmann:
uglify-js is not longer supported. Supported fork is https://www.npmjs.com/package/terser

hkollmann:
should be backported to 6.0

PR #10328, opened 2022-01-07
Move to terser and fix paths in source map files

hkollmann:
see https://github.com/qooxdoo/qooxdoo/issues/10145 and https://github.com/qooxdoo/qooxdoo/issues/10220


source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -168,15 +168,6 @@
...
             "$ref": "#/properties/parts"
-          },
-          "minify": {
-            "$ref": "#/properties/targets"
-          },
-          "save-source-in-map": {
-            "$ref": "#/properties/targets"
-          },
-          "save-unminified": {
-            "$ref": "#/properties/targets"
```
cboulanger, line 179: Wouldn't this require a change in version - it no longer is 1.0.0 if we remove a range of keys

johnspackman: changing the version number of the schema is a nightmare because it effectively breaks everything and means that you must have a specific framework to compile, meaning that compiler and all packages have to be upgraded at once, or not at all

hkollmann: and it will not break anything - the tags are just unsupported.

derrell (approved):

source/resource/qx/tool/schema/compile-1-0-0.json
cboulanger: Ok I see, glad we don#t have to change the version number for the reasons that @johnspackman mentioned.

cboulanger (approved):

johnspackman (approved):

source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -128,6 +128,11 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
         type: "boolean",
         default: false
       },
+      "source-map-relative-paths": {
+        "describe": "If true, the source file will be saved in the map file if the target supports it. Can be overridden on a per application basis.",
```
oetiker, line 132: is it the source file or the source map file?

source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -269,6 +260,10 @@
             "description": "If true, the source file will be saved in the map file if the target supports it. Can be overridden on a per application basis.",
             "type": "boolean"
           },
+          "source-map-relative-paths": {
+            "description": "If true, the source map files will be saved with relative paths in the map file if the target supports it. Can be overridden on a per application basis.",
```
oetiker, line 264: is it the source map file or the source file ?

oetiker (changes requested):

source/resource/qx/tool/schema/compile-1-0-0.json
hkollmann: It's the path to the source file. 

{
  "version":3,
  "sources":[
    "../../usr/lib/node_modules/@qooxdoo/compiler/node_modules/@qooxdoo/framework/source/class/qx/Bootstrap.js",
   ...
}

level420 (approved):

source/resource/qx/tool/schema/compile-1-0-0.json
oetiker: @hkollmann the two descriptions do not align!

oetiker (approved):

Issue #10330, opened 2022-01-11
qx es6ify bugs


hkollmann:
1. .prettierignore is ignored.

reproduce:

run qx es6ify . in qooxdoo root.

/bin/ is processed.

2. parameter  exclude is not handled as array:
   if (exclude && filename.startsWith(exclude)) {

johnspackman:
Also the ex lint no longer understands our syntax because we misuse super()

What do you mean about 2.? I don't understand

hkollmann:
Parameter description says that it is type array. Handling is done as single string

hkollmann:
> Also the ex lint no longer understands our syntax because we misuse super()
> 
> What do you mean about 2.? I don't understand 

Could be handled by an eslint rule, could'nt it?

johnspackman:
Possibly, the reason it exists is because it is possible to provide paths on the command line which are inclusion, so exclude is supposed to balance that

johnspackman:
>>Also the ex lint no longer understands our syntax because we misuse super()
>Could be handled by an eslint rule, could'nt it?

no, the problem is that the underlying babel parser sees `super` and rejects it because it is not in an es6 `class` so is illegal; the solution is to produce a custom eslint parser, based on the babel eslint parser, which turns on the switch to tell babel to allow super anywhere.  I had the same problem in es6ify, the fix was this one line: https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/Es6ify.js#L161

PR #10333, opened 2022-01-12
es6ify: fix ignore, see https://github.com/qooxdoo/qooxdoo/issues/10330

hkollmann:
add some paths to ignore

### DIFF
--- a/source/class/qx/tool/cli/commands/Es6ify.js
+++ b/source/class/qx/tool/cli/commands/Es6ify.js
@@ -58,8 +58,12 @@ qx.Class.define("qx.tool.cli.commands.Es6ify", {
       await this.base(arguments);
       const ignoreFileName = ".prettierignore";
       const ig = ignore();
-      if (fs.existsSync(ignoreFileName)) {
-        ig.add(fs.readFileSync(ignoreFileName).toString());
+      try {
+        ig.add((await fs.promises.readFile(ignoreFileName)).toString());
+      } catch (err) {
+        if (err.code !== 'ENOENT') {
+          throw err;
+        }
       }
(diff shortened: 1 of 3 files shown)


johnspackman (approved):

cboulanger (approved):

source/class/qx/tool/cli/commands/Es6ify.js
```diff
@@ -55,13 +56,23 @@ qx.Class.define("qx.tool.cli.commands.Es6ify", {
   members: {
     async process() {
       await this.base(arguments);
+      const ignoreFileName = ".prettierignore";
+      const ig = ignore();
+      if (fs.existsSync(ignoreFileName)) {
+        ig.add(fs.readFileSync(ignoreFileName).toString());
```
derrell, line 62: Maybe we don't care in this app, but instead of the Sync variants, this could await the async variants of `exists` and `readFile`

oetiker (approved):

source/class/qx/tool/cli/commands/Es6ify.js
```diff
@@ -55,13 +56,27 @@ qx.Class.define("qx.tool.cli.commands.Es6ify", {
   members: {
     async process() {
       await this.base(arguments);
+      const ignoreFileName = ".prettierignore";
+      const ig = ignore();
+      try {
+        ig.add((await fs.promises.readFile(ignoreFileName)).toString());
+      } catch (err) {
+        if (err.code !== 'ENOENT') {
```
derrell, line 64: Much better! Avoids the race condition, too!

derrell (approved):

level420 (approved):

Issue #10133, opened 2021-02-16
COMPILER: add mangle-privates parameter to config.json, application


hkollmann:
as discussed here:

Henner Kollmann @hkollmann 14:15
Any idea how to write unit tests for private methods? Switch off mangle-privates? Other ideas?

Dietrich Streifert @level420 14:18
How about deriving from that class and using a public method there to test the private member?
Arg! Should read more carefully: this would only work for protected members.

Henner Kollmann @hkollmann 14:36
The problem is the mangling of private methods. Protected methods are not mangled - so it runs out of the box.

Dietrich Streifert @level420 15:57
@hkollmann  yes, you're right.

Christian Boulanger @cboulanger 16:02
@hkollmann Maybe have a special compilation target for testing that does not mangle privates?

Henner Kollmann @hkollmann 16:08
mangle-privates is just a command line parameter in the moment. Not setable via compile.js. What's about adding it as application setting?

Dietrich Streifert @level420 16:37
that would be cool. with the generator this could be achieved by omitting privatesin the optimizeconfiguration. See https://qooxdoo.readthedocs.io/en/latest/pages/tool/generator/generator_config_ref.html#pages-tool-generator-generator-config-ref-compile-options

Christian Boulanger @cboulanger 16:37
👍

hkollmann:
done with https://github.com/qooxdoo/qooxdoo/commit/1a4387409e5196d73062176e180dc25cf14a4e80

PR #10339, opened 2022-01-19
https://github.com/qooxdoo/qooxdoo/issues/10133: move mangle-privates to compile.json

hkollmann:


source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -247,29 +268,34 @@
             "type": ["boolean","string"],
             "description": "When true, output the typescript definition into ``./compiled/source/qooxdoo.d.ts`. If you use a string instead of true, the string is the name of the file which is generated inside the target output directory"
           },
+          "mangle-privates": {
+            "description": " Whether and how to mangle private identifiers. Can be overridden on a per application basis. Can be overridden on a per application basis.",
```
derrell, line 272: Repeated sentence "Can be overridden on a per application basis"

Also, it should be "per-application" (with a dash) here and other places where that phrase is used

derrell (changes requested):

cboulanger (approved):

johnspackman (approved):
Approving this change (thank you @hkollmann ) , my only thought is that perhaps in future command line options which change the output should be removed altogether and only allowed in compile.js?

Years ago I think the idea was that you could compile just with command line and no compile.js at all, but that's obviously not viable (not any more anyway), and I question the benefit of leaving them in.

hkollmann:
@johnspackman : You can compile without compile.js! Most of my projects do so.

johnspackman:
> @johnspackman : You can compile without compile.js! Most of my projects do so.

Oh wow!  OK I take it all back! :) 👍

oetiker:
according to semantic versioning, the version of the schema should change when it ismodified .. since this change is compatible, the version number would be 1-1-0 ...

hkollmann:
@oetiker : this changes are introduced here https://github.com/qooxdoo/qooxdoo/commit/1b6c45e5b2a4658030e52608c1b26e9f6c50670d#diff-2c50702c8b657fb45f0b13b2b56b8f8ec10ad4a87a76ed43d2b7292ef4152795. But where lost.

We discussed the version change sometime ago the result that it will bring us into an update hell...

oetiker:
> @oetiker : this changes are introduced here [1b6c45e#diff-2c50702c8b657fb45f0b13b2b56b8f8ec10ad4a87a76ed43d2b7292ef4152795](https://github.com/qooxdoo/qooxdoo/commit/1b6c45e5b2a4658030e52608c1b26e9f6c50670d#diff-2c50702c8b657fb45f0b13b2b56b8f8ec10ad4a87a76ed43d2b7292ef4152795). But where lost.
> 
> We discussed the version change sometime ago the result that it will bring us into an update hell...

oh ... missed that one ... well can't help you there ... I can't do more than provide all the tools for sustainability if the mode is to ignore it ... well ..

oetiker (approved):

johnspackman:
>> We discussed the version change sometime ago the result that it will bring us into an update hell...

>if the mode is to ignore it ... well ..

It's a temporary issue - the compiler requires the current version of compile.json and Manifest.json, so if we change the version number updates then everything needs to be changed (eg Manifest.json means that every plugin will need to have the schema version changed).  For users who switch between compiler versions, this means that the plugin versions have to be switched back and forth also, for no good reason. 

One day there would be a simple mapping between version numbers, eg a v1.0.0 can be translated into v2.0.0 on the fly so that the compiler can work

derrell (approved):

level420 (approved):

cboulanger (approved):

PR #10346, opened 2022-01-25
prepare doc for v7

hkollmann:


docs/core/classes.md
```diff
@@ -262,78 +262,41 @@ ability to override methods in derived classes.
...
+constructor, but note that you have to specify the method name:
 
 ```javascript
 qx.Class.define("qx.test.Cat", {
   extend: qx.test.Animal,
   members: {
-    makeSound: function () {
-      this.base(arguments);
+    makeSound() {
+      super.makeSound(arguments);
```
derrell, line 289: `arguments` here should be unnecessary (wrong), I think.

docs/core/classes.md
```diff
@@ -345,22 +308,20 @@ qx.Class.define("qx.test.Cat", {
...
-`this.constructor.someStaticVar`. Please note, for `this.constructor` to be
-available, the class must be a derived class of `qx.core.Object` , which is
-usually the case for regular, non-static classes.
-
-Instead of `this.constructor` you can also use the alternative syntax
-`this.self(arguments)`.
+`qx.test.Animal.someStaticVar`.  Please note that `this.constructor` is not
+recommended because it always refers to the subclass of the object, and therefore
+the implementation can change unexpectedly at run time.  Always specify class names
+explicitly in static methods.
```
derrell, line 321: This is way too strong of a statement. `this.constructor` is, IMO, often the preferred form for one's own code because (a) it's far shorter, and (b) development is simpler, particularly when needing to change class names, because one needn't search/replace for all occurrences of the old class name. I would much prefer that we describe the problem (even leave the "not recommended" text if you want) as is done in the "Please note" sentence, but exclude the "Always" sentence.

docs/core/classes.md
```diff
@@ -345,22 +308,20 @@ qx.Class.define("qx.test.Cat", {
...
 
-In purely static classes for calling a static method from another static method,
-you can directly use the `this` keyword, e.g. `this.someStaticMethod(x)`.
+You should also not use the `this` keyword in purely static classes, because static
```
derrell, line 323: s/You should also not/Unless you know how the static method is called in all cases, you should also not/

docs/core/environment.md
```diff
@@ -206,8 +206,7 @@ those keys which have **not** been given a value in the configuration.
 
 This isn't always practical - for example, you may wish to use the configuration
 to set the default value for an environment setting - to prevent the compiler 
-eliminating the code, use the "preserveEnvironment" setting (or in the generator,
-disable the `variants` optimisation and set `qx.allowUrlSettings` to true).
+eliminating the code, use the "preserveEnvironment" setting.
```
derrell, line 209: s/to prevent the compiler eliminating/to prevent the compiler from eliminating/

docs/desktop/gui/customwidgets.md
```diff
@@ -99,7 +99,7 @@ look like:
           this._add(control);
           break;
       }
-      return control || this.base(arguments, id);
+      return control || super(id);
```
derrell, line 102: ok, good. This is *different* than ES6 `super`, which require that `super()` be called before `this` is available. We have much code that depends on access to `this` before the superclass method is called, so I'm really glad it didn't try to emulate ES6 exactly.

docs/desktop/gui/themes.md
```diff
@@ -420,7 +420,7 @@ result is identical to the real alias
 
 When extending themes the so-named `base` flag can be enabled to include the
 result of this selector of the derived theme into the local selector. This is
-quite comparable to the `this.base(arguments, ...)` call in member functions of
+quite comparable to the `super(...)` call in member functions of
```
derrell, line 423: Line 421 is talking about `base` so I'm not sure this makes sense simply replacing `this.base` with `super` in this case...

docs/release-notes.7.0.md
```diff
@@ -0,0 +1,53 @@
...
+
+- `qx.ui.form.Slider` now works correctly in vertical orientation.
+The maximum value is at the top of the slider and the minimal value
+at the bottom. Before that, the 2 values were reversed 
+
+- all flash supporting classes are removed - flash is dead since January 2020.
+
+- `qx.ui.command.Group` fixed a bug where new `qx.ui.command.Command` added was
+not set the `active` status of the group, thus staying active even if the group
+was inactive.
```
derrell, line 25: s/was not set/did not set/

docs/release-notes.7.0.md
```diff
@@ -0,0 +1,53 @@
...
+
+  ```
+
+
+## Accessibility
+
+Accessibility is an important topic in nowadays webdevelopment in order to allow assistive technologies to better help persons with disabilities. Qooxdoo now supports better integration of [WAI-ARIA](https://www.w3.org/TR/wai-aria-1.1/) and more keyboard accessibility.
```
derrell, line 32: s/nowadays webdevelopment/modern web development/
s/help persons with/help people with/

derrell (changes requested):

hkollmann:
@derrell : Could you do your fixes direct in the branch? Many thanks!

derrell:
@hkollmann If I put the changes directly into the branch, they don't get reviewed. Why do we want to move to such a model?

docs/core/classes.md
hkollmann: Do you have any suggestions?
@johnspackman : What do you think about?

docs/desktop/gui/themes.md
hkollmann: @johnspackman : What do you think?

hkollmann:
Only for this documentation changes - would be less work for me as not english native speaker

docs/core/classes.md
johnspackman: I'll amend it, but the reason why I put it in there is because there have been a couple of questions recently (in gitter and on SO) where users thought that `this.constructor` will return the class of the method in which it appears, and did not understand why it would not work when they extended or mixed in.

docs/desktop/gui/themes.md
johnspackman: Line 421 is talking about the `base` keyword in theme definitions, which is still called `base`

johnspackman:
> @hkollmann If I put the changes directly into the branch, they don't get reviewed. Why do we want to move to such a model?

They are being reviewed, in this PR.  Henner's just suggesting that we all share this one branch rather than multiple PRs, and if there are minor tweaks that you suggest such as replacing "has not " with "did not", it's probably quicker and easier if you just do it via the github editor

derrell (approved):

johnspackman (approved):

oetiker (approved):

johnspackman (approved):

level420 (approved):

cboulanger (approved):

PR #10347, opened 2022-01-26
fix performance slowdown between 6 and 7

### LINKED ISSUE #10345: Performance degradation with Virtual DOM and Material Icons
> Can you get this into a reproducible example? One which shows the huge hits your taking would be really useful, but if not then just a comparative example where 6.x vs master has a noticeable difference

I finally found the time to put a minimal example together:
```js
function makeSelectBox(){
  var box = new qx.ui.form.SelectBox();
  for (var i=0; i<50; i++) {
    box.add(new qx.ui.form.ListItem("Free text", "@MaterialIcons/text_fields/12"));
  }
  return box;
}

// Create a button
var button1 = new qx.ui.form.Button("Click me", "pull10200/test.png");

var doc = this.getRoot();

doc.add(button1, {left: 100, top: 50});

// Add an event listener
button1.addListener("execute", function() {
  for(var i=0; i<3; i++){
    mainSelect.setSelection([mainSelect.getChildren()[i]]);
  }
});

var mainSelect = makeSelectBox();
doc.add(mainSelect, {left: 100, top: 100});

var vBox = new qx.ui.container.Composite(new qx.ui.layout.VBox());
doc.add(vBox, { left: 100, top: 130 });
// Add an event listener
mainSelect.addListener("changeSelection", ()=>{
  var startTime = performance.now()
  vBox.removeAll();
  /* eslint no-alert: "off" */
  for (let i = 0; i < 20; i++) {
    vBox.add(makeSelectBox());
  }
  var endTime = performance.now();
  alert(endTime - startTime);
});
```

The same code above shows a huge difference in performance between qx 6 and 7, I host both version here:
Demo qooxdoo v7: https://adrelino.github.io/pull10200
Demo qooxdoo v6: https://tinyurl.com/yc94hqv3

Just click the button or change the topmost selectbox' selection to start the measurements.

with qooxdoo v6 I get consistent timings of around 200 msec at most whenever I change the selection of the topmost selectbox.
with qooxdoo v7, the first time I get a timing of around 500msec, but which each subsequent selection change the time doubles. So in addition to beeing less performant, there seems to be a garbage collection problem or something like that.

_Originally posted by @adrelino in https://github.com/qooxdoo/qooxdoo/issues/10200#issuecomment-1021184712_

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo/issues/10345


cboulanger (approved):

derrell (approved):

hkollmann (approved):

level420:
@johnspackman lot of tests are failing with `setUp failed: object[property] is not a constructor`

johnspackman:
> @johnspackman lot of tests are failing with `setUp failed: object[property] is not a constructor`

I don't understand why - I'm running them in the qxl.testrunner and they are passing.

@cboulanger please can you have a look?  it seems that in qxl.testtapper, the `qx.test.io.jsonrpc.Client.setupFakeTransport` is saying that the function to call is `_createTransport`, but that function returns an instance of `qx.bom.request.Xhr` whereas `qx.dev.unit.MMock.injectStub` expects a class, and that's why it fails.

cboulanger:
@johnspackman I can have a look but the same error is shown for tests that have nothing to do with `qx.test.io.jsonrpc.*` (which has passed before), so the problem must be more general.

hkollmann:
I will have a look - some problem with sinon

oetiker (approved):

adrelino:
I just checked out https://github.com/johnspackman/qooxdoo/tree/performance-fix and verified that the timings are down to about 200ms consistently, just as they were before in v6.

Thanks for the prompt fix, I could not have done it myself since I still don't really understand the initial reason behind the slowdown ;-)

johnspackman:
you're welcome :+1:  The problem was caused by a long standing bug, but along the way I found a series of other little tweaks

johnspackman:
@hkollmann please can you try running esx6ify on this branch and let me know if it creates loads of changes?  I'm getting different whitespace which is messing with my merge

hkollmann:
es6ify is inserting a lot of blank lines.

hkollmann:
[test.zip](https://github.com/qooxdoo/qooxdoo/files/7951190/test.zip)
See diff

hkollmann:
I used the fixed version

hkollmann:
This is not an effect of your last fix - i tested with an older branch.

derrell:
```
es6ify()
{
  for (let i = 0; i < 3; i++) {
    es6ify_internal();
  }
}
```

hkollmann:
LOL

derrell:
I was actually only half kidding. That may not be the appropriate solution, but the output of `es6ify` should be idempotent. Running it multiple times should not, IMO, change the result. If that silly code I posted is the easiest way to solve it, then maybe it's appropriate as a first cut at solving the problem. Better would be to find out why it requires three passes before it stabalizes.

hkollmann:
We need to find out - could be more or less then three 

Von meinem iPhone gesendet

level420 (approved):

PR #10348, opened 2022-01-27
fixes a bug (which caused qx.dev.unit.Sinon to crash es6ify)

johnspackman:


cboulanger (approved):

hkollmann:
And how can we fix the framework?

johnspackman:
> And how can we fix the framework?

Do you mean the performance PR?

oetiker (approved):

level420 (approved):

hkollmann:
Yes. I mean we run es6ify on the whole code base and do not know which parts are affected by this bug. I think it is neccessary to es6ify again. Or what do you think?

hkollmann (approved):

johnspackman:
No it should be fine - the bug is that when an ES6 Symbol is used as a key in an object literal where the value is a function, so it was just that es6ify is being tripped up by the updated Sinon.js file only.

hkollmann:
So test failures https://github.com/qooxdoo/qooxdoo/pull/10347 has another reason

johnspackman:
yes :(

hkollmann:
assuming your permission i fixed it directly in your branch.

derrell (approved):

PR #10349, opened 2022-01-27
make es6ify idempotent

hkollmann:
The problem is the call of babelCore.transform. This will change output during repeated calls.
Implement an modfied idea from @derrel. 

while (true) {
        result = babelCore.transform(src, config);
        if (result.code === src) {
          break;
        }
        src = result.code;
      }

### DIFF
--- a/source/class/qx/tool/compiler/Es6ify.js
+++ b/source/class/qx/tool/compiler/Es6ify.js
@@ -162,14 +162,21 @@ qx.Class.define("qx.tool.compiler.Es6ify", {
         },
 
         generatorOpts: {
-          retainLines: true
+          retainLines: true,
+          compact: false
         },
 
         passPerPreset: true
       };
 
       let result;
+      let cycleCount = 0;
       while (true) {
+        cycleCount++;
+        if (cycleCount > 10) {
+          qx.tool.compiler.Console.warn(`Can not find a stable format for ${this.__filename}`);
+          break;
+        }
         result = babelCore.transform(src, config);
         if (result.code === src) {
           break;
--- a/source/class/qx/tool/compiler/Es6ify.js
+++ b/source/class/qx/tool/compiler/Es6ify.js
@@ -168,7 +168,14 @@ qx.Class.define("qx.tool.compiler.Es6ify", {
         passPerPreset: true
       };
 
-      let result = babelCore.transform(src, config);
+      let result;
+      while (true) {
+        result = babelCore.transform(src, config);
+        if (result.code === src) {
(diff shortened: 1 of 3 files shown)


johnspackman (approved):
Brilliant :)

Approving, just wondered if it works on the pre-es6ified source?

cboulanger (approved):

source/class/qx/tool/compiler/Es6ify.js
```diff
@@ -168,7 +168,14 @@ qx.Class.define("qx.tool.compiler.Es6ify", {
...
 
-      let result = babelCore.transform(src, config);
+      let result;
+      while (true) {
+        result = babelCore.transform(src, config);
+        if (result.code === src) {
+          break;
+        }
+        src = result.code;
+      }
```
derrell, line 178: This looks good... but is there any chance that it could get into a state where it toggles back and forth between two results, or otherwise doesn't stabilize? Maybe add a maximum number of iterations (10? 50?) as a "just in case"? I'll approve this, but I do have a bit of concern.

derrell (approved):

source/class/qx/tool/compiler/Es6ify.js
derrell: Maybe even add a printed warning if it ever does hit the maximum count, so that we can look into what's causing it?

oetiker: yep, having an escape hatch sounds like a good thing ...

source/class/qx/tool/compiler/Es6ify.js
```diff
@@ -168,7 +168,14 @@ qx.Class.define("qx.tool.compiler.Es6ify", {
         passPerPreset: true
       };
 
-      let result = babelCore.transform(src, config);
+      let result;
+      while (true) {
```
oetiker, line 172:
```suggestion
      let cycleCount = 0;
      while (true) {
```

derrell: lol. I'll assume you pressed the green button before you were finished typing in your suggestion...

hkollmann:
@johnspackman : Yes it did. Tested with framework code.

hkollmann:
Thought about an exit strategy after launching the PR...

oetiker (approved):

hkollmann:
After merging this we have rerun the es6ify process...

derrell (approved):

level420 (approved):

PR #10353, opened 2022-02-04
Changed jsdoc of find method qx.data.Array

goldim:
Wonder if need to change undefined into null for return value. Could do it in this PR too.

### DIFF
--- a/source/class/qx/data/Array.js
+++ b/source/class/qx/data/Array.js
@@ -1017,7 +1017,7 @@ qx.Class.define("qx.data.Array", {
      *   <li><code>index</code>: the index of the current item</li>
      *   <li><code>array</code>: The native array instance, NOT the data array instance.</li>
      * @param self {var?undefined} The context of the callback.
-     * @return {Boolean} <code>true</code>, if any element passed the test function.
+     * @return {var | undefined} The found item.
      */
     find(callback, self) {
       return this.__array.find(callback, self);


source/class/qx/data/Array.js
```diff
@@ -1017,7 +1017,7 @@ qx.Class.define("qx.data.Array", {
      *   <li><code>index</code>: the index of the current item</li>
      *   <li><code>array</code>: The native array instance, NOT the data array instance.</li>
      * @param self {var?undefined} The context of the callback.
-     * @return {Boolean} <code>true</code>, if any element passed the test function.
+     * @return {var | undefined} The found item.
```
johnspackman, line 1020: We don't tend to use `var`, and optional results are indicated with a trailing `?` - I think that the correct jsdoc would be `@return {*?}` or `@return {Object?}` - but not sure which is correct/better

johnspackman (changes requested):

source/class/qx/data/Array.js
goldim: Agree, but nevertheless it is still much better than Boolean and the wrong description of return which confuses a documentation reader. In other methods of qx.data.Array I noticed there is used for example for removeAt method. I doubt that Object? is valid jsdoc. Especially an undefined value is needed to be highlighted/emphasized.

derrell: We use `?` to indicate optional parameters. I've never seen, and would find confusing, a return value indicated that way. It's not optional; it can just be of various types (including undefined).

We use `var` at multiple places in the framework, e.g., in qx.bom.Event and qx.core.MLogging. I think the PR's description of the `{var | undefined}` is exactly appropriate. (We have discussed possibly moving to `*` particularly as it relates to arrays (I think) but I don't know that we ever came to a conclusion on that or decided to begin to use it. If we want to do so, this return value could be documented as `{* | undefined}`.

Regarding changing `undefined` to `null`... See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find. `undefined` is correct.

derrell (approved):

hkollmann (approved):

source/class/qx/data/Array.js
goldim: @derrell I am not sure about "| undefined" because other typical methods which can return undefined/null value have just "var" in JSDoc. But another thing I've noticed in ApiViewer in min/max methods. JSDoc of these methods have Number | null type but in ApiView they are shown as | null. Is it bug of ApiViewer or not correct notation?

derrell: That sounds like a bug in ApiViewer. Can you maybe provide an example class documented that way, but shows up incorrectly? (A new PR with that would be better than here.)

johnspackman:
@derrell 

> We use ? to indicate optional parameters. I've never seen, and would find confusing, a return value indicated that way. It's not optional; it can just be of various types (including undefined).

There are 74 instances where we use `?` in `@return` and only 9 instances where we use `undefined`
There are 1076 instances where we use `?` in `@param` and only 20 instances where we use `undefined`

> We use var at multiple places in the framework, e.g., in qx.bom.Event and qx.core.MLogging. 

There are 145 uses of `var` in `@return` and 529 in `@param`
There are 30 uses of `*` in `@return` and 103 uses in `@param`

My regex's could probably do with some work (ie those numbers are a little high) but I think that we can agree that we are quite consistently inconsistent, probably because some of this stuff dates back so far.

Personally, I like the brevity of `?` and would expect that it's logical to be consistent between the `@param` and `@return`

I can't remember discussing `*` for arrays, I think I tend to use `*` to mean `var` because one of my IDE's started proposing it and I had no other opinion on it.

I don't want to derail @goldim 's PR with a debate, but just flag it up as something to think harder about - I will be working on coding type checking based on JSDOC expressions shortly.

source/class/qx/data/Array.js
goldim: @derrell I'll do it probably like issue for **ApiViewer**. The class is the same one:  `qx.data.Array`. Last detail I would like to clarify  `var | undefined` is fine, no need to change onto just `var` ?

derrell:
lol. You did a far more detailed investigation than I did. Yup, we are decidedly consistently inconsistent.

I already approved this PR, and I'm fine with any of the above moving forward. Consistency would be nice. (I do find `@return {*?}` unclear, though, so favor `| undefined`. I'll be fine with the prevailing opinion.)

derrell (approved):
I guess my approval got removed with a new commit? Anyway, re-approving.

source/class/qx/data/Array.js
goldim: https://github.com/qooxdoo/qxl.apiviewer/issues/48

goldim:
@derrell I am sorry, I did it. Thank you.
@johnspackman I could create new issue related with such kind of problem for future decision about style and format. This way the problem won't be forgotten.

oetiker (approved):

level420 (approved):

cboulanger (approved):

johnspackman (approved):

Issue #10355, opened 2022-02-08
bugs in mixin with derived classes


hkollmann:
Add this tests to qx.test.Mixin:
```
    testPatchOverwrittenDerived_1() {
      qx.Class.define("qx.Patch", {
        extend: qx.core.Object,

        members: {
          sayJuhu() {
            return "Juhu";
          }
        }
      });

      qx.Class.define("qx.PatchDerivative", {
        extend: qx.Patch,

        members: {
          sayJuhu() {
            return super.sayJuhu() + " Derivative";
          },

          foo() {
            return "foo";
          }
        }
      });

      qx.Mixin.define("qx.MPatch", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " Mixin";
          },
        }
      });
      qx.Class.patch(qx.Patch, qx.MPatch);

      var o = new qx.PatchDerivative();
      this.assertEquals("Juhu Mixin Derivative", o.sayJuhu());
      o.dispose();

    },
```
Here qx.MPatch.sayJuhu is never called. Instead qx.PatchDerivative.sayJuhu called qx.Patch.sayJuhu directly.

```
    testPatchOverwrittenDerived_2() {
      qx.Class.define("qx.Patch", {
        extend: qx.core.Object,

        members: {
          sayJuhu() {
            return "Juhu";
          }
        }
      });

      qx.Class.define("qx.PatchDerivative", {
        extend: qx.Patch,

        members: {
          sayJuhu() {
            return super.sayJuhu() + " Derivative";
          },

          foo() {
            return "foo";
          }
        }
      });

      qx.Mixin.define("qx.MPatch", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " Mixin";
          },
        }
      });
      qx.Class.patch(qx.PatchDerivative, qx.MPatch);

      var o = new qx.PatchDerivative();
      this.assertEquals("Juhu Derivative Mixin", o.sayJuhu());
      o.dispose();

    }
```
This gives an stackoverflow in call to qx.PatchDerivative.sayJuHu.

PR #10357, opened 2022-02-09
 fix for  Mixin in derivated classes 

hkollmann:
fix for  https://github.com/qooxdoo/qooxdoo/issues/10355: Mixin in derivated classes


derrell (approved):

johnspackman (approved):

oetiker (approved):

cboulanger (approved):
Trusting you here, I have no idea what this does

hkollmann:
.superclass.prototype.func calls the actual function of the base class whereas .prototype.base.func calls just acopy of base class function. This copy reflects the state during creation oft the class.
If you add an mixin this copy not longer actual so the changed function is not called.
This is done by call to .superclass.prototype.func

johnspackman:
although I approved this yesterday, I'm having second thoughts that it's the right solution  - 

I'm trying to figure out why the methods have a `base` property at all, because this fix makes it look like it's not needed (it may have only been needed for the generator). 

Does this work if the superclass has a method with the same name which is added by a different mixin?

EG if A is the base class, B derives from A, and C derives from B; MA, MB, and MC are all mixins which all define a method `doSomething`, and where MA is included in A, MB in B, and MC in C.

Each mixin's `doSomething()` can call `super()`, so all three methods are called.  Same if A has `doSomething` and MA is not used

hkollmann:
base is needed in Mixin.baseClassMethod to find original method if you use super() in an mixin.

hkollmann:
Could you provide a test case for this?

Von meinem iPhone gesendet

hkollmann:
testPatchMultiOverwrittenDerived() {
      qx.Class.define("qx.A", {
        extend: qx.core.Object,
        members: {
          sayJuhu() {
            return "A";
          }
        }
      });
      qx.Class.define("qx.B", {
        extend: qx.A,
        members: {
          sayJuhu() {
            return super() + " B";
          }
        }
      });
      qx.Class.define("qx.C", {
        extend: qx.B,
        members: {
          sayJuhu() {
            return super() + " C";
          }
        }
      });
      qx.Mixin.define("qx.MA", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " MA";
          },
        }
      });
      qx.Mixin.define("qx.MB", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " MB";
          },
        }
      });
      qx.Mixin.define("qx.MC", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " MC";
          },
        }
      });
      qx.Class.patch(qx.A, qx.MA);
      qx.Class.patch(qx.B, qx.MB);
      qx.Class.patch(qx.C, qx.MC);

      var o = new qx.C();
      this.assertEquals("A MA B MB C MC", o.sayJuhu());
      o.dispose();

    }

hkollmann:
@johnspackman: This test is running. Is it what you ask for?

level420 (approved):

PR #10359, opened 2022-02-11
Fix uglify

hkollmann:
This PR fixes 2 problems:

qx compile --target=build --save-unminified results in empty files
When bundeling a lot of files where written because of package change.
Now only one file for each part and bundle is written


derrell (approved):
Blind acceptance.

cboulanger (approved):

hkollmann:
changes in target did not work

PR #10370, opened 2022-02-20
TreeVirtual - Add options for expandAll, collapseAll that suppress redraws

rad-pat:
Also adds expandBeneath, collapseBeneath for expand/collapse tree below a node.

This has a considerable performance benefit when expanding a large tree and maintains backwards compatibility if new parameter `suppressRedraw` is not passed.

### DIFF
--- a/source/class/qx/ui/treevirtual/MNode.js
+++ b/source/class/qx/ui/treevirtual/MNode.js
@@ -165,7 +165,33 @@ qx.Mixin.define("qx.ui.treevirtual.MNode", {
     },
 
     /**
-     * Set the opened state for a node.  (Note that this method has no effect
+     * Opens all nodes in the tree with minimal redraw
+     */
+    nodeOpenAll() {
+      var model = this.getTableModel();
+      model.getData().forEach(node => {
+        if (node) {
+          model.setState(node.nodeId, {bOpened: true}, true);
+        }
+      });
+      model.setData();
+    },
+
+    /**
+     * Closes all nodes in the tree with minimal redraw
+     */
+    nodeCloseAll() {
+      var model = this.getTableModel();
+      model.getData().forEach(node => {
+        if (node) {
+          model.setState(node.nodeId, {bOpened: false}, true);
+        }
+      });
+      model.setData();
+    },
+
+    /**
+     * Internal call to set the opened state for a node. (Note that this method has no effect
      * if the requested state is the same as the current state.)
      *
      * @param nodeReference {Object | Integer}
@@ -173,11 +199,16 @@ qx.Mixin.define("qx.ui.treevirtual.MNode", {
      *   represented either by the node object, or the node id (as would have
      *   been returned by addBranch(), addLeaf(), etc.)
      *
-     * @param b {Boolean}
+     * @param opened {Boolean}
      *   The new opened state for the specified node.
      *
+     * @param cascade {Boolean}
+     *   Whether to descend the tree changing opened state of all children
+     *
+     * @param isRecursed {Boolean?}
+     *   For internal use when cascading to determine outer level and call setData
      */
-    nodeSetOpened(nodeReference, b) {
+    _nodeSetOpenedInternal(nodeReference, opened, cascade, isRecursed) {
       var node;
 
       if (typeof nodeReference == "object") {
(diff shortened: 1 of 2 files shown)


source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -795,7 +795,7 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
      * Sets the whole data en bulk, or notifies the data model that node
      * modifications are complete.
      *
-     * @param nodeArr {Array | null}
+     * @param [nodeArr] {Array | null}
```
cboulanger, line 798: This syntax is not supported, AFAIK

derrell, line 798: Does the API Vieiwer understand this syntax? We've talked about updating what we allow inside of the curly braces, but not brackets around the variable name...

cboulanger (approved):
Please fix the param documentation, other than that, very useful contribution, thanks!

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -1099,6 +1104,76 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
       return (
         this.__tree.getAllowNodeEdit() && this.getNodeFromRow(rowIndex).bCanEdit
       );
+    },
+
+    /**
+     * Expands all nodes in the tree with minimal redraw
+     */
+    expandAll() {
```
derrell, line 1112: What do you think of moving the changes from here onwards into MNode.js, which already has similar utility methods such as `nodeToggleOpened`? If doing so, `collapseBeneath` could be renamed to match the naming in MNode, to `nodeCollapseBeneath`. Similarly, `expandBeneath` -> `nodeExpandBeneath`.

derrell (changes requested):

derrell:
BTW, these look to be very useful additions!

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
derrell: In fact, there's already a `nodeSetOpened`. Maybe add a `recursive` parameter to it, in lieu of `nodeCollapseBeneath` and `nodeExpandBeneath`? The only question, both in this case and in your original, is whether the specified node itself should be collapsed or expanded. Just need to make sure it's documented what it does.

oetiker (approved):

rad-pat:
As commented by @derrell, I'll reorganise to add a cascade/recurse flag to nodeSetOpened.

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
derrell: The most common way, in qooxdoo code, of expressing this with an optional array that assumes a default value of null, would be:
```
@param nodeArr {Array?null}`
```

Does that stop your IDE from complaining?

johnspackman: that's not correct syntax.  It should be:

```
@param nodeArr {Array?}
```

rad-pat: Great, that works, thanks

derrell: I think @johnspackman's comment about incorrect syntax was pertaining to what was in the PR, not my suggestion. Throughout qooxdoo, there are examples of what I suggested... but if you don't care that an optional parameter is replaced by null (you only care that it's optional), then John's syntax is more common.

johnspackman: yes, that's right sorry for the confusion...

rad-pat:
It's all been moved over to MNode now, hopefully this looks ok.

source/class/qx/ui/treevirtual/MNode.js
```diff
@@ -173,11 +199,16 @@ qx.Mixin.define("qx.ui.treevirtual.MNode", {
...
      *   The new opened state for the specified node.
      *
+     * @param cascade {Boolean}
+     *   Whether to descend the tree changing opened state of all children
+     *
+     * @param isRecursed {Boolean?}
+     *   For internal use when cascading to determine outer level and call setData
      */
-    nodeSetOpened(nodeReference, b) {
+    nodeSetOpened(nodeReference, opened, cascade, isRecursed) {
```
derrell, line 211: I see the need for `isRecursed` but it seems ugly to expose that to the user. What do you think about having `nodeSetOpened` call an internal method, e.g., `_nodeSetOpenedInternal`, that adds/uses that additional argument?

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -907,9 +907,12 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
      *   {@link SimpleTreeDataModel}.  Each property value will be assigned
      *   to the corresponding property of the node specified by nodeId.
      *
+     * @param suppressRedraw {Boolean}
+     *    If true then prevents redraw upon expand/contract. Useful if expanding/contracting many
```
derrell, line 911: This doc string is a bit restrictive, I think. This change is making this new parameter available for general use, so how about, `If true then prevents redraw; it becomes the caller's responsibility to call `setData()` subsequently, to cause a redraw.` ?

derrell (changes requested):

source/class/qx/ui/treevirtual/MNode.js
rad-pat: Added private method

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
rad-pat: Amended param description

derrell (approved):
Looks good. Thanks!

johnspackman (approved):

hkollmann (approved):

Issue #10371, opened 2022-02-20
`qx lint` does not honour the `-f` format flag


rad-pat:
With Qooxdoo v7 Beta 10
When running command `qx lint -f=checkstyle`, the output is not the XML checkstyle format as expected.

This was running fine with Qooxdoo 6
Looking at the code, it's likely that the cause it that it is forced to compact mode in the case when there are many errors. However, we are expecting to parse the checkstyle format in our Jenkins build to give us a lovely chart.
https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/cli/commands/Lint.js#L162..L164

hkollmann:
Could you check if this is true with current version: 
// If there are too many errors, the pretty formatter is appallingly slow

rad-pat:
When the comment was made, it looks like it was reporting 51,000 `Errors` in the Qooxdoo codebase (with --warnAsError flag set https://github.com/qooxdoo/qooxdoo/pull/10123). Running on the Qooxdoo codebase currently, there are 1234 errors found. This takes ~25s to run locally for me with any of the output formats selected (and those lines of code removed to prevent forcing to compact mode).

hkollmann:
What's about a PR to change?

rad-pat:
Yep, will do

PR #10372, opened 2022-02-21
Don't switch lint output format

rad-pat:
This looks to have been added to improve performance of linting the qx source when there were 51,000 errors. That same operation currently only reports 1234 issues and takes about 25 seconds for me regardless of output format. Switching the format based on the number of errors is causing issues for our CI build which expects "checkstyle" XML format.

To address https://github.com/qooxdoo/qooxdoo/issues/10371

### DIFF
--- a/source/class/qx/tool/cli/commands/Lint.js
+++ b/source/class/qx/tool/cli/commands/Lint.js
@@ -157,13 +157,23 @@ qx.Class.define("qx.tool.cli.commands.Lint", {
         }
         if (report.errorCount > 0 || report.warningCount > 0) {
           let outputFormat = this.argv.format || "codeframe";
-
-          // If there are too many errors, the pretty formatter is appallingly slow
-          if (report.errorCount + report.warningCount > 150) {
-            outputFormat = "compact";
-          }
           const formatter = await linter.loadFormatter(outputFormat);
           const s = formatter.format(report);
+          // If there are too many errors, the pretty formatter is appallingly slow so if the
+          // user has not specified a format, change to compact mode
+          const maxDefaultFormatErrorCount = 150;
+          if (report.errorCount + report.warningCount > maxDefaultFormatErrorCount) {
+            if (!this.argv.format) {
+              qx.tool.compiler.Console.info(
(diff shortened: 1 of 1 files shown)


oetiker (approved):

hkollmann (approved):

derrell (approved):

source/class/qx/tool/cli/commands/Lint.js
```diff
@@ -157,11 +157,6 @@ qx.Class.define("qx.tool.cli.commands.Lint", {
         }
         if (report.errorCount > 0 || report.warningCount > 0) {
           let outputFormat = this.argv.format || "codeframe";
-
-          // If there are too many errors, the pretty formatter is appallingly slow
-          if (report.errorCount + report.warningCount > 150) {
-            outputFormat = "compact";
-          }
```
johnspackman, line 164: shouldn't this be protected by a command line switch?  EG to always force the output format?  That condition was added specifically because of a massive performance hit (AFACR it made the compiler look like it had hung) so I'm not keen on having it removed.

johnspackman (changes requested):

source/class/qx/tool/cli/commands/Lint.js
rad-pat: Isn't asking for the specific format a good enough command line switch? Perhaps it should say, if the format is default and the errors are in this range then use the compact format? Even so, if you expect the results back in a certain format, getting a different format is not ideal. Perhaps even force the format if an output file is specified?

rad-pat:
I've changed this around so that it will only switch format if the user has not requested a specific format. What do you think of this?

derrell (approved):

johnspackman (approved):

johnspackman:
> I've changed this around so that it will only switch format if the user has not requested a specific format. What do you think of this?

that's what I was thinking of :)

hkollmann (approved):

rad-pat:
Rebased against latest master

level420 (approved):

PR #10377, opened 2022-03-03
Raise error if table is editing when trying to modify the data in the model

rad-pat:
As discussed briefly with @derrell in Gitter, this PR adds the throwing of an error if trying to set the data for the table model when the table is still being edited. The model *could* change to a dataset that no longer includes the row being edited at which point an error is raised when the user attempts to navigate from the cell being edited.

This would represent a change in behaviour. Currently, if the table is being edited and the table model is updated, the displayed table will update and there would be an error in the console (unlikely to be seen by the user) when navigating from the edited cell. With this change, the table will not be updated.

### DIFF
--- a/source/class/qx/ui/table/model/Abstract.js
+++ b/source/class/qx/ui/table/model/Abstract.js
@@ -66,11 +66,22 @@ qx.Class.define("qx.ui.table.model.Abstract", {
     this.__columnIndexMap = {};
   },
 
+  statics: {
+    /**
+     * Member to control if a table should throw an error when you try to change the
+     * data model data whilst there is an incomplete edit. It could possibly break
+     * current implementations so only introduce the change from QX v8.
+     * Ref: https://github.com/qooxdoo/qooxdoo/pull/10377#discussion_r818697343
+     */
+    THROW_ON_MODEL_CHANGE_DURING_EDIT: parseInt(qx.core.Environment.get("qx.version"), 10) >= 8
+  },
+
   members: {
     __columnIdArr: null,
     __columnNameArr: null,
     __columnIndexMap: null,
     __internalChange: null,
+    __table: null,
 
     /**
      * Initialize the table model <--> table interaction. The table model is
@@ -84,7 +95,17 @@ qx.Class.define("qx.ui.table.model.Abstract", {
      *   The table to which this model is attached
      */
     init(table) {
-      // default implementation has nothing to do
+      // store a reference back to the table
+      this.__table = table;
+    },
+
+    /**
+     *
+     *
+     * @returns table {qx.ui.table.Table}
+     */
+    getTable() {
+      return this.__table;
     },
 
     /**
@@ -292,6 +313,15 @@ qx.Class.define("qx.ui.table.model.Abstract", {
       }
 
       this.setColumnNamesByIndex(columnNameArr);
+    },
+
+    _checkEditing() {
+      if (!qx.ui.table.model.Abstract.THROW_ON_MODEL_CHANGE_DURING_EDIT) {
+        return;
+      }
+      if (this.getTable() && this.getTable().isEditing()) {
+        throw new Error("A cell is currently being edited. Commit or cancel the edit before setting the table data");
+      }
     }
   },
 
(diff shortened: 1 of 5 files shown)


cboulanger (approved):

cboulanger:
Would you modify https://github.com/qooxdoo/qooxdoo/blob/master/CHANGELOG.md to add this breaking change (not in API but in behavior) - when you're at it, please rename "upcoming" to "v7.0.0" and add a new "Upcoming" section for your entry. Thank you!

johnspackman (approved):

rad-pat:
I forgot to look at the tests - will take a look at those now

source/class/qx/ui/table/model/Abstract.js
```diff
@@ -292,6 +303,12 @@ qx.Class.define("qx.ui.table.model.Abstract", {
       }
 
       this.setColumnNamesByIndex(columnNameArr);
+    },
+
+    _checkEditing() {
+      if (this.getTable().isEditing()) {
+        throw new Error("A cell is currently being edited. Commit or cancel the edit before setting the table data");
```
derrell, line 310: I just wonder if this is going to break anyone's app? I suspect there are cases of editing that would not have caused a failure. What do you think of adding a static `qx.ui.table.model.Abstract.THROW_ON_MODEL_CHANGE_DURING_EDIT` which defaults to `false` for now (and your app would set it to `true`), until qx v8 where we change the default to `true`, thereby maintaining backward compatibility and allowing an easy "fix" even in v8, for those with a valid and working use case?

derrell (changes requested):

source/class/qx/ui/table/model/Abstract.js
rad-pat: Yep, that's fine by me

rad-pat: Is this a valid thing to do for a static?
```
  statics: {
    THROW_ON_MODEL_CHANGE_DURING_EDIT: qx.core.Environment.get("qx.version") >= "8.0.0"
  },
```

derrell: I think so. It should probably have a comment, maybe even referencing the PR discussion about it, though.

source/class/qx/ui/table/model/Abstract.js
```diff
@@ -66,11 +66,22 @@ qx.Class.define("qx.ui.table.model.Abstract", {
...
   },
 
+  statics: {
+    /**
+     * Member to control if a table should throw an error when you try to change the
+     * data model data whilst there is an incomplete edit. It could possibly break
+     * current implementations so only introduce the change from QX v8.
+     * Ref: https://github.com/qooxdoo/qooxdoo/pull/10377#discussion_r818697343
+     */
+    THROW_ON_MODEL_CHANGE_DURING_EDIT: qx.core.Environment.get("qx.version") >= "8.0.0"
```
derrell, line 76: Just realized this won't work. As soon as we get to v10, it'll fail because as a string, "10.0.0" < "8.0.0". I do believe there's a way to do this with numbers or proper version number comparision, though. @johnspackman @hkollmann isn't there a way to do it?

cboulanger: But maybe this should be done in the `defer()` magic method?

derrell: Isn't `qx.version` sometimes not a number, e.g., "7.1.0-beta.1" or something? Doesn't it require a semver comparison?

johnspackman: `parseInt` stops at the first non-digit, so `parseInt(qx.core.Environment.get("qx.version"), 10)` will be the first number.

derrell: @johnspackman 's suggestion of using `parseInt` solves the problem in this case. I suggest we go with that.

derrell (approved):
Thanks!

oetiker (approved):

level420 (approved):

hkollmann (approved):

PR #10378, opened 2022-03-09
Automatically detect and browserify CommonJS modules

derrell:
This implements a working proof of concept. It requires that the
compile.json target indicates `application-types : [ "browser" ]`
exclusively (i.e., it doesn't yet build browser and node apps
separately, bringing in the browserified files only in the browser
version.

Usage:
In a target, specify the application type as exclusively "browser",
and add an array of CommonJS modules to be browserified. For example:
```
      "application-types": [
        "browser"
      ],
      "commonjs-modules": [
        "semver"
      ],
```

You may also add a similar `commonjs-modules` array in an
`application` entry, in which case those CommonJS modules will be
browserified for all targets which indicate "browser" as the only
application type. The resulting set of modules to be browserified is
the union of those indicated in `applciations` and those indicated in
the `target`.

### DIFF
--- a/source/class/qx/tool/cli/commands/Compile.js
+++ b/source/class/qx/tool/cli/commands/Compile.js
@@ -915,6 +915,7 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
        */
       let targetOutputPaths = {};
       let makers = [];
+      let commonjsModules = [];
       targetConfigs.forEach(targetConfig => {
         if (!targetConfig.appConfigs) {
           qx.tool.compiler.Console.print(
@@ -1310,6 +1311,42 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
           (targetConfig.defaultAppConfig.type || "browser") === "browser"
         ) {
           targetConfig.defaultAppConfig.app.setWriteIndexHtmlToRoot(true);
+
+          // Merge the application's (common to targets) commonjs
+          // modules list and the current target's commonjs modules
+          // list
+          commonjsModules =
+            mergeArray(
+              [],
+              targetConfig.defaultAppConfig["commonjs-modules"] || [],
+              targetConfig["commonjs-modules"] || []);
+
+          // If there are any commonjs modules, browserify them
+          //
+          // TODO: Move this to someplace where it can be generated
+          // directly in the output directory. There's no need,
+          // really, for it to be placed in the source directory.
+          if (commonjsModules.length > 0) {
+            let command;
+            let output;
+            let analyser = maker.getAnalyser();
+            const childProcess = require("child_process");
+            command =
+              `browserify -r ${commonjsModules.join(" -r ")}` +
+              ` -o source/resource/script/qx-commonjs-browserified.js`;
+            fs.mkdirSync("source/resource/script", { recursive : true });
+            try {
+              output = childProcess.execSync(command);
+            } catch (e) {
+              qx.tool.compiler.Console.log(`Failed: ${output}`);
+              throw e;
+            }
+
+            // Don't show "unresolved" error for `require` since the
+            // browserified code defines it as a global
+            analyser.setIgnores(
+              mergeArray([], analyser.getIgnores(), [ "require" ]));
+          }
         } else {
           qx.tool.utils.files.Utils.safeUnlink(
             target.getOutputDir() + "index.html"
--- a/source/class/qx/tool/cli/commands/Compile.js
+++ b/source/class/qx/tool/cli/commands/Compile.js
@@ -1321,26 +1321,12 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
               targetConfig.defaultAppConfig["commonjs-modules"] || [],
               targetConfig["commonjs-modules"] || []);
 
-          // If there are any commonjs modules, browserify them
-          //
-          // TODO: Move this to someplace where it can be generated
-          // directly in the output directory. There's no need,
-          // really, for it to be placed in the source directory.
+          // If there are any commonjs modules, create the list of modules to later be browserified
           if (commonjsModules.length > 0) {
-            let command;
-            let output;
             let analyser = maker.getAnalyser();
-            const childProcess = require("child_process");
-            command =
-              `browserify -r ${commonjsModules.join(" -r ")}` +
-              ` -o source/resource/script/qx-commonjs-browserified.js`;
-            fs.mkdirSync("source/resource/script", { recursive : true });
-            try {
-              output = childProcess.execSync(command);
-            } catch (e) {
-              qx.tool.compiler.Console.log(`Failed: ${output}`);
-              throw e;
-            }
+
+            // Save the list of required CommonJS modules
+            analyser.setCommonjsModules(commonjsModules);
 
             // Don't show "unresolved" error for `require` since the
             // browserified code defines it as a global
--- a/source/class/qx/tool/cli/commands/Compile.js
+++ b/source/class/qx/tool/cli/commands/Compile.js
@@ -915,7 +915,7 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
        */
       let targetOutputPaths = {};
       let makers = [];
-      let commonjsModules = [];
+
       targetConfigs.forEach(targetConfig => {
         if (!targetConfig.appConfigs) {
           qx.tool.compiler.Console.print(
@@ -1311,28 +1311,6 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
           (targetConfig.defaultAppConfig.type || "browser") === "browser"
         ) {
           targetConfig.defaultAppConfig.app.setWriteIndexHtmlToRoot(true);
-
-          // Merge the application's (common to targets) commonjs
-          // modules list and the current target's commonjs modules
-          // list
-          commonjsModules =
-            mergeArray(
-              [],
-              targetConfig.defaultAppConfig["commonjs-modules"] || [],
-              targetConfig["commonjs-modules"] || []);
-
-          // If there are any commonjs modules, create the list of modules to later be browserified
-          if (commonjsModules.length > 0) {
-            let analyser = maker.getAnalyser();
-
-            // Save the list of required CommonJS modules
-            analyser.setCommonjsModules(commonjsModules);
-
-            // Don't show "unresolved" error for `require` since the
-            // browserified code defines it as a global
-            analyser.setIgnores(
-              mergeArray([], analyser.getIgnores(), [ "require" ]));
-          }
         } else {
           qx.tool.utils.files.Utils.safeUnlink(
             target.getOutputDir() + "index.html"
--- a/source/class/qx/tool/compiler/ClassFile.js
+++ b/source/class/qx/tool/compiler/ClassFile.js
@@ -223,6 +223,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
     };
 
     this.__externals = [];
+    this.__commonjsModules = [];
 
     this.__taskQueueDrains = [];
     this.__taskQueue = async.queue(function (task, cb) {
@@ -281,6 +282,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
     __privates: null,
     __blockedPrivates: null,
     __externals: null,
+    __commonjsModules: null,
 
     _onTaskQueueDrain() {
       var cbs = this.__taskQueueDrain;
@@ -496,6 +498,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
       delete dbClassInfo.translations;
       delete dbClassInfo.markers;
       delete dbClassInfo.fatalCompileError;
+      delete dbClassInfo.commonjsModules;
       for (var key in this.__dbClassInfo) {
         dbClassInfo[key] = this.__dbClassInfo[key];
       }
@@ -665,6 +668,11 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
       // Errors
       if (this.__fatalCompileError) {
         dbClassInfo.fatalCompileError = true;
+      }
+
+      // CommonJS modules
+      if (this.__commonjsModules.length) {
+        dbClassInfo.commonjsModules = qx.lang.Array.clone(this.__commonjsModules);
       }
 
       return dbClassInfo;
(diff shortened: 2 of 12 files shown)


oetiker:
I guess the json schema would need updateding too .. https://github.com/qooxdoo/qooxdoo/pull/10379

derrell:
Thanks @oetiker. I'm going to work on automating this through the compiler simply seeing calls to `require` rather than the user stating modules to include in compile.json... although I'm not yet sure that's a better implementation than having the user call them out explicitly. Still thinking through it.

oetiker:
well, understanding the require would be more nodish and as such probably more readily understood by most users.

source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -1310,6 +1311,42 @@ Framework: v${await this.getQxVersion()} in ${await this.getQxPath()}`);
...
+              `browserify -r ${commonjsModules.join(" -r ")}` +
+              ` -o source/resource/script/qx-commonjs-browserified.js`;
+            fs.mkdirSync("source/resource/script", { recursive : true });
+            try {
+              output = childProcess.execSync(command);
+            } catch (e) {
+              qx.tool.compiler.Console.log(`Failed: ${output}`);
+              throw e;
+            }
+
```
johnspackman, line 1344: No compile / build tasks should be done here.  `qx.tool.cli.*` is all about configuring the compiler API, and it is definitely not intended that compiler functionality should be baked into the CLI.  

The `commonjs-modules` in the **target** definition in `compile.json` should be copied into the `qx.tool.compiler.targets.Target` as a property; the `commonjs-modules` in the **application** definition should be copied into `qx.tool.compiler.app.Application` as a property.

As explained in Gitter, in order to include the `commonjs-modules` (which as you say, would be a union of those listed in the target and the application), you need to write an implementation of `qx.tool.compiler.targets.meta.AbstractJavascriptMeta` which would perform the code above which runs browserify.

Once you have a meta class, you need to modify `qx.tool.compiler.targets.Target.generateApplication` to optionally include the new meta class `if (application.getType() == “browser") { …`

johnspackman (changes requested):

johnspackman:
So that it doesn't get lost, here is the explanation I posted in Gitter about how to add this feature.  I've divided into auto-detection and output, because I agree that auto-detection isn't necessary, just a nice-to-have.

## Auto detection of `require()`

>Derrell Lipman @derrell 
>>@johnspackman What are your thoughts about how to identify require to ensure that it's the global require function and not a local variable declared by the user's app. Also, how do you envision deciding whether to browserify the require's arguments[0].value or whether it will be done at runtime, via NodeJS?

>Derrell Lipman @derrell [Mar 04 16:25](https://gitter.im/qooxdoo/qooxdoo?at=62223d6f9a09ab24f38b2bba)
>>For the latter, compile.json has application-types which can be "browser" or "node" (anything else?). Does ClassFile.js have access to that information?
Another question for @johnspackman: would you expect that the user has already done npm install <package> prior to qx-compiling a browser app that does require("<package>"), or that the compiler should run npm install <package> if it finds require("<package>")? I'm inclined towards the user issuing the npm install prior to compiling, as the user may want a non-standard release or something, but interested in your thoughts.

>John Spackman @johnspackman [Mar 05 09:38](https://gitter.im/qooxdoo/qooxdoo?at=62232f8a9a09ab24f38ce944)
>>@derrell the compiler maintains a stack of scopes which you can search upwards looking for variables; a similar example to what you’re after is where the compiler takes each variable and goes up the stack looking to see if the variable is just an alias for this, eg:

```
{
  abc: function() {
    var that = this;
    this.callSomething(function() {
      that.doSomethingElse();
    });
  }
}
```

>>This is the code that works it out: [https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/ClassFile.js\#L1667-L1692](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/ClassFile.js/#L1667-L1692)

>>As for looking at the argument, there is also a similar issue with qx.core.Environment.get; first, literal arguments are resolved so that qx.core.Environment.get(“abc” + "def”) appears as just qx.core.Environment.get(“abcdef”), but there is no attempt to resolve more complicated expressions, eg including variables

>>it’s a requirement therefore that if you want the compiler to figure it out at compile time, you must use literals, otherwise it must be runtime

>>Here’s an example on getting the literal value: https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/ClassFile.js#L1774-L1775
You can get the supported application types from the qx.tool.compiler.Analyser, which you can access via the this.__analyser in qx.tool.compiler.ClassFile

>>As for installing packages, I agree that it is something that the user should do, however at some point this is something that we would take account of - for example, if you install a package it could have a series of npm module requirements, so we would like to be able to track what is required. You will need to store the list of require-d modules somewhere, and the place to do this is in qx.tool.compiler.ClassFile._compileDbClassInfo, which is where various complied states are copied into the “database” object - see https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/ClassFile.js#L507-L671

## Output of browserify code

>>When it comes to generating the browserify code, this will be a matter of intercepting the generation of the application from the list of required classes, after all the dependency calculation and other analysis work is complete; this is handled by the classes in qx.tool.compiler.targets.meta, which is a set of classes that are capable of producing a stream of code, and which can be plugged together - this allows the build process to decide whether to merge bits of JS together (eg for bundling), whether to wrap code in a beatifier or uglifier, etc. You will need to write an implementation of qx.tool.compiler.targets.meta.AbstractJavascriptMeta which gets the list of classes required for the application, reads the list of requires that ClassFile previously stored in the database for each class, and creates the browserify code and then output the browserify code as a stream.

>>The simplest example of this is [qx.tool.compiler.targets.meta.Polyfill](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/targets/meta/PolyfillJs.js)
Then, in [qx.tool.compiler.targets.Target.generateApplication](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/targets/Target.js#L334-L580) you need to add your new meta processor to inject the browserify code into the application - probably around [line 510](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/compiler/targets/Target.js#L510)


## Manually listing `commonjs-modules` in `compile.json`

>Derrell Lipman @derrell [Mar 05 14:26](https://gitter.im/qooxdoo/qooxdoo?at=62237331c61ef0178e5b95a9)
Thanks for all the hints, @johnspackman . This will be a non-trivial effort, it seems, but a good independent project with which I can gain some understanding of the compiler.

>@johnspackman I'm thinking that an easy-to-implement, reasonably straightforward for the user, implementation would be to add the option in a target in compile.json, an array commonjs-modules which is an array of strings, each containing either a module to be found in node_modules or a relative path to the directory containing package.json, for local modules. In either case, the dependencies would be expected to be found in node_modules or be further relative paths, such that browserify can find them. Implementing this implies that the compiler would run browserify passing each of the members of this array, and automatically include (as an "external"), the browserified script. require would then automatically "just work" and the keyword require could be added to globalSymbols to avoid the warning about unresolved use of it. Any objection? If we then, later, want to more automatically create that list of required modules, we could add it to ClassFile, but I'm not sure it's really even necessary.

>I envision commonjs-modules being used only in a target that also specifies an application-type of browser, but in any case, the script would only be added as an "external" when browser is in the list of application types.

>(Actually, maybe you'd want to add the script as an "external" only if browser is the only application type? I think otherwise, a NodeJS app would get the browserified file...???

>Derrell Lipman @derrell [Mar 10 15:37](https://gitter.im/qooxdoo/qooxdoo?at=622a1b33257a357825223fca)
>@johnspackman The default application-types is [ "browser", "node" ]. What does that really mean? I've only ever seen a single compiled output from that configuration, but the app is written either as a browser app, or as a node app. There are places in the compiler that ask whether "browser" is in the array, and do certain things, but I don't believe there is currently anything done differently when both "browser" and "node" are specified, such as generating two different compiled outputs. Can you explain how application-types as an array is meant to work?
(For the browserify branch I'm working on, I do need to somehow generate separate compiled output for browser and node, if we want to continue to allow application-types to contain both.)

>>John Spackman @johnspackman [Mar 10 17:09](https://gitter.im/qooxdoo/qooxdoo?at=622a30c299d94f5f0c44c8ec)
as you mention, the compiler checks the type and adjusts its behaviour; it will output different environment variables, and if the variables can differ between applications on the same target then they are determined at runtime instead of during transpilation (meaning that code elimination and dependency calculations will not happen); the main way of using is it is to define a target for “browser” and a target for “node”, and this allows you to enable babel optimisations (eg node >= 17 because you know what your server is running, but browser might include IE11 and therefore ES6 transpilation is necessary). 

>>By saying that the application is for the browser or node, allows the compiler to select the correct target.
EG if you have some code which can be used on client or server but uses if (qx.core.Environment.get(“qx.headless”)) or if (qx.core.Environment.get(“qx.compiler.targetType”) == …) then the compiler cannot elminate the code unless the environment variable can be determined at compile time. If the code contained in that if test brings in dependencies, you need the code elimination (otherwise you could load qx.ui.* into node causing a crash during startup, or vice versa)

source/class/qx/tool/cli/commands/Compile.js
johnspackman: There's a key question in your gitter that I missed previoously -

>Derrell Lipman @derrell Mar 05 17:35
(Actually, maybe you'd want to add the script as an "external" only if browser is the only application type? I think otherwise, a NodeJS app would get the browserified file...???

the "external" is really used for adding external scripts in the `source/resources` section which are provided by the user; they are expected to be loaded by the boot loader as either resources or from a URL.  

To answer your question about how to prevent the nodejs app getting the browserified file, that's what the `qx.tool.compiler.targets.Target` figures out, and in order to make it so that the browserified code can be easily stacked and used, the `qx.tool.compiler.targets.meta` classes are used.

I'll add a better explanation about why `.meta.` is important below

johnspackman:
## `qx.tool.compiler.targets.meta` classes

When compiling an application (any application, for browser or node) means taking various transpiled JS code files, plus some standard code, plus generating other bits of code (eg a boot loader generated from a template, or a list of resources along with image sizes), and wrapping it up together.

On top of this, source and build targets support parts where certain selected files are removed from the list of files to be output, joined together into a single temporary file, and then that temporary file is inserted into the build.  Source targets also support bundling.

Build targets have a goal of a single output file, and so they wrap the entire generation into a single file.

Also, it is possible to wrap parts or all of the output with other processes such as beautifier or uglifier.

Exactly how these pieces of code are assembled, and what processes are applied to each, is determined at runtime and can vary a lot depending on minor configuration changes.  Early versions of the compiler tried to take a rule-based approach to this, but it resulted in complex and difficult to understand code with subtle bugs that was very hard to add features into.

This is where `qx.tool.compiler.targets.meta.*` comes in - it is just a way of streaming some code, either as source code or as a file reference, and often that stream will come from the disk (eg copy a set of files) but sometimes it generates code (eg generate the boot loader from a template) or wraps some code (eg uglify/beautify the code which is streamed into it)

The key thing is that it then becomes much easier to assemble the metas as the Target needs, and also that the Target can modify that assembly for certain features.  The result is readable and maintainable code, with flexible and easily testable implementation.

derrell:
@johnspackman This is getting pretty close, and is ready for a review. My biggest problem is that when I find a missing module at the time of browserifying, I haven't yet found a clean way to display the missing module, not show any type of thrown error, and continue with generation so that any other missing modules can be displayed as well... and then have the compiler exit with a non-zero status. I found a way to do that during compiling of each class, but I haven't found a mechanism that's available while generating the target output. How can I accomplish it?

source/class/qx/tool/compiler/ClassFile.js
```diff
@@ -1648,6 +1656,60 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
...
 
+            // Are we looing at the Identifier `require`, and is it a
+            // function call (identified by having
+            // `path.node.arguments`? (A better implementation might
+            // look for a type of `CallExpression` at the top level of
+            // the node...) If so, we'll add
+            // qx-commonjs-browserified.js as an External so it gets
+            // pulled in automagically, but only if application-types
+            // is "browser" (and nothing else)
+            //
```
johnspackman, line 1667: That's a rational concern (esp given all the warnings in babel about not assuming context) but what you're doing here is correct IMHO.  Because you are in a CallExpression, you're just checking that the callee is an Identifier as opposed to a MemberExpression (or whatever).  So `require` is an Identifier but `someObject.require` is a MemberExpression which has two Identifiers within it.  

BTW I don't know if you've seen https://astexplorer.net/ ?  It's not 100% up to date with some ES6/Babel but it is fantastically useful in visualising the AST.

source/class/qx/tool/compiler/ClassFile.js
```diff
@@ -1648,6 +1656,60 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
...
+            // TODO: Move this to someplace that does it only if the
+            // current build target is "browser" even if there are
+            // multiple application types being built.
```
johnspackman, line 1670: is this comment out of date?

source/class/qx/tool/compiler/TargetError.js
```diff
@@ -0,0 +1,27 @@
+/* ************************************************************************
+ *
+ *    qooxdoo-compiler - node.js based replacement for the Qooxdoo python
+ *    toolchain
+ *
+ *    https://github.com/qooxdoo/qooxdoo-compiler
+ *
+ *    Copyright:
+ *      3033 Derrell Lipman
```
johnspackman, line 9: let me know when I can borrow your delorean 😉

johnspackman (approved):
Great, I'm really glad you opted for auto-detection, I think it's a really nice addition to the user experience 👍 

> I haven't yet found a clean way to display the missing module, not show any type of thrown error, and continue with generation 

I think that you're right that there isn't a good way of handling that except for a catastrophic exception or easily-overlooked text message.  While the classes support a persistent list of errors and warnings with machine readable, well-defined message IDs, the final target stage of linking it all together does not have any such thing.  Definitely something that needs to be sorted out properly, because then there's the possibility of adjusting severity by config.

In  the mean time, just take a view on whether to report an error message or throw an exception.

source/class/qx/tool/compiler/ClassFile.js
derrell: It was, yes.

source/class/qx/tool/compiler/TargetError.js
derrell: :smile:

hkollmann:
Documentation of the new feature where great. eg a short chapter in Development/The Qooxdoo Compiler and a sentence in the release docs.

derrell:
Thanks for the reminder @hkollmann. It was in my plans. I'll certainly do that.

derrell:
I have removed the DRAFT tag. This is now ready for full-team review.

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):
grand!

johnspackman:
please can we hang on before merging this, i'm getting some strange errors

johnspackman:
@derrell sorry, this was my mistake.  The problem is fixed, but I think it would be a good idea to not throw an error if the `require` cannot be found, because that will kill `qx compile --watch`

source/class/qx/tool/compiler/targets/meta/Browserify.js
```diff
@@ -0,0 +1,144 @@
...
+              message.push("  required from:");
+              [ ...references[id] ].forEach(refs => {
+                refs.forEach(ref =>
+                  {
+                    message.push(`    ${ref}`);
+                  });
+              });
+
+              qx.tool.compiler.Console.error(message.join("\n"));
+              throw new qx.tool.compiler.TargetError(`ERROR: could not locate require()d module: "${id}"`);
```
johnspackman, line 121: please can we not throw the error?  it's catastrophic to the build process

johnspackman (changes requested):

source/class/qx/tool/compiler/targets/meta/Browserify.js
derrell: Done.

johnspackman (approved):

level420 (approved):

PR #10379, opened 2022-03-10
update json schema

oetiker:


hkollmann:
Then we can close it?

derrell:
I don't see why not. Closing.

PR #10386, opened 2022-03-18
Allow dispose of __scrollerParent to dispose of panescrollers

rad-pat:
Remove `_cleanUpMetaColumns` in favour of disposing the `__scrollerParent` which disposes it's children immediately. `_cleanUpMetaColumns` calls `destroy()` which is delayed and the mouse events do still fire, specifically the pointermove which calls `getTableColumnModel()` which attempts to recreate a columnModel if its already disposed, leading to the call stack below.
Luckily I had a very reproducible example in our application despite being unable to build a simple replica.
```
Uncaught TypeError: Cannot read properties of null (reading 'getChildren')
    at wrapper._getPaneScrollerArr (Table.js:1118:36)
    at wrapper._onColVisibilityChanged (Table.js:1490:30)
    at Direct.js:133:37
    at Function._true [as then] (Utils.js:149:22)
    at Direct.js:132:26
    at Array.forEach (<anonymous>)
    at wrapper.dispatchEvent (Direct.js:107:19)
    at wrapper.wrappedFunction [as dispatchEvent] (Interface.js:531:31)
    at wrapper.dispatchEvent (Manager.js:958:22)
    at Registration.js:361:40
    at Function._true [as then] (Utils.js:149:22)
    at Object.fireEvent (Registration.js:360:22)
    at wrapper.fireDataEvent (MEvent.js:292:36)
    at wrapper.init (Basic.js:204:14)
    at wrapper.getTableColumnModel (Table.js:860:21)
    at wrapper._getResizeColumnForPageX (Scroller.js:2006:43)
_getPaneScrollerArr	@	Table.js:1118
_onColVisibilityChanged	@	Table.js:1490
(anonymous)	@	Direct.js:133
_true	@	Utils.js:149
(anonymous)	@	Direct.js:132
dispatchEvent	@	Direct.js:107
wrappedFunction	@	Interface.js:531
dispatchEvent	@	Manager.js:958
(anonymous)	@	Registration.js:361
_true	@	Utils.js:149
fireEvent	@	Registration.js:360
fireDataEvent	@	MEvent.js:292
init	@	Basic.js:204
getTableColumnModel	@	Table.js:860
_getResizeColumnForPageX	@	Scroller.js:2006
_onPointermovePane	@	Scroller.js:1076
(anonymous)	@	EventHandler.js:298
```

### DIFF
--- a/source/class/qx/ui/table/Table.js
+++ b/source/class/qx/ui/table/Table.js
@@ -2193,7 +2193,6 @@ qx.Class.define("qx.ui.table.Table", {
       );
     }
 
-    this._cleanUpMetaColumns(0);
     this.getTableColumnModel().dispose();
     this._disposeObjects(
       "__selectionManager",


source/class/qx/ui/table/Table.js
```diff
@@ -2193,7 +2193,6 @@ qx.Class.define("qx.ui.table.Table", {
       );
     }
 
-    this._cleanUpMetaColumns(0);
```
derrell, line 2196: It's not clear to me how this fixes the described problem. `__scrollerParent` was already being disposed at line 2200, so even if `_cleanUpMetaColumns` caused a delayed action, wouldn't those event listeners be cleaned up when `__scrollerParent` is disposed anyway?

derrell (changes requested):

source/class/qx/ui/table/Table.js
rad-pat: @derrell - I think when `_cleanUpMetaColumns` is called, and `destroy` is called for each Scroller, any event listeners are immediately removed from the Scroller itself, but not from it's child components. The `_paneClipper` child has this particular pointermove listener and so that would not get removed until disposal. Also, I think `destroy` removes the Scroller from it's parent `__scrollerParent` so when it is actually disposed, the Scrollers are still delayed. In our app, it is very easy to replicate the error, there is an "are you sure" dialog appears in the foreground to delete an item, when you click yes there is a backend call and then a call to destroy the table. There is a small delay in the dialog disappearing and in that time, you can wiggle the mouse over the table and 2 out of 3 times you could generate the error. With the change, I can't generate the error at all.

derrell: My concern is that `_cleanUpMetaColumns` was freeing the memory for each of `__scrollerParent`'s children, which is no longer being done. (`__scrollerParent` is a `qx.ui.container.Composite` which itself does not have a `destruct` method.) Now it may be unnecessary to explicitly clean that up, assuming that modern memory management will handle the cleanup. Is my concern unfounded? I just want to ensure we're not adding a memory leak here... 

@johnspackman You implemented the changes a year or so ago that negated the need for many `destruct` calls. What is the criteria, these days, for when it's needed and when it's not?

derrell: Oh! However, https://github.com/qooxdoo/qooxdoo/blob/b4e50a92b39b71cb5b49619694fa5c6121c4d1cd/source/class/qx/ui/core/Widget.js#L3669 disposes the children. I think we're good.

derrell (approved):

oetiker (approved):

level420 (approved):

hkollmann (approved):

johnspackman (approved):

PR #10395, opened 2022-05-05
Allow including local (not-in-node_modules) CommonJS or ES6 modules

derrell:
With this change, local modules, either CommonJS format or ES6 modules, can now be included in a compiled app. To do so, include a `localModules` map in an element of `applications` map. The keys in the `localModules` map are the names by which the local module may be `require()`d in qooxdoo code. The value for each of those keys is the path, from where `compile.json` is located, to the module to be included. For example:

compile.json:
```
...
  "applications": [
    {
      "class": "xxx.Application",
      "theme": "xxx.theme.Theme",
      "name": "xxx",
      "bootPath": "source/boot",
      "localModules": {
        "testes6": "./test/testes6.js",
        "testcommonjs": "./test/testcommonjs.js"
      }
    }
  ]
...
```

./test/testes6.js:
```
export default "hello world";
```

./test/testcommonjs.js
```
module.exports = "hi there";
```

Application.js:
```
qx.Class.define("xxx.Application",
{

  members :
  {
    main : function()
    {
      let test;

      this.base(arguments);

      test = require("testes6").default;
      console.log("test es6=", test);

      test = require("testcommonjs");
      console.log("test commonjs=", test);

      const semver = require("semver");
      console.log("expect '1.2.3': " + semver.valid("1.2.3"));
      console.log("expect null: " + semver.valid("a.b.c"));
      console.log("expect '1.2.3': " + semver.clean(" =v1.2.3 "));
    }
  }
});
```

@adrelino This should solve your issue requested on gitter. Would you please test it and report back?

### DIFF
--- a/docs/development/compiler/configuration/README.md
+++ b/docs/development/compiler/configuration/README.md
@@ -34,10 +34,10 @@ It is the developer's responsibility to ensure that any modules referenced in a 
 In addition to the automatic detection of calls to `require()` to
 bundle CommonJS modules, local modules -- either CommonJS modules or
 ES6 modules -- may be bundled and included in the application. To do
-so, include a `localModules` map in an element of the applications map
+so, include a `localModules` map in an element of the applications array
 in `compile.json`. The keys in the `localModules` map are the names by
-which the local module may be require()d in qooxdoo code. The value
-for each of those keys is the path, from where compile.json is
+which the local module may be `require()`d in qooxdoo code. The value
+for each of those keys is the path, from where `compile.json` is
 located, to the module to be included. The following example shows how
 this can be used:
 
@@ -116,7 +116,7 @@ qx.Class.define("xxx.Application",
 });
 ```
 
-Notice that `compile.json` includes only local modules `testes6.js` and `testcommon.js`. It does not include `included.js` but since `testes6.js `import`s `included.js`, and `testes6.js` is explicitly listed in `localModules`, `included.js` will be bundled as well.
+Notice that `compile.json` includes only local modules `testes6.js` and `testcommon.js`. It does not include `included.js` but since `testes6.js` `import`s `included.js`, and `testes6.js` is explicitly listed in `localModules`, `included.js` will be bundled as well.
 
 ## Configuration Files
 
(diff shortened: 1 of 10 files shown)


oetiker:
hmmm I wonder, how can you add new keys to config.json without changing the json schema, is this not enforced anymore by the compiler ?

johnspackman:
> But I wonder, how can you add new keys to config.json without changing the json schema, is this not enforced anymore by the compiler ?

Unfortunately we have to right now, because there is no compatibility matrix in the compiler.  EG if the config.json is changed to v1.0.1, the compiler will require every compile.json to be changed before it will compile.  

Some kind of transparent upgrade transformation when loading compile.json is on the todo list

johnspackman (approved):

oetiker:
the purpose of the json schema was to ensure that all is documented and validated ... by abandoning enforcement we go back to the old ways :(

The idea would be to have semantic versioning on the schema

The compiler can then decide how 'old' a version it would still support ... if the version is older then it aborts and points out the need to upgrade ...

We probably need rather little code for the 'transformation' as long as we do not change property names ...

hkollmann (changes requested):
Could you please update the schema file too?

johnspackman:
I completely agree, that it is sensible to have validation and would be a small amount of code to transform (eg) 1.0.0 into 1.0.1 (all the way up to whatever is the current version).  It's just not done yet

However, it's not too terrible in practice, provided that we do not make changes to the compile.json schema which are not 100% backwards compatible.  

AIUI the problem is that the JSON schema validator is rigid and says that 1.0.0 !== 1.0.1, and it does not follow semver.  Sounds pretty awful now I say it and it may be that we just need to tweak an option or something ...

oetiker (approved):

derrell:
@adrelino With this PR, you don't need to list every imported file in compile.json. It automagically locates other imported modules and includes them. To see this in action, use the above example I provided, but with the following changes:

Replace file `testes6.js with this:
```
export default "hello world";

import { included } from "./included.js";
console.log("testes6: included=", included);
```

Add new file `included.js`:
```
export let included = "INCLUDED";
```

`compile.json` is not modified, so still includes only local modules `testes6.js` and `testcommonjs.js`. It does not include `included.js` but you'll see when you run this that `included.js` is being imported.

I think you could replace your `main.js` that assigns all of the imports to `globalThis`, with a `main.js` which exports a map with each of the other imported modules. That allows you to assign that map to any variable when you import `main.js`, instead of having to pollute the global namespace.

Thoughts?

hkollmann (approved):

cboulanger (approved):

level420 (approved):

PR #10396, opened 2022-05-10
fix compile.json schema validation; enable it during `qx compile`

derrell:
There were multiple issue:

- the correct property banning extra keys was missing, and an obsolete
one was in its place.

- with the correct property there, it tries to use better-ajv-errors
but it was not properly required, so crashed when trying to generate
the error.

- schema validation was not implemented for `qx compile`


oetiker (approved):
great!

hkollmann (approved):

cboulanger (approved):

level420 (approved):

johnspackman (approved):

cboulanger (approved):

PR #10397, opened 2022-05-17
Allow parallel script loading in Firefox once more.

rad-pat:
As reported in Gitter, loading source compiled apps in Firefox is incredibly slow. It's because the parallel script load was disabled many years ago due to a bug in Firefox debugger. https://github.com/qooxdoo/qooxdoo/issues/8946

Given that the bug was fixed in 2015, it's probably safe to now remove the restriction.


cboulanger (approved):

johnspackman (approved):

level420 (approved):

derrell (approved):

oetiker (approved):

hkollmann (approved):

hkollmann:
Thanks @rad-pat !

PR #10399, opened 2022-05-18
 Convert `import` statements to `require()` in `require()`d dependencies. Fixes #10398

### LINKED ISSUE #10398: const ol = require("ol") gives exception
**Describe the bug**
import do not work for ol

**To Reproduce**
1. unpack attached test project
2. run npm install
3.  run npx qx compile

See exception:

Failed: Error: Parsing file D:\Projekte\github.windows\qooxdoo\test\node_modules\ol\index.js: 'import' and 'export' may only appear at the top level (5:0)

[test.zip](https://github.com/qooxdoo/qooxdoo/files/8710925/test.zip)

derrell:
This applies to the automatically discovered use of `require()` in
    browser apps, that causes those `require()`d modules to be bundled and
    included in the app. The problem was that `import` statements need to
    be at top level, but due to the wrapping generated by the qooxdoo
    compiler, they ended up not at the top level.
    
    This adds use of browserify plugin `esmify` that converts `import` to
    `require()`.

### DIFF
--- a/docs/development/compiler/configuration/README.md
+++ b/docs/development/compiler/configuration/README.md
@@ -34,10 +34,10 @@ It is the developer's responsibility to ensure that any modules referenced in a 
 In addition to the automatic detection of calls to `require()` to
 bundle CommonJS modules, local modules -- either CommonJS modules or
 ES6 modules -- may be bundled and included in the application. To do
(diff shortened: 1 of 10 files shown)


derrell:
Closing. Was based on a different PR branch. Will resubmit based on master.

PR #10400, opened 2022-05-18
Fix 10398: const ol = require("ol") gives exception

hkollmann:


derrell (approved):

cboulanger (approved):

johnspackman (approved):
no problem with the PR, but just wanted to ask if it was intended to have only this change in it?  IE does this fix https://github.com/qooxdoo/qooxdoo/issues/10398 ?  nI can't find any docs for the `global` config option

derrell:
The docs, limited as they are, for `global` are in https://www.npmjs.com/package/browserify. Search for `opts.global` to skip a bunch of unrelated hits. And yes, it fixes #10398.

derrell:
I had not comprehended the meaning of that option, so had not tried it.

hkollmann:
Found it here https://stackoverflow.com/a/56608843 and just tried it

hkollmann:
Here is the explantion of global: https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed 
not very helpfull

derrell:
lol @hkollmann. With https://github.com/qooxdoo/qooxdoo/pull/10400/commits/af7b932a526870defeabfabcb18c7b5e073a2520, you're acting like the US Congress, and sneaking in a completely unrelated clause to a "bill", in hopes that no one will notice :-)

hkollmann:
Sorry.
I found this issue during compiling my multi target project. It's just in the same workflow for me.

johnspackman (approved):

derrell (approved):

oetiker (approved):

cboulanger (approved):

level420 (approved):

PR #10425, opened 2022-07-01
add member declarations for previously undeclared members

derrell:
These ought to be documented and have more appropriate initializers than `undefined`. The new class system detects and warns of undeclared member variables. We'd be good to have these all declared soon/now so that when we later add JSDoc type definitions to the member declaration, they'll all already be there.

This is not an all-inclusive list of changes. It includes only those I've yet detected, i.e., classes I've yet run through the new class system.

### DIFF
--- a/source/class/qx/event/dispatch/Direct.js
+++ b/source/class/qx/event/dispatch/Direct.js
@@ -56,6 +56,8 @@ qx.Class.define("qx.event.dispatch.Direct", {
   */
 
   members: {
+    _manager : undefined,
+
     /*
     ---------------------------------------------------------------------------
       EVENT DISPATCHER INTERFACE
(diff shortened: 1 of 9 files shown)


cboulanger (approved):

oetiker:
it would be great to have docu strings on these too

level420 (approved):

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

hkollmann:
@derrel: is this ready for production?

derrell:
There are many, many more occurrences of missing declarations that the new class system identifies. I'm going to close this one, and will fix all of them in a single commit, either after the new class system is incorporated into master, or soon before (TBD exactly when)

PR #10432, opened 2022-08-07
Some refactoring and correction to FileSelectorButton widget

evandropoa:
- The internal file input element doesn't start hidden in Chrome 102.0.5005.61 64bits linux
- Refactored the element ID counter from a global variable to a static member, created an internal id generator

### DIFF
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -147,7 +147,7 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
         null,
         { id: id }
       ));
-      input.hide();
+      input.exclude();
       let label = new qx.html.Element("label", {}, { for: id });
       label.addListenerOnce(
         "appear",
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -64,12 +64,12 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
   extend: qx.ui.form.Button,
 
   statics: {
-    _FileInputElementIdCounter: 0,
-    _FileInputElementIdPrefix: "qxFileSelector_",
+    _fileInputElementIdCounter: 0,
+    _fileInputElementIdPrefix: "qxFileSelector_",
 
     _generateElementId: function(){
-      ++this._FileInputElementIdCounter;
-      return this._FileInputElementIdPrefix + this._FileInputElementIdCounter;
+      ++this._fileInputElementIdCounter;
+      return this._fileInputElementIdPrefix + this._fileInputElementIdCounter;
     }
   },
 
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -21,11 +21,11 @@
  * files from the local filesystem. A FileList is returned which allows access
  * to the content of the selected files from javascript. The file(s) can now be
  *  processed in javascript, or it/they can be uploaded to a server.
- * 
+ *
  * *Example*
  *
  * Post the content of the file to the server.
- * 
+ *
  * ```javascript
  * let button = new qx.ui.form.FileSelectorButton("Select File");
  * button.addListener('changeFileSelection',function(e){
@@ -40,9 +40,9 @@
  *   });
  * });
  * ```
- * 
+ *
  * Process the file directly in javascript using the FileReader API.
- * 
+ *
  * ```javascript
  * let button = new qx.ui.form.FileSelectorButton("Select File");
  * button.addListener('changeFileSelection',function(e){
@@ -134,9 +134,10 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
       let id = "qxFileSelector_" + InputElementIdCounter++;
       let input = (this.__inputObject = new qx.html.Input(
         "file",
-        { display: "none" },
+        null,
         { id: id }
       ));
+      input.hide();
       let label = new qx.html.Element("label", {}, { for: id });
       label.addListenerOnce(
         "appear",
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -157,10 +157,10 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
       );
       input.addListenerOnce("appear", e => {
         let inputEl = input.getDomElement();
-       // since qx.html.Node does not even create the
-       // domNode if it is not set to visible initially
-       // we have to quickly hide it after creation.
-       input.setVisible(false);
+        // since qx.html.Node does not even create the
+        // domNode if it is not set to visible initially
+        // we have to quickly hide it after creation.
+        input.setVisible(false);
         inputEl.addEventListener("change", e => {
           this.fireDataEvent("changeFileSelection", inputEl.files);
           inputEl.value = "";
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -55,8 +55,8 @@
  *   const file = fileList[0];
  *   reader.readAsText(file.slice(0,4));
  * });
-
-*
+ * ```
+ *
  * [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications)
  */
 
--- a/source/class/qx/ui/form/FileSelectorButton.js
+++ b/source/class/qx/ui/form/FileSelectorButton.js
@@ -60,9 +60,19 @@
  * [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications)
  */
 
-let InputElementIdCounter = 0;
 qx.Class.define("qx.ui.form.FileSelectorButton", {
   extend: qx.ui.form.Button,
+
+  statics: {
+    _FileInputElementIdCounter: 0,
+    _FileInputElementIdPrefix: "qxFileSelector_",
+
+    _generateElementId: function(){
+      ++this._FileInputElementIdCounter;
+      return this._FileInputElementIdPrefix + this._FileInputElementIdCounter;
+    }
+  },
+
   events: {
     /**
      * The event is fired when the file selection changes.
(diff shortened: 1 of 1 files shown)


source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -60,9 +60,19 @@
  * [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications)
  */
 
-let InputElementIdCounter = 0;
 qx.Class.define("qx.ui.form.FileSelectorButton", {
   extend: qx.ui.form.Button,
+
+  statics: {
+    _FileInputElementIdCounter: 0,
+    _FileInputElementIdPrefix: "qxFileSelector_",
```
derrell, line 68: Good catch, that it was referencing a global variable. This is a good change.

By convention, qooxdoo uses an uppercase first letter to indicate that the variable contains a class reference, e.g., one might do, `Object : qx.core.Object`... but variables not containing a class reference have a lower-case first letter. Would you please change these to `_fileInput...` instead of `_FileInput...`

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -131,12 +141,13 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
     },
 
     _createContentElement: function () {
-      let id = "qxFileSelector_" + InputElementIdCounter++;
+      let id = qx.ui.form.FileSelectorButton._generateElementId();
       let input = (this.__inputObject = new qx.html.Input(
         "file",
-        { display: "none" },
+        null,
```
derrell, line 147: Why do you prefer to have the element hidden after initially created, rather than initially?

derrell (changes requested):

source/class/qx/ui/form/FileSelectorButton.js
derrell: This is actually a behavior change. In the original case, it would consume no space, equivalent to a call to `this.exclude()`. `this.hide()` leaves the layout as if the widget were shown, but doesn't show the widget.

evandropoa: It's not a preference, actually it doesnt work on my Chrome 102.0.5005.61 64bits linux using display none as style parameter on instantiation, still showing and I just dint deep dive on it.

I'll trye using exclude and not hide.

evandropoa: Yes, using exlude works too, I just looking qoodoo for the first time, so I dont know all its nuances.

evandropoa: Actually I found some problems with this, if exluded or hidden before first appear event (creation I think) it just doesnt create the dom element, and if using exclude during appear event the file dialog doesnt appear

derrell: I'll await @oetiker's comments, as he implemented this recently. I think, though, that there is likely some other issue, as providing attributes "should" work. (I know, theory vs. reality, but still, it is supposed to work...)

BTW, I just noticed that `InputElementIdCounter` wasn't global. It was declared private to the file, with `let InputElementIdCounter = 0;`, which is legitimate. Your change makes it more "traditional qooxdoo"-like, but it wasn't polluting the global namespace.

@oetiker During your work on this, did you encounter any problems in Chrome (specifically, I guess, Chrome 102.0.5005.61)?

evandropoa:
Yes there was some problems, it still showing the input field like in this
using master branch.

[image: image.png]


*Evandro Bitencourt - Gerente*
*Sistema GP-Web LTDA - ME.*
CNPJ: 14659881/0001-61
*Av. Antônio Giudice, 124 - Casa - Alto Petrópolis*
*Porto Alegre-RS - CEP 91260-040*
(51) 9852-48484



Em dom., 7 de ago. de 2022 às 14:54, Derrell Lipman <
***@***.***> escreveu:

source/class/qx/ui/form/FileSelectorButton.js
evandropoa: Yes there was some problems, it still showing the input field like in this print taken using master branch.

image

evandropoa: And you are right, using let in file scope, now its up to you what implementation to use :)

oetiker: @derrell there could well be issues with chrome 102 ... I have not tested this

oetiker: The whole 'dom element does not get created if 'display' is set to 'none', would indicate that there is some optimization on the qx side at work which we should get to the bottom of I guess. @evandropoa is this whole thing really 'chrome dependent' ?

evandropoa: It's not a Chrome only thing, the same happens on my FireFox 95.0.2 (64-bits) the imput element is rendered over the button when using the master branch.

About doesnt being created when **display** is set to **none** could be an optimization 4 sure, I just didnt deep dive on element creation yet.

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -60,9 +60,19 @@
  * [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications)
  */
 
-let InputElementIdCounter = 0;
 qx.Class.define("qx.ui.form.FileSelectorButton", {
```
oetiker, line 63:
```suggestion
let InputElementIdCounter = 0;
qx.Class.define("qx.ui.form.FileSelectorButton", {
```

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -60,9 +60,19 @@
...
+  statics: {
+    _fileInputElementIdCounter: 0,
+    _fileInputElementIdPrefix: "qxFileSelector_",
+
+    _generateElementId: function(){
+      ++this._fileInputElementIdCounter;
+      return this._fileInputElementIdPrefix + this._fileInputElementIdCounter;
+    }
+  },
+
```
oetiker, line 75:
```suggestion
```
this seems overyly complex for what we try to achieve here

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -131,12 +141,13 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
     },
 
     _createContentElement: function () {
-      let id = "qxFileSelector_" + InputElementIdCounter++;
+      let id = qx.ui.form.FileSelectorButton._generateElementId();
```
oetiker, line 144:
```suggestion
      let id = "qxFileSelector_" + InputElementIdCounter++;
```

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -146,6 +157,7 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
       );
       input.addListenerOnce("appear", e => {
         let inputEl = input.getDomElement();
+        input.setStyle('display', 'none');
```
oetiker, line 160:
```suggestion
       // since qx.html.Node does not even create the
       // domNode if it is not set to visible initially
       // we have to quickly hide it after creation.
       input.setVisible(false);
```

oetiker (changes requested):
No idea why this was working for me ... in any event, the proper fix is to the the 'visible' property of the Node to false after the dom node has been created.

source/class/qx/ui/form/FileSelectorButton.js
evandropoa: Correct me if I'm wrong, but if we transpile using babel preset to ES5 (IE11) it will became a globar (var InputElementIdCounter=0)

source/class/qx/ui/form/FileSelectorButton.js
derrell: @oetiker Is there a `visible` property with boolean value? Do you mean `input.setVisiibility("excluded")`? (Equivalent to `input.exclude()`).

From Widget.js:

```
    /**
     * Make this widget visible.
     *
     */
    show() {
      this.setVisibility("visible");
    },

    /**
     * Hide this widget.
     *
     */
    hide() {
      this.setVisibility("hidden");
    },

    /**
     * Hide this widget and exclude it from the underlying layout.
     *
     */
    exclude() {
      this.setVisibility("excluded");
    },
```

oetiker:
@derrell the html.Node has a visible property ... it is NOT a widget ...

source/class/qx/ui/form/FileSelectorButton.js
```diff
@@ -157,7 +157,10 @@ qx.Class.define("qx.ui.form.FileSelectorButton", {
       );
       input.addListenerOnce("appear", e => {
         let inputEl = input.getDomElement();
-        input.setStyle('display', 'none');
+       // since qx.html.Node does not even create the
+       // domNode if it is not set to visible initially
+       // we have to quickly hide it after creation.
+       input.setVisible(false);
```
oetiker, line 163: fix indentation

```suggestion
        // since qx.html.Node does not even create the
        // domNode if it is not set to visible initially
        // we have to quickly hide it after creation.
        input.setVisible(false);
```

oetiker (changes requested):

source/class/qx/ui/form/FileSelectorButton.js
evandropoa: I agree, could be just a static variable of course, but I really dont like to relly on the let in field scope, becouse of transpilling to ES5 in global var

derrell (approved):
We've discussed it to death. All good by me. Defer to @oetiker who worked on this.

level420 (approved):

source/class/qx/ui/form/FileSelectorButton.js
oetiker: @evandropoa I am pretty sure it does not, since the individual files get wrapped ... but please test if you are unsure.

source/class/qx/ui/form/FileSelectorButton.js
oetiker: yes this has nothing todo with chrome ...

oetiker (approved):
ok, lets use the static ... :)

johnspackman (approved):

cboulanger (approved):

source/class/qx/ui/form/FileSelectorButton.js
evandropoa: Hi, you're right, the qx compiler will wrap a file declaration inside an IIFE function, I didn't know that, but after looking at the compiler source, I can see it.
Just to note that only Babel transpilling will produce a global variable, but the qx compiler makes it safer, WELL DONE.

Issue #10411, opened 2022-06-08
qx.ui.core.Blocker.blockContent() misplaces the blocker element.


lucapivato:
The ctor in qx.ui.core.Blocker attaches to the "move" event of the target widget to move the blocker div along with the blocked widget. But the call to blocker.blockContent() calls _block passing true to the blockContent argument which creates the blocked div as child of the widget to block, causing the blocker to be displaced by the same x,y position of the target widget.

https://tinyurl.com/26nop7av

![image](https://user-images.githubusercontent.com/37640802/172652422-0f940cfd-ace7-4ca1-a173-bc1e453d9309.png)

```javascript
// Create a button
var button1 = new qx.ui.form.Button("First Button", "icon/22/apps/internet-web-browser.png");

// Document is the application root
var doc = this.getRoot();

// Add button to document at fixed coordinates
doc.add(button1,
{
  left : 10,
  top  : 10
});

// Add an event listener
button1.addListener("execute", function(e) {
  alert("Hello World!");
});
var blocker = new qx.ui.core.Blocker(button1);
blocker.getBlockerElement(button1).setStyle("background-color", "red");
blocker.blockContent(10);

```

I can do a PR but wanted to make sure I'm not misreading this.

PR #10434, opened 2022-08-16
Wrong blocker positioning

evandropoa:
Blocker's are added as child of a root or a layout content element and have it's position set to absolute, so just needs to be positioned to the top/left of its parent with 100% of its width/height to cover it (#10411) and dont need move/resize

https://tinyurl.com/mpfk2yvk

### DIFF
--- a/source/class/qx/html/Blocker.js
+++ b/source/class/qx/html/Blocker.js
@@ -38,6 +38,10 @@ qx.Class.define("qx.html.Blocker", {
     var styles = {
       position: "absolute",
       opacity: opacity || 0,
+      width: '100%',
+      height: '100%',
+      top: '0px',
+      left: '0px',
       backgroundColor: backgroundColor
     };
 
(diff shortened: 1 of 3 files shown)


lucapivato:
Would it be possible to change the literals to double quotes?

oetiker:
Looking at the code which adds the blocker: 

https://github.com/qooxdoo/qooxdoo/blob/326fba0ccfcc5c30e58ddbca0504dab4e8ab7cf1/source/class/qx/ui/core/Blocker.js#L151

it seems that it gets added into the same PARENT as the widget it is blocking ... so it seems odd that the parent should have exactly the same size as the widget it is blocking ... at least not in a general sense ... 

in the sample code the content element may be nested and that is why it works ?

evandropoa:
> 

I see what you are talking about, I just dont follow why it should be added to the parent if I want to block a specific widget.

evandropoa:
> > 
> 
> I see what you are talking about, I just dont follow why it should be added to the parent if I want to block a specific widget. Absolute position its absolute in relation to its container, in the button example, the content element its the button content, if you move the blocker (already in container relation) it will be offset what is wrong.

Maybe what's wrong is that it's not getting the parent container element, it should get container element window body or document root in the examples?
```
 this._widget
          .getLayoutParent()
          .getContentElement()
```

PR #10435, opened 2022-08-17
automatic es6ify

johnspackman:
This contains the result of `qx es6ify` and also adds a commit hook so that git will automatically run `qx es6ify` during a commit.

The git hook is stored in `./github/hooks/` so that it is included in version control, but as it is a non-standard location it is necessary to tell git about it - that will happen with `npm i`

This does not include a windows version of the commit hook unfortunately.

It is possible to add this hook (or a similar one) to github, but as a first pass this is probably easier to try out

### DIFF
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -1,0 +1,20 @@
+#!/bin/sh -vx
+
+#
+# This pre-commit hook can work when you use `git add` if you run this command:
+#
+#   git config core.hooksPath .github/hooks
+#
+#
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | grep '^source/class' | grep '.js$' | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
+# Prettify all selected files
+echo $FILES
+echo "$FILES" | xargs qx es6ify
+
+# Add back the modified/prettified files to staging
+echo "$FILES" | xargs git add
+
+exit 0
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/sh -vx
+#!/bin/sh
 
 #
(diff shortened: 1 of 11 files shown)


.github/hooks/pre-commit
```diff
@@ -0,0 +1,20 @@
...
+#   git config core.hooksPath .github/hooks
+#
+#
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR | grep '^source/class' | grep '.js$' | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
+# Prettify all selected files
+echo $FILES
+echo "$FILES" | xargs qx es6ify
```
oetiker, line 15: this assumes for qx to be installed globally ...

hkollmann: What's about using a node script instead of the shell scrIpt? This should work for windows too.

johnspackman: > this assumes for qx to be installed globally ...

this is for development of qooxdoo so we already assume a global compiler, and `npx qx` won't work because of our bootstrapping.

however, in your own projects, you would choose to whether to use `npx`, depending on how you added qooxdoo to your project

hkollmann (approved):

source/class/qx/tool/cli/commands/Es6ify.js
```diff
@@ -58,6 +62,7 @@ qx.Class.define("qx.tool.cli.commands.Es6ify", {
 
   members: {
     async process() {
+      debugger;
```
derrell, line 65: This shouldn't be here, right?

derrell (changes requested):

derrell (approved):

level420 (approved):

cboulanger (approved):

oetiker (approved):

PR #10437, opened 2022-08-17
changes the semantics of `invalidMessage` to allow for null

johnspackman:


source/class/qx/ui/form/validation/Manager.js
```diff
@@ -482,7 +482,11 @@ qx.Class.define("qx.ui.form.validation.Manager", {
               continue;
             }
 
-            tooltip.setLabel(item.getInvalidMessage());
+            let msg = item.getInvalidMessage();
+            if (qx.core.Environment.get("qx.debug")) {
+              this.assertTrue(msg != null && msg.length > 0);
```
derrell, line 487: Why did you make it nullable but then disallow null?

Maybe you mean `this.assertTrue(msg === null || msg.length > 0)` ?

But in that case, why do you care if someone uses an empty string? (In fact, that may be a big BC break as it's likely, since the init value used to be an empty string, that many apps currently use an empty string...

source/class/qx/ui/form/validation/Manager.js
```diff
@@ -531,12 +535,20 @@ qx.Class.define("qx.ui.form.validation.Manager", {
       for (var i = 0; i < this.__formItems.length; i++) {
         var formItem = this.__formItems[i].item;
         if (!formItem.getValid()) {
-          messages.push(formItem.getInvalidMessage());
+          let msg = formItem.getInvalidMessage();
+          if (qx.core.Environment.get("qx.debug")) {
+            this.assertTrue(msg !== null && msg.length > 0);
```
derrell, line 540: ditto

source/class/qx/ui/form/validation/Manager.js
```diff
@@ -531,12 +535,20 @@ qx.Class.define("qx.ui.form.validation.Manager", {
...
+          messages.push(msg);
         }
       }
       // add the forms fail message
-      if (this.getInvalidMessage() != "") {
-        messages.push(this.getInvalidMessage());
+      if (!this.isValid()) {
+        let msg = this.getInvalidMessage();
+        if (qx.core.Environment.get("qx.debug")) {
+          this.assertTrue(msg !== null && msg.length > 0);
```
derrell, line 549: ditto

derrell (changes requested):

source/class/qx/ui/form/validation/Manager.js
johnspackman: because the field is invalid, this is requiring that there is an explanation of why the field is failing its validation.  Without this check, forms will be unable to complete but unable to give the user any feedback

derrell: So with a `null` initial value, it looks like an invalid user entry will cause an error to be thrown. But an error being thrown is a developer-solvable problem, not a user-solvable problem. Maybe a better solution might be to revert to the original, and then change the default error message from empty string to "Invalid entry", and disallow setting the property to an empty string with `check : (v) => typeof v == "string" && v.length > 0` or similar?

johnspackman: > So with a null initial value, it looks like an invalid user entry will cause an error to be thrown

Yes; the idea is that this helps you find the field which does not have a valid description.

The reason that this change proposal came about was because I've been experimenting with a number of canned validation routines which can be attached at various points to the UI - for example, a field which is a `qx.ui.form.TextField` can have a data type of "email address" or "integer" etc

Where validation is handed off to generic library tools, it is typically an error if a field is flagged as invalid without a corresponding invalid message, on the basis that the task of a generic validation routine includes providing good descriptions.

In one-off validation, then assuming a default message eg "Invalid field" is a convenience for the developer although a bit ugly for the user, and I wanted to trap all those occasions where a decent message had not been provided.

How about if we change this to (a) allow `null` in invalid message, because the `""` is functionally equivalent to `null` in this case and means that it is not necessary to always map `null` to `""`, and (b) encapsulate any enforcement in an environment setting?  I've adjusted the proposal to do this

derrell (changes requested):
That looks like an appropriate change, I think. Don't forget, though, that there are two other locations which need the change to be applied (marked with my "ditto" comment) in my initial review.

johnspackman:
sorry, i thought i'd done those ... fixed now though

derrell (approved):

level420 (approved):

hkollmann (approved):

oetiker (approved):

cboulanger (approved):

source/class/qx/ui/form/validation/Manager.js
hkollmann: I agree with your argumentation.
The problem is the current implementation. Form becomes invalid if one of it's widgets are invalid. So you always get the invalidMessage from the form. This might not be what you expect. At least this is a breaking change - so we should move it to v8.

PR #10439, opened 2022-08-17
Adds the ability for multiple events with promises to be tracked and …

johnspackman:
…awaited on

### DIFF
--- a/source/class/qx/core/MEvent.js
+++ b/source/class/qx/core/MEvent.js
@@ -173,7 +173,7 @@ qx.Mixin.define("qx.core.MEvent", {
     __promiseWaitForPendingEvents: null,
 
     /**
-     * Internal helper method to track promises returned from evcent handlers
+     * Internal helper method to track promises returned from event handlers
      *
      * @param {var} result the result from the event handler
      * @returns {qx.Promise|var} the value to return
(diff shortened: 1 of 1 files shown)


derrell (changes requested):
Can you explain a bit of what this is doing? It's in a mixin that will likely cause it to have wide-ranging consequences so I'd like to fully understand it.

johnspackman:
events can be asynchronous, and can themselves trigger actions which are asynchronous - when they do so, they return a Promise.  If you're just firing one event (and there are no side effects) then it is easy to await firing, but lots of code does not do that.  This fix tracks any promises returned from an event being fired, and allows the caller to await for them to all be resolved.

derrell (approved):

source/class/qx/core/MEvent.js
```diff
@@ -166,6 +166,69 @@ qx.Mixin.define("qx.core.MEvent", {
...
     },
 
+    /** @type{Object<String,qx.Promise>} list of pending events, indexed by hash code */
+    __pendingEvents: null,
+
+    /** @type{qx.Promise} promise that callers are waiting on, ready for when all events are finished */
+    __promiseWaitForPendingEvents: null,
+
+    /**
+     * Internal helper method to track promises returned from evcent handlers
```
cboulanger, line 176: s/evcent/event/

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman !

PR #10441, opened 2022-08-17
improve error messages

johnspackman:


source/class/qx/core/MProperty.js
```diff
@@ -28,7 +28,7 @@ qx.Mixin.define("qx.core.MProperty", {
      * As a fallback, if no generated property setter could be found, a
      * handwritten setter will be searched and invoked if available.
      *
-     * @param data {Object | String} a map of property values. The key is the name of the property.
+     * @param data {Map | String} a map of property values. The key is the name of the property.
```
derrell, line 31: Now that there is an actual type called `Map`, I think we're poorly served by using the old meaning of `Map` (for `Object`). I agree that `Object` is insufficiently specific, and that it's an Object being used as a Map, but we should either find another name to indicate an Object used as a Map (`Dictionary`?) or continue to use `Object` for these cases.

derrell (changes requested):

derrell (approved):

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman !

PR #10444, opened 2022-08-17
minor fixes for simple theme and virtual widgets

johnspackman:


hkollmann (approved):

derrell (commented):
I'm wondering if these are going to be non-BC changes to users apps? It's changing the default appearance, and apps may not want those changes...???

derrell:
If documented, and released in v8, that'd be fine. For a v7.x point release, I'm not so sure.

johnspackman:
the appearances are missing, so the virtual components just have an absence of styling when using the simple theme- this just gives them something

cboulanger (approved):

oetiker (approved):

hkollmann (approved):

derrell (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman !

PR #10446, opened 2022-08-17
adding support for proxy classes

johnspackman:


derrell (changes requested):
The changes look fine, but the PR title says "proxy classes" but there's nothing in this PR about proxy classes. What is a proxy class, and what have these changes to do with it?

johnspackman:
I've just realised that the docs for this were included some time ago, but the functionality wasn't!  It's documented here: https://github.com/qooxdoo/qooxdoo/blob/master/docs/development/compiler/configuration/compile.md

Basically, `proxySourcePath` allows an alternative source path to be searched when compiling; this allows "proxy" classes to be generated, which have a special meaning on a given target.  For example, if you have a class that exists on the server, you might generate a "proxy" version of that for use on the client, and that "proxy" version would use async I/O to connect to the server.

cboulanger (approved):

oetiker (approved):

derrell (approved):

level420 (approved):

hkollmann (approved):

hkollmann:
Thanks @johnspackman !

PR #10448, opened 2022-08-17
allows tabview pages to be scrollable (adds ScrollablePage)

johnspackman:


source/class/qx/ui/tabview/Page.js
```diff
@@ -24,7 +24,11 @@
  * @childControl button {qx.ui.tabview.TabButton} tab button connected to the page
  */
 qx.Class.define("qx.ui.tabview.Page", {
-  extend: qx.ui.container.Composite,
+  extend: qx.ui.core.Widget,
+  include: [
+    qx.ui.core.MRemoteChildrenHandling,
+    qx.ui.core.MRemoteLayoutHandling
+  ],
```
derrell, line 31: I'm concerned about BC with this change. Are you sure that users' apps can't be doing anything with a `Composite` that isn't included into `Widget` with these two mixins?

derrell (changes requested):

cboulanger (approved):

source/class/qx/ui/tabview/Page.js
johnspackman: It's because the [Composite.defer](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/ui/container/Composite.js#L123-L126) calls `MChildrenHandling.remap` and `MLayoutHandling.remap` which overwrite the members of the class, which breaks the normal inheritance of methods so we can't inherit from Composite.  

That's pretty damn ugly IMHO and took some debugging to figure out WTF was going on - I presume that it was there to try and tweak a tiny bit more performance out of IE6.  I think it would be better to just prefer to deprecate the `.remap` calls and not call them from Composite at all, and then Page can inherit from Composite as you'd expect

derrell (approved):

oetiker (approved):

hkollmann (approved):

level420 (approved):

PR #10451, opened 2022-08-24
fixes a bug where the context is missing from Slider event handlers

johnspackman:


cboulanger (approved):

hkollmann (approved):

oetiker (approved):

derrell (changes requested):
There is a call to `removeListener` that also needs to be updated with the context, or I believe the listener won't be removed.

johnspackman:
done

derrell (approved):

level420 (approved):

hkollmann:
Thanks. @johnspackman !

PR #10457, opened 2022-09-05
fix a couple of bugs in compiler

johnspackman:
fix bug where classes which are no longer dependencies still have their `require` classes included in browserify output;
fix for better checking mtime when determining whether to recompile classes

### DIFF
--- a/source/class/qx/tool/compiler/targets/meta/Browserify.js
+++ b/source/class/qx/tool/compiler/targets/meta/Browserify.js
@@ -56,9 +56,7 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
         "browser"
       ) {
         // Get a Set of unique `require`d CommonJS module names from classes needed by the application
-        let classnames = this.__appMeta__P_44_0
-          .getApplication()
-          .getDependencies();
+        let classnames = this.__appMeta.getApplication().getDependencies();
         for (let className of classnames) {
           let classInfo = db.classInfo[className];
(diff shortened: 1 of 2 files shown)


source/class/qx/tool/compiler/targets/meta/Browserify.js
```diff
@@ -55,9 +55,11 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
         this.__appMeta.getEnvironmentValue("qx.compiler.applicationType") ==
         "browser"
       ) {
-        // Get a Set of unique `require`d CommonJS module names from
-        // all classes
-        for (let className in db.classInfo) {
+        // Get a Set of unique `require`d CommonJS module names from classes needed by the application
+        let classnames = this.__appMeta__P_44_0
```
hkollmann, line 59: Is this what you mean? Will change with each compile.

hkollmann (changes requested):

source/class/qx/tool/compiler/targets/meta/Browserify.js
johnspackman: doh!  copied and pasted from the debugger ... 😊

hkollmann (approved):

oetiker (approved):

derrell (approved):

cboulanger (approved):

level420 (approved):

hkollmann:
Thanks @johnspackman !

PR #10462, opened 2022-09-22
DateChooser with min and max values

goldim:
The standard `input="date"` html element does have _min_ and _max_ attributes which restrict user from setting "invalid" dates. DateChooser and DateField (which contains DateChooser as a child) don't support this functionality. I decided add it via properties _MinValue_ and _MaxValue_. If MinValue is set then dates less _MinValue_ are not pickable and the same about _MaxValue_. If code I added not efficient let me know. I will fix it.


derrell (changes requested):
This looks good, code-wise. Please add doc strings to the two new properties so that they show up in the API viewer.

cboulanger (approved):
Cool, thanks!

johnspackman (approved):

oetiker (approved):

goldim:
@derrell Hello, thanks for response, I did. About props I don't sure is it worth to add event? how usefull will it be? For my goals it was needed. I think if somebody wants to have them it's not difficult to add them anytime.

derrell (approved):

level420 (approved):

hkollmann (changes requested):
immediately after the `=>` (arrow-body-style) at source/class/qx/test/core/Property.js:1066:43:
81
  1064 | 
82
  1065 |     testPromises() {
83
> 1066 |       const promiseDelay = (delay, fn) => {
84
       |                                           ^
85
  1067 |         return new qx.Promise(resolve => {
86
  1068 |           setTimeout(async () => {
87
  1069 |             await fn();
88



Could you fix the lint error first and have a look at the not running datechooser tests

Thanks,
Hennrt

hkollmann:
Could you run `npm test` in your local copy and check if this test failures are github speficic?

goldim:
@hkollmann okay, will do it, thx.

goldim:
@hkollmann I checked local tests and errors the same. The errors with DataChooser were before my changes. I can fix some tests but have I to create separate PR or commit to this one?

hkollmann:
Commit to this one

goldim:
@hkollmann I decresed amount of failed tests. All failed tests are not related with my changes of this PR. About some failed tests I created  issues and fixed lint error.

hkollmann (approved):

PR #10467, opened 2022-09-29
Up/Down keys are inverted.

lucapivato:
When the slider is vertical Up should increase the value and Down should decrease it. Same when horizontal or vertical, PgUp should increase and PgDown decrease the value. Currently it's inverted.


derrell (changes requested):
I agree with this change, but it is not backward compatible, so should not go into any point release; only into version 8.0. We should create an 8.0 branch to merge this into, instead of into master.

hkollmann:
from core-team dicussion: As a general rule, PageDown means "move towards the end"; PageUp means "move towards the beginning". That's fairly universal, I think...
So the behavior I'm seeing in the widgetbrowser is consistent with that.

PR #10473, opened 2022-10-28
adding a git hook should only be done in the qooxdoo project.

hkollmann:
installation fails if there is no hook dir

### DIFF
--- a/addGitHook
+++ b/addGitHook
@@ -1,7 +1,6 @@
 #!/usr/bin/env node
 const utils = require("./bin/tools/utils");
 const fs = require("fs");
-const { exit } = require("process");
 (async function() {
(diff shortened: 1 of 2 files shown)


cboulanger (approved):

oetiker (approved):

johnspackman (approved):

addGitHook
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+const utils = require("./bin/tools/utils");
+const fs = require("fs");
+const { exit } = require("process");
```
derrell, line 4: `exit` isn't used

derrell (approved):
You've marked this as a hotfix so I won't hold it up, but see my one comment about `exit` being unused.

PR #10474, opened 2022-10-28
Add arguments class and method to Test command

### LINKED ISSUE #10465: Unknown arguments class and method for qx test
**Describe the bug**
The command `qx test` with arguments class and method doesn't work and produces an error with a message"Unknown arguments".
But there is in documentations said:

> To see the test results as output of a command line script (in the syntax of the TAP protocol), execute npx qx test. This will run the server similarly to npx qx serve, and run the tests in a headless browser. To run a single test, you can use qx test --class myapp.test.MyTest --method testOne

**Expected behavior**
Single test has to be run

 - OS: Windows and Linux
 - Version qx 7.2.1

goldim:
Fixes #10465


cboulanger (approved):

hkollmann (approved):

source/class/qx/tool/cli/commands/Test.js
```diff
@@ -56,6 +56,20 @@ qx.Class.define("qx.tool.cli.commands.Test", {
         describe: "Disables the start of the webserver",
         default: false,
         type: "boolean"
+      },
+
+      class: {
```
johnspackman, line 61: It would be better to use a different word eg testClass because in upcoming release we will support native properties and as class is a reserved word, that will break

johnspackman (changes requested):

source/class/qx/tool/cli/commands/Test.js
goldim: @johnspackman Does using the argument class in double quotes solve the problem? and the another nuance in docs there is next usage of the command test:

> To see the test results as output of a command line script (in the syntax of the [TAP protocol](https://node-tap.org/tap-protocol/)), execute npx qx test. This will run the server similarly to npx qx serve, and run the tests in a headless browser. To run a single test, you can use qx test --class myapp.test.MyTest --method testOne

If I change names of the arguments there should be changes applied to docs too.

cboulanger: Quoting it should be enough, no?

johnspackman:
No, sorry that won’t be enough.The issue is that with native properties, you will be able to use ‘myObj.propertyName’ instead of ‘myObj.getPropertyName()’This means that property names cannot be reserved wordsSent from my iPhoneOn 29 Oct 2022, at 13:09, Dmitry Zolotov ***@***.***> wrote:﻿
@goldim commented on this pull request.



In source/class/qx/tool/cli/commands/Test.js:
> @@ -56,6 +56,20 @@ qx.Class.define("qx.tool.cli.commands.Test", {
         describe: "Disables the start of the webserver",
         default: false,
         type: "boolean"
+      },
+
+      class: {

@johnspackman Does using the argument class in double quotes solve the problem? and the another nuance in docs there is next usage of the command test:

To see the test results as output of a command line script (in the syntax of the TAP protocol), execute npx qx test. This will run the server similarly to npx qx serve, and run the tests in a headless browser. To run a single test, you can use qx test --class myapp.test.MyTest --method testOne

If I change names of the arguments there should be changes applied to docs too.

goldim:
@johnspackman ok, then what's better className (--class-name) and methodName (--method-name) or as you mentioned before testClass (--test-class) and testMethod(--test-method)?

johnspackman:
Personally id go for className as it's a string, but it's your PR, your choice :)

source/class/qx/tool/cli/commands/Test.js
johnspackman: That is true ... maybe I'm being pedantic:) I like not having to quote things though, it seems like an unnecessary collision

goldim:
@johnspackman hurried request review. Need to check something. Probably just renaming won't work.

goldim:
Yeah, just renaming don't work. Sorry, have to find another way to transform args to yargs command.

hkollmann:
Changing the name means changing the name in testtapper too.  The argument is just passed through.
And changing this is not backwards compatible - so it shouldbe done with version 8.

We should but it on the list for v8.

hkollmann:
Is it working with your first version?

hkollmann (changes requested):
This version is a breaking change!

goldim:
@hkollmann After you approved changes I made one commit where I renamed arguments due to @johnspackman asked . Yes the first version worked.

hkollmann:
And the rename did not work because of tettapper etc. 
IMHO we should use class and method now and change this in the test packages and here in version 8.

@johnspackman : Could you agree?

goldim:
I've reverted last change

hkollmann (approved):

hkollmann:
This fix i not necessary. class and method parameter are provided by the testtapper package.

hkollmann (changes requested):
not needed - handled by testtapper package

goldim:
All works fine without PR changes. Tested on Windows and Linux. _qooxdoo_ version since **7.2.1** and _qxl.testtapper_ **2.0.0**.

PR #10475, opened 2022-10-30
fix unit tests

hkollmann:
for details see commit messages

### DIFF
--- a/source/class/qx/ui/form/validation/Manager.js
+++ b/source/class/qx/ui/form/validation/Manager.js
@@ -483,8 +483,13 @@ qx.Class.define("qx.ui.form.validation.Manager", {
             }
 
             let msg = item.getInvalidMessage();
-            if (!msg && qx.core.Environment.get("qx.ui.form.validation.Manager.allowDefaultInvalidMessage")) {
-              msg = "Invalid field";
+            if (
+              msg &&
+              qx.core.Environment.get(
+                "qx.ui.form.validation.Manager.allowDefaultInvalidMessage"
+              )
+            ) {
+              msg = qx.locale.Manager.tr("Invalid field");
             } else if (qx.core.Environment.get("qx.debug")) {
               this.assertTrue(msg != null && msg.length > 0);
             }
@@ -538,8 +543,13 @@ qx.Class.define("qx.ui.form.validation.Manager", {
         var formItem = this.__formItems[i].item;
         if (!formItem.getValid()) {
           let msg = formItem.getInvalidMessage();
-          if (!msg && qx.core.Environment.get("qx.ui.form.validation.Manager.allowDefaultInvalidMessage")) {
-            msg = "Invalid field";
+          if (
+            !msg &&
+            qx.core.Environment.get(
+              "qx.ui.form.validation.Manager.allowDefaultInvalidMessage"
+            )
+          ) {
+            msg = qx.locale.Manager.tr("Invalid field");
           } else if (qx.core.Environment.get("qx.debug")) {
             this.assertTrue(msg !== null && msg.length > 0);
           }
@@ -550,12 +560,9 @@ qx.Class.define("qx.ui.form.validation.Manager", {
       // add the forms fail message
       if (!this.isValid()) {
         let msg = this.getInvalidMessage();
-        if (!msg && qx.core.Environment.get("qx.ui.form.validation.Manager.allowDefaultInvalidMessage")) {
-          msg = "Invalid field";
-        } else if (qx.core.Environment.get("qx.debug")) {
-          this.assertTrue(msg !== null && msg.length > 0);
+        if (msg != "") {
+          messages.push(msg);
         }
-        messages.push(msg);
       }
 
       return messages;
@@ -665,7 +672,7 @@ qx.Class.define("qx.ui.form.validation.Manager", {
   },
 
   environment: {
-    // Whether to assume a default "Invalid Field" message for invalid fields; if false, an 
+    // Whether to assume a default "Invalid Field" message for invalid fields; if false, an
     //  exception will be raised if invalid fields do not have an `invalidMessage`
     "qx.ui.form.validation.Manager.allowDefaultInvalidMessage": true
   }
--- a/source/class/qx/ui/form/validation/Manager.js
+++ b/source/class/qx/ui/form/validation/Manager.js
@@ -71,7 +71,7 @@ qx.Class.define("qx.ui.form.validation.Manager", {
     },
 
     /**
-     * The invalid message should store the message why the form validation
+     * The invalid message stores the message why the form validation
      * failed. It will be added to the array returned by
      * {@link #getInvalidMessages}.
      */
--- a/.github/workflows/test-framework.yml
+++ b/.github/workflows/test-framework.yml
@@ -26,7 +26,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows]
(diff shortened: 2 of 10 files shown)


oetiker (approved):

oetiker:
great!

derrell (approved):

level420 (approved):

cboulanger (approved):

source/class/qx/ui/form/validation/Manager.js
```diff
@@ -550,12 +560,9 @@ qx.Class.define("qx.ui.form.validation.Manager", {
       // add the forms fail message
       if (!this.isValid()) {
         let msg = this.getInvalidMessage();
-        if (!msg && qx.core.Environment.get("qx.ui.form.validation.Manager.allowDefaultInvalidMessage")) {
-          msg = "Invalid field";
-        } else if (qx.core.Environment.get("qx.debug")) {
-          this.assertTrue(msg !== null && msg.length > 0);
```
johnspackman, line 556: Why is this removed?  It should remain, although with the locale changes for `msg = qx.locale.Manager.tr("Invalid field");`

johnspackman (changes requested):

source/class/qx/ui/form/validation/Manager.js
hkollmann: You did not change the invalidmessage handling of the form. So the default is '' as before.
This check you always get an invalid field error even if there is no other error.

johnspackman: the previous code was to assume a default message if there is none but when the form is still invalid.  By removing this code, then invalid forms will not have a message even if they are invalid - unless they have widgets which are also invalid.  AIUI the code should go back in

johnspackman (changes requested):

source/class/qx/ui/form/validation/Manager.js
hkollmann: Form validation works differnt from formItem validation.
The invalidMessage is not the message you send to the user it's the set from exception message, see:

Manager.js, line 354ff

```
// reset the invalidMessage
      this.setInvalidMessage("");

      if (this.__isAsyncValidator(formValidator)) {
        this.__asyncResults[this.toHashCode()] = null;
        formValidator.validateForm(items, this, context);
        return null;
      }

      try {
        var formValid = formValidator.call(context, items, this);
        if (formValid === undefined) {
          formValid = true;
        }
      } catch (e) {
        if (e instanceof qx.core.ValidationError) {
          formValid = false;

          if (e.message && e.message != qx.type.BaseError.DEFAULTMESSAGE) {
            var invalidMessage = e.message;
          } else {
            var invalidMessage = e.getComment();
          }
          this.setInvalidMessage(invalidMessage);
        } else {
          throw e;
        }
      }
      return formValid;
    },
```

hkollmann: With this the test of empty invalidMessage of the form fails for ever,

hkollmann: Same name for different functionality - should we change this in v8?

hkollmann:
But the invalidateNessage is not the default message used when form validation fails. It holds the current invalidateMessage generated elsewhere - see above.
Here is the same property name used for different purpose.

IMHO we should change the name in v8 to get this clearer.

hkollmann:
The form is set invalid if one of the widgets is invalid.
```
      // check the form validator (be sure to invoke it even if the form
      // items are already false, so keep the order!)
      var formValid = this.__validateForm(items);
      if (qx.lang.Type.isBoolean(formValid)) {
        this.__syncValid = formValid && this.__syncValid;
      }
      valid = formValid && valid;

      this._setValid(valid);
```
With the check for empty invalidMessage it's not possible to have an empty invalidMessage for the form in the case when only  the items are invalid.
Changing this is a breaking change. The required invalidMessage too.

I suggest for v8:
  - change the meaning of the forms invalidMessage to the same as invalidMessage in the widget.
  - add check for empty invalidMessage
  - set form invalid only if the formValidator says that the form is invalid.
  
After approving of this PR i can make a PR with this changes and mark it for v8.

What do you think?

derrell:
@hkollmann I'm not following quite what the problem is at present, and what the difference will be with your change. When you create the subsequent issue or PR, maybe you can better explain the problem and the proposed fix? It's hard to follow that code since there are multiple variables with similar by different names.

derrell:
I believe this is a breaking change. Any app that previously had an empty message will now show a (default) message. Whether or not we believe it _should_ show a message, we don't know whether the app has arranged to do something of its own, expecting no invalid message to show up, that this change will compete with. It sounds to me like this change belongs in 8.0, not in 7.3.

derrell:
And if someone has implemented something of their own, is there still a way to have no invalid message, e.g., bu setting it to null instead of empty string? There probably should be.

johnspackman:
The motivation for the change in 7.3 was because the `isValid` of a form or widget could be `false` but without there being a message to display; it's rather odd and unexpected behaviour if a widget is invalid but incapable of reporting that it is invalid to the user.  The other rather odd thing is that the message was set to an empty string, and not null, which seems quite inconsistent with how we would normally write code.

Just as widgets need to explain why they are invalid if they are in fact, invalid, the form also has the same requirement.  A form can be invalid even if none of its widgets are invalid (which is reasonable, because you might not allow certain combinations of widget values).  Again, if the form is invalid, it must provide a message otherwise the form cannot be submitted and there is no feedback to the user.

https://github.com/qooxdoo/qooxdoo/pull/10437#discussion_r953349252

hkollmann:
I agree with your argumentation.
The problem is the current implementation. Form becomes invalid if one of it's widgets are invalid. So you always get the invalidMessage from the form. This might not be what you expect. At least this is a breaking change - so we should move it to v8.

hkollmann:
Could we find some compromise here? 
e.g. merge thid and change it back for v8? I would create the PR for this.

johnspackman:
Why not just roll back the PR that caused this in the first place?

hkollmann:
Do you mean this one? Then we will get back the failing tests.
Or do you mean https://github.com/qooxdoo/qooxdoo/commit/407842fdda7af3a4060cd962b9d9b5c0d76f4f75 ? The part for the mixin is great.

johnspackman:
Let's just go with your change to the Manager and get the tests working, although I was thinking to revert https://github.com/qooxdoo/qooxdoo/commit/407842fdda7af3a4060cd962b9d9b5c0d76f4f75 when I wrote that.

From my point of view, I've already changed this once in [the last PR](https://github.com/qooxdoo/qooxdoo/commit/407842fdda7af3a4060cd962b9d9b5c0d76f4f75) in order to avoid potential BC changes by adding a default message, which I wasn't too happy about because I saw the PR as fixing a bug, but did it to compromise.  

OTOH the PR was not perfect because how those errors are reported should be handled carefully, eg there is no point popping a message on a widget saying "I'm invalid" and also on the form saying "I'm invalid" because it would look like there are two errors.

The form validation is broken IMHO, so maybe the best way would be to put together a separate use case for a general revamp as a single separate PR and do a really good upgrade.

We still don't have a mechanism for collecting a set of changes together into an ongoing pre-release (eg 8.0-beta)

hkollmann:
I flagged the v8 PR with "do not merge - v8" and set them to draft. not ideal. We can think about a v8-prerelease branch.
Can i get your appreoval then?

Thanks!

johnspackman (approved):

Issue #10476, opened 2022-11-01
Property targetType is not expected to be here


goldim:
**Describe the bug**
The command `npx qx package install --save=0 qooxdoo/qxl.testtapper` doesn't work and fails with a message: 
"**Error: Error validating data for compile.json: 1)  Property targetType is not expected to be here** . 

**To Reproduce**
1. `npx qx create "project name" --type package --out`
2. `npx qx package install --save=0 qooxdoo/qxl.testtapper`
3. See error

**Expected behavior**
The package _qooxdoo/qxl.testtapper_ is installed.

**Desktop (please complete the following information):**
 - OS: Linux (Debian), Windows 10
 - qooxdoo v7.3.3

PR #10477, opened 2022-11-02
move config.targetType to Compile property.

hkollmann:
using config.targetType do change the config object so that later schema checks will fail

fix for https://github.com/qooxdoo/qooxdoo/issues/10476

### DIFF
--- a/source/class/qx/tool/config/Compile.js
+++ b/source/class/qx/tool/config/Compile.js
@@ -31,13 +31,5 @@ qx.Class.define("qx.tool.config.Compile", {
 
   construct() {
     super(this.self(arguments).config);
-  },
-  properties: {
-    /**
-     * calculated target type
-     */
-    targetType: {
-      check: "String"
-    }
   }
 });
--- a/source/class/qx/tool/config/Compile.js
+++ b/source/class/qx/tool/config/Compile.js
@@ -31,5 +31,13 @@ qx.Class.define("qx.tool.config.Compile", {
 
   construct() {
     super(this.self(arguments).config);
+  },
+  properties: {
+    /**
+     * calculated target type
+     */
+    targetType: {
+      check: "String"
+    }
   }
 });
--- a/compile.js
+++ b/compile.js
@@ -159,7 +159,7 @@ qx.Class.define("qx.compiler.CompilerApi", {
               path.join(
                 __dirname,
                 "bin",
-                qx.tool.config.Compile.getInstance().getTargetType(),
+                command.getTargetType(),
                 "qx"
               ),
               "lint",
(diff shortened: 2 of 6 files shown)


cboulanger (approved):

source/class/qx/tool/config/Compile.js
```diff
@@ -31,5 +31,13 @@ qx.Class.define("qx.tool.config.Compile", {
...
     super(this.self(arguments).config);
+  },
+  properties: {
+    /**
+     * calculated target type
+     */
+    targetType: {
+      check: "String"
+    }
   }
```
johnspackman, line 42: I don't understand the reason for this change - because the bug in https://github.com/qooxdoo/qooxdoo/issues/10476 seems to be related to a badly formed `compile.json`, and not to whether `targetType` is global or not

Also, why is it necessary to have `targetType` as a global state?  Although it is unlikely that we would compile source and build targets at the same time, it is not entirely unreasonable and AFAIK it is achievable (up until this PR).  

There are presumably lots of other examples where data could be held globally, but aren't because of side effects - what makes `targetType` special?

johnspackman (changes requested):

source/class/qx/tool/config/Compile.js
hkollmann: It's not a badly formed compile.json. The problem is the on the change of the config object. After this change you can not verify the json object again against the schema.
That's the reason for this change

hkollmann: TargetType is needed global as before. Before it was part of the global config object

derrell (approved):

source/class/qx/tool/config/Compile.js
johnspackman: > The problem is the on the change of the config object

Ah ISWYM - the code is misusing the config to add state information, and the state information breaks the validation.

> TargetType is needed global as before. Before it was part of the global config object

Wouldn't it be much simpler to just get it from the `command` instead of copying it out into an independent property, added to the class responsible for managing schema validation?  Especially as it's CLI state and has nothing to do with the schema itself.  

IE where the [original compile.js code](https://github.com/qooxdoo/qooxdoo/blob/master/compile.js#L162) used: `command._getConfig().targetType,`, the `command` is the instance of `qx.tool.cli.commands.Compile`, which already knows how to get the parsedArgs of the CLI, and it's the parsedArgs which determine the targetType.

hkollmann: targetType is calculated:
qx.tool.config.Compile.getInstance().setTargetType(parsedArgs.target || config.defaultTarget || "source");

i prefered to calculate it once.

johnspackman: >targetType is calculated:

It makes no sesne to do it this way - you're copying it out of CLI into a global (where it was not global before), storing it with the precompiled schema (which it has nothing to do with), and then pulling it back in from that global when you have access to the state that it should have been (and previously was) part of.

It should remain where it was, which is part of the state of the command.  parsedArgs should have it added as a property, and then the Compile class would pick that up and provide a `getTargetType()` that returns it - just like any other state information

johnspackman (approved):

level420 (approved):

oetiker (approved):

PR #10486, opened 2022-11-20
Speedup npm modules

hkollmann:
Browserify of the used npm modules is a time cosuming process. This PR checks if browserify is neccesary:
It needed when:
- commonjs-browserify.js do not exits
- The list of modules changed
- One of the localModules are newer compared to the last run


cboulanger (approved):
Makes sense, thanks.

level420 (approved):

oetiker (approved):

johnspackman (approved):

derrell (approved):

PR #10491, opened 2022-11-24
Scrolling Scroll container and list via dragging

### LINKED ISSUE #10489: Scroll doesn't work by dragging content inside qx.ui.container.Scroll
**Describe the bug**
I have `qx.ui.container.Scroll` widget and there is content inside it. I want scroll content inside by draggin space with mouse. And in most browsers and OS this feature doesn't work but interesting thing somewhere it DO work out of box and it is cool and awesome.  The scrolling via scroll button works everywhere.
Below I described where it works and doesn't.

**To Reproduce**
1. Go to Demo Browser
2. Type in a search field: ScrollContainer
3. Click it and wait for loading demo.
4. Try to drag content inside scroll widget.

**Expected behavior**
I am not sure but possible it has to be implemented and supported.

**Screenshots**

 - OS: Debian and Debian derivative distros
 - Browser Chrome
 - Version ^100
 - Result: works fine 

 - OS: Debian and Debian derivative distros
 - Browser Firefox
 - Version 102
 - Result: DOESN'T work

 - OS: Windows (^10), Ubuntu (20 version)
 - Browser Chrome, Firefox, Opera
 - Result: DOESNT work

**Additional context**
I try to figure out why it works only in Debian Chrome and appriciate any advice about what to check and where dig into. There is `qx.ui.core.scroll.AbstractScrollArea`  widget which has mixin **MDragDropScrolling**. I am not sure but maybe it is what I need to explore.

goldim:
Resolved #10489

### DIFF
--- a/source/class/qx/ui/core/scroll/MRoll.js
+++ b/source/class/qx/ui/core/scroll/MRoll.js
@@ -58,11 +58,6 @@ qx.Mixin.define("qx.ui.core.scroll.MRoll", {
      * @param e {qx.event.type.Roll} Roll event
      */
     _onRoll(e) {
-      // only wheel and touch
-      if (e.getPointerType() == "mouse") {
-        return;
-      }
-
       if (this._cancelRoll && e.getMomentum()) {
         e.stopMomentum();
         this._cancelRoll = null;


johnspackman (approved):

hkollmann (approved):

level420 (approved):

oetiker (approved):

cboulanger (approved):

source/class/qx/ui/core/scroll/MRoll.js
```diff
@@ -58,11 +58,6 @@ qx.Mixin.define("qx.ui.core.scroll.MRoll", {
      * @param e {qx.event.type.Roll} Roll event
      */
     _onRoll(e) {
-      // only wheel and touch
-      if (e.getPointerType() == "mouse") {
```
derrell, line 62: Do we understand the intent of this original code? Is it really safe to just delete it? If we're not sure of the original intent, I'd be more comfortable adding an additional pointer type that's allowed, rather than allowing anything at all.  I think you'd said, in some conversation, that you were seeing a pointer type of "pen"? I'm not familiar with that pointer type, but if so, and it fixes the problem, then how about something like this?

```javascript
if ( [ "mouse", "pen" ].includes(e.getPointerType())) {
```

derrell (changes requested):

source/class/qx/ui/core/scroll/MRoll.js
goldim: I would like to know why the original code allows all type events except mouse. Judging by comment which is there right now code should work only for wheel and touch but nothing about pen (stylus). Probably pen and touch events are the same (pen like I think its just very small area of touch). As I said in the issue pen event generated in debian chrome (and chromium too) and I could drag inner content of container.Scroll. This mixin influences only several widgets: container.Scroll and List. I checked behaviour of list in other gui widget frameworks (Windows Forms .NET) and there you cant scroll content of a list by dragging. Maybe it is only reason. But for me it is nothing bad to scroll list by dragging (any disadvantages?).

goldim: @derrell your code forbidden another event, on the contrary I want allow mouse (pen already allowed).

derrell: Yes, sorry about that. I really wonder why mouse was excluded in the first place...

derrell (approved):

PR #10497, opened 2022-12-07
Fix - Prevent multiple redraws on cascading open/close node

rad-pat:
Fixes a problem from when I introduced this code some while back. The private method should have been called in the node loop to prevent redraws whilst cascading the open/close nodes. It had the correct parameters, but was erroneously calling the public method.

### DIFF
--- a/source/class/qx/ui/treevirtual/MNode.js
+++ b/source/class/qx/ui/treevirtual/MNode.js
@@ -226,7 +226,7 @@ qx.Mixin.define("qx.ui.treevirtual.MNode", {
       }
       if (cascade) {
         node.children.forEach(child =>
-          this.nodeSetOpened(child, opened, cascade, true)
+          this._nodeSetOpenedInternal(child, opened, cascade, true)
         );
       }
       if (!cascade || !isRecursed) {


oetiker (approved):

cboulanger (approved):

goldim:
Method with 3 arguments get 4 - obviously an error. Logically, supposed other method to call. Didn't find any usage of this mixin nether in API Viewer nor in qooxdoo source via search.

goldim (approved):

hkollmann:
It's used in the DemoBrowser:
e.g. 
```
qx.Class.define("qxl.demobrowser.demo.treevirtual.TreeVirtual_Events", {
  extend: qx.application.Standalone,

  members: {
    main() {
      super.main();
      // We want to use some of the high-level node operation convenience
      // methods rather than manually digging into the TreeVirtual helper
      // classes.  Include the mixin that provides them.
      qx.Class.include(qx.ui.treevirtual.TreeVirtual, qx.ui.treevirtual.MNode);
```

hkollmann (approved):

derrell (approved):

level420 (approved):

hkollmann:
Thanks @rad-pat !

PR #10502, opened 2022-12-13
revert PR 10491 because it breaks drag and drop

johnspackman:
The PR in https://github.com/qooxdoo/qooxdoo/pull/10491 enabled the mouse to scroll the container, but this breaks the drag and drop behaviour if the item being dragged is inside a list which is on a scrollable container.

What happens is that as the user tries to drag an item (eg to reorder the contents of a list), then as they move the mouse the entire container scrolls, and this makes it impossible to position the drop target - because as you move closer to where you want to drag it, the entire container scrolls and moves the list.

Sometimes this causes the screen to jitter as the repositioning causes the drag target as the mouse moves due to automatic scrolling.

I have not investigated a fix which could allow auto-scroll and drag and drop to not conflict, but as this is a relatively recent change it seems that a first response would be to stop the problem from occurring

### DIFF
--- a/source/class/qx/ui/core/scroll/MRoll.js
+++ b/source/class/qx/ui/core/scroll/MRoll.js
@@ -58,6 +58,11 @@ qx.Mixin.define("qx.ui.core.scroll.MRoll", {
      * @param e {qx.event.type.Roll} Roll event
      */
     _onRoll(e) {
+      // only wheel and touch
+      if (e.getPointerType() == "mouse") {
+        return;
+      }
+
       if (this._cancelRoll && e.getMomentum()) {
         e.stopMomentum();
         this._cancelRoll = null;


derrell:
It seems that an appropriate fix that accommodates both goals is to replace this original patch deleting this code:
```
     // only wheel and touch
      if (e.getPointerType() == "mouse") {
```

with code that does that original thing if a `drag` is in process.

derrell (changes requested):
See my immediately prior comment proposing a fix that solves both problems

goldim (approved):
I'll try to find another way. But with Drag&Drop it is impossible to scroll list for example. Sad.

johnspackman:
> It seems that an appropriate fix that accommodates both goals is to replace this original patch deleting this code:
> with code that does that original thing if a drag is in process.

Yes, that would obviously be ideal - but unfortunately I do not have time to figure out how that would work, and this PR fixes the immediate issue where D&D is broken.

>  if a drag is in process.
Drag and drop is implemented across a number of files, and can involve events firing up and down the stack of widgets/elements.  It would be great if there is a way to determine whether d&d is operating, but that "if" might take some work to add that in.

If you can point me at the code that should be used to reliably test whether drag and drop is operating, I will gladly add it into this PR.

I suggest that we accept this PR in the short term, and return to add the new feature of auto scrolling later.

@derrell in the previous PR [you said](https://github.com/qooxdoo/qooxdoo/pull/10491#discussion_r1032937625):
"I really wonder why mouse was excluded in the first place..."

I guess this is why.

derrell (approved):
Yup, it certainly seems to now be explained. My concern was valid, I guess.

Ok, I'll approve this to revert, and hopefully @goldim is able to update the original PR to accommodate drag&drop. (@goldim, let me know if you'd like some help. I haven't implemented drag&drop for for quite some time, but IIRC, it is easy to ascertain whether a drag operation is in progress.)

hkollmann (approved):

Issue #10507, opened 2022-12-20
Calling widget.dispose() doesn't remove the widget from the layout parent.


lucapivato:
Not sure if this is a bug or the intended behavior. I'm going to change it in our branch.

In the destroy method in qx.ui.core.Widget it checks if the widget is already disposed and exists if it is. However, calling dispose() on a widget it sets the this.$$disposed flag to true and then calls the destructors second, and since parent_remove(this) in Widget.destroy() comes after the $$disposed check calling dispose() leaves disposed widget as invalid children.

Calling destroy() on the widget remove it but queues the disposal(). It seems to me that it would be more consistent to remove the widget from the layout parent also if calling dispose().

hkollmann:
Could you create a PR please?

lucapivato:
Looks like the Destroy test doesn't like it.

BTW this is a repro simple case: https://tinyurl.com/bdcss7db

PR #10508, opened 2022-12-20
Fix for https://github.com/qooxdoo/qooxdoo/issues/10507

lucapivato:


hkollmann:
Seems that this change kill all the tests. Any idea why?

PR #10510, opened 2022-12-21
fix problem that wait will always restart if you change files during compile

hkollmann:
Up to now wait starts before the first build was finished. So if you change files during build, e.g. generated version infos, build was restartet infinite. Now the wait starts after the first build is finished


derrell (approved):

cboulanger (approved):

goldim (approved):
Great!

level420 (approved):

oetiker (approved):

PR #10512, opened 2023-01-05
Fix IndigoDark tooltip

rad-pat:
Remove label appearance which was preventing the tooltip textcolor cascading through.
I can't see that this has upset any other label usage in the Widget Browser

Did look like this:
image

Now looks like this:
image

### DIFF
--- a/source/class/qx/theme/indigo/ColorDark.js
+++ b/source/class/qx/theme/indigo/ColorDark.js
@@ -85,8 +85,8 @@ qx.Theme.define("qx.theme.indigo.ColorDark", {
     "text-placeholder": "#cbc8cd",
 
     // tooltip
-    "tooltip": "#dddddd",
-    "tooltip-text": "#000000",
+    "tooltip": "#666666",
+    "tooltip-text": "#dddddd",
 
     // table
     "table-header": "#f2f2f2",
--- a/source/class/qx/theme/indigo/AppearanceDark.js
+++ b/source/class/qx/theme/indigo/AppearanceDark.js
@@ -34,14 +34,6 @@ qx.Theme.define("qx.theme.indigo.AppearanceDark", {
   extend: qx.theme.indigo.Appearance,
 
   appearances: {
-    "label": {
-      style(states) {
-        return {
-          textColor: states.disabled ? "text-disabled" : "text"
-        };
-      }
-    },
-
     "listitem": {
       alias: "atom",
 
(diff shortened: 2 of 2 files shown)


oetiker:
hmm not sure, but from looking at the patch I would worry that it affects other things too

sknick:
> hmm not sure, but from looking at the patch I would worry that it affects other things too. Schouldn't a tooltip in a dark theme be dark anyway ?

That's debatable I guess. I think what rad-pat's commit does is fix another case of a hard-coded color in the Appearance theme, which I had to do several times in making the dark theme.

sknick:
Since a version of qx hasn't even been released yet that contains the new theme, there is no chance of regressions for existing qx users, so it probably won't hurt to throw caution to the wind and just merge it. Just my $0.02.

hkollmann (approved):

derrell (approved):
`qx.theme.indigo.Appearance`, which this inherits from does not have a `label` appearance, so this was overwriting the super-superclass `label` in `qx.theme.simple.Appearance` The difference is that when `states.disabled` is falsy, `textColor` is set to `undefined` instead of "text".

The `AppearanceDark` theme, to my knowledge, has never yet seen the light of day in a release, so this can as easily as not simply be called a bug in the original commit of this new appearance.

Good to go. Ship it!

cboulanger (approved):

level420 (approved):

oetiker (approved):

goldim (changes requested):
I checked out other two dark themes and a tooltip has black/dark background color and white text color there.  My opinion is IndigoDark tooltip also should be like those.

goldim:
A theme select button in Widget Browser now has black text color on black background with the change too.

rad-pat:
Updated tooltip to be black with light text. Removed redundant appearance for listitem.
The black text in the selectbox of widget browser will likely need fixing there https://github.com/qooxdoo/qxl.widgetbrowser/blob/a4ad62842f5208b40b2aa3a2283acfda4fbedcf1/source/class/qxl/widgetbrowser/view/Header.js#L68

source/class/qx/theme/indigo/ColorDark.js
```diff
@@ -85,8 +85,8 @@ qx.Theme.define("qx.theme.indigo.ColorDark", {
     "text-placeholder": "#cbc8cd",
 
     // tooltip
-    "tooltip": "#dddddd",
-    "tooltip-text": "#000000",
+    "tooltip": "#000000",
```
goldim, line 88: I think less harsh color than a pure black color for background would be nice some kind of gray. Tried #666666 - probably looks better than black.

rad-pat: Updated to gray.

goldim (approved):

Issue #10511, opened 2023-01-01
testSelf is not defined


goldim:
**Describe the bug**
Noticed in console of fresh new app an error since framework 7.4.1

**To Reproduce**
1. Create new project
2. Install @qooxdoo/framework@7.4.1
3. `qx create`
4. `qx serve`
5. Open dev tools -> console
6. See the error

**Expected behavior**
No error

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Yandex Browser, Edge
 - Version latest

**Additional context**
Stacktrace:
```
index.js:11380 [Violation] 'setTimeout' handler took 163ms
Function.js:200 Uncaught ReferenceError: testSelf is not defined
    at Function.js:200:11
(анонимная) @ Function.js:200
Объект setTimeout (асинхронный)
schedule @ DeferredCallManager.js:64
schedule @ DeferredCall.js:74
__addToBlacklist__P_62_14 @ Manager.js:1010
removeListener @ Manager.js:707
removeListener @ Registration.js:138
removeListener @ MEvent.js:116
_removeWindow @ MDesktop.js:204
_afterRemoveChild @ MDesktop.js:189
__removeHelper__P_24_10 @ Widget.js:1995
_remove @ Widget.js:1839
__addHelper__P_24_9 @ Widget.js:1944
_add @ Widget.js:1722
add @ MChildrenHandling.js:83
main @ Application.js:49
ready @ BaseInit.js:77
(анонимная) @ Direct.js:133
_true @ Utils.js:149
(анонимная) @ Direct.js:132
dispatchEvent @ Direct.js:107
wrappedFunction @ Interface.js:529
dispatchEvent @ Manager.js:958
(анонимная) @ Registration.js:361
_true @ Utils.js:149
fireEvent @ Registration.js:360
__fireReady__P_39_5 @ Application.js:170
onScriptLoaded @ Application.js:93
done @ index.js:11334
signalStartup @ index.js:11357
(анонимная) @ index.js:11384
Объект setTimeout (асинхронный)
(анонимная) @ index.js:11380
checkForEnd @ index.js:11593
onLoad @ index.js:11601
elem.onreadystatechange.elem.onload @ index.js:11471
Объект load (асинхронный)
loadScript @ index.js:11467
flushScriptQueue @ index.js:11606
begin @ index.js:11378
init @ index.js:11392
(анонимная) @ index.js:11694
```

rad-pat:
[...]

PR #10515, opened 2023-01-06
Move scope of testSelf so it is available where used

rad-pat:
To address https://github.com/qooxdoo/qooxdoo/issues/10511, move scope of testSelf function

### DIFF
--- a/source/class/qx/lang/Function.js
+++ b/source/class/qx/lang/Function.js
@@ -173,30 +173,27 @@ qx.Bootstrap.define("qx.lang.Function", {
         return func;
       }
 
-      if (qx.core.Environment.get("qx.debug")) {
-        function testSelf(self) {
-          if (
-            qx.core.Object &&
-            self &&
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

hkollmann (approved):

goldim (approved):
Using this change the problem is gone but still can't understand why it does happen. Didn't find any usage of `testSelf` in compiled version of framework (v7.4.2) via IDE or MS Code editor except 2 ones in `qx.lang.Function.create`.

level420 (approved):

oetiker (approved):

oetiker:
> Using this change the problem is gone but still can't understand why it does happen. Didn't find any usage of `testSelf` in compiled version of framework (v7.4.2) via IDE or MS Code editor except 2 ones in `qx.lang.Function.create`.

if the code is not running in debug mode

PR #10521, opened 2023-01-08
Fixed dry run argument for publish command

goldim:
Also doc is fixed


cboulanger:
In order to make this backward-compatible, you could add the following code at the appropriate place:

```javascript
{
// snip
dryrun: {
    describe: 'Deprecated. Use --dry-run',
    type: 'boolean'
  }
//snip
}

// line 148
if (this.argv.dryrun) {
  console.log('The "--dryrun" option is deprecated. Please use "--dry-run" instead.');
  this.argv['dry.-run'] = true
}
```

And you need to update the use of the option as  well: https://github.com/qooxdoo/qooxdoo/blob/89e3699b5a08dee45e50154e32bee4d3a35ec200/source/class/qx/tool/cli/commands/package/Publish.js#L380

cboulanger (approved):

source/class/qx/tool/cli/commands/package/Publish.js
```diff
@@ -79,6 +79,12 @@ qx.Class.define("qx.tool.cli.commands.package.Publish", {
           },
 
           dryrun: {
+            type: "boolean",
+            alias: "d",
```
derrell, line 83: I think this is a problem. There are two options with the same alias, 'd'. Since `dryrun` is deprecated, it needn't have an alias.

derrell (changes requested):

source/class/qx/tool/cli/commands/package/Publish.js
goldim: @derrell Do I need  just remove an alias for the deprecated option?

derrell: Yes, I believe that's the correct solution. Leave the 'd' alias on `--dry-run` but remove it from `--dryrun`. I do not think it's required that an option has an alias.

derrell (approved):
Looks good. Thanks!

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

level420 (approved):

PR #10526, opened 2023-01-21
Templates from es5 to es6 syntax

goldim:
Moving to ES6 syntax:

- `var` -> `const` or `let`
- using `super` instead `this.base`
- using `super` for calling parent method
- removing `function` from method syntax


derrell (approved):

hkollmann (approved):

cboulanger (approved):
👍

oetiker (approved):

johnspackman (approved):

PR #10527, opened 2023-01-22
Fixes for locale bugs 

### LINKED ISSUE #10522: Compiler crash with sr_Latn locale
**Describe the bug**
When adding the "sr_Latn" string to the locales array, the compiler crash:
```
Error: TypeError: Cannot read properties of undefined (reading 'length')
    at /tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:55697:47
    at find (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:55501:26)
    at /tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:55696:15
    at find (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:55501:26)
    at /tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:55695:13
    at async /tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:47965:22
From previous event:
    at Function.__P_143_8 (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:20570:69)
    at Function.all (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:20146:27)
    at wrapper._writeLocales (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:47969:26)
    at wrapper._writeTranslations (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:47887:20)
    at wrapper.generateApplication (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:47787:98)
    at async wrapper.make (/tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:49898:11)
    at async /tmp/testlocale/node_modules/@qooxdoo/framework/lib/compiler/index.js:46429:20
```

**To Reproduce**
You can use this project [crash_locale.zip](https://github.com/qooxdoo/qooxdoo/files/10373577/crash_locale.zip). Created by following the tutorial [The First Application](https://qooxdoo.org/documentation/v7.4/#/README?id=the-first-application) and by adding sr_Latn to the locales array in compile.json (but any project will crash)

1. Unzip the crash_locale.zip
2. npm i && npx qx compile

**Expected behavior**
No compiler crash

 - OS: Mac OS Monterey
 - Browser NA
 - Version  NA

goldim:
Fixed #10522 

1. Compiling such locales as en_GB and sr_Latn now works on Linux (case-sensitive OS)
2. sr_Latn has root.xml (parent locale dependency) which doesn't have some content like month abbreviations.

If there is rejection for readdir promise in case when file for searched locale is not found I do nothing (except trace there will be error after compilation). I don't know how properly handle it. Any suggestion.

### DIFF
--- a/source/class/qx/tool/compiler/app/Cldr.js
+++ b/source/class/qx/tool/compiler/app/Cldr.js
@@ -29,6 +29,7 @@ var xml2js = require("xml2js");
 const CLDR = require("cldr");
 const { promisify } = require("util");
 const readFile = promisify(fs.readFile);
+const readDir = promisify(fs.readdir);
 
 var log = qx.tool.utils.LogManager.createLog("cldr");
 
@@ -59,9 +60,24 @@ qx.Class.define("qx.tool.compiler.app.Cldr", {
       }
       log.debug("Loading CLDR " + cldrPath);
 
-      return readFile(path.join(cldrPath, data_path, locale + ".xml"), {
-        encoding: "utf-8"
+      const fullDir = path.join(cldrPath, data_path);
+
+      return readDir(fullDir).then(names => {
+        return new Promise((resolve, reject) => {
+          const searchedName = locale.toLowerCase() + ".xml";
+          const realName = names.find(name => name.toLowerCase() === searchedName);
+          if (realName){
+            resolve(realName);
+          } else {
+            reject(new Error("Cannot find XML file for locale \"" + locale + "\" in CLDR folder"));
+          }
+        });
       })
+        .then(fileName =>
+          readFile(path.join(fullDir, fileName), {
+            encoding: "utf-8"
+          })
+        )
         .then(data =>
           qx.tool.utils.Utils.promisifyThis(parser.parseString, parser, data)
         )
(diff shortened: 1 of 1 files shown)


source/class/qx/tool/compiler/app/Cldr.js
```diff
@@ -59,9 +60,24 @@ qx.Class.define("qx.tool.compiler.app.Cldr", {
...
       log.debug("Loading CLDR " + cldrPath);
 
-      return readFile(path.join(cldrPath, data_path, locale + ".xml"), {
-        encoding: "utf-8"
+      const fullDir = path.join(cldrPath, data_path);
+
+      return readDir(fullDir).then(names => {
+        return new Promise((resolve, reject) => {
+          const searchedName = locale.toLowerCase() + ".xml";
+          const realName = names.find(name => name.toLowerCase() === searchedName);
```
derrell, line 68: This looks pretty liberal, allowing any mixed case. Maybe that's ok, but might it be more appropriate to check for the two known options (fully uppercase, or capitalized)?

source/class/qx/tool/compiler/app/Cldr.js
```diff
@@ -564,29 +580,28 @@ qx.Class.define("qx.tool.compiler.app.Cldr", {
...
-                }
+            }
+
+            const parseMonthContext = (what, where) => {
+              find(monthContext, "type", "format", function (row) {
+                find(row.monthWidth, "type", what, function (row) {
+                  parseMonth(row.month, where);
+                });
               });
-            });
```
derrell, line 583: This change looks like it's lost the effect of `cldr["cldr_month_format_wide_" + m["$"].type] = getText(m);`. Is that no longer needed in this version?

derrell (changes requested):

source/class/qx/tool/compiler/app/Cldr.js
goldim: I replaced 2 code duplication by a function. There are two calls of the function for wide and abbreviated:
```
 parseMonthContext("abbreviated","cldr_month_format_abbreviated");
 parseMonthContext("wide","cldr_month_format_wide");
 ```
 Didn't understand a problem.

derrell: I see what you did. Sorry. Yes, should be fine.

source/class/qx/tool/compiler/app/Cldr.js
goldim: @derrell How am I supposed to check these cases: ff_Latn_SN.xml, be_TARASK.xml, en_US_POSIX.xml?
I mean parse string for 2 or 3 parts and make on them some transformations (into full or capitalized). In examples the second part could be or fully uppercase or capitalized. The checking code would be difficult. Does it worth it? There are around 600 locales.

derrell: Do we want to allow any case? Is there a specification that dictates what the case should be? I think you'd mentioned previously that there were at least two valid options of case. It's fine to be liberal as long as that isn't going to cause any problems. Is there ever a time where two different mixes of case of the same text are *supposed* to be different? If not, then no, it's likely not worth it at all.

derrell: I'll approve on the assumption that there are no such times. If you can think of any, we can discuss how to best implement it.

derrell (approved):

cboulanger (approved):

source/class/qx/tool/compiler/app/Cldr.js
goldim: @derrell I was wrong when said there are two cases. Actually there are more than 2 choices. You may see all file names [here](https://github.com/papandreou/node-cldr/tree/master/3rdparty/cldr/common/main). At the moment in Windows if you set up for example a locale name in 'compile.json' like sR_laTin it will work.

oetiker (approved):

source/class/qx/tool/compiler/app/Cldr.js
goldim: I refactor this duplication bc I had to add the same check 3 times.

johnspackman (approved):

hkollmann (approved):

PR #10531, opened 2023-01-27
Catch block for no xml file for locale

goldim:


hkollmann (approved):

cboulanger (approved):

source/class/qx/tool/compiler/app/Cldr.js
```diff
@@ -88,6 +90,10 @@ qx.Class.define("qx.tool.compiler.app.Cldr", {
             encoding: "utf-8"
           })
         )
+        .catch(function (err) {
```
derrell, line 93: Only for consistency with the above code, this could be an arrow function rather than a traditional function. I will approve anyway, but for your consideration...

derrell (approved):

level420 (approved):

oetiker (approved):

johnspackman (approved):

PR #10534, opened 2023-02-04
Static method with reference value

goldim:
Static methods value of which is other qooxdoo class method full name are displayed as constants in API Viewer. Investigating the problem I found out that detection of static method requires another check. This issue is related to [this](https://github.com/qooxdoo/qxl.apiviewer/issues/46).


hkollmann (approved):

derrell (approved):

johnspackman (approved):

cboulanger (approved):

oetiker (approved):

hkollmann:
Thanks @goldim

PR #10538, opened 2023-02-14
Add background for IndigoDark menu button

### LINKED ISSUE #10517: Wrong colors in IndigoDark for some widgets and states
Looking through Widget Browser I've noticed there are some widget has bad/not suitable colors for theme. Firstly it is a tooltip. There is black color of text and white background. Need to be vice versa. Secondly it is a hovered MenuBar button: both text and background colors during hover are the same (white).

goldim:
Fixed #10517

### DIFF
--- a/source/class/qx/theme/indigo/DecorationDark.js
+++ b/source/class/qx/theme/indigo/DecorationDark.js
@@ -10,19 +10,21 @@
 
    Authors:
      * Scott Knick (sknick)
+     * Dmitrii Zolotov (goldim)
 
 ************************************************************************ */
-/**
- * IndigoDark Theme
- */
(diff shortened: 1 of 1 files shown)


derrell (approved):

derrell:
Hmmm... This change will affect existing apps in a non-BC way. May require a major version upgrade...

oetiker (approved):

oetiker:
> Hmmm... This change will affect existing apps in a non-BC way. May require a major version upgrade...

I think this should be considered a bugfix :)

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

level420 (approved):

PR #10539, opened 2023-02-15
long running compile processes do  not exit

hkollmann:
This is problem with newer version of yargs which calls the process handler to early. 
This fix stores the needed command and do the processing after initialisation. 
Furthermore process.exit is added to end the process after finishing the promise.

### DIFF
--- a/source/class/qx/tool/cli/Application.js
+++ b/source/class/qx/tool/cli/Application.js
@@ -34,9 +34,9 @@ qx.Class.define("qx.tool.cli.Application", {
         await new qx.tool.cli.Cli().run();
       } catch (e) {
         qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-        process.exitCode = -1;
+        process.exitCode = 1;
       } finally {
-        process.exit(process.exitCode);
+        process.exit();
       }
     }
   },
--- a/source/class/qx/tool/cli/Application.js
+++ b/source/class/qx/tool/cli/Application.js
@@ -28,13 +28,16 @@ qx.Class.define("qx.tool.cli.Application", {
      * This method contains the initial application code and gets called
      * during startup of the application
      */
-    async main() {
-      try {
-        await new qx.tool.cli.Cli().run();
-      } catch (e) {
-        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-        process.exit(1);
-      }
+    main() {
+      new qx.tool.cli.Cli()
+        .run()
+        .then(() => {
+          process.exit(0);
+        })
+        .catch(e => {
+          qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
+          process.exit(1);
+        });
     }
   },
 
--- a/source/class/qx/tool/cli/Application.js
+++ b/source/class/qx/tool/cli/Application.js
@@ -28,16 +28,16 @@ qx.Class.define("qx.tool.cli.Application", {
      * This method contains the initial application code and gets called
      * during startup of the application
      */
-    main() {
-      new qx.tool.cli.Cli()
-        .run()
-        .then(() => {
-          process.exit(0);
-        })
-        .catch(e => {
-          qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-          process.exit(1);
-        });
+    async main() {
+      process.exitCode = 0;
+      try {
+        await new qx.tool.cli.Cli().run();
+      } catch (e) {
+        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
+        process.exitCode = -1;
+      } finally {
+        process.exit(process.exitCode);
+      }
     }
   },
 
--- a/source/class/qx/tool/cli/Cli.js
+++ b/source/class/qx/tool/cli/Cli.js
@@ -25,6 +25,11 @@ const semver = require("semver");
  */
 qx.Class.define("qx.tool.cli.Cli", {
   extend: qx.core.Object,
+  properties: {
+    command: {
+      apply: "__applyCommand"
+    }
+  },
 
   construct() {
     super();
@@ -55,11 +60,13 @@ qx.Class.define("qx.tool.cli.Cli", {
     /** @type {Object} Parsed arguments */
     _parsedArgs: null,
 
-    /** @type {Promise} Promise that resolves to the _parsedArgs, but only when completely finished parsing them */
-    __promiseParseArgs: null,
-
     /** @type {Boolean} Whether libraries have had their `.load()` method called yet */
     __librariesNotified: false,
+
+    __applyCommand(command) {
+      command.setCompilerApi(this._compilerApi);
+      this._compilerApi.setCommand(command);
+    },
 
     /**
      * Creates an instance of yargs, with minimal options
@@ -241,8 +248,6 @@ Version: v${await qx.tool.config.Utils.getQxVersion()}
      */
     async processCommand(command) {
       qx.tool.compiler.Console.getInstance().setVerbose(this.argv.verbose);
-      command.setCompilerApi(this._compilerApi);
-      this._compilerApi.setCommand(command);
       await this.__notifyLibraries();
       try {
         const res = await command.process();
@@ -260,8 +265,8 @@ Version: v${await qx.tool.config.Utils.getQxVersion()}
      *
      * @return {Object}
      */
-    async getParsedArgs() {
-      return await this.__promiseParseArgs;
+    getParsedArgs() {
+      return this.__parsedArgs;
     },
 
     /**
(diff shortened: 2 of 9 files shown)


source/class/qx/tool/cli/Cli.js
```diff
@@ -356,8 +361,7 @@ Version: v${await qx.tool.config.Utils.getQxVersion()}
             (await qx.tool.utils.Json.loadJsonAsync(name)) || lockfileContent;
         } catch (ex) {
           // Nothing
-        }
-        // check semver-type compatibility (i.e. compatible as long as major version stays the same)
+        } // check semver-type compatibility (i.e. compatible as long as major version stays the same)
```
derrell, line 364: This looks like an accidental joining of the two lines. I don't believe the comment belongs after the ending brace, right?

derrell (changes requested):

source/class/qx/tool/cli/Application.js
```diff
@@ -28,13 +28,16 @@ qx.Class.define("qx.tool.cli.Application", {
...
+      new qx.tool.cli.Cli()
+        .run()
+        .then(() => {
+          process.exit(0);
+        })
+        .catch(e => {
+          qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
+          process.exit(1);
+        });
     }
```
johnspackman, line 41: why is this switching from `async/await` to old manual promises?

source/class/qx/tool/cli/Cli.js
```diff
@@ -283,8 +288,8 @@ Version: v${await qx.tool.config.Utils.getQxVersion()}
       var args = qx.lang.Array.clone(process.argv);
       args.shift();
       process.title = args.join(" ");
-      this.__promiseParseArgs = this.__parseArgsImpl();
-      await this.__promiseParseArgs;
+      await this.__parseArgsImpl();
```
johnspackman, line 291: `__promiseParsedArgs` was used to avoid a race condition, I can't remember exactly why it came about, but it I expect it would be because something in `__parseArgsImpl` can refer recursively back to the `Cli.getParsedArgs()`.  Recursion will fail now, which is fine if that's been eliminated.   

Possibly add a check that `__parsedArgs` is non-null in `getParsedArgs`?

johnspackman (changes requested):

source/class/qx/tool/cli/Application.js
oetiker, line 41: how about this?

```suggestion
    async main() {
      try {
        await new qx.tool.cli.Cli().run();
        process.exit(0);
      } catch (e) {
        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
        process.exit(1);
      }
```

goldim: also there is finally keyword in which could be done process.exit(<<your variable value>>) to get rid of duplicate

source/class/qx/tool/cli/Application.js
```diff
@@ -28,13 +28,16 @@ qx.Class.define("qx.tool.cli.Application", {
...
      * during startup of the application
      */
-    async main() {
-      try {
-        await new qx.tool.cli.Cli().run();
-      } catch (e) {
-        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-        process.exit(1);
-      }
+    main() {
```
goldim, line 31:
```suggestion
   async main() {
      let exitCode = 0;
      try {
        await new qx.tool.cli.Cli().run();
      } catch (e) {
        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
       exitCode = 1;
      }
      finally {
         process.exit(exitCode);
      }
```

hkollmann:
I switched to old promise style because main is not called as promise function without await 

So it's a sychron function.
IHMO this style is clearer in reading

source/class/qx/tool/cli/Cli.js
hkollmann: Done

source/class/qx/tool/cli/Application.js
hkollmann: Switched back to await

source/class/qx/tool/cli/Cli.js
hkollmann: Only usage is in Lint. There null value is handled

source/class/qx/tool/cli/Application.js
hkollmann: Done

source/class/qx/tool/cli/Application.js
```diff
@@ -28,16 +28,16 @@ qx.Class.define("qx.tool.cli.Application", {
...
-          qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-          process.exit(1);
-        });
+    async main() {
+      process.exitCode = 0;
+      try {
+        await new qx.tool.cli.Cli().run();
+      } catch (e) {
+        qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
+        process.exitCode = -1;
```
goldim, line 37: failure code -1, maybe 1?

source/class/qx/tool/cli/Application.js
```diff
@@ -29,11 +29,14 @@ qx.Class.define("qx.tool.cli.Application", {
      * during startup of the application
      */
     async main() {
+      process.exitCode = 0;
       try {
         await new qx.tool.cli.Cli().run();
       } catch (e) {
         qx.tool.compiler.Console.error("Error: " + (e.stack || e.message));
-        process.exit(1);
+        process.exitCode = -1;
```
goldim, line 37: Failure code is 1, didn't find -1 value in nodejs docs

source/class/qx/tool/cli/Application.js
```diff
@@ -29,11 +29,14 @@ qx.Class.define("qx.tool.cli.Application", {
...
+      } finally {
```
goldim, line 38: if process.exitCode was set before there is no need to pass argument for exit method

source/class/qx/tool/cli/Application.js
derrell: I missed this before. There are a few issues here. Firstly, this `process.exitCode = -1;` will not even be executed. `process.exit(1)` is the last line of JS code that will be executed.

Exit codes are, by definition in some environments, and by convention in others, 8-bit unsigned values, with 0 meaning success and 1-255 meaning various types of error exits. Using negative values as exit codes is typically discouraged, and in most environments, simply maps to the unsigned equivalent. `process.exit(-1)`, for example, will usually yield an exit code of `255` as seen by the parent process.

`process.exitCode` is typically used when there are multiple places that could fail. You can set `process.exitCode` to a unique value for each of those failure locations, and then do `process.exit()` (with no argument) to exit with the current value of `process.exitCode`

goldim:
@hkollmann  Is PR finished for checking and testing? And I would like to know which situations to look at and check. Does the problem with yargs already happen? only on master?

hkollmann:
The problem is that node do not stop with very large projects.
So i added the process.exit code.

During my tests qx test and qx serve need some extra handling because they do not wait for processing before.
Now everything should work again.

goldim (changes requested):
Lint doesnt work. Deploy, compile, serve seems fine.
```
goldim@goldim:/var/www/html/git/scadaclient/app$ ../../qooxdoo/bootstrap/qx lint
One or more libraries not found - trying to install them from library repository...
Error: TypeError: Cannot read properties of undefined (reading 'eslintConfig')
    at wrapper.process (/var/www/html/git/qooxdoo/bootstrap/compiled/node/build/compiler/index.js:56121:34)
    at async wrapper.processCommand (/var/www/html/git/qooxdoo/bootstrap/compiled/node/build/compiler/index.js:31704:21)
    at async wrapper.main (/var/www/html/git/qooxdoo/bootstrap/compiled/node/build/compiler/index.js:31254:11)
```

hkollmann:
Lint testet in a new project?

goldim:
@hkollmann I just run usual global installed qx lint - it works without errors.

hkollmann:
Lint is fixed!

oetiker (approved):

goldim (approved):

cboulanger (approved):
Cannot really comment, trusting your judgement here

johnspackman (approved):

source/class/qx/tool/cli/Cli.js
derrell: This was marked as resolved, but doesn't appear to have changed. The comment is still showing as joined onto the line with the closing brace.

derrell (changes requested):

hkollmann:
Can'nt find the  requested change?

derrell (approved):

PR #10560, opened 2023-03-14
Clear memory for array in DateFormat

### LINKED ISSUE #10558: Missing destruct warning in qx.util.format.DateFormat
**Describe the bug**
Get a warning:
```
004395 qx.util.format.DateFormat: Missing destruct definition for '__formatTree__P_37_6' in qx.util.format.DateFormat[undefined]: [object Object]
```

**To Reproduce**
1. Add this environment settings at compile.json:
```
"environment": {
    "qx.debug": true,
    "qx.debug.dispose.level": 3,
    "qx.debug.dispose": true
},
```
2. Create an instance of  `qx.util.format.DateFormat`
3. Call some methods of it
4. Open console
5. See the warning

**Expected behavior**
No warning

goldim:
Fixed #10558

### DIFF
--- a/source/class/qx/util/format/DateFormat.js
+++ b/source/class/qx/util/format/DateFormat.js
@@ -109,6 +109,10 @@ qx.Class.define("qx.util.format.DateFormat", {
     }
   },
 
+  destruct() {
+    this.__formatTree = null;
+  },
+
   /*
   *****************************************************************************
      STATICS
--- a/source/class/qx/util/format/DateFormat.js
+++ b/source/class/qx/util/format/DateFormat.js
@@ -110,6 +110,7 @@ qx.Class.define("qx.util.format.DateFormat", {
   },
 
   destruct() {
+    // has to be cleared explicitly as it is pointing to a complex datastructure
     this.__formatTree = null;
   },
 


oetiker:
any idea why this is necessary for __formatTree but not for __format ? Or is it just some compilation artifact ?

goldim:
@oetiker as I could say reading the code of class __format is just a string. But __formatTree is a complex object. Btw I compiled with high level of debug and dispose. In my app there is a storm of these warnings related with DateFormat.

goldim:
Btw in qooxdoo docs is following said:

> As a rule of thumb
> 
> - native Javascript types (Number, String, Object, ...) usually don't need to be disposed.
> - for Qooxdoo objects (e.g. [qx.util.format.DateFormat](http://qooxdoo.org/apps/apiviewer/#qx.util.format.DateFormat), testgui.Report, ...) use _disposeObjects
> - for arrays or maps of Qooxdoo objects use _disposeArray or _disposeMap.
> - be sure to cut all references to the DOM because garbage collection can not dispose object still connected to the DOM. This is also true for event listeners for example.

I think this case is the first point of list: String, Object, **Array** . The array `__formatTree` is an array with JS Objects (non qooxdoo).

oetiker (approved):

source/class/qx/util/format/DateFormat.js
```diff
@@ -109,6 +109,10 @@ qx.Class.define("qx.util.format.DateFormat", {
     }
   },
 
+  destruct() {
+    this.__formatTree = null;
```
oetiker, line 113:
```suggestion
    // has to be cleared explicitly as it is pointing to a complex datastructure
    this.__formatTree = null;
```

source/class/qx/util/format/DateFormat.js
```diff
@@ -109,6 +109,11 @@ qx.Class.define("qx.util.format.DateFormat", {
     }
   },
 
+  destruct() {
+    // has to be cleared explicitly as it is pointing to a complex datastructure
+    this.__formatTree = null;
```
derrell, line 114: Although it doesn't hurt to have this here, it's unnecessary. `__formatTree` is simply an array of objects. There is nothing in it that can't be easily garbage-collected by JavaScript's normal memory management. I have no problem with leaving the destructor here, but the comment is inaccurate, as it indicates that this cleanup is required, when it is not.

derrell (changes requested):

source/class/qx/util/format/DateFormat.js
goldim: Revert last commit with @oetiker changes? There is no problem with memory deallocation but I think a lot of compiler warnings which comes out from qooxdoo framework is not good thing if someone wants to find out bad weak places of his app code.

derrell: What compiler warnings? Is there some warning that was being shown, that this fixes?

goldim: @derrell these ones:
```
004395 qx.util.format.DateFormat: Missing destruct definition for '__formatTree__P_37_6' in qx.util.format.DateFormat[undefined]: [object Object]
```
I described it in attached issue. But please notice, I set dispose level 3.

goldim: DateFormat is not only what has this destruct warning. At this moment I also noticed the same problem with disposing `qx.ui.table.Table`  there are a lot of warnings about internal members. Probably we should discuss it on Gitter.

johnspackman:
WRT the dispose warnings - my TL;DR is that the warnings should only be if the object implements `qx.core.IDisposable`

-------------

This is probably something that came in when the automatic memory management was introduced, ie when we no longer needed to dispose most objects and could rely on the default garbage collection.

But reading the code it only needs to complain if the object implements `qx.core.IDisposable` - which is how a class advertises that it *must* be manually `.dispose()`-d.

The issue is that destructors are never called when an object is garbage collected - this isn't a big deal, because even C++ (where the destructor is always called by the GC) the destructors are almost entirely useless, because they are called in any order and must never look outside the class (unless you want catastrophic side effects).

So `dispose()` is a way of providing an API to the equivalent of a destructor, but it must be manually triggered.

I suspect that many of the `dispose()` or `destruct()` methods can be entirely eliminated because they were only written because of when every object was globally registered.

Removing the destructors/dispose would clean up the code a lot, because when you see that code you have to wonder whether you need to dispose or not - ie it's very misleading.  But that takes a lot of work to deeply evaluate each destructor and decide whether to delete it or not.

I'm guessing again, but mahbe that's why I wrote the `qx.core.IDisposable` interface  - so we had an obvious way of saying "this needs to be disposed manually" but doesn't mean that we delete the destructor code and lose it forever.

derrell:
So this destructor (`desturct` method) is never called, since there is no call to `dispose()` in this class.

It seems the proper solution to this problem is that the debug dispose level 3 code should not flag missing destructors (or members not disposed) unless the class implements `qx.core.IDisposable`. Would you agree, @johnspackman?

johnspackman:
> So this destructor (`desturct` method) is never called, since there is no call to `dispose()` in this class.

the destructor would be called, but only if someone called `myFormater.dispose()` - but it's pointless to do so.  This class is an easy example of where the destructor can be deleted.

> It seems the proper solution to this problem is that the debug dispose level 3 code should not flag missing destructors (or members not disposed) unless the class implements `qx.core.IDisposable`. Would you agree, @johnspackman?

Yes - IMHO it's a bug that it does not do that at the moment

We could go a bit further and say that it is deprecated to have a destructor unless the class implements `qx.core.IDisposable` (and issue a warning) ... and in the next major version, throw an exception

derrell:
@goldim Would you like to take a stab at implementing that change? If not, I can look into it, although that may not be immediate.

In any case, I think this whole newly added destructor  method can be reverted. It's not necessary.

derrell:
For reference, when implementing this change:

```
find source/class -type f -exec grep --color=auto -nH --null -e 'Missing destruct definition' \{\} +
source/class/qx/bom/rest/Resource.js1031:                    "Missing destruct definition for '" +
source/class/qx/bom/rest/Resource.js1047:                    "Missing destruct definition for '" +
source/class/qx/core/Object.js381:                  "Missing destruct definition for '" +
```

goldim:
@derrell I will check it out in spare time, not easy task i think. Also will close this PR and create another issue in which assign myself to solve it. If cant handle it assign it to you, ok?

derrell:
@goldim That's perfect.

I pointed you to the places that may require change. Also, there is a static method `qx.Class.hasOwnInterface` that can be used to determine if the class implements `qx.core.IDisposable`.

```
    /**
     * Whether a given class directly includes an interface.
     *
     * This function will only return "true" if the interface was defined
     * in the class declaration ({@link qx.Class#define}) using the "implement"
     * key.
     *
     * @param clazz {Class} class or instance to check
     * @param iface {Interface} the interface to check for
     * @return {Boolean} whether the class includes the mixin directly.
     */
    hasOwnInterface(clazz, iface)
```

PR #10568, opened 2023-03-29
Obtain rendered row height to calculate row positions

### LINKED ISSUE #10505: Strange table cell boarder behaviour during zooming in Firefox
**Describe the bug**
In Mozilla Firefox after zooming (scaling) in table you can see border gaps which overlap adjacent cells.

**To Reproduce**
1. Go to Widget Browser
2. Click on Table
3. Start Zooming in (till 125% for example)

**Expected behavior**
The same like in chrome no border gaps should be visible

**Screenshots**
image
Peek 2022-12-15 12-13

 - OS: [Windows, Debian]
 - Browser [firefox]
 - Version [102.6.0esr 64 bit]
 
 **Additional context**
Noticed by reader of my article about qooxdoo.

rad-pat:
An attempt to fix https://github.com/qooxdoo/qooxdoo/issues/10505 as this has been driving me nuts on a 4k monitor with Windows 11 and scaling set to 150%
This seems to fix the issue, but I don't know if there's a better way of doing this but I'm happy to fix it up now I know it can be fixed.
Also, for the FocusIndicator.js, is it correct that the `userHeight` and `userTop` values are calculated using some left and right decoration widths?

### DIFF
--- a/source/class/qx/ui/table/pane/FocusIndicator.js
+++ b/source/class/qx/ui/table/pane/FocusIndicator.js
@@ -124,9 +124,9 @@ qx.Class.define("qx.ui.table.pane.FocusIndicator", {
               wl = deco.getWidthLeft();
             }
           }
-          var userHeight = rowHeight + (wl + wr - 2);
+          var userHeight = rowHeight + (wt + wb - 2);
           var renderedRowHeight = this.__scroller.getTablePane().getRenderedRowHeight();
-          var userTop = Math.floor((row - firstRow) * renderedRowHeight) - (wr - 1);
+          var userTop = Math.floor((row - firstRow) * renderedRowHeight) - (wt - 1);
           if (
             editing &&
             this.__scroller.getMinCellEditHeight() &&
@@ -140,9 +140,9 @@ qx.Class.define("qx.ui.table.pane.FocusIndicator", {
           }
 
           this.setUserBounds(
-            paneModel.getColumnLeft(col) - (wt - 1),
+            paneModel.getColumnLeft(col) - (wl - 1),
             userTop,
-            columnModel.getColumnWidth(col) + (wt + wb - 3),
+            columnModel.getColumnWidth(col) + (wl + wr - 3),
             userHeight
           );
 
--- a/source/class/qx/ui/table/pane/FocusIndicator.js
+++ b/source/class/qx/ui/table/pane/FocusIndicator.js
@@ -125,7 +125,8 @@ qx.Class.define("qx.ui.table.pane.FocusIndicator", {
             }
           }
           var userHeight = rowHeight + (wl + wr - 2);
-          var userTop = (row - firstRow) * rowHeight - (wr - 1);
+          var renderedRowHeight = this.__scroller.getTablePane().getRenderedRowHeight();
+          var userTop = Math.floor((row - firstRow) * renderedRowHeight) - (wr - 1);
           if (
             editing &&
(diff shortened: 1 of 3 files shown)


source/class/qx/ui/table/pane/FocusIndicator.js
```diff
@@ -125,7 +125,8 @@ qx.Class.define("qx.ui.table.pane.FocusIndicator", {
             }
           }
           var userHeight = rowHeight + (wl + wr - 2);
-          var userTop = (row - firstRow) * rowHeight - (wr - 1);
+          var renderedRowHeight = this.__scroller.getTablePane().getRenderedRowHeight();
+          var userTop = Math.floor((row - firstRow) * renderedRowHeight) - (wr - 1);
```
derrell, line 129: It looks like you are correct that the original code, and now your new code too, uses decoration widths from left and right rather than top and bottom as they should. To confirm that, maybe try modifying the decorator to have drastically different left and right widths than the top and bottom widths, and see how it looks. My suspicion is that it has worked to date only because those widths are all the same.

derrell (changes requested):

source/class/qx/ui/table/pane/FocusIndicator.js
rad-pat: So, as suspected this did not work well when the widths were changed in the decorator. I have added a commit amending the errors.

zaucker:
> Also, for the FocusIndicator.js, is it correct that the `userHeight` and `userTop` values are calculated using some left and right decoration widths?

Cool catch!

rad-pat:
It appears I've removed review equests from everyone else by re-requesting a review from @derrell, apologies.

derrell (approved):
Thanks!

derrell:
I fixed the reviewer list.

goldim (changes requested):
I tried out in chrome based browser. Do next:
1. Open Widget Browser with table example
2. Set scale to 150%
3. Select first line
4. Scroll down to for example 50-100 line and select
5. Repeat 4.
See, when you click on line, the next 4-5 line below will be selected.

rad-pat:
@goldim, thanks for the test. I think I see the problem there.

rad-pat:
It seems this is going to be more complicated. Whilst the last code change does seem to rectify the selection when scrolled, it does make me wonder about use of rowheight in scrollbar calculations. In my development case, zoomed in at 150%, the table specified rowheight is 20px, the "row" div is given a height of 19px and a border-bottom of 1px solid.  As can be seen from the snip, the rowheight is rendered as 19.67, is there a way to ensure the rendered height of this div is 20? 
image

goldim:
@rad-pat looks like you fixed it with last commit. The problem is very complicated at least for me bc I abandoned  it. Maybe I forgot to point in the issue that there's a border size which changes by some rule during zooming. The rule is unknown and depends on a browser implementation. There is still a little gap seen but for PR I think alright I will approve it. I decided to check how resolved this problem in other apps for example google sheets (online excel) and notice there is no usage of border (I think another way of implementation - div inside another div and diff is a border). But in qooxdoo to apply this approach is huge work and there are many concepts depends on border css property like decorations.

goldim (approved):

hkollmann (approved):

oetiker (approved):

level420 (approved):

hkollmann:
@rad-pat Could you update your branch please?

rad-pat:
Branch updated

PR #10575, opened 2023-04-04
Added flag for static methods to be shown in api viewer

goldim:
Not sure if it is done for static variables too.

### DIFF
--- a/source/class/qx/tool/compiler/ClassFile.js
+++ b/source/class/qx/tool/compiler/ClassFile.js
@@ -884,6 +884,9 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
             node.value.type === "MemberExpression"
           ) {
             meta.type = "function";
+            if (sectionName === "statics"){
+              meta.isStatic = true;
+            }
           } else {
             meta.type = "variable";
           }


johnspackman:
Do we need this flag?  The list of statics are output separately to the members so the API viewer can already distinguish between the two.  Doing it that way is essential because you can have a static method and a member method with the same name (it would be bad practice IMHO but it's allowed).

goldim:
@johnspackman in API viewer checked against this flag for static methods but this flag is absent for static methods in meta right now. I added the `isStatic` flag to each static method there in API viewer to work around the problem.

derrell (approved):

oetiker (approved):

johnspackman:
I see what you mean now about isStatic in https://github.com/qooxdoo/qxl.apiviewer/blob/master/source/class/qxl/apiviewer/dao/Class.js#L129

I prefer your original version - because the API viewer is already mapping that data into the meta data, which is what *it* needs rather than the compiler needs (ie from the compiler's point of view the data is already correct and sufficient for its purpose, and it feels a little wrong to modify the output structure because it's slightly more convenient for an app to consume); this is also a duplication of information in the output, and we can't resolve it by merging the list of methods into one list to remove that duplication.

hkollmann (approved):

goldim:
@johnspackman I agree excessive info is bad then I just close this PR?

PR #10576, opened 2023-04-04
Sometimes getWindow is absent what leads call of this method to an error

goldim:
The error is detected in API viewer and described in https://github.com/qooxdoo/qxl.apiviewer/issues/54 .

### DIFF
--- a/source/class/qx/event/handler/GestureCore.js
+++ b/source/class/qx/event/handler/GestureCore.js
@@ -472,7 +472,7 @@ qx.Bootstrap.define("qx.event.handler.GestureCore", {
       if (
         (Math.abs(deltaY) < 1 && Math.abs(deltaX) < 1) ||
         this.__stopMomentum[oldTimeoutId] ||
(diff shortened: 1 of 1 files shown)


oetiker (approved):

derrell (changes requested):
I wonder if this is covering up the problem rather than fixing it? When is `getWindow` ever defined? It's not used anyplace else in this file, and this isn't a mixin...??? Is a better solution to delete that entire line 475?

goldim:
@derrell as I get it is derived from JS Object. Sometimes it has this method sometimes doesn't I don't know why. Removing this line the same as my edit but I think more risky bc I don't know what is behind this change will happen.

derrell (approved):

hkollmann (approved):

johnspackman (approved):

cboulanger (approved):

level420 (approved):

Issue #10410, opened 2022-06-06
Properties rewrite proposal


johnspackman:
This is a proposal for a complete rewrite of the properties mechanism in Qooxdoo classes, it is 100% BC and adds a number of really useful features; a big motivation for this complete rewrite is that the previous system relied on code generation, which was great for IE6 but in modern JS engines this increases memory usage and circumvents JIT compiler technology.  Worse of all, the code generation code has grown to be virtually unmaintainable.

The second big motivation is to improve the ease of use, both when defining properties (simpler specification boilerplate) and using them (native properties), and to add better type checking and control over storage and other semantics.

Finally, this is to allow weak references to be built into Qooxdoo in a way that is transparent and can be managed automatically with garbage collection.

New Property Features:
* Native properties
* Properties are first-class objects
* Storage can be completely replaced
* Eliminate the need for psuedo-properties
* Eliminate the need for property sniffing/detection
* Support for private and protected properties
* More advanced, and reusable, type checking
* Immutability and mutation detection
* Fast property definition
* Integration with References

References are included in this PR because property support for References is integral to making References transparent and useful.  Reference features are:
* Hard, Weak, Soft, and On Demand
* Transparent, asynchronous, object destructors and garbage collection

## Native properties
Add native property getters/setters so that it is possible to use `myObject.someValue` instead of `myObject.getSomeValue()` / `myObject.setSomeValue(…)`.  

### Compiler Changes
This should also work for arrays backed by `qx.data.Array`, so that `myObject.getMyArray().getItem(0)` becomes `myObject.myArray[0]` but this requires compiler changes to translate all array references into a function, eg `qx.data.Array.get(myObject.myArray, 0)` which tests for the type of `myObject.myArray`

## Eliminate Psuedo Properties
There are lots of places where objects have to define something that looks like a property, but does not list it in the `properties` section.  The most common example is usually because the property stores an instance of `qx.data.Array`, and when you call `object.setArrayProperty()` with a new value, the widget only wants to change the contents of the array and does *not* want to change the actual instance of the array.

For example, see [qx.ui.form.MModelSelection](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/ui/form/MModelSelection.js) and the `modelSelection` property.

[...]

PR #10592, opened 2023-05-13
8 0 beta

derrell:
New class/property system


johnspackman:
@derrell I'm getting this message when compiling:
```
Writing application www.spar.co.uk/demoapp
ERROR: could not locate require()d module: "jsdoctypeparser"
  required from:
    qx.Class:1294
```

Is the browserify extension is trying to include the jsdoctypeparser from my project?

derrell:
That has to do with the future `check` feature to check against a JSDoc string. The code is not executed (it's inside of `if (FUTURE)` which is `false`), but the compiler picks it up anyway and complains if that module isn't found. I'm not sure the best way to handle it, while leaving the code intact. Maybe I could just comment it out but leave it there, but I don't much like that option either... But the message you're seeing will return later (when that code is in fact executed), unless we can figure out how to get `require` to use qooxdoo's node_modules directory, or we'll have to require each user application to `npm install jsdoctypeparser`.

hkollmann:
During compile the modules from compiler should be used. have you added jsdoctypeparser to qooxdoo package.json?

johnspackman:
> The code is not executed (it's inside of if (FUTURE) which is false), but the compiler picks it up anyway and complains if that module isn't found

I think that if you do that as an environment check it should eliminate the code, and then the compiler's browserify won't see the `require`.  Later

> During compile the modules from compiler should be used. have you added jsdoctypeparser to qooxdoo package.json?

That would mean that in order to use the compiler, you have to initialise a package.json, add an unrelated package, and `npm i` - currently you don't have to do any of that.

The compiler's browserify implementation could be modified to search inside the compiler's own node_modules

derrell:
jsdoctypeparser was in npm-shrinkwrap but had gotten lost in package.json at some point, probably with a merge conflict.

I replaced the `if FUTURE(` with an environment test.

There should be no need to `npm install jsdoctypeparser` in desktop apps as it should be found and browserified. In node apps, it'll be required to install that library in the app, per the current implementation, once it is  completed. For now, the code will never be executed due to the environment test. Figuring out the best way of dealing with it can be deferred until we implement that feature. Right now, the frame for JSDoc type checks has been constructed, but no walls, doors, or windows yet exist to complete it.

johnspackman:
i'm stuck on testing this branch because of an issue with Rhino; I'll try and sort out a work around for that, but in the mean time can you confirm that you implemented the feature where `setXxx` accepts a promise?

derrell:
> can you confirm that you implemented the feature where `setXxx` accepts a promise?

Yes. All preexisting tests pass, as do the extras I asked you to create for me, proving that promises are handled properly. It should be fully BC in that regard.

johnspackman:
> Yes

Bill; I'll try and get testing ASAP, but Rhino does not support Proxy so I have to sort out a work around and compile that with 7 and the UI in 8.x

hkollmann:
from classs.md:
t the moment, it is not possible to mix Qooxdoo classes and ECMAScript 2015 classes; this is planned for a future release. In any case, as shown below, the Qooxdoo class system is still much more powerful than the native classes.

Is this true any longer?

hkollmann (approved):

ChristianMayer:
Making Qx classes compatible with ES classes is a feature I'm really looking forward to :+1: 

A quick look at the code change makes me believe that you are sticking with `__foo` nomenclature for private stuff. Wouldn't it make more sense to also switch to the hash syntax `#foo` that ES is using and which is available in all major environments since some time? (-> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields )

derrell:
That's a really good idea, I think. It is unrelated to this PR, though. That would be a change to the compiler, whereas this PR touches the compiler only for the minor changes needed to adjust it for use with the new class/property system.

Your suggestion could be implemented in a backward-compatible fashion, i.e., without a major version bump, because currently qooxdoo class members and properties can not use the '#' prefix.

I recommend that you create a "new feature" issue, with your suggestion!

CHANGELOG.md
```diff
@@ -1,11 +1,27 @@
...
+# v8.0.0-beta
+
 ## Breaking changes
 
-- (>=v8.0.0) Setting model data for a `qx.ui.table.Table` when the table is still editing will
+- Setting model data for a `qx.ui.table.Table` when the table is still editing will
 now raise an error as this could have lead to an invalid edit. To prevent any errors, ensure
 that the table edits are completed or cancelled before refreshing table model data.
 
+- Properties and members are now in the same namespace. Formerly, a class could have a member variable and a property with the same name, and there was no conflict. Now, since properties are first-class and can be manipulated as normal members, the properties and members use the same namespace, so a single name can not be defined in both.
```
oetiker, line 11: Tested one of my apps and promptly ran into this problem. My class `callbackery.ui.Busy` uses 'show' as a method which collides with the 'show' property of the `qx.ui.basic.Atom` class.

For catching such issues it would be extremely helpful if the compiler could flag them.


```javascript
qx.Class.define("callbackery.ui.Busy", {
    extend : qx.ui.basic.Atom,
    type : "singleton",

    construct : function() {
        this.base(arguments);
        this.set({
            center : true,
            show: 'both',
            iconPosition: 'top',
            visibility: 'excluded',
            zIndex: 10000001
        });
        this.getApplicationRoot().add(this,{top:0,bottom:0,left:0,right:0});
        this.__blocker = new qx.ui.core.Blocker(this.getApplicationRoot()).set({
            opacity: 0.7,
            color   : '#fff',
            keepBlockerActive: true
        });
        this.__blocker.getBlockerElement().setStyle('zIndex',10000000);
    },
    members : {
        __blocker : null,
        show: function(label){
            this.setVisibility('visible');
            this.setLabel(label);
            this.__blocker.block();
        },
        hide: function(){
            this.setVisibility('excluded');
            this.fadeOut(10);
            this.__blocker.unblock();
        }
    }
});
```
All I get now, when using my `Busy` class 

```javascript
let mybusy = callbackery.ui.Busy.getInstance();
mybusy.show('loading data');
```


is an error message in the browser console at runtime with the complaint:

```
Uncaught TypeError: mybusy.show is not a function
```

docs/core/defining_properties.md
```diff
@@ -122,21 +122,45 @@ when property values are being reset.
 
 There are two ways to set an init value of a property.
 
-#### Init value in declaration
+#### Init value, calculated at class instantiation time
+The _preferred_ way for reference vaues such as objects or arrays,
```
oetiker, line 126:
```suggestion
The _preferred_ way for reference values such as objects or arrays,
```

docs/core/defining_properties.md
```diff
@@ -122,21 +122,45 @@ when property values are being reset.
...
+  myOtherProperty : { initFunction : () => 42 }
 }
 ```
 
 #### Init value in constructor
 
-Alternatively, you could set the init value of the property in the constructor
+Before the introduction of the `initFunction` key in qooxdoo version
+8.0, it was necessary to initialize reference types in the
+constructor. The following remains as an available option, although it
```
oetiker, line 160:
```suggestion
constructor. This can still be done, although it
```

docs/core/defining_properties.md
```diff
@@ -122,21 +122,45 @@ when property values are being reset.
...
+should be rarely necessary any longer, for use in new code.
+
+You could set the init value of the property in the constructor
 of the class. This is only recommended for cases where a declaration of an init
 value as explained above is not sufficient.
```
oetiker, line 165:
```suggestion
```

docs/core/defining_properties.md
```diff
@@ -274,14 +302,15 @@ the multi-value idea behind the dynamic properties.
```
oetiker, line 302:
```suggestion
If you are using an `initFunction` as described above, any apply function will be called automatically.

```

docs/core/defining_properties.md
```diff
@@ -274,14 +302,15 @@ the multi-value idea behind the dynamic properties.
...
-(re-)definition: `init` and `refine`. `refine` is a simple boolean flag which
-must be configured to true.
+Derived classes can refine the init value of a property defined by
+their super class. This is however the only modification which is
+allowed through inheritance. To refine a property just define two keys
+inside the property (re-)definition: either `init` or `initFunction`;
+and `refine`. `refine` is a simple boolean flag which must be
+configured to true.
 
 Normally properties could not be overridden. This is the reason for the `refine`
```
oetiker, line 312:
```suggestion
Normally properties can not be overridden. This is the reason for the `refine`
```

docs/core/defining_properties.md
```diff
@@ -656,6 +686,110 @@ support inheritance may be a good orientation:
...
+  this[`$$variant_${prop}`] = variant;
+},
+```
+
+Note the `variant` handling. This pertains to internals of the Class
+implementation. The key point here is that when
+`this[`$$variant_${prop}`]` is not `immediate`, all of the property
+handling such as validation, transform, etc., may occur if
+`this[prop]` is changed. To ensure that no overhead of additional
+processing is incurred, the default storage implementation saves the
```
oetiker, line 764:
```suggestion
`this[prop]` is changed. To avoid any processing overhead, the default storage implementation saves the
```

docs/core/property_behavior.md
```diff
@@ -100,5 +100,5 @@ Because only in that scenario, one of the buttons need to be disabled.
 
 Currently, we don't plan to change it because it can have some hard to track side
 effects as seen in the example above and we don't have any deprecation strategy.
-Maybe it can be changed on a major version like 2.0 but currently there are no
+Maybe it can be changed on a major version, but currently there are no
 plans to do so.
```
oetiker, line 104: We are not aware of a way to change this behavior in a backward compatible way. So for now, it is up to you to be aware of the problem and work around it!

docs/core/understanding_properties.md
```diff
@@ -178,6 +178,29 @@ provide the following convenience methods:
 - `isPropertyName()`: Identical to `getPropertyName()`.
 - `togglePropertyName()`: Toggles between true and false.
 
+## First-class nature of properties
+Properties are first-class members of a class. They can be manipulated
+as if they were declared in the `member` section of the configuration
+-- they can be assigned to, or their value read as if it were a member
+variable -- but they are treated as if the functional setter or
+getter, were called. All operations desscribed below, that would occur
```
oetiker, line 186:
```suggestion
getter, were called. All operations described below, that would occur
```

docs/core/understanding_properties.md
```diff
@@ -342,21 +365,45 @@ when property values are being reset.
 
 There are two ways to set an init value of a property.
 
-#### Init value in declaration
+#### Init value, calculated at class instantiation time
+The _preferred_ way for reference vaues such as objects or arrays,
```
oetiker, line 369:
```suggestion
The _preferred_ way to initialize reference vaues such as objects or arrays,
```

docs/core/understanding_properties.md
```diff
@@ -342,21 +365,45 @@ when property values are being reset.
...
+that instance. You can use this key standalone or in combination with
+`nullable` and/or `inheritable`.
 
-The _preferred_ way for regular init values is to simply declare them by an
-`init` key in the property configuration map. You can use this key standalone or
-in combination with `nullable` and/or `inheritable`.
+```javascript
+properties : {
+  myProperty : {
+    initFunction : function() { return [ 1, 2, 4, 8 ]; }
```
oetiker, line 380:
```suggestion
    initFunction : () => { return [ 1, 2, 4, 8 ]; }
```

source/class/qx/Bootstrap.js
```diff
@@ -15,723 +16,505 @@
...
-      return typeof this == "undefined";
-    };
-
-    if (!config) {
-      config = { statics: {} };
-    }
+      // Process environment
+      let environment = config.environment || {};
+      for (let key in environment) {
+        qx["core"]["Environment"].add(key, environment[key]);
```
oetiker, line 91:
```suggestion
        qx.core.Environment.add(key, environment[key]);
```

source/class/qx/Bootstrap.js
```diff
@@ -15,723 +16,505 @@
...
+        }
 
-    qx.Bootstrap.setDisplayNames(config.statics, name);
+        implicitType = true;
+        config.type = "static";
+      }
 
-    if (config.members || config.extend) {
-      qx.Bootstrap.setDisplayNames(config.members, name + ".prototype");
+      if (qx["core"]["Environment"].get("qx.debug")) {
```
oetiker, line 116:
```suggestion
      if (qx.core.Environment.get("qx.debug")) {
```

docs/core/defining_properties.md
```diff
@@ -169,6 +193,10 @@ qx.Class.define("qx.MyClass", {
 
 ### Applying an init value
 
+The following applies if using the methodology describe immediately
+above, in 'Init value in constructor'. It is not necessary when using
+the `initFunction` key.
+
```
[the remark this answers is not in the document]
derrell: I think this is an important distinction, so I'd prefer to leave this here.

docs/core/property_behavior.md
derrell: added in later commit

source/class/qx/Bootstrap.js
derrell: We can't do that in qx.Bootstrap because it would bring in qx.core.Environment before it's ready. I originally intended to merge Bootstrap and Class, and thus Class had a bunch of these too. I did correct those recently, but the ones in Bootstrap have to remain as is.

docs/core/defining_properties.md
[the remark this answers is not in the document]
oetiker: I agree, but if someone reads this at this point they have no idea what you are talking about ...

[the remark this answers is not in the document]
derrell: lol. You and I are coming at it from different directions. My viewpoint is based on my research for my dissertation, where I learned that People Don't Read. I was evaluating how people treat error messages generated by a compiler. Even with short error messages, if there are three parts to the error (what happened, where it occurred, what you might do to fix it), people tended to read only the beginning. To have students learn from the error message and be able to correct the problem, I had to switch around the order (what you might do to fix it, where it occurred, what happened). In that way, students actually learned what they could do to correct their programs.

Anyway, that's a long story to reflect why I put important stuff out front. At least it tells them that they may not care about what's upcoming, even though they don't yet know what's upcoming. With text only appended to the end, saying they don't need to do that stuff, they're unlikely to read that far. I did, however, leave in the text you added at the end, so now they get two chances to learn that they don't need to do the stuff described in that section.

CHANGELOG.md
```diff
@@ -1,11 +1,27 @@
...
+
+- Refining a property in a subclass used to modify it in place. It now adds it to the subclass' prototype, so it ends up in the prototype chain twice.
+
+- The predefined instance.name variable is no longer predefined because, with first-class properties, it conflicts with the commonly used property name name. Use instance.classname instead.
```
goldim, line 15:
```suggestion
- The predefined instance.name variable is no longer predefined because, with first-class properties, it conflicts with the commonly used property name "name". Use `instance.classname` instead.
```

docs/core/classes.md
```diff
@@ -411,3 +411,98 @@ qx.Class.define("qx.test.Cat", {
...
+instance[2] = 42;
+value = instance[2];
+delete instance[2];
+```
+
+To allow this, the `delegate` key of the class configuration can be
+provided. The `delegate` key should have, as its value, an object with
+two or three keys. At a minimum, the `get` and `set` keys must be
+provided. Optionally, a `delete` key can also be provided. Each is a
+function that replaces the default operation on the objec. `get`
```
goldim, line 440:
```suggestion
function that replaces the default operation on the object. `get`
```

docs/core/defining_properties.md
```diff
@@ -656,6 +683,109 @@ support inheritance may be a good orientation:
...
+}
+```
+
+### set
+A storage implementation's `set` key defines how to store a value for the property in its storage. The default storage implementation stores the value within the instance object, in a property of the given name:
+
+```
+set(prop, value)
+{
+  let             variant = this[`$$variant_${prop}`];
```
goldim, line 747:
```suggestion
  let variant = this[`$$variant_${prop}`];
```

docs/core/defining_properties.md
```diff
@@ -656,6 +683,109 @@ support inheritance may be a good orientation:
...
+
+### init
+
+A storage implementation's `init` key defines how to initialize the
+property's value in its storage. The default storage implementation
+uses the property's `init` value (or `undefined` if the property
+doesn't define an `init` key) and stores it into the instance. The
+default storage's `init` function looks something like this:
+
+```
```
goldim, line 725:
```suggestion
```javascript
```

docs/core/defining_properties.md
```diff
@@ -656,6 +683,109 @@ support inheritance may be a good orientation:
...
+      configurable : false,
+      enumerable   : false
+    });
+}
+```
+
+### set
+A storage implementation's `set` key defines how to store a value for the property in its storage. The default storage implementation stores the value within the instance object, in a property of the given name:
+
+```
```
goldim, line 744:
```suggestion
```javascript
```

docs/core/defining_properties.md
```diff
@@ -656,6 +683,109 @@ support inheritance may be a good orientation:
...
+the value in its storage location, and then restores the variant.
+
+### get
+
+A storage implementation's `get` key defines how to retrieve the
+property's value from its storage. The default storage implementation
+simply retrieves the instance object's value of the given property
+name:
+
+```
```
goldim, line 771:
```suggestion
```javascript
```

docs/core/defining_properties.md
```diff
@@ -656,6 +683,109 @@ support inheritance may be a good orientation:
...
+{
+  return this[prop];
+}
+```
+
+### dereference
+
+If the property configuration includes `dereference : true`, then the storage implementation's `dereference` function is called just before the instance's destructor. The default storage implementation deletes the property from the instance:
+
+```
```
goldim, line 782:
```suggestion
```javascript
```

docs/core/defining_properties.md
```diff
@@ -729,11 +859,49 @@ obj.setPadding( 10, 20 );
...
+property to manipulate it, rather than manipulating it through the
+instance object.
+
+The property descriptor for a property can be obtained, through the
+instance object, by calling the `getPropertyDescriptor` method, and
+passing as an argument, the name of the property for which the
+property descriptor is desired. For example, to obtain the property
+descriptor for the property `myProp`:
+
+```
```
goldim, line 895:
```suggestion
```javascript
```

docs/core/defining_properties.md
```diff
@@ -729,11 +859,49 @@ obj.setPadding( 10, 20 );
...
+let propDesc = myClassInstance.getPropertyDescriptor("myProp");
+```
+
+With that property descriptor in hand, you can now manipulate the
+property with it, e.g.:
+
+```
```
goldim, line 902:
```suggestion
```javascript
```

docs/core/understanding_properties.md
```diff
@@ -178,6 +178,29 @@ provide the following convenience methods:
...
+Properties are first-class members of a class. They can be manipulated
+as if they were declared in the `member` section of the configuration
+-- they can be assigned to, or their value read as if it were a member
+variable -- but they are treated as if the functional setter or
+getter, were called. All operations described below, that would occur
+as a result of calling the functional setter, are also called by
+assigning to the property as a variable.
+
+By way of example, for property `myProp`, these two getter statements are equivalent:
+
```
goldim, line 191: Same with these code snippet, add js lang

docs/release-notes-8.0.md
```diff
@@ -0,0 +1,159 @@
+# Qooxdoo Release Notes
+
+## Notable changes and new features in v8.0
+- Qooxdoo version 8.0 includes a nearly completely rewriten class and
```
goldim, line 4:
```suggestion
- Qooxdoo version 8.0 includes a nearly completely rewritten class and
```

docs/core/understanding_properties.md
derrell: fixed

source/class/qx/core/Object.js
```diff
@@ -476,35 +483,29 @@ qx.Class.define("qx.core.Object", {
...
-                this[storeInit[name]] =
-                  undefined;
-            }
-          }
-        }
-
-        clazz = clazz.superclass;
-      }
-    }
+    // // only of properties are available
```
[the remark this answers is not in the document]
derrell: fixed

goldim:
@derrell I don't like the name "showFeatures". Previous variant (show) was not clear and obvious and this one even more make harder to understand what for this prop is. Looking at documentation there is said that this prop make CHILDREN VISIBLE then maybe `visibleChildren` or `visibleParts` something like that? The getter `getShow` looks strange but `getVisibleChildren` looks nice imho.

goldim:
@derrell There is a plenty commented code and code snippets without language set. Could you fix them?

source/class/qx/util/PropertyUtil.js
```diff
@@ -187,8 +199,8 @@ qx.Class.define("qx.util.PropertyUtil", {
      * @param value {var} The value to set
      */
     setThemed(object, propertyName, value) {
-      var styler = qx.core.Property.$$method.setThemed;
-      object[styler[propertyName]](value);
+      let             propertyFirstUp = qx.Bootstrap.firstUp(propertyName);
```
goldim, line 202: why space between two words so long?

source/class/qx/util/PropertyUtil.js
```diff
@@ -187,8 +199,8 @@ qx.Class.define("qx.util.PropertyUtil", {
...
+      object[`setThemed${propertyFirstUp}`](value);
```
goldim, line 203: same

source/class/qx/util/PropertyUtil.js
```diff
@@ -198,8 +210,8 @@ qx.Class.define("qx.util.PropertyUtil", {
      * @param propertyName {String} The name of the property
      */
     resetThemed(object, propertyName) {
-      var unstyler = qx.core.Property.$$method.resetThemed;
-      object[unstyler[propertyName]]();
+      let             propertyFirstUp = qx.Bootstrap.firstUp(propertyName);
```
goldim, line 213: same long space

source/class/qx/test/event/GlobalError.js
```diff
@@ -198,7 +198,7 @@ qx.Class.define("qx.test.event.GlobalError", {
       window.onerror = function (msg, uri, lineNumber) {
         wasNativeHandled = true;
 
-        self.assertMatch(msg, "Doofer Fehler");
+        self.assertMatch(msg, "Doofer Fehler 2");
```
goldim, line 201: I don't understand this change

source/class/qx/data/SingleValueBinding.js
```diff
@@ -857,6 +857,7 @@ qx.Class.define("qx.data.SingleValueBinding", {
      */
     __setTargetValue(targetObject, targetPropertyChain, value) {
       // get the last target object of the chain
+      var result;
```
goldim, line 860: what's point to declare variable far from where it is used?

source/class/qx/core/propertystorage/Default.js
```diff
@@ -0,0 +1,72 @@
...
+      },
+
+      get(prop)
+      {
+        return this[prop];
+      },
+
+      set(prop, value)
+      {
+        let             variant = this[`$$variant_${prop}`];
```
goldim, line 57: long space

source/class/qx/test/event/GlobalError.js
derrell: It was the same message as used in a different test, and I needed to distinguish between the two when one of them was failing.

source/class/qx/data/SingleValueBinding.js
derrell: I much prefer all variables declared at the top of the block to which they apply, so that it's not necessary to search for a declaration; you always know where to look. This is particularly true with `var` (vs. `let`), where I prefer all variables declared at the top of the function which ends up happening by "hoisting" anyway.

source/class/qx/util/PropertyUtil.js
derrell: My personal style is to align all variable names in declarations at the top of the function, indented so that the length of the "type" of the variable doesn't affect the indentation of all of the variable names. Granted in JS the "type" doesn't exist, and either `var` or `let` consumes the same amount of space in all cases. It's just habit. I removed the extra indentation in the places you identified.

derrell:
> @derrell There is a plenty commented code and code snippets without language set. Could you fix them? Or should I?

Likely, I want to keep some of them there, so please don't simply "fix" them. If there is code commented out that doesn't indicate why it's commented out, let me know and I'll decide whether to better comment why it's retained, or remove it.

derrell:
> @derrell I don't like the name "showFeatures". Previous variant (show) was not clear and obvious and this one even more make harder to understand what for this prop is. Looking at documentation there is said that this prop make CHILDREN VISIBLE then maybe `visibleChildren` or `visibleParts` something like that? The getter `getShow` looks strange but `getVisibleChildren` looks nice imho.

The purpose of that property is to specify whether to display the label, the icon, or both. The label and icon are "features" of the widget. `showChildren` or `showVisibleChildren` is no more clear to me. If we really want to identify what it is in the name, we need a much longer property name, e.g., `showLabelIconBoth` and in any case, one needs to look at the documentation to see that there are three options allowed for the property: "label", "icon", "both". I'm not wedded to `showFeatures` but it seems to me at least as good as `showVisibleChildren` and the decision is just a personal preference. If one didn't need to look at the documentation to know how to use it, I'd agree that the property name should tell you all you need, but I'm not seeing that that's the case here...

derrell:
> > @derrell There is a plenty commented code and code snippets without language set. Could you fix them? Or should I?
> 
> Likely, I want to keep some of them there, so please don't simply "fix" them. If there is code commented out that doesn't indicate why it's commented out, let me know and I'll decide whether to better comment why it's retained, or remove it.

Ok, I've gone through and either removed commented code that is obsolete and need not be retained; or ensured that comments indicate why still-retained commented code is there. Most are to demonstrate what used to happen in cases of BC breaks.

source/class/qx/core/propertystorage/ImmutableDataArray.js
```diff
@@ -0,0 +1,51 @@
...
+        return;
+      }
+
+      // Otherwise, they're providing a new value. Instead of
+      // replacing the qx.data.Array itself, we replace the contents
+      // of the existing qx.data.Array.
+      //
+      // FIXME: Is this the correct way to replace the contents of a
+      // data array???
+      this[prop].replace(value);
```
goldim, line 46: Looking at API Viewer it is the correct and "efficient" way for `qx.data.Array`

derrell: Updated docu. Thanks.

cboulanger (approved):
Exciting! I hope I will have the opportunity to test it in the near future. Until then- as it passed the tests, I trust it's ready as it is for a breaking release

derrell:
Yes, deep debugging requires a bit of getting used to, but once you get used to it, you learn to quickly run-until-function-exit on some of the calls. It's not that each is wrapped in a function; rather, the proxy wraps the entire object. Every access to the object runs through the proxy. That includes access to methods, so first, you'll see a `get` of the method itself, and then you'll see the method invocation. That fist `get` of the method is one that you learn to ignore and immediately step out of. As I worked with this system, I found it far, far easier to debug properties than the original generated code, once I got used to working with it and learned to quickly step past the non-property accesses.

derrell:
BTW, the vast majority of those object properties already existed in the old property system, but you typically had no reason to notice them. I added a few as necessary features of the new property system.

johnspackman:
> Yes, deep debugging requires a bit of getting used to, but once you get used to it, you learn to quickly run-until-function-exit on some of the calls

Every other debugging experience I can think of just steps into the function.  It looks really odd, and I can't imagine new users taking to it very easily.

> BTW, the vast majority of those object properties already existed in the old property system

I don't remember seeing them; they were presumably in the prototype or in a closure before?  My debugger now shows 210 properties when there were approx 30 before.

IMHO that is confusing and makes it difficult to see whats going on.

johnspackman:
also there seems to be a problem, maybe connected to the function being an async function - I cannot step into my function, the `value` variable is the function, not the result of the function:

image

johnspackman (changes requested):
This PR has quite an impact on debuggability - every single method call (not just properties) now goes via the proxy handler code; IMHO that's a bit confusing and distracting.  Also, every property now causes 7 fields to be added to each object, which adds pages of irrelevant details to an object.

AIUI `Object.defineProperty` can be used to achieve the desired effect (ie native property values) without using `Proxy`, meaning that debugging is much simpler.

Here's a proof of concept that native properties can be added to the existing v7 `qx.Class` classes: 

https://tinyurl.com/2gk2jpvy

johnspackman (changes requested):
Pseudo properties need to be supported also, eg `qx.ui.form.List` has a `selection` property but it only appears as `getSelection` and not as `selection`:

image

derrell:
I don't know what a pseudo property is. Something is either defined in the `properties` configuration, or it is not. There are a few places that kind of simulate properties, but I'm not sure what you'd have the code do here. *Guess* that when there is a `getX` method, that means there should be an `x` property automatically added? That doesn't work if the getter method is actually retrieving the value from someplace else, e.g., some sub-object as is commonly done. I don't think I'm comfortable with guessing...

johnspackman:
> I don't know what a pseudo property is. 

we discussed it here: https://github.com/qooxdoo/qooxdoo/issues/10410

> There are a few places that kind of simulate properties

they do this because the property system did not allow custom storage

> I don't think I'm comfortable with guessing...

It's what qooxdoo does and how binding works.  It sniffs, basically deciding that if there is a get, set, reset, and change event, it is a property.

derrell:
Ok, I've added support for pseudo-properties. I excluded the test for the `reset` method, as it seems it's not required: the example of MModelSelection's `selected` pseudo-property doesn't have it, so I'm assuming it's not a required component of pseudo-properties.

I also added tests for it, as there apparently weren't any previously.

source/class/qx/Class.js
```diff
@@ -305,1400 +657,2442 @@ qx.Bootstrap.define("qx.Class", {
...
+            value,
+            qx.Interface.getByName(property.check),
+            "Expected value to implement " + property.check
+          );
+        } else {
+          if (qx.core.Environment.get("qx.Class.futureCheckJsDoc")) {
+            // Next  try to parse the check string as JSDoc
+            let bJSDocParsed = false;
+            try {
+              const { parse } = require("jsdoctypeparser");
```
johnspackman, line 1409: I appreciate that this is currently switched off, but IMHO we should not plan on including require in browser based builds because it forces the user to use npm and have a package; years ago, one of the 1&1 team added grunt and it was very frustrating to have to create a package.json etc just to be able to compile.

CHANGELOG.md
johnspackman, line 11: Technically I don't think they are first class, just native properties - eg a function is called "first class" because it is an object of type `Function` which has properties, as well as having native functionality as a function

docs/core/classes.md
```diff
@@ -411,3 +411,98 @@ qx.Class.define("qx.test.Cat", {
...
+`instance.setItem(2, 42);`; to retrieve it with `value =
+instance.getItem(2);`; and to delete it with `instance.removeAt(2);`.
+
+For usability, the developer might like for those statements to be
+accessed using the array index operator, as such:
+
+```
+instance[2] = 42;
+value = instance[2];
+delete instance[2];
```
johnspackman, line 433: please can you explain how `delete` is implemented?  EG does this mean that you could have two objects of the same class, but one has `undefined` for the property because it has been deleted but the other does not, even if the definition of the property says that an uninitialised value is impossible?

source/class/qx/data/Array.js
```diff
@@ -31,6 +31,51 @@ qx.Class.define("qx.data.Array", {
...
+        // ... then use the method for modifying an item
+        this.setItem(property, value);
+      } else {
+        // otherwise, use the default action
+        this[property] = value;
+      }
+    },
+
+    delete(property) {
+      // If the property is a number or string representing a number...
```
johnspackman, line 68: `delete dataArray[2]` is not the same as `dataArray.removeAt(2)` because `delete` will simply create a hole in the array where as `removeAt` will shuffle the elements down.  Does qx.data.Array support sparse arrays?

johnspackman (changes requested):

source/class/qx/Class.js
derrell: Remember that `require` in a browser-based app includes the browserified code. There *should* be o need for the user to use npm and have a package; the compiler should be browserifying that code from qooxdoo's node_modules.

That said, I don't have any strong need for this feature. We could strip it out completely. We could, later, implement it with our own code to parse JSDoc comments. Or whatever. Let me know if you'd like for me to entirely remove that code.

CHANGELOG.md
derrell: I was using the terminology you used in https://github.com/qooxdoo/qooxdoo/issues/10410, but I just want it to be clear what the new feature does. I don't really care what term we use. Would you prefer that I change the docu to refer to them as native, rather than first-class?

docs/core/classes.md
derrell: Calling `delete instance[2]` is identical to calling `instance.removeAt(2)` because the `delete` is trapped and `removeAt` is actually called in its stead.

source/class/qx/data/Array.js
derrell: `delete dataArray[2]` in qx.data.Array is implemented as `dataArray.removeAt(2)`. I don't believe that qx.data.Array supports sparse arrays, so we probably don't want `dataArray.removeAt(2)` to leave an `undefined` hole. The implementation is trivially changeable, though. Do you have suggestions for a different implementation? (It's marked as experimental, so what we do now can be changed later.)

johnspackman: But it's different semantics - calling `delete myNativeArray[2]` is not the same as `myDataArray.removeAt(2)`

CHANGELOG.md
johnspackman: yes please; the terminolgoy in #10410 was referring to creating a `qx.Property` object that would describe the entire property (eg much like `Function` describes a function)

source/class/qx/data/Array.js
derrell: Understood. What would you have it do instead?

source/class/qx/Class.js
johnspackman: I think that the jsdoc parser would be useful, but at the moment browserify does not work as you're propising (although I think it would be good if it did).  I havn't specifically tested, but the compiler was complaining that it could not find the module

CHANGELOG.md
derrell: Done

source/class/qx/data/Array.js
johnspackman: throw an exception?

derrell: Ok. It now throws an exception by default. If environment variable `qx.data.Array.deleteAsRemoveAt` is `true` then it acts as I had implemented it originally.

source/class/qx/Class.js
derrell: > I think that the jsdoc parser would be useful, but at the moment browserify does not work as you're propising (although I think it would be good if it did). I havn't specifically tested, but the compiler was complaining that it could not find the module

I just pushed a fix to this. The compiler now looks in both the app's node_modules directory and qooxdoo's node_modules directory for browserifyable modules. This means that `jsdoctypeparser`, used by `Class.js` (when enabled by the environment variable, later), is found regardless of whether there is a node_modules directory in the user's app.

johnspackman:
@derrell would it be possible to progress this PR?  I appreciate you're busy, but it would be good to see it come to a conclusion

ChristianMayer:
There wasn't any action for nearly a year now. Is this still active?

johnspackman:
> There wasn't any action for nearly a year now. Is this still active?

Yes - I have a version which seems to be complete but I need to merge the current master into it so that I can do the tests; I'm just stupidly busy at work at the moment, sorry for the delay - I'm really keen to get it out

hkollmann:
@johnspackman : Are you still active this this?

patryk-m-malinowski:
Myself and @johnspackman are currently working on this. We will push out some more commit on this soon!

hkollmann:
@johnspackman @patryk-m-malinowski Any news?

johnspackman:
Yes - @patryk-m-malinowski has been back on it this week; properties are done, as are async properties, and a new implementation of bindings - Bindings are now a first class object which support sync and async properties.

I think this implementation will remove ambiguities in how async is handled wrt bindings and properties also.

Included is a rudimentary framework for load on demand and we'll be able to add garbage collection to properties in the future (ie to expire large object graphs, when we know that we can reload them on demand)

Hope to have something  that we use daily in pre production builds very soon, and to announce this properly to the group

johnspackman:
@patryk-m-malinowski has been working hard on this branch this week and it's now ready for early beta; the key features are:

- Properties are implemented as first class objects
- - async properties are supported
- - storage of data can be customised
- Bindings has been re-written
- - Bindings are first class objects, and can be `destroy()`ed
- - Support async properties

There are a lot of changes under the hood - we're doing some final prep now on our own code in preparation for integrating this branch into our long-term development branches.

There are only limited BC issues, mostly to do with issues around `init` values

hkollmann:
What's about the conflicts?

cboulanger (approved):
I trust you guys with this. Thanks so much for the gigantic work that was involved, @derrell, @johnspackman, @patryk-m-malinowski !

hkollmann:
@johnspackman @patryk-m-malinowski Is the documentation also adapted to the latest changes?

johnspackman:
> @johnspackman @patryk-m-malinowski Is the documentation also adapted to the latest changes?

good question; it should be but this has been a long time coming so docs will need a good review

hkollmann (approved):

derrell:
These recent comments are all attached to my v8.0 code PR that was rejected. @johnspackman Where is the code your team has been working on?

johnspackman:
> These recent comments are all attached to my v8.0 code PR that was rejected. @johnspackman Where is the code your team has been working on?

It was just pushing into the public branch and was based on your original work.  

If you check out the branch, you'll see the code.

Happy to restart with a fresh branch if you think that would be preferable

derrell (commented):
I can't "Request changes" on my own (original) pull request, but this should be assumed to be a "Request changes" response. due to the few issues I've flagged.

I spent a couple of hours reviewing significant portions of this more recent work. It looks like good work and a reasonable alternate implementation. I haven't tested any projects with it. I have indicated a few changes needed. My biggest concern is that it appears that the `apply` method is called in all cases of initialization now, when there were cases in v7 and prior where that was not the case. It could be *really* difficult to locate and alter code in old projects that depend on the prior behavior. I know well what a PITA it is to emulate the old behavior, from my earlier implementation, but I consider it absolutely essential that we maintain backward compatibility there even though this is a major version bump. It's just not possible for developers of projects using v7 and prior to figure out where they depend on the old behavior. We need to maintain that old behavior, i.e., all v7 tests must pass. (That may already be the case, in which case I stand corrected.)

source/class/qx/core/property/Property.js
```diff
@@ -0,0 +1,1252 @@
...
+      if (value === undefined) {
+        throw new Error(`${this}: init() called without a value`);
+      }
+
+      this.__storage.set(thisObj, this, value);
+      this._setMutating(thisObj, true);
+      thisObj["$$init_" + this.__propertyName] = value;
+
+      try {
+        this.__callFunction(thisObj, this.__apply, value, null, this.__propertyName);
```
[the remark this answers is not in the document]
patryk-m-malinowski: As far as I'm aware, the old implementation did not call `apply` when there was an `init` in the prop definition but did call `apply` when `deferredInit` was set to `true` and `this.initProperty` was called. Based on my observations, the new implementation keeps this old behaviour. Do you have an example use case where the new implementation behaves differently?

source/class/qx/core/property/PsuedoPropertyStorage.js
```diff
@@ -0,0 +1,92 @@
...
+
+  members: {
+    /** @type{String} capitalised name of the property */
+    __upname: null,
+
+    /** @type{Boolean} whether this is for async storage */
+    __isAsyncStorage: null,
+
+    /**
+     * @Override
```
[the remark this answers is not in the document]
johnspackman: I've always used @override for implementations defined by interfaces, IMHO this is the correct way to do it

johnspackman:
> I can't "Request changes" on my own (original) pull request, but this should be assumed to be a "Request changes" response. due to the few issues I've flagged.

Apologies, I didn't realise the consequences of adding onto your original branch, I was trying to keep the commit history going (especially as I did start with your branch)

I don't mind creating a new PR from my own branch (eg into a virgin v8 branch that @hkollmann talked about creating in gitter the other day) if that helps?  The huge number of commits in this branch probably make tracking individual commits a bit pointless now, so it would be an opportunity to squash...
 
> My biggest concern is that it appears that the `apply` method is called in all cases of initialization now, when there were cases in v7 and prior where that was not the case. It could be _really_ difficult to locate and alter code in old projects that depend on the prior behavior. I know well what a PITA it is to emulate the old behavior, from my earlier implementation, but I consider it absolutely essential that we maintain backward compatibility there even though this is a major version bump. 

I agree, except to note that there are a couple of issues which are arguably bugs in the original implementation, usually around the concept of init values.

EG issue https://github.com/qooxdoo/qooxdoo/issues/10704 lead to PR https://github.com/qooxdoo/qooxdoo/pull/10707 which was abandoned because it was impossible to fix.

> It's just not possible for developers of projects using v7 and prior to figure out where they depend on the old behavior. We need to maintain that old behavior, i.e., all v7 tests must pass. (That may already be the case, in which case I stand corrected.)

We've tried hard to be 100% BC - you wont be surprised to hear we also found it quite tricky to get a clear idea of what was happening vs supposed to happen! 🙄

The existing (ie v7) unit tests have "ok" coverage, and this code passed them all without changing the tests.  But I strongly suspect that those unit tests are not 100% coverage (probably demonstrated by the fact that issues like #10704 exist, or just how hard it was to figure out the last tiny details of how the damn thing worked)

source/class/qx/core/property/Property.js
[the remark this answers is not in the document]
patryk-m-malinowski: Hold on, just realized the new implementation calls init on all the properties when the object is constructed! That is a breaking change.

source/class/qx/core/property/PsuedoPropertyStorage.js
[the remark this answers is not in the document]
derrell: If the api viewer is able to show the Interface's documentation for the function when `@Override` is used, as with a true method override, I'm find with that. Otherwise, these functions should be properly documented here, at their definition.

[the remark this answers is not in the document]
johnspackman: The definition of the function is in the interface, not in each and every implementation of that method, surely?  It would just be duplication to repeat the jsdoc.  The purpose of `/** @override */` is to say "i am not documenting this method because it is defined in a parent class or interface"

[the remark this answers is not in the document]
derrell: Noted, and I agree. It's just not what was traditionally done. Traditionally, it was noted as a comment, e.g.,
```
// interface definition
```

The traditional way is actually better, in that it tells you not to bother looking for a super method that documents it, but rather, the interface. I guess the following would be the best, but may not be JSDoc conformant:

```
@override interface
```

I drop my objection to using `@override` for these.

derrell:
It just occurred to me that I saw where the change was made so that class instantiation is no longer proxied, but I didn't see the alternate implementation that allows:

```
this.color = "red";
```

act on the property `color` so that it is identical to calling:

```
this.setColor("red");
```

Where/how is that implemented?

patryk-m-malinowski:
> Where/how is that implemented?

https://github.com/qooxdoo/qooxdoo/blob/517cbeab1e01efbdbf786aa3bf1308fb0cb70d20/source/class/qx/core/property/Property.js#L417-L433

hkollmann:
@johnspackman @patryk-m-malinowski : Are you working on the open issues?

oetiker:
ping ?

patryk-m-malinowski:
Sorry for the long response. I have been testing this branch against our proprietary software and there were a lot of minor error edge cases which weren't caught by the unit tests which I've had to fix and write unit tests for. I have taken in considerations all the mentioned issues, and in cases where there will be BC breaking changes that won't be trivial to fix, I have added environment settings which can enable the old behaviour. 

I have also realised that there's been a significant decrease in performance following major changes to the property system and bindings which I'm gonna have to fix.

Once I get those issues sorted I will push out more commits.

source/class/qx/core/property/Property.js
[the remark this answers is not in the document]
patryk-m-malinowski: There will be an environment settings which allows one to toggle between the old and new behaviour.

hkollmann:
Any progress here?

patryk-m-malinowski:
Yes; my [fork](https://github.com/patryk-m-malinowski/qooxdoo/tree/8_0_beta) contains the most recent changes. Unfortunately I don't have the permission to merge them into here.

Backwards compatibility now seems to be working well as far as I'm concerned. All unit tests pass. I've also tested with some of our proprietary software and it works great.

Performance with properties has been improved also. It's still slightly slower than the old version, but I think it's adequate. If people notice that it's too slow then we can look at optimizing it more later.

The behaviour of automatically initting properties at construction is implemented but it's currently disabled because it causes bugs which are not trivial to fix. I think we should deprecate the old behaviour in this upcoming version, and slowly rewrite our code to make it compatible, and then drop the old behaviour in the next major version (9). This behaviour is controlled via the environment variable `qx.core.property.Property.applyDuringConstruct`. If that is set to true, you can also set `qx.core.property.Property.excludeAutoApply` to an array of patterns for class names of which properties you don't want auto-applied. This is useful if you want to slowly migrate your code but don't have time to do it all at once.

hkollmann:
So i can create a v8_beta from your fork?

derrell:
> The behaviour of automatically initting properties at construction is implemented but it's currently disabled because it causes bugs which are not trivial to fix. I think we should deprecate the old behaviour in this upcoming version, and slowly rewrite our code to make it compatible, and then drop the old behaviour in the next major version (9).

In my version of the new Property implementation, I spent a lot of time getting the init code to work exactly as previous even though the original behavior is somewhat convoluted, because it will create very hard-to-find bugs; it will be very, very difficult for people to know what to change in their apps when they suddenly fail to work properly with v8. It would require reviewing every single property to ascertain whether the cases of init/apply require changes. It is my feeling that releasing any new version (whether v8, v9, or some future version) with changes to initialization of properties, will force people to remain at old versions, so we'd either be forced to continue to maintain v7 indefinitely, or create much anger within the community.

I'm also concerned about the described slowdown of this Property implementation, but depending on how much slowdown, that could be less relevant. What is the difference in time between a build of qooxdoo and running the tests, using v7 vs using the new Property system (this branch)?

johnspackman:
> > The behaviour of automatically initting properties at construction is implemented but it's currently disabled because it causes bugs which are not trivial to fix. I think we should deprecate the old behaviour in this upcoming version, and slowly rewrite our code to make it compatible, and then drop the old behaviour in the next major version (9).
> 
> In my version of the new Property implementation, I spent a lot of time getting the init code to work exactly as previous even though the original behavior is somewhat convoluted, because it will create very hard-to-find bugs; it will be very, very difficult for people to know what to change in their apps when they suddenly fail to work properly with v8. It would require reviewing every single property to ascertain whether the cases of init/apply require changes. It is my feeling that releasing any new version (whether v8, v9, or some future version) with changes to initialization of properties, will force people to remain at old versions, so we'd either be forced to continue to maintain v7 indefinitely, or create much anger within the community.
> 
> I'm also concerned about the described slowdown of this Property implementation, but depending on how much slowdown, that could be less relevant. What is the difference in time between a build of qooxdoo and running the tests, using v7 vs using the new Property system (this branch)?

I think what Patryk's saying here is the same as the point you are making - that changing the behaviour of init could cause issues, and so the changes are protected by environment settings (and are disabled by default) to allow people to evaluate the alternative implementation, if they choose to.  

This is *not* making sweeping, backwards incompatible changes in init behaviour or anywhere else.

As @derrell found, the current init behaviour is counter-intuitive (and has lead to bug reports in the past) and its easy to see that if the curent behaviour was introduced today as new code, it could be considered a bug.  Replicating the current (v7) behaviour is a delicate and tricky thing to do.

It may be the case that the proposed change to init behaviours do cause significant, and hard to track down, bugs; personally, Im not convinced as to how likely/bad that would be, but the environment settings will allow us to find out if that is the case or not without impacting the existing user base.  

We have two big projects (one has 12 apps on node, browser, and rhino with just over 2,00 registered business users, the other has 9 apps on 3 websites with over 7,000 users) and the next minor version of both will be using 8.x beta and the new init code.

WRT to performance - this needs to be evaluated and optimised.  We are switching from a code-generation solution so the performance is bound to change, I havnt looked in detail myself but AIUI the new v8 performance is not massively far from the current v7

johnspackman:
> So i can create a v8_beta from your fork?

yes

oetiker:
@hkollmann I guess you could PR like this to update the branch 
https://github.com/qooxdoo/qooxdoo/compare/8_0_beta...patryk-m-malinowski:qooxdoo:8_0_beta

hkollmann:
I merged all changes from https://github.com/patryk-m-malinowski/qooxdoo/tree/8_0_beta into this branch!

hkollmann:
Could you please *document* the change of init in changelog.md?

hkollmann:
Any progress with the performance issue?

oetiker:
I fixed the merge conflicts to get the test running

patryk-m-malinowski:
> Any progress with the performance issue?

Like I mentioned before, performance has been improved. I think we should let other people test the release out and evaluate performance first and if it still is too slow, then we'll worry about further optimizing.

hkollmann:
So it's time to merge into v_8_0_beta and release a beta version?

goldim:
@patryk-m-malinowski  @johnspackman  I get an error "The constructor of class 'qxWeb' returned a different instance than 'this'. This is highly suspect, not recommended, and should be removed." It is just shown in console. No negative effect I see so far. Don't you know what is it?

goldim:
This code does work for qooxdoo v7 but in this PR it fails with the next error:
```
000420 : Assertion error! Binding property username of object qx.data.model.Username not possible. No event available.: Expected value not to be null but found null!
```
The code:
```javascript
qx.Class.define("test.Form", {
    extend: qx.ui.form.Form,
    
    construct(){
        super();
        this.__createUserNameField();
    },    

    members: {
        __createUserNameField(){
            const host = new qx.ui.form.TextField();
            this.add(host, "Username", null, "Username");
        }
    }
});

      let form = new test.Form();//qx.ui.form.Form();
      const controller = new qx.data.controller.Form(null, form);
      // const field = new qx.ui.form.PasswordField();
      // form.add(field, "test", null, "test");
      controller.createModel();
```
If I use qooxdoo form and add field outside of constructor everything is fine for both versions. Is it BC?
Update: actually I got why it doesn't work anymore. My property named with uppercase symbol in method `add`. That's why.

hkollmann:
fully integrated in branch_8_0_beta

goldim:
@hkollmann where now can I write issues about this prop system?

hkollmann:
Open issues with V8: .. and label v8?

goldim:
@hkollmann Ok, I get it. Thank you.

Issue #10596, opened 2023-05-27
Setting modal to false on window creation adds the "modal" state incorrectly.


lucapivato:
Create a window and set modal to false, then check the "modal" state, it returns true. This is caused by the lack if this.initModal() and the _applyModal() method adding the state incorrectly.

https://tinyurl.com/2gp5qob7

lucapivato:
Looks like window._applyActive() has the same mistake. 

```
      if (old) {
        this.removeState("active");
      } else {
        this.addState("active");
      }
```

Adds "active" when value is false and old is false, which happens without the init call. IMO it should be the opposite:

```
      if (value) {
        this.addState("active");
      } else {
        this.removeState("active");
      }
```

goldim:
Merged PR

PR #10597, opened 2023-05-27
Fixed active and modal state initialization.

lucapivato:
See #10596

### DIFF
--- a/source/class/qx/ui/window/Window.js
+++ b/source/class/qx/ui/window/Window.js
@@ -87,9 +87,11 @@ qx.Class.define("qx.ui.window.Window", {
     // Automatically add to application root.
     qx.core.Init.getApplication().getRoot().add(this);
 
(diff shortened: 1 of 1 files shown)


goldim (changes requested):
`initProperty` leads to `removeState` but there is no state. Not sure if anything wrong happens in this case. I think checking old value on null value will exclude this case.

lucapivato:
Calling removeState on a state that is not set is quite common. States are just fields of the __states map.

goldim (approved):

oetiker (approved):

derrell (approved):

cboulanger (approved):

hkollmann (approved):

level420 (approved):

hkollmann:
Thanks @lucapivato !

PR #10599, opened 2023-05-30
Method of Mobile Popup placeTo works

### LINKED ISSUE #9423: mobile popup.placeTo() doesn't work as expected
Hello,
the popup.placeTo() is not placing the popup in the specified coordinates.
Please check http://tinyurl.com/y9u2nvmz

Or use this code in playground:

```
var page = new qx.ui.mobile.page.NavigationPage();
page.setTitle("Show popup");
page.addListener("initialize", function() {
  var button = new qx.ui.mobile.form.Button("First Button");
  page.getContent().add(button);

  button.addListener("tap", function() {
    var cont = new qx.ui.mobile.container.Composite(new qx.ui.mobile.layout.VBox());
		cont.add(new qx.ui.mobile.basic.Label("<p><b>Helloooo</b></p>"));
		var popup = new qx.ui.mobile.dialog.Popup(cont);
		popup.placeTo(20,20);
		popup.show();
  }, this);
},this);

this.getManager().addDetail(page);

page.show();
```

goldim:
Fixed #9423  . 
Please check it carefully with attention. Some of code difficult to understand.
For example I don't get next code (what reason):
```javascript
// Move outside of viewport
this.placeTo(-1000, -1000);
```
I haven't noticed any bad side effects removing it. It generated extra `updatePosition`.
In docs said there is 3rd case of using popup - it is attaching to certain an absolute point but forgotten to process this case during `updatePosition` and it is was just centered with first case of using (see docs about class Popup).
And another thing: I remember that replaceTo was called and don't reset it. Should it be reset in some case?


source/class/qx/ui/mobile/dialog/Popup.js
```diff
@@ -239,9 +239,6 @@ qx.Class.define("qx.ui.mobile.dialog.Popup", {
 
         this.__registerEventListener();
 
-        // Move outside of viewport
-        this.placeTo(-1000, -1000);
-
```
oetiker, line 244: I think it is fine to remove this. I assume it was put there by someone who thought that the screen would potentially be updated during the show() call, and wanted this to happen off screen as it had the potential of looking ugly or something ... but since the screen does not get updated while js code is running AFAIK this precaution is not necessary and this the code can be removed.

source/class/qx/ui/mobile/dialog/Popup.js
```diff
@@ -223,9 +223,9 @@ qx.Class.define("qx.ui.mobile.dialog.Popup", {
 
           this.placeTo(computedPopupPosition.left, computedPopupPosition.top);
         }
-      } else if (this.__childrenContainer) {
-        // No Anchor
-        this._positionToCenter();
+      } else if (this.__childrenContainer && !this.__placed) {
+          // No Anchor and not placed to point manually
+          this._positionToCenter();
```
oetiker, line 228: what if the dialog has been placed but not in the center (yet)?

goldim: if to call `show` before "placeTo" popup will have margin values and if after it wont assign margin value.

source/class/qx/ui/mobile/dialog/Popup.js
```diff
@@ -323,6 +320,10 @@ qx.Class.define("qx.ui.mobile.dialog.Popup", {
      * @param top {Integer} - the value the will be set to container's top style property
      */
     placeTo(left, top) {
+      this.__placed = true;
+      var container = this.getContainerElement();
+      container.style.marginLeft = "0px";
+      container.style.marginTop = "0px";
```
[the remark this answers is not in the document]
goldim: if to call show before "placeTo" popup will have margin values and if after it wont assign margin value.

source/class/qx/ui/mobile/dialog/Popup.js
goldim: @oetiker if the dialog has been placed what's point to place it again? I think in this case there is no needed actions to do. If the dialog was not placed at all it will be centered, if it has an anchor it will be placed to coords of the anchor widget.

oetiker (approved):

hkollmann (approved):

johnspackman (approved):

derrell (changes requested):

source/class/qx/ui/mobile/dialog/Popup.js
```diff
@@ -323,6 +320,10 @@ qx.Class.define("qx.ui.mobile.dialog.Popup", {
      * @param top {Integer} - the value the will be set to container's top style property
      */
     placeTo(left, top) {
+      this.__placed = true;
+      let style  = this.getContainerElement().style;
+      style.marginLeft = "0px";
+      style.marginTop = "0px";
```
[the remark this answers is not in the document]
goldim: @derrell I removed this edits but without them everything works. Maybe there is some case which I forgot. I regret I removed my test program with all possible cases. Btw there is another question should I somewhere reset `__placed` flag: for example I set widget anchor and then reset anchor to undefined. The popup will get position which was set via `placeTo`.

derrell:
What problem were you trying to fix with this PR? Was there some problem caused by moving the popup offscreen by default with

```
// Move outside of viewport
this.placeTo(-1000, -1000);
```

? It's been working, I believe, for a very long time...

source/class/qx/ui/mobile/dialog/Popup.js
```diff
@@ -323,6 +320,7 @@ qx.Class.define("qx.ui.mobile.dialog.Popup", {
      * @param top {Integer} - the value the will be set to container's top style property
      */
     placeTo(left, top) {
+      this.__placed = true;
```
derrell, line 323: `__placed` should be declared in the `members` section

derrell: It probably should be reset someplace, but without the default of placing the popup offscreen, I'm not sure where it would be reset...

goldim:
@derrell this issue - https://github.com/qooxdoo/qooxdoo/issues/9423

source/class/qx/ui/mobile/dialog/Popup.js
goldim: @derrell I think one place where it should be reset is where an anchor is set.

goldim:
@derrell I declared the member in the beginning of a member section as you asked. Also renamed it and added in destructor of class.
I don't get why I have to reset it via time. I try to explain there are 3 cases of positioning: 

1. constructor - you just don't  pass the second argument - an anchor
2. method `setAnchor` btw this method calculates position of widget-anchor and calls `placeTo`
3. directly via `placeTo`

There is no method `resetAnchor` and only way to get rid of anchor - to pass `undefined/null` value in `setAnchor` but this method will save the position. I thought maybe it would be right to position to center bc the anchor is removed.

derrell:
There's no BC issue, and if one really wanted to reset it, it looks like placing it offscreen, e.g. with `popup.placeTo(-1000, -1000)`, would do it. I suppose you could reset `__isPlaced` near line 211:

```
        if (isOutsideViewPort) {
          this.__isPlaced = false;
          this._positionToCenter();
        } else {
```

although I'm not sure it's really necessary.

goldim:
@derrell I would try it and check it out.

goldim:
@derrell I can not reproduce offscreen case neither in original code not mine. 
Also about setting margins to zero px or null: There is case when firstly popup is moved to center and margins are left and if I want to `placeTo` then the popup will be shifted - It is unexpected behaviour. I think resetting them is good choice. What do you think?
Btw I found more dangerous bug (infinite recursive calls) debugging code of Popup but about it I will make another PR.

goldim:
@derrell I think the method `qx.util.placement.Placement` and its options keeps inside of viewport and situation where it is outside is imposible. About margins: it is possible that someone can set margins for popup via theming for example? In this case resetting margin before `placeTo` is bad choice.

derrell:
@goldim I don't think resetting things is worth the trouble. Popups are not issued in a fast loop; rather, they are presented only occasionally. I think if someone wanted to show a previously shown popup and restore its centered behavior, they could simply regenerate a new popup, discarding the old one. I don't want to discourage you from searching for a way to reset the position, even adding a new method to do it, but it does not appear to be necessary for BC, nor IMO, is it necessary for usability.

derrell (approved):

goldim:
@cboulanger @level420 Could you approve or reject this PR?

cboulanger (approved):

PR #10601, opened 2023-06-01
Add cross compiler to 8.x beta

johnspackman:
This adds support for cross compiling so that older runtimes can still work with whatever version the current compiler is.

### DIFF
--- a/source/class/qx/tool/utils/QooxdooVersions.js
+++ b/source/class/qx/tool/utils/QooxdooVersions.js
@@ -1,0 +1,87 @@
+qx.Class.define("qx.tool.utils.QooxdooVersions", {
+  extend: qx.core.Object,
+
+  construct() {
+    super();
+  },
+
+  members: {
+    __githubTags: null,
+
+    /**
+     * Gets the tags from Github so we can find releases
+     *
+     * @returns {*} see GitHub API
+     */
+    async _getTags() {
+      if (!this.__githubTags) {
+        this.__githubTags = qxGithubTags = await qx.tool.utils.Http.getJson(
+          "https://api.github.com/repos/qooxdoo/qooxdoo/tags"
+        );
+      }
+      return this.__githubTags;
+    },
+
+    /**
+     * Finds the best version of Qooxdoo which is available, whether it is at Github as a release, or the
+     * default Qooxdoo that is provided by this compiler instance
+     *
+     * @param {String} versionToMatch semver satisfaction sequence, eg "^7.0.0"
+     * @returns {String} the directory where the Qooxdoo installation is
+     */
+    async findBestVersion(versionToMatch) {
+      let defaultVersion = await qx.tool.config.Utils.getQxVersion();
+      if (semver.satisfies(version, versionToMatch)) {
+        let qxpath = await this.getQxPath();
+        return qxpath;
+      }
+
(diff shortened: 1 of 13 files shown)


source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -299,6 +299,10 @@
             "type": "object",
             "description": "Options given to @babel/preset-env. With this options the output type of babel can be defined. For details see here: <https://babeljs.io/docs/en/babel-preset-env#options>. They can be overridden per target."
           },
+          "qooxdooVersion": {
+            "type": "string",
+            "description": "The semver that must be satisifed when selecting the Qooxdoo version to compile against; compiler will download a version automatically if the default Qooxdoo does not satisfy this semver"
+          },
```
derrell, line 305: It seems wrong to have a file name with 3 levels, including minor revision, but then to be making changes to the schema without updating the file name. Can we either update the filename to `compile-1-0-1`, or change the filename to one that doesn't include the extra 0's, i.e., `compile-1.json` if we plan to make changes to it without updating its name?

source/class/qx/tool/utils/QooxdooVersions.js
```diff
@@ -0,0 +1,190 @@
...
+const semver = require("semver");
+const path = require("upath");
+const fs = qx.tool.utils.Promisify.fs;
+
+qx.Class.define("qx.tool.utils.QooxdooVersions", {
+  extend: qx.core.Object,
+
+  construct() {
+    super();
+  },
```
derrell, line 28: Superfluous constructor, since all it does it call the superclass constructor, but not wrong.

derrell (changes requested):

source/resource/qx/tool/schema/compile-1-0-0.json
johnspackman: that's the problem we've had with updating the manifest schema, it makes all existing .json files incompatible, because the compiler does not have the ability to consider multiple, compatible, schemas.  Modifying the 1.0.0 is just a practical solution

johnspackman:
@derrell BTW Im happy for you to choose to include this into your PR or not as you choose, with or without reviews from other people - I meant it as a polite way to suggest and keep track of changes that are essential to my testing the 8 beta

source/resource/qx/tool/schema/compile-1-0-0.json
derrell: Can we simply rename it to `compile-1.json` so it's clear that the filename isn't modified with each change? It's not a big deal to me, but it seems like it would be more clear. There would then be a single incompatibility with this change, and then no further ones.

derrell:
Yes, I'm happy to include this in the PR. It seems very appropriate.

source/resource/qx/tool/schema/compile-1-0-0.json
johnspackman: the problem is that all the existing `compile.json` refer to the schema as the `compile-1-0-0`, so if we rename it then they will all fail validation.

I suspect that the solution would be to code a simple transform from 1.0.0 to 1.0.1 to 1.0.2 etc, so that teh compiler is coded to use the latest, but while loading the compile .json will upgrade it if at all possible

PR #10605, opened 2023-06-21
Removed missing apply methods

lucapivato:
_applyOpenInterval and _applyCloseInterval don't exist.


goldim (approved):
When you set values to these props you get an error about missing method. Looks like nobody sets new value for these props.

johnspackman (approved):

hkollmann (approved):

oetiker (approved):

derrell (approved):

level420 (approved):

PR #10606, opened 2023-06-22
Treevirtual - reselect all selections, not just the last one

rad-pat:
I discovered a bug with the Treevirtual whereby it does not reselect all the previous selections when refreshed (in multi-selection mode). This is already apparent in the DemoBrowser where 2 selections are made in the set-up, but upon setting the datamodel, only one selection is actually displayed.

Demo Browser where it fails to select 2 nodes
http://qooxdoo.org/qxl.demobrowser/#treevirtual~TreeVirtual_Selections.html

The fix uses the `batchMode` of the selectionModel to add each of the selections before ultimately firing the `changeSelection` event once all selections are added. For single selection mode, the effect will be the same as before where the last of the selections would be selected.

This is a link to a Playground example of the above Demo Browser with my fix applied:
https://tinyurl.com/y2jdc8v7

### DIFF
--- a/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
+++ b/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
@@ -831,9 +831,16 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
       // Set selections in the selection model now
       var selectionModel = this.getTree().getSelectionModel();
       var selections = this._selections;
-      for (var nodeId in selections) {
-        var nRowIndex = this.getRowFromNodeId(nodeId);
-        selectionModel.setSelectionInterval(nRowIndex, nRowIndex);
+
+      selectionModel.setBatchMode(true);
+      try {
+        selectionModel.resetSelection();
+        for (var nodeId in selections) {
+          var nRowIndex = this.getRowFromNodeId(nodeId);
+          selectionModel.addSelectionInterval(nRowIndex, nRowIndex);
+        }
+      } finally {
+        selectionModel.setBatchMode(false);
       }
     },
 


oetiker (approved):

johnspackman (approved):

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -831,9 +831,16 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
...
+
+      selectionModel.setBatchMode(true);
+      try {
+        selectionModel.resetSelection();
+        for (var nodeId in selections) {
+          var nRowIndex = this.getRowFromNodeId(nodeId);
+          selectionModel.addSelectionInterval(nRowIndex, nRowIndex);
+        }
+      } finally {
+        selectionModel.setBatchMode(false);
```
derrell, line 843: Is it possible that batch mode was already `true` initially, and this code should retrieve the current batch mode before line 835, and then restore that prior batch mode here rather than always assuming it was `false`?

derrell (changes requested):

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
rad-pat: @derrell , you're right, it could be true already. Someone has implemented reference counting in the batchMode which should suffice, I think?

rad-pat: https://github.com/qooxdoo/qooxdoo/blob/84400c57ae4fce6dd01169b5a1b7adfe89b7ed1c/source/class/qx/ui/table/selection/Model.js#L140

derrell: I suppose. That's hidden, rather than clean and obvious code -- it requires looking at that setter to know that -- so my inclination would be to save the current value and restore that value when I'm done. You're right, it's not strictly necessary. I'll approve and let you decide whether to make it obvious or leave it as a more hidden side-effect.

derrell (approved):

goldim (approved):
Well Done! One notice: example from the playground doesn't work locally bc there is a signature like _P_1_3.

hkollmann (approved):

level420 (approved):

PR #10610, opened 2023-07-01
Mousedown instead of tap event for abstract select box

### LINKED ISSUE #9023: Focus problem with qx.application on touch devices
Hello,
I see a weird Focus problem when using qx.application on touch devices:
Every time a select box item is selected, the underlying widget gets the focus afterwards.
This not only applies to select boxes but happens also when updating the window with a different set of widgets after a button press --> the widget that appears under the button gets the focus.
If the new focus widget is a textfield, the virtual keyboard gets activated, which is disrupting.
For me it is reproducible with android tablets on http://demo.qooxdoo.org/current/showcase/#Form:
When selecting Item 0 of ComboBox the VirtualComboBox gets the focus, when selecting Item 1 the DateField gets the focus.
Since I'm not familiar with the qooxdoo internals I have neither a workaround nor a pull request available.
Regards, Johannes

goldim:
Fixed #9023.
I tested and noticed it doesn't depends on what browser is used but depends on what device is used. You could check this behaviour in Firefox dev tools changing device. After selecting some item SelectBox is focused. Maybe it shouldn't? If yes there is a simpler solution than changing event. I did check my changes for emulated devices but not for real so far.

### DIFF
--- a/source/class/qx/ui/form/AbstractSelectBox.js
+++ b/source/class/qx/ui/form/AbstractSelectBox.js
@@ -144,7 +144,9 @@ qx.Class.define("qx.ui.form.AbstractSelectBox", {
           );
 
           control.addListener("pointerdown", this._onListPointerDown, this);
-          control.getChildControl("pane").addListener("tap", this.close, this);
+          control
+            .getChildControl("pane")
+            .addListener("mousedown", this.close, this);
           break;
         }
         case "popup":


cboulanger (approved):

hkollmann:
This is not reproducable on ios. Could you check the device? Are you sure that mousedown is working on ios?

goldim:
@hkollmann checked on real iphone. Mousedown solution doesn't work for them but for android devices the problem is gone. I'll try to find other solution.

PR #10611, opened 2023-07-03
Change var to let for babel compilation in qxl.apiviewer

goldim:
A compilation of qxl.apiviewer is failed due to:
```
TypeError: /var/www/html/test/node_modules/@qooxdoo/framework/source/class/qx/tool/compiler/targets/Target.js: Property left of ForInStatement expected node to be of a type ["VariableDeclaration","LVal"] but instead got undefined
qx.tool.compiler.targets.Target: error: FATAL Syntax error: /var/www/html/test/node_modules/@qooxdoo/framework/source/class/qx/tool/compiler/targets/Target.js: Property left of ForInStatement expected node to be of a type ["VariableDeclaration","LVal"] but instead got undefined
```


hkollmann (approved):

oetiker (approved):

johnspackman (changes requested):
It looks like the bug that is causing the crash is a bug in the compiler, where is does not properly handle the `for (var x in y)`; this should be fixed in `qx.tool.compiler.ClassFile` and not in `Target`

goldim:
> It looks like the bug that is causing the crash is a bug in the compiler, where is does not properly handle the `for (var x in y)`; this should be fixed in `qx.tool.compiler.ClassFile` and not in `Target`

Do you mean replacing for/in loops in ClassFile? It doesn't do a thing. Or you mean more difficult  change don't you?

goldim:
@johnspackman this PR doesn't do anything bad at all. Replacing var -> let fast resolve the problem not diving to compiler implementation and details.

hkollmann:
This is only happening in build version of compiler!

hkollmann:
And it happens only with version from npm. 
So i really suggest to approve this changes.

And the code get cleaner by this change :-)

cboulanger (approved):

level420 (approved):

hkollmann:
@johnspackman: Could you agree t the argumentation and approve the PR?

johnspackman:
I can reproduce this, but only on an `npm` based compile; the problem is *not* a problem with `qx.tool.compiler.targets.Target`, but it seems that it is a problem with the build and deployment of qooxdoo itself (ie the generation of the npm release)

My guess ATM is that the bug solved in https://github.com/qooxdoo/qooxdoo/pull/10619 may be responsible, because what was happening there was that the build was not a repeatable compile - which would cause bootstrap-compiler to use the very latest of npm modules and not the known-good versions.

On the same VM I used to reproduce this error, I also checked out qooxdoo/master, bootstrapped, and compiled api viewer again using the newly-made build version of the compiler ... and it works just fine.

I do not agree that this PR should be merged, because if we do then we will simply be hiding a problem that will bite someone at some point in the future, and we might not find it so easy to reproduce it and work on the problem (eg if it is an end-user's application that is failing to compile).  At least now we have a reproducible example of the bug.

goldim:
@johnspackman in this case I cancel the PR, right?

Issue #10136, opened 2021-02-19
unexpected scrolling behavior in table widget 


nanexo:
**Describe the bug**
There is an issue with the scrolling in the table widget, where it is scrolling more than it should.

**To Reproduce**
Case 1:
This is easily verifiable in the demo for the table widget: http://qooxdoo.org/qxl.demobrowser/#table~Table.html
1. Select the second to last fully visible row.
2. Press Arrow Down.
4. See error: the table scrolled.

Case 2:
See this playground link: https://tinyurl.com/ycoqq5tq
This is essentially the same demo, but with the rowHeight set to 50.
1. Drag the Scrollbar. If you drag slowly the effect is VERY noticable.
2. See error: The scrollbar is really janky - it is jumping up and down.

**Expected behavior**
I would expect both cases to behave like 5.0.2
Playground link with the same example on qx5.0.2 for convenience 
https://tinyurl.com/y9ym2uaa
http://archive.qooxdoo.org/current/demobrowser/#table~Table.html

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 88

**Additional context**
This was introduced with https://github.com/qooxdoo/qooxdoo/issues/9538.

hkollmann:
This is a problem with doc handling and table scrolling not a problem of the table. If you modify the playground example to use rowHeight: 50 everything works.
See https://tinyurl.com/y9vk8dr8

level420:
@hkollmann I've taken your playground sample and replaced 
```javascript
tableModel.setData(createRandomRows(1000));
```
with
```javascript
tableModel.setData(createRandomRows(50));
```
and now the behaviour described by @nanexo is reproducible!
I think there is a problem when having `rowHeight` set to something which leads to sort of rounding issues on scroll bar position to scroll view position calculations.

goldim:
@hkollmann Tried to fix it but appears to me too difficult. There was your PR which fixed scroll buttons (scroll step equals to row height). But the current bug imho more uncomfortable for users and looks terrible. I would roll back changes of your PR. What do you think?

hkollmann:
do you mean this https://github.com/qooxdoo/qooxdoo/pull/9829 ?

goldim:
@hkollmann yes, it is.

goldim:
Fixed by PR

PR #10613, opened 2023-07-06
fix for issue https://github.com/qooxdoo/qooxdoo/issues/10136: unexpected scrolling

hkollmann:
do scrolling only if it should move one line. This elimintes the flicker.


johnspackman (approved):

oetiker (approved):

derrell (approved):

goldim:
@hkollmann the slider is still not smooth. I may approve but the issue will be open. Btw in you example maybe to increase number of random rows to 50? I would like somehow separate buttons' scroll and slider slow handler. Haven't found a way to do it.

hkollmann:
What do you mean with not smooth? Flicker is gone.
Seperating button and scroll is hard work and mus start in Slider.js line 1041, _fireValue. There is the change value event for scrolling generated.

cboulanger (approved):

goldim (approved):
@hkollmann I am sorry. Yes, it fixes the flicker.

PR #10619, opened 2023-08-15
Resolve issue 10618 Scss name conflict

### LINKED ISSUE #10618: Bootstrap failing on linux
I try to compile the compiler on Linux, but fail.

# Steps to reproduce:
`$ git clone https://github.com/qooxdoo/qooxdoo.git`
`$ cd qooxdoo`
`$ npm ci`
`$ ./bootstrap-compiler`

Unfortunately, this is what I get after running the last command:
[out.txt](https://github.com/qooxdoo/qooxdoo/files/12328033/out.txt)

Based on what I have found out, the `rem` function in CSS requires two arguments but only one was given (https://developer.mozilla.org/en-US/docs/Web/CSS/rem). Adding the value '1' as the second argument for every call of `rem` across the whole source of Qx makes the compiler compile. I don't understand how this code is supposed to work if there are many instances in the Qx source where `rem` is passed one argument instead of two?

# System info:
- OS: Arch Linux 64-bit
- NodeJs version: 16.20.2 (downloaded from nodejs.org)

### LINKED ISSUE #10621: qooxdoo/qxl.apiviewer compiler error
Description:
1. the getting started project "test1"  works fine with no errors
2. after the apiviewer package installation i get a TypeError.

$ cd test1
$ npx qx create test1 --type desktop --noninteractive --out=.
$ npx qx compile

$ npx qx pkg install --save=0 qooxdoo/qxl.apiviewer
$ npx qx compile

Compiler message / FATAL Syntax error: 
qx.tool.compiler.resources.ScssConverter: [30,13] warning: Unres ......
qx.tool.compiler.resources.ScssFile: [31,13] warning: Unresolved ......
TypeError: C:_Projekte\qooxdoo\test1\node_modules@qooxdoo\framework\source\class\qx\tool\compiler\targets\Target.js: Property left of ForInStatement expected node to be of a type ["VariableDeclaration","LVal"] but instead got undefined
qx.tool.compiler.targets.Target: error: FATAL Syntax error: C:_Projekte\qooxdoo\test1\node_modules@qooxdoo\framework\source\class\qx\tool\compiler\targets\Target.js: Property left of ForInStatement expected node to be of a type ["VariableDeclaration","LVal"] but instead got undefined

Versions:
 - qooxdoo 7.6.1
 - qooxdoo_qxl_apiviewer_v2_0_0
 - node v14.21-3
 - OS: Windows 11
 - Browser [chrome, firefox]

WillsterJohnsonAtZenesis:
Fixes #10618 

Previously the known good compiler was always the latest available `^7.0.0` on npm. It has now been pinned to v7.3. The `node_modules` for known good is now removed when the bootstrap is run to ensure dependencies are installed correctly. The `package-lock.json` has been removed as it previously was ignored by tooling, and is unnecessary for installing a pinned version of the compiler. 

For some reason I could not determine or resolve, v7.4 has different behaviour with it's dependencies, allowing the latest version of Sass to be installed (vs v1.43.4). [Since v1.65](https://github.com/sass/dart-sass/releases/tag/1.65.0), Sass supports [the new CSS builtin `rem()` function](https://developer.mozilla.org/en-US/docs/Web/CSS/rem).

Pinning Sass to v1.64.2 for now will resolve any conflicts with the buitin `rem()` when compiling Sass, and may allow for letting the known good compiler increase to the version this change is added in (eg, v7.6.2 vs v7.3).

The documentation for themes has been modified to note the deprecation notice and to reference the new function instead.


johnspackman (approved):

cboulanger (approved):

hkollmann:
When you update package.json you need to update npm-shrinkwrap.json too!

Why do you use fixed version of sass 1.64.2 and not the newest one ^1.65.1?

hkollmann (changes requested):
When you update package.json you need to update npm-shrinkwrap.json too!

Why do you use fixed version of sass 1.64.2 and not the newest one ^1.65.1?

hkollmann (changes requested):
When you update package.json you need to update npm-shrinkwrap.json too!

Why do you use fixed version of sass 1.64.2 and not the newest one ^1.65.1?

WillsterJohnsonAtZenesis:
> When you update package.json you need to update npm-shrinkwrap.json too!

Ah my bad, I'll commit that shortly.

> Why do you use fixed version of sass 1.64.2 and not the newest one ^1.65.1?

In Sass v1.65.0 support was added for the built-in `rem()` function in CSS, as specified by Values and Units Level 4. This conflicts with the function [defined in Qooxdoo](https://github.com/qooxdoo/qooxdoo/pull/10619/files#diff-9edee7ca18cf35229df225d2e982a5fbda167212d59f4b17bced23eeb208ed05) and is the reason for #10618 (the builtin `rem()` always takes priority in Sass ^1.65.0, and it has very different functionality)

The use of Sass ^1.65 is causing these errors, and is also the reason why I had to drop the known good down to v7.3, as using newer versions would install Sass ^1.65, which blocks `bootstrap-compiler` from working - v7.3 is pinned at v1.43.4.

That is to say that every version after and including v7.4 may now be broken, or partly broken, due to the combination of the naming conflict and the permissive dependency version. As Sass 1.65.0 was only released 5 days ago, this hasn't been an issue until now.

hkollmann:
> > When you update package.json you need to update npm-shrinkwrap.json too!
> 
> Ah my bad, I've added a commit for that above.
> 
> > Why do you use fixed version of sass 1.64.2 and not the newest one ^1.65.1?
> 
> In Sass v1.65.0 support was added for the built-in `rem()` function in CSS, as specified by Values and Units Level 4. This conflicts with the function [defined in Qooxdoo](https://github.com/qooxdoo/qooxdoo/pull/10619/files#diff-9edee7ca18cf35229df225d2e982a5fbda167212d59f4b17bced23eeb208ed05) and is the reason for #10618 (the builtin `rem()` always takes priority in Sass ^1.65.0, and it has very different functionality)
> 
> The use of Sass ^1.65 is causing these errors, and is also the reason why I had to drop the known good down to v7.3, as using newer versions would install Sass ^1.65, which blocks `bootstrap-compiler` from working - v7.3 is pinned at Sass v1.43.4.
> 
> That is to say that every version after and including v7.4 may now be broken, or partly broken, due to the combination of the naming conflict and the permissive dependency version. As Sass 1.65.0 was only released 5 days ago, this hasn't been an issue until now.

hkollmann:
Got it. So clamp version 1.64.2 and deprecate rem() as you did. In version 8 we can delete rem from our scss source and move to actual sass version again.

hkollmann (approved):

oetiker (approved):

goldim (approved):

Issue #10628, opened 2023-10-18
iOS Safari not always firing `change` events breaks `valueChange`


oetiker:
When closing the keyboard on IOS Safari, the input element fires a `blur` event, but no `change` event. This is problematic for the qooxdoo implementation of the `AbstractField` which relies on `change` events to detect changed content.

https://github.com/qooxdoo/qooxdoo/blob/2625902f10ae3c60e14ad878aba328d215e6caa2/source/class/qx/ui/form/AbstractField.js#L129

and 

https://github.com/qooxdoo/qooxdoo/blob/2625902f10ae3c60e14ad878aba328d215e6caa2/source/class/qx/ui/form/AbstractField.js#L722-L724

See also https://stackoverflow.com/questions/72002932/why-onchange-function-not-working-in-safari

PR #10629, opened 2023-10-19
Update AbstractField.js

oetiker:
Make `changeValue` event working on ios. There is no `change` event when just closing the keyboard, only a `blur`. 

Fix for #10628

### DIFF
--- a/source/class/qx/ui/form/AbstractField.js
+++ b/source/class/qx/ui/form/AbstractField.js
@@ -126,12 +126,13 @@ qx.Class.define("qx.ui.form.AbstractField", {
       this.setValue(value);
     }
     let el = this.getContentElement();
+    // change does not fire on ios webkit when just the keyboard is closed
+    // blur does fire though ... 
     // since ios allows no other html engines, checking for ios should do
-    // also the blur trick should work on other browsers too, but lets not rock the boat.
     if (qx.core.Environment.get("os.name") == "ios") {
-      el.addListener("blur", (e) => {
+      el.addListener("blur", () => {
         this._onChangeContent(new qx.event.type.Data(this.getValue()));
-      });
+      }, this);
     }
     else {
       el.addListener("change", this._onChangeContent, this);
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

source/class/qx/ui/form/AbstractField.js
```diff
@@ -125,9 +125,18 @@ qx.Class.define("qx.ui.form.AbstractField", {
...
-
-    this.getContentElement().addListener("change", this._onChangeContent, this);
-
+    let el = this.getContentElement();
+    // change does not fire on ios webkit when just the keyboard is closed
+    // blur does fire though ... 
+    // since ios allows no other html engines, checking for ios should do
+    if (qx.core.Environment.get("os.name") == "ios") {
+      el.addListener("blur", () => {
+        this._onChangeContent(new qx.event.type.Data(this.getValue()));
```
hkollmann, line 134: You should only fire this if the valkue has really changed!

hkollmann (changes requested):

goldim (approved):

derrell (approved):

level420 (approved):

source/class/qx/ui/form/AbstractField.js
```diff
@@ -131,7 +131,9 @@ qx.Class.define("qx.ui.form.AbstractField", {
     // since ios allows no other html engines, checking for ios should do
     if (qx.core.Environment.get("os.name") == "ios") {
       el.addListener("blur", () => {
-        this._onChangeContent(new qx.event.type.Data(this.getValue()));
+        let ev = new qx.event.type.Data();
```
hkollmann, line 134: As before: Should only be fired when content is changed. This is documented here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event

hkollmann (changes requested):

source/class/qx/ui/form/AbstractField.js
oetiker: yes I agree, except that in this case, code further down is already taking care of this ...

hkollmann (approved):

oetiker:
can we have a release ? this bug is pretty evil on ios ... people who submit a form after filling a field may end up not submitting the fields value because the changeValue event does not get triggered ..

goldim:
I wouldn't mind. It will be 7.6.3.

PR #10633, opened 2023-10-25
Desktop dispose fix

### LINKED ISSUE #10622: Dispose desktop widget leads to an error
**Describe the bug**
Flushing `qx.ui.window.Manager` is broken.

**To Reproduce**
1. Go to Playground
2. Open browser console
3. Paste next code:
```javascript
var mainwin = new qx.ui.window.Window("Main Window");
const mainDes = new qx.ui.window.Desktop(new qx.ui.window.Manager());
mainDes.add(mainwin);
this.getRoot().add(mainDes, {left:20, top:20});
mainDes.dispose();
```
4. Press Run
5. See error

 - Qooxdoo Version [7.6.1]

**Error message**
```
003753 qx.ui.core.queue.Widget: Error in the 'Widget' queue:TypeError: this.__P_558_0 is null TypeError: this.__P_558_0 is null
```
Instead of `this.__P_558_0` locally got `__desktop` in my app.

goldim:
There are tightly coupled classes: `qx.ui.window.Desktop` and `qx.ui.window.Manager`. In destructor of Desktop there is disposing of Manager and vice versa. There is a case when desktop could be disposed firstly BUT `syncWidget` of Manager uses desktop member and leads to an error.

It is a fast and straightforward solution If you know better one just let me know. Thanks.

### DIFF
--- a/source/class/qx/ui/window/Manager.js
+++ b/source/class/qx/ui/window/Manager.js
@@ -81,7 +81,7 @@ qx.Class.define("qx.ui.window.Manager", {
      * {@link qx.ui.core.queue.Widget widget queue}.
      */
     syncWidget() {
-      if (!this.__desktop){
+      if (this.isDisposed()){
         return;
       }
       this.__desktop.forceUnblock();
--- a/source/class/qx/ui/window/Manager.js
+++ b/source/class/qx/ui/window/Manager.js
@@ -81,6 +81,9 @@ qx.Class.define("qx.ui.window.Manager", {
      * {@link qx.ui.core.queue.Widget widget queue}.
      */
     syncWidget() {
+      if (!this.__desktop){
+        return;
+      }
       this.__desktop.forceUnblock();
 
       var windows = this.__desktop.getWindows();


source/class/qx/ui/window/Manager.js
```diff
@@ -81,6 +81,9 @@ qx.Class.define("qx.ui.window.Manager", {
      * {@link qx.ui.core.queue.Widget widget queue}.
      */
     syncWidget() {
+      if (!this.__desktop){
```
derrell, line 84: A much more typical way of handling this type of issue in qooxdoo, is:

```
      if (this.isDisposed()) {
        return;
      }
```

derrell (changes requested):

source/class/qx/ui/window/Manager.js
goldim: It looks much better! Tested it, does work.

hkollmann (approved):

johnspackman (approved):

level420 (approved):

cboulanger (approved):

derrell (approved):

oetiker (approved):

PR #10638, opened 2023-11-16
Allow clearData to be called for a TreeVirtual without a re-render of the tree

rad-pat:
When pruning all nodes from a TreeVirtual SimpleTreeDataModel `prune(0)`, the nodes are recursively "deleted" by replacing the node array entries with null and removing these from any selections. When the tree is re-filled, the amount of data in the node array is increased. If this happens for a large tree several times over then the performance of the tree starts to degrade which is something noticed in our application.

Pruning at zero (root node) is a special case because all the nodes except root node are removed. The node array can be reset to a single root node and still maintain a nodeId == index correspondence for all new nodes as noted just below the amended code within the `prune` method. This allows prune and re-population to take place without growing the node array.

There is a method `clearData` of the datamodel which also does this, but causes the tree to be re-rendered which was not desirable in our use-case and the reason we were pruning in the first instance. Arguably that method could also be amended to suppress the re-render and used instead.  I'm happy to take direction and amend as required.

### DIFF
--- a/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
+++ b/source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
@@ -809,10 +809,13 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
      *   has completed building or modifying a tree by issuing a series of
      *   calls to {@link #addBranch} and/or {@link #addLeaf}.
      *
+     * @param bRerender {Boolean?true}
+     *  Rerender the tree data after setting the data. If set false it becomes the caller's responsibility to
+     *    call setData() subsequently to cause a redraw.
      *
      * @throws {Error} If the parameter has the wrong type.
      */
-    setData(nodeArr) {
+    setData(nodeArr, bRerender = true) {
       this._checkEditing();
       if (nodeArr instanceof Array) {
         // Save the user-supplied data.
@@ -825,8 +828,10 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
         );
       }
 
-      // Re-render the row array
-      this.__render();
+      // Re-render the row array, if so requested
+      if (bRerender) {
+        this.__render();
+      }
 
       // Set selections in the selection model now
       var selectionModel = this.getTree().getSelectionModel();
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

oetiker (approved):

goldim (approved):

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
```diff
@@ -518,6 +518,14 @@ qx.Class.define("qx.ui.treevirtual.SimpleTreeDataModel", {
         throw new Error("Expected node object or node id");
       }
 
+      // Optimisation for removing all the nodes, node array can be reset
+      // rather than setting all the values except the root to null
+      if (nodeId === 0) {
+        this._clearSelections();
+        this._nodeArr = [qx.ui.treevirtual.MTreePrimitive._getEmptyTree()];
```
derrell, line 525: This is a nice idea, but it is not backward compatible. Your app requires no re-rendering, but we don't know that others would appreciate losing their rerendering. I suspect that in most cases, it's wanted.

I'd propose this (untested) alternative. First, alter `clearData` as follows:

```
    clearData(bRerender = true) {
      this._checkEditing();
      this._clearSelections();
      this.setData([qx.ui.treevirtual.MTreePrimitive._getEmptyTree()], bRerender);
    },
```

Then alter `setData` signature and where it does the rerendering:

```
    setData(nodeArr, bRerender = true) {
    ...
    // Re-render the row array, if so requested
    if (bRerender) {
      this.__render();
   }

```

And of course, add documentation for `bRerender` in both functions' headers.

What do you think of this as an alternative?

derrell (changes requested):

source/class/qx/ui/treevirtual/SimpleTreeDataModel.js
rad-pat: I think this will probably work fine, I'll try it out tomorrow. 

The change I proposed here doesn't prevent re-rendering itself, just the prune method didn't re-render anyway. I'm just preventing the node array getting larger every time I empty and refill the model.

rad-pat:
Updated with the requested changes to `clearData` that seem to work well.

cboulanger (approved):

hkollmann (approved):

derrell (approved):
Thanks.

PR #10642, opened 2023-12-11
`qx.ui.tabview.TabButton` does not conform to the API of superclass `qx.ui.basic.Atom`

WillsterJohnsonAtZenesis:
Due to the following code in TabButton;

https://github.com/qooxdoo/qooxdoo/blob/b3553384ed4a5194e67d4addd41b1795ea73704f/source/class/qx/ui/tabview/TabButton.js#L37-L41

A conflict arises with the following property apply methods in the Atom superclass;

https://github.com/qooxdoo/qooxdoo/blob/b3553384ed4a5194e67d4addd41b1795ea73704f/source/class/qx/ui/basic/Atom.js#L280-L282
https://github.com/qooxdoo/qooxdoo/blob/b3553384ed4a5194e67d4addd41b1795ea73704f/source/class/qx/ui/basic/Atom.js#L291-L293
https://github.com/qooxdoo/qooxdoo/blob/b3553384ed4a5194e67d4addd41b1795ea73704f/source/class/qx/ui/basic/Atom.js#L296-L298

This is due to `qx.ui.basic.Atom` prusuming that it's layout will always be a `qx.ui.layout.Atom`, which is untrue and unenforced.

This restricts the use of included appearance entries, as if `TabButton` includes one of it's superclass' appearances, that superclass cannot safely define these properties in it's appearance.

**To Reproduce**

1. Add a TabView to a project
2. Target the TabView button in appearances and set a `gap` value (or include a superclass' appearance which has a `gap` value)
3. Observe the error

**Expected behavior**

The error should not be thrown, and the application of the `gap` property should be ignored. This mirrors the way in which the vast majority of layout-specific properties work in CSS.

This could be done by either detatching `TabButton` from the inheritance chain to `Atom`, or by modifying `Atom` to be permissive of alternative layouts (latter option implemented in this PR).

**This PR**

Provides a simple error prevention mechanism within the property applies in `qx.ui.basic.Atom` which ensures the layout supports the given property before attempting to set it, provides a warning in development environments if the layout does not support the property.

### DIFF
--- a/source/class/qx/test/ui/basic/Atom.js
+++ b/source/class/qx/test/ui/basic/Atom.js
@@ -78,6 +78,39 @@ qx.Class.define("qx.test.ui.basic.Atom", {
       );
 
       a.destroy();
+    },
+
+    testLayoutOptionalProperties() {
+      let atomLayout;
+      const hasAtomLayout = new (qx.Class.define(null, {
+        extend: qx.ui.basic.Atom,
+        construct() {
+          super("test");
+          atomLayout = this._getLayout();
+        }
+      }))();
+
+      hasAtomLayout.setGap(10);
+      hasAtomLayout.setIconPosition("right");
+      hasAtomLayout.setCenter(true);
+
+      this.assertEquals(10, atomLayout.getGap());
+      this.assertEquals("right", atomLayout.getIconPosition());
+      this.assertEquals(true, atomLayout.getCenter());
+
+      const hasGridLayout = new (qx.Class.define(null, {
+        extend: qx.ui.basic.Atom,
+        construct() {
+          super("test");
+          this._getLayout().dispose();
+          this._setLayout(new qx.ui.layout.Grid());
+        }
+      }))();
+
+      // expect these to not throw an error (should log in debug env)
+      hasGridLayout.setGap(10);
+      hasGridLayout.setIconPosition("right");
+      hasGridLayout.setCenter(true);
     }
   }
 });
--- a/source/class/qx/test/ui/basic/Atom.js
+++ b/source/class/qx/test/ui/basic/Atom.js
@@ -111,6 +111,10 @@ qx.Class.define("qx.test.ui.basic.Atom", {
       hasGridLayout.setGap(10);
       hasGridLayout.setIconPosition("right");
       hasGridLayout.setCenter(true);
+      // the values will be set to the Atom, they will not be set (or attempt to be set) on the layout of the Atom
+      this.assertEquals(10, hasGridLayout.getGap());
+      this.assertEquals("right", hasGridLayout.getIconPosition());
+      this.assertEquals(true, hasGridLayout.getCenter());
     }
   }
 });
--- a/source/class/qx/ui/basic/Atom.js
+++ b/source/class/qx/ui/basic/Atom.js
@@ -283,9 +283,6 @@ qx.Class.define("qx.ui.basic.Atom", {
     },
 
     __safeSetPropertyOnLayout(value, propertyName) {
-      if (value === undefined) {
-        return;
-      }
       const layout = this._getLayout();
       const propertySetter = `set${qx.lang.String.firstUp(propertyName)}`;
       if (layout[propertySetter]) {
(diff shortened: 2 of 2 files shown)


oetiker (approved):

hkollmann (approved):

goldim (changes requested):

source/class/qx/ui/basic/Atom.js
```diff
@@ -277,8 +277,22 @@ qx.Class.define("qx.ui.basic.Atom", {
...
 
     // property apply
-    _applyGap(value, old) {
-      this._getLayout().setGap(value);
+    _applyGap(value) {
+      if (value === undefined) {
+        return;
+      }
+      const layout = this._getLayout();
+      if (layout.setGap) {
```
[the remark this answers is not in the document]
WillsterJohnsonAtZenesis: I can't see any reason not to, I'll refactor to do that.

johnspackman (approved):

source/class/qx/ui/basic/Atom.js
```diff
@@ -276,25 +276,44 @@ qx.Class.define("qx.ui.basic.Atom", {
...
     },
 
+    __safeSetPropertyOnLayout(value, propertyName) {
+      if (value === undefined) {
+        return;
+      }
+      const layout = this._getLayout();
+      const propertyGetter = `get${qx.lang.String.firstUp(propertyName)}`;
+      if (layout[propertyGetter]) {
+        layout[propertyGetter](value);
```
derrell, line 292: Something seems off here. The original code was calling `layout.setGap(value)` or `layout.setCenter(value)`, etc. The new code is calling `layout.getGap(value)` or `layout.getCenter(value)`, etc., which ignores the parameter and doesn't actually do anything, if I'm reading this correctly...???

derrell (changes requested):

source/class/qx/ui/basic/Atom.js
goldim: Exactly. It is a mistake.

hkollmann: Seems that this is not tested. Could you enhance the tests please?

hkollmann (changes requested):
Please add a test for this setters!

source/class/qx/ui/basic/Atom.js
WillsterJohnsonAtZenesis: Sorry, that's a complete lack of focus on my part. I'll get that corrected and add some tests later today.

source/class/qx/test/ui/basic/Atom.js
```diff
@@ -78,6 +78,39 @@ qx.Class.define("qx.test.ui.basic.Atom", {
...
+      const hasGridLayout = new (qx.Class.define(null, {
+        extend: qx.ui.basic.Atom,
+        construct() {
+          super("test");
+          this._getLayout().dispose();
+          this._setLayout(new qx.ui.layout.Grid());
+        }
+      }))();
+
+      // expect these to not throw an error (should log in debug env)
```
goldim, line 110: I would somehow check these next three lines via assertions. There is `fail` method and `assertException`. Give a look at them.

hkollmann: Or check the result...

source/class/qx/ui/basic/Atom.js
```diff
@@ -277,8 +277,22 @@ qx.Class.define("qx.ui.basic.Atom", {
     },
 
     // property apply
-    _applyGap(value, old) {
-      this._getLayout().setGap(value);
+    _applyGap(value) {
+      if (value === undefined) {
```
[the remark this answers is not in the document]
WillsterJohnsonAtZenesis: The unnecessary undefined check has been removed

source/class/qx/test/ui/basic/Atom.js
WillsterJohnsonAtZenesis: I've added assertions for these setter calls

goldim (approved):

oetiker (approved):

hkollmann (approved):

cboulanger (approved):

derrell (approved):

PR #10645, opened 2023-12-15
Implement cached objects top-level section

### LINKED ISSUE #10641: Proposal: New cached object system (think getQxObject)
# Context
This proposes a new system in which cached objects for Qooxdoo classes are defined (getQxObject, getChildControl, etc). It is 100% BC and solves the major problems present in the current system, which include:

- No checking for return statement after each case block. This can lead to recursion errors, which are a pain to debug, like in the example below:
```
_createQxObjectImpl(id) {
    switch (id) {
        case "selectBox1":
            var sb = new qx.ui.form.SelectBox();
            //I forget to put a return statement here, or I accidentally delete it
        case "ctlrSelectbox1":
            var sb = this.getQxObject("selectBox1");
            var ctlr = new qx.data.controller.List(model, sb, "title");
            return ctlr;
        ...

    }
}
```
- Forced use of the old 'var' instead of let or const because you cannot re-declare variables with the same name in the same scope. The compiler will complain in this case:
```
_createQxObjectImpl(id) {
    switch (id) {
    case "sb1":
        let sb = new qx.ui.form.SelectBox();
        return sb;
    case "sb2":
        let sb = new qx.ui.form.SelectBox();
        return sb;
    }
    ...
}
```    

- Having to remember to call `super._createQxObjectImpl(id)` (or similar to createWidget or createChildControl) at the end, which we may forget to do or accidentally delete.

# Design
- We create a new configuration key in the Qooxdoo class, just like 'extend' or 'members', but we name it to something like qxObjects. It contains an object of functions named like the widget ids, which return the object to be cached.
- The data is merged with the superclass, with subclass taking priority.

Example:

    qx.Class.define("com.myApp.MyWidget",{
        extend: qx.ui.core.Widget,

        qxObjects: {
            selectBox1() {
                const numItems = 5;
                let sb = new qx.ui.form.SelectBox();
                for (let i = 0; i < numItems; i++)
                    sb.add(new qx.ui.form.ListItem("Item: " + i));            
                return sb;
            },

            ctlrSelectBox1() {
                let sb = this.getQxObject("selectBox1");
                let ctlr = new qx.data.controller.List(this.getQxObject("modelSelectbox1"), sb, "name");
                return ctlr;
            },

            btnClose() {
                let btn = super._createQxObjectImpl("btnClose");//if we want to execute the inherited function
                btn.setBackgroundColor("red");
                return btn;
            },

            edtName: () => new qx.ui.form.TextField(),

            edtSurname: "edtName" //just takes the function that is defined for 'edtName'
        }
    });

# Some points

As well as solving the above-mentioned problems, this solution offers many additional benefits, including:
- Better code organisation
- Ability to navigate in outline (VSCode currently does not work with putting switch statements in the outline). Helpful when you are reading through new code or are looking for a widget but can't remember the name.

# Things to consider
- We have the problem of the situation where the programmer both defines the qxObjects key (new way) and overrides `_getQxObject`` (old way). In this case, I think it is sensible that the old way trumps the new way because this allows the user to have finer control over the behaviour. We shouldn't have this thing happen anyway.

patryk-m-malinowski:
This PR implements the proposal in issue #10641 (a better way of defining objects referenced by getQxObject)


johnspackman:
Hi,

Thank you for your email - I am unavailable, returning on January 8th 2024, but my calendar is booked for the following week to catch up from the holidays.
  
 For customers with extended support, please contact your IT department in the first instance, who will be able to pass on any questions to the appropriate person to handle your query in my absence.
  
Regards
John Spackman

source/class/qx/Class.js
```diff
@@ -1281,6 +1287,37 @@ qx.Bootstrap.define("qx.Class", {
...
+            "Invalid cached object definition for " +
+              key +
+              " in " +
+              clazz.classname
+          );
+        }
+
+        if (typeof key != "string") {
+          throw new Error(
+            "Invalid cached object key for " + key + +" in " + clazz.classname
```
cboulanger, line 1303: remove the double `+`

cboulanger (commented):
Otherwise, looks good to me.

goldim (changes requested):
Test have to pass.

source/class/qx/tool/compiler/ClassFile.js
```diff
@@ -1243,6 +1258,37 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
         defer: null
       };
 
+      function ensureCreateQxObjectImpl(membersPropertyPath) {
+        const membersPropertyNode = membersPropertyPath.node;
+        const memberNames = membersPropertyNode?.value?.properties?.map(
+          it => it.key.name
+        );
+
+        if (!memberNames) throw new Error("Members section is not an object");
```
johnspackman, line 1267: you'll need curly braces etc

docs/core/classes.md
```diff
@@ -185,6 +185,92 @@ aware that programs where private members are accessed from outside of their own
 class will work in the source version, but will fail to work in the build
 version!_
 
+### Cached Objects
```
johnspackman, line 188: Isn't this a change in terminology?  Currently we refer to this as [Universal] Object ID, eg https://qooxdoo.org/documentation/v7.6/#/core/object_id

I've just seen that page is not linked to in the navigation, please can you add that too

johnspackman (changes requested):

docs/core/classes.md
hkollmann: Please remember that private variables are mangled in source build too,

hkollmann:
Tests has to be parsed!

hkollmann (changes requested):
Could you add some tests for the new feature?Thanks!

hkollmann:
tests should run:

# run unit tests via qxl.testnode
Assertion error! : Called assert with 'false'
 Stack trace: 
qx.core.AssertionError:54:48,qx.Bootstrap:49:24,qx.core.Assert:80:21,qx.core.Assert:123:35,qx.core.MAssert:50:24,qx.test.compiler.ClassFile:61:14
not ok 2 - qx.test.compiler.ClassFile:test issue 633 - [0.43 ms] - : Called assert with 'false'
Assertion error! : Called assert with 'false'
 Stack trace: 
qx.core.AssertionError:54:48,qx.Bootstrap:49:24,qx.core.Assert:80:21,qx.core.Assert:123:35,qx.core.MAssert:50:24,qx.test.compiler.ClassFile:75:14
not ok 4 - qx.test.compiler.ClassFile:test issue 524 - [1.38 ms] - Error in asynchronous test: resume() called before wait()
Assertion error! : Called assert with 'false'
 Stack trace: 
qx.core.AssertionError:54:48,qx.Bootstrap:49:24,qx.core.Assert:80:21,qx.core.Assert:123:35,qx.core.MAssert:50:24,qx.test.compiler.ClassFile:82:14
not ok 8 - qx.test.compiler.ClassFile:test issue 726 - [0.21 ms] - : Called assert with 'false'

oetiker:
@pmalinowski-vmn it would be great if you could make your change such that the tests will run, as this also goes a long way towards avoiding backward compatibility issues

p9malino26:
Sorry for the long reply; was on a break.

I will fix the code so the tests pass.

oetiker (approved):

cboulanger (approved):

hkollmann (approved):

source/class/qx/Class.js
hkollmann, line 1303: Remove double + +

hkollmann (changes requested):
Please remove double + in Class.js!

derrell:
Just a note that this PR will need to be reimplemented in the new class/property system if/when that PR is merged.

hkollmann (approved):

goldim (approved):

johnspackman (approved):

cboulanger:
This is a cool addition, thanks alot - @p9malino26 it would be great if you could write something for https://news.qooxdoo.org/ on it. BTW - @johnspackman I think you/your company also added some features that should be covered. the good thing is that you can just reuse parts of the docs for a blog post.

johnspackman:
@cboulanger good idea - we'll get onto it in the next week 👍

p9malino26:
I've written a draft of the blog post: 
https://medium.com/@pmalinowski/new-feature-top-level-objects-section-63dab192d7e0

Not sure if I have the privileges to publish it on the Qooxdoo channel though.

oetiker:
@p9malino26 I will gladly give you these ... please tell me your medium username

p9malino26:
@pmalinowski

oetiker:
done

PR #10648, opened 2024-01-10
Detect duplicate namespaces in libraries of Qooxdoo project

patryk-m-malinowski:
This PR makes the compiler complain if there are two libraries within a Qx project which have the same namespace. This is because having a duplicate namespace causes some problems, such as the compiler not responding to changed source files.

### DIFF
--- a/source/class/qx/tool/compiler/Analyser.js
+++ b/source/class/qx/tool/compiler/Analyser.js
@@ -1415,14 +1415,16 @@ qx.Class.define("qx.tool.compiler.Analyser", {
      * @param library
      */
     addLibrary(library) {
-      if (this.__librariesByNamespace[library.getNamespace()]) {
+      const existingLibrary =
+        this.__librariesByNamespace[library.getNamespace()];
+      if (existingLibrary) {
         throw new Error(
           "Multiple libraries with namespace " +
             library.getNamespace() +
             " found " +
             library.getRootDir() +
             " and " +
-            this.__librariesByNamespace[library.getNamespace()].getRootDir()
+            existingLibrary.getRootDir()
         );
       }
       this.__libraries.push(library);
--- a/source/class/qx/tool/compiler/Analyser.js
+++ b/source/class/qx/tool/compiler/Analyser.js
@@ -1415,6 +1415,16 @@ qx.Class.define("qx.tool.compiler.Analyser", {
      * @param library
      */
     addLibrary(library) {
+      if (this.__librariesByNamespace[library.getNamespace()]) {
+        throw new Error(
+          "Multiple libraries with namespace " +
+            library.getNamespace() +
+            " found " +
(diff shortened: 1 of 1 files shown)


cboulanger (approved):

johnspackman (approved):

source/class/qx/tool/compiler/Analyser.js
```diff
@@ -1415,6 +1415,16 @@ qx.Class.define("qx.tool.compiler.Analyser", {
      * @param library
      */
     addLibrary(library) {
+      if (this.__librariesByNamespace[library.getNamespace()]) {
```
goldim, line 1418: Multiple calling of method `getNamespace`. Maybe add some local variable for it?

source/class/qx/tool/compiler/Analyser.js
```diff
@@ -1415,6 +1415,16 @@ qx.Class.define("qx.tool.compiler.Analyser", {
...
+        throw new Error(
+          "Multiple libraries with namespace " +
+            library.getNamespace() +
+            " found " +
+            library.getRootDir() +
+            " and " +
+            this.__librariesByNamespace[library.getNamespace()].getRootDir()
```
goldim, line 1425: The same about `this.__librariesByNamespace[library.getNamespace()]`. You called it twice.

hkollmann (approved):

source/class/qx/tool/compiler/Analyser.js
hkollmann: IMHO it could be at it is....

johnspackman: yes, i agree,  the `.getNamespace` is only a property

oetiker (approved):

source/class/qx/tool/compiler/Analyser.js
```diff
@@ -1415,6 +1415,18 @@ qx.Class.define("qx.tool.compiler.Analyser", {
...
+          "Multiple libraries with namespace " +
+            library.getNamespace() +
+            " found " +
+            library.getRootDir() +
+            " and " +
+            existingLibrary.getRootDir()
+        );
+      }
       this.__libraries.push(library);
       this.__librariesByNamespace[library.getNamespace()] = library;
```
goldim, line 1431:
```suggestion
    addLibrary(library) {
      const namespace = 
      const existingLibrary = library.getNamespace();
        this.__librariesByNamespace[namespace];
      if (existingLibrary) {
        throw new Error(
          "Multiple libraries with namespace " +
            namespace +
            " found " +
            library.getRootDir() +
            " and " +
            existingLibrary.getRootDir()
        );
      }
      this.__libraries.push(library);
      this.__librariesByNamespace[namespace] = library;
```

goldim (approved):

PR #10655, opened 2024-02-08
add cycle behaviour to ToggleButton

patryk-m-malinowski:
This PR gives the programmer the ability to control the behaviour of tristate checkboxes when they are clicked. Currently, they only toggle between the enabled/disabled state, but it is impossible for the end user to enter the undetermined state.

This adds a new property to ToggleButton, clickBehaviour, which can either be set to "toggle" (reflects current behaviour), or "cycle" (cycle between disabled, undetermined, enabled).

### DIFF
--- a/source/class/qx/ui/form/ToggleButton.js
+++ b/source/class/qx/ui/form/ToggleButton.js
@@ -114,6 +114,16 @@ qx.Class.define("qx.ui.form.ToggleButton", {
       apply: "_applyTriState",
       nullable: true,
       init: null
+    },
+
+    /**
+     * The behavior when the button is clicked. Only useful for tri-state checkboxes.
+     * If "cycle" is set, the button cycles through the states "disabled", "undetermined", and "enabled"
+     * If "toggle" is set, the button toggles between "disabled" and "enabled".
+     */
+    clickBehavior: {
+      check: ["cycle", "toggle"],
+      init: "toggle"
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

source/class/qx/ui/form/ToggleButton.js
```diff
@@ -114,6 +114,16 @@ qx.Class.define("qx.ui.form.ToggleButton", {
...
       nullable: true,
       init: null
+    },
+
+    /**
+     * The behavior when the button is clicked. Only useful for tri-state checkboxes.
+     * If "cycle" is set, the button cycles through the states "disabled", "undetermined", and "enabled"
+     * If "toggle" is set, the button toggles between "disabled" and "enabled".
+     */
+    clickBehavior: {
```
derrell, line 124: This property name is a bit of a misnomer, since it is applied whether the button is clicked or toggled elsewise, e.g., via the keyboard. Can we rename it to `executeBehavior` instead (and fix the doc comment too)?

BTW, this'll be a nice new feature.

derrell (changes requested):

oetiker (approved):

cboulanger (approved):

derrell (approved):

hkollmann (approved):

PR #10657, opened 2024-02-12
fix bug with tabview button appearances

patryk-m-malinowski:
This PR fixes an issue with appearances in a qx.ui.tabview.TabView, where if the first tab has its `tabVisibility` set to `excluded`, the first visible tab will not have the state `firstTab`, which will break appearance.


cboulanger (approved):
Looks good to me, but I can't test it visually at the moment.

oetiker (approved):
same here .,.. looks sensible, not testetd

hkollmann (approved):

goldim (changes requested):
I disagree that it is a **bug**. The first tab is first tab. Why it should move own state to other button otherwise if the first or last button is removed but in case of removal everything is fine. Also could you create an issue next time for discussion please and code example attached to it also would be great bc it is time for other devs to reproduce the issue.

johnspackman:
> Why it should move own state to other button

Because it looks weird from the users point of view - if the first tab's visibility is excluded, then it means that the user's "first" tab does not look like the first when the theme's appearance needs to style that "first" tab differently (eg due to borders etc).  

When a tab widget has a state like "first", this is only something which is passed to the appearance, it seems reasonable that this state is saying which tab "appears first"

goldim:
@johnspackman For me first tab is tab which first in array of tabs and not in visible ones. Maybe users' apps are based on old behavior and I may consider this like broken BC, am I right? I would like hear everyone's opinion on this problem. Maybe to create separate discussion?

goldim:
@p9malino26 In old implementation if I have 3 buttons, removed first and third, central button will have both states. In yours only lastTab.

johnspackman:
by all means create a separate discussion, or we can carry on here; I guess my point is that the state of "first" is only for the appearance, and if the tab is excluded then the appearance will never see a "first" tab (without this PR), in which case what would be the point of a "first" state.

source/class/qx/ui/tabview/TabView.js
```diff
@@ -637,6 +623,32 @@ qx.Class.define("qx.ui.tabview.TabView", {
...
+          if (!firstPage) {
+            firstPage = page;
+          }
+          lastPage = page;
+        }
+        page.removeState("firstTab");
+        page.removeState("lastTab");
+      }
+      firstPage?.addState("firstTab");
+      firstPage?.removeState("lastTab");
```
goldim, line 648: You already remove this state in cycle.

source/class/qx/ui/tabview/TabView.js
```diff
@@ -637,6 +623,32 @@ qx.Class.define("qx.ui.tabview.TabView", {
...
+
+      lastPage?.addState("lastTab");
+      lastPage?.removeState("firstTab");
```
goldim, line 651: The same

source/class/qx/ui/tabview/TabView.js
```diff
@@ -637,6 +623,32 @@ qx.Class.define("qx.ui.tabview.TabView", {
...
+
+    __onPageChangeTabVisibility() {
+      this.__updateFirstLastTabStates();
+    },
+
+    __updateFirstLastTabStates() {
+      let firstPage = null;
+      let lastPage = null;
+
+      for (let pages = this.getChildren(), i = 0; i < pages.length; i++) {
```
goldim, line 636: Could you move `let pages = this.getChildren()` out of cycle on separate line, please? Several instructions on one line - looks like a bad smell.

goldim (changes requested):

source/class/qx/ui/tabview/TabView.js
johnspackman: we use that pattern in a lot of places - I'm counting 290 `for` loops with multiple variables declared, one for the array and one for the index.  My regex is `for \((var|let) [^;]+,`

goldim: Ok, I won't insist. Just noticed. My view on good code is obviously different. May fix it, it is up to you.

goldim:
@johnspackman I can see that it is a bug actually. Classic Theme:
image

goldim:
@p9malino26 just fix what I suggested above and I gonna approve it.

goldim (approved):

derrell (approved):

PR #10662, opened 2024-02-19
Removed qx.locale.String

### LINKED ISSUE #10661: qx.locale.String methods
The methods of the class don't work and don't return what expected.
Also they are not used in qooxdoo framework code at all. I doubt they are used in users' apps too.
There are no tests for these methods too.
Maybe to remove them?

[Playground](https://tinyurl.com/24eh8kop).

qooxdoo version is 7.6.3.

goldim:
Tested output of methods of this class in v5-v7. The result is wrong. There are no tests for methods of this class. I may suggest that nobody uses these methods. These methods could be implemented but does it worth it?


derrell (approved):

cboulanger (approved):

johnspackman (approved):

hkollmann (approved):

oetiker (approved):
ok, if they did not work anyway ...

PR #10668, opened 2024-03-05
Replace cldr by intl

### LINKED ISSUE #10660: Intl instead of cldr
In qooxdoo we use a node package  "CLDR" for internalization. This package requires cldr-data which weights around 250 MG. And so dependencies are duplicated twice because of known-good folder this size will be 500 MB.

I've checked there is [Internalization API](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Intl) already exists for this purposes.

I think about moving to this API. What do you think? I checked it has introduced in 2011-2014 maybe there already was discussion about it.

### LINKED ISSUE #10664: Wrong narrow stand-alone locale de_DE
During implementing cldr functionality via Intl API. I am stucked with some tests of method `getDayNames`. 
What we have:
```javascript
var narrowDays = ["So", "Mo", "Di", "Mi", "Do", "Fr", "Sa"];
this.assertJsonEquals(
  narrowDays,
  Date.getDayNames("narrow", useLocale, "stand-alone").map(function (v) {
    return v + "";
  })
);
```
But if we have a look at [cldr data](https://github.com/papandreou/node-cldr/blob/master/3rdparty/cldr/common/main/de.xml) we will see that for narrow both as "format" and "stand-alone" case there are words with only one letter.
If it is a bug then how does it hand

goldim:
This is my first try to move qooxdoo from CLDR to Intl API and not final I think. This allows us to get rid of big dependency which weights too much (500 MB). During migration I encountered bunch of problems, questions and ambiguities which I tried to solve on my own. And that's why I am asking you help me to change some of my solutions if that's required. Also I've created a [discussion](https://github.com/qooxdoo/qooxdoo/discussions/10663) but didn't see a big involvement. No problem, I think the PR is right place.
I've added tests for most methods which have not been tested.
I would like to emphasize that some solutions via Intl API are elegant if you look at methods like `getStartWeek`, `isWeekend`, `getWeekendStart`, `getWeekendEnd` and will notice that there are no hardcoded values and exceptions.
But Intl API has some disadvantages. There are following moments which I would like to clarify with you:

- [ ] 1. There is no quarter part like Q, QQQ and so on. Fortunately in qooxdoo there are two cases: `yQ` and `yQQQ`. First one has only one exception and I just hardcoded. The second one is much more difficult and has a lot of different variants. I also hardcoded them but didn't count **child** and **parent** locale. If it is required I will added them too. For example `de_DE` has a parent `de` and `de_DE` may have different `yQQQ` format for example.
- [ ] 2. The calendar type, in current qooxdoo implementation there are next algorithm we just take first child in array (0-index) but I doubt if that's right way because for example if I take chinese locale `zh` and check what it takes:  "Buddhist calendar". To be honest I didn't find a solution to detect what calendar is most suitable for a locale. There is some functionality but it doesn't work in Firefox (`getCalendars` function). I would like to just use Gregorian calendar by default and add calendar method argument.
- [ ] 3. I removed most of CLDR mentions in the code and renamed methods but not all. Most of left ones are in compiler files. If required I rename and remove them too just tell me.
- [x] 4. There is `cldr_` prefix I decided to leave it because of BC. There is a `translate` method which uses the prefix. I can't remove it but may rename it for example to `intl_`.
- [x] 5. I changed some tests for examples: [this issue](https://github.com/qooxdoo/qooxdoo/issues/10664)
- [ ] 6. Some expected values of `getDateTimeFormat` differ from CLDR implementation. For example for `yM` format of `de_DE` there is `M/y` but Intl API results only `MM/y`. I will try to find why it is. Other cases work like in CLDR implementation.
- [x] 7. For example Intl methods get locale `de-DE` as argument but qooxdoo methods get `de_DE` (dash instead of underline). I introduced a private method in two classes which transforms locale (they are duplications). I would like to get rid of the duplication but don't know what common class or entity will have this method. Maybe `qx.locale.Manager`.
- [x] 8. Didn't rebuild and committed `npm-shrinkwrap.json`. Should I?
- [x] 9. I added myself as author in source head but didn't change it overall.
- [x] 10. `Intl.Locale` doesn't work in 123 version of Firefox. For BC should be used old implementation.


cboulanger (approved):
Can't really comment- thank you for all the work you've put into this! As long as the tests pass, I am happy.

oetiker (approved):
wow, thank you!

hkollmann (approved):

hkollmann:
8. Yes you should rebuild and add npm-shrinkwrap!

goldim:
@hkollmann Fixed.

johnspackman (approved):

goldim:
Just noticed that `Intl.Locale` is only supported since 124 version of Firefox and got failed in 123 of Ubuntu. I think to back old implementation of some methods.

derrell:
Unfortunately, I have no way to test this, as I have almost no experience with localization, and only the most basic of usage of it in my apps. I'm concerned, however, about the list of variants that aren't covered by this change. It's very likely to affect some users. My suggestion is to mark as deprecated, in the upcoming release of 7.x, as many of the things that are known to be issues with the new implementation as possible, and then actually release this change as part of 8.0. Some users will likely encounter non-BC changes that will require altering their code.

Since this PR is marked as "do not merge - v8", I will approve this PR as is. I added the additional "do not merge" tag just in as an added warning, probably not necessary, but doesn't hurt.

derrell (approved):

goldim:
@derrell it could be tested with our apps like widget browser for example. To see if the localization works set other locale and see how a datefield calendar is changed.

hkollmann:
@goldim Could you please write a short text in changelog.md with the breaking changes? Thanks!

goldim:
@hkollmann yes, I will do it near soon.

goldim:
@hkollmann Done. Is it enough?

hkollmann:
Did you test with firefox too? Seems that firefox and chrome have different settings in locale.

See https://github.com/qooxdoo/qooxdoo/actions/runs/18307570301/job/52132851557 the test/locale/Date  in branch_8_0_beta is not working for firefox 

Any idea?

goldim:
@hkollmann No, I didn't see it because firefox tests we have introduced after this PR was passed. I will check why tests are failed later.

hkollmann:
Could you do your work in the branch_8_0_beta please? It was a lot of work to fix all the conflicts. 
Therefore i would like to close all the merged branches as i suggested in the developers chat

goldim:
@hkollmann if you need this urgently then let's go. If it can wait till next week then I could try to fix it here and then to merge all.

hkollmann:
It's not urgently. You can fix it here and we can do a cherry pick. This pr is already merged into branch_8_0_beta

goldim:
@hkollmann hello, could you tell me how to rerun tests in this PR please?

hkollmann:
That was my problem too. I do not know. 
Tests will run in branch_8_0_beta

goldim:
@hkollmann wait, please. Don't close it. Gonna do one more fix

goldim:
@hkollmann I think I fixed tests for both browsers. You can move commits to beta branch.

hkollmann:
chromium: not ok 2870 - qx.test.locale.Date:testGetDateTimeFormat_GermanLocale - [22.20 ms] - 
Error: Expected 'M/y' but found 'MM.y'!
    at Error.construct (http://localhost:8080/transpiled/qx/type/BaseError.js:43:24)
    at Error.constructor (http://localhost:8080/transpiled/qx/Bootstrap.js:50:24)
    at Error.construct (http://localhost:8080/transpiled/qx/core/AssertionError.js:53:25)
    at new <anonymous> (http://localhost:8080/transpiled/qx/Bootstrap.js:50:24)
    at Object.__fail__P_177_1 (http://localhost:8080/transpiled/qx/core/Assert.js:80:21)
    at Object.assertEquals (http://localhost:8080/transpiled/qx/core/Assert.js:164:35)
    at wrapper.assertEquals (http://localhost:8080/transpiled/qx/core/MAssert.js:90:24)
    at wrapper.testGetDateTimeFormat_GermanLocale (http://localhost:8080/transpiled/qx/test/locale/Date.js:145:16)
    at http://localhost:8080/transpiled/qx/dev/unit/TestFunction.js:115:29
    at wrapper.run (http://localhost:8080/transpiled/qx/dev/unit/TestResult.js:217:38)
    at wrapper.run (http://localhost:8080/transpiled/qx/dev/unit/TestFunction.js:111:20)
    at wrapper.runTests (http://localhost:8080/transpiled/qx/dev/unit/MTestLoader.js:153:26)
    at next (http://localhost:8080/transpiled/qxl/testtapper/Application.js:204:27)
chromium: not ok 2871 - qx.test.locale.Date:testGetDateTimeFormat_ChineseLocale - [48.20 ms] - 
Error: Expected 'y/M' but found 'y年M月'!
    at Error.construct (http://localhost:8080/transpiled/qx/type/BaseError.js:43:24)
    at Error.constructor (http://localhost:8080/transpiled/qx/Bootstrap.js:50:24)
    at Error.construct (http://localhost:8080/transpiled/qx/core/AssertionError.js:53:25)
    at new <anonymous> (http://localhost:8080/transpiled/qx/Bootstrap.js:50:24)
    at Object.__fail__P_177_1 (http://localhost:8080/transpiled/qx/core/Assert.js:80:21)
    at Object.assertEquals (http://localhost:8080/transpiled/qx/core/Assert.js:164:35)
    at wrapper.assertEquals (http://localhost:8080/transpiled/qx/core/MAssert.js:90:24)
    at wrapper.testGetDateTimeFormat_ChineseLocale (http://localhost:8080/transpiled/qx/test/locale/Date.js:178:16)
    at http://localhost:8080/transpiled/qx/dev/unit/TestFunction.js:115:29
    at wrapper.run (http://localhost:8080/transpiled/qx/dev/unit/TestResult.js:217:38)
    at wrapper.run (http://localhost:8080/transpiled/qx/dev/unit/TestFunction.js:111:20)
    at wrapper.runTests (http://localhost:8080/transpiled/qx/dev/unit/MTestLoader.js:153:26)
    at next (http://localhost:8080/transpiled/qxl/testtapper/Application.js:204:27)

should we change the test to MM.y?

goldim:
@hkollmann No, I have to figure out why it happens.

goldim:
@hkollmann Interesting thing that there are no tests for `getDateTimeFormat` in master. I've checked output of this method in master and in new implementation and they both are equal `M/y`. The second thing is a decimal point instead of slash. Third thing is that doesn't work only this case. I get different result. What OS, node version you get error? Is it on autobuild server?

hkollmann:
see https://github.com/qooxdoo/qooxdoo/actions/runs/18383728020 it's from the github checks. let's see if it's the same in linux.

goldim:
The result on both OS are the same. On my local linux machine I get equal result which I committed yesterday. Gonna play with different environment. I think the problem is different calendar. In current CLDR implementation is gregorian if I got it right by program code.

hkollmann:
got the same error in ubuntu node. And Windows node too. Both in the github enviroment.

hkollmann:
merged into branch_8_0_beta

PR #10671, opened 2024-03-11
keyboard: always fireInputEvent whenever a charcode is provided

patryk-m-malinowski:
This PR fixes an issue in qx.event.handler.Keyboard where, if both a non-0 keycode and non-0 charcode are passed into _idealKeyHandler, this._fireInputEvent will not run, even though a text character key has been pressed. This will not fire a "keyinput" event when it should be fired. This scenario only occurs in some browsers, such as Firefox but not Chrome.

I know that the current keyboard system uses things which are deprecated, so it will be necessary to rewrite it in the future, but do you all think that this is a good fix for now?

### DIFF
--- a/source/class/qx/event/handler/Keyboard.js
+++ b/source/class/qx/event/handler/Keyboard.js
@@ -590,13 +590,12 @@ qx.Class.define("qx.event.handler.Keyboard", {
       var keyIdentifier;
 
       // Use: keyCode
-      let returnPromise = null;
       var tracker = {};
 
       if (keyCode || (!keyCode && !charCode)) {
         keyIdentifier = qx.event.util.Keyboard.keyCodeToIdentifier(keyCode);
 
-        returnPromise = qx.event.Utils.track(tracker, () =>
+        let returnPromise = qx.event.Utils.track(tracker, () =>
           this._fireSequenceEvent(domEvent, eventType, keyIdentifier)
         );
 
@@ -605,6 +604,7 @@ qx.Class.define("qx.event.handler.Keyboard", {
             this._fireInputEvent(domEvent, charCode)
           );
         }
+        return returnPromise;
       }
 
       // Use: charCode
@@ -620,8 +620,6 @@ qx.Class.define("qx.event.handler.Keyboard", {
           this._fireInputEvent(domEvent, charCode)
         );
       }
-
-      return returnPromise;
     },
 
     /*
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

cboulanger (approved):
I cannot test whether this has any side effects, so trusting you on this

oetiker (approved):

goldim:
@p9malino26 Thank you for the PR but could you attach some code example next time please to ease check your request?

source/class/qx/event/handler/Keyboard.js
```diff
@@ -590,27 +590,38 @@ qx.Class.define("qx.event.handler.Keyboard", {
...
 
       // Use: keyCode
+      let returnPromise = null;
+      var tracker = {};
+
       if (keyCode || (!keyCode && !charCode)) {
         keyIdentifier = qx.event.util.Keyboard.keyCodeToIdentifier(keyCode);
 
-        return this._fireSequenceEvent(domEvent, eventType, keyIdentifier);
+        returnPromise = qx.event.Utils.track(tracker, () =>
```
hkollmann, line 599: Why did you change the return statements to returnPromise? IMHO the return statements are much more readable. And    you did not change it in loine 619. This is a little bit inconsistent.

p9malino26:
To test, go to branch keyboard-test of [zx.ui.theme.avocado](https://github.com/ZenesisUK/zx.ui.theme.avocado/) type some stuff into the textfield and check if the correct stuff is printed in the console.

hkollmann (approved):

source/class/qx/event/handler/Keyboard.js
p9malino26: You're kind of right, it is better to declare and return `returnPromise` in the true branch of the if statement only. File updated.

However, because the `if (charcode)` statement may or may not execute (although it should execute 90% of the time), the `returnPromise` way is the only way to ensure that that promise only resolves when all the listeners have fully executed.

goldim:
>  check if the correct stuff is printed in the console

@p9malino26 Tbh I don't know what to check. What do you mean under "the correct stuff"?
I've tried both browsers: firefox (>124 version) and chrome-like browser. Used master of qooxdoo and your branch with fix and haven't noticed difference between outputs. I tried to press different keyboard buttons.
Thank you for the separate branch for testing but still be good decision to make an issue and fill the issue template.

p9malino26:
Just realised that this bug has already been fixed but my team didn't merge those changes into our fork of Qx yet! Apologies for brining up this PR.

PR #10680, opened 2024-03-28
fixes a bug where `qx.html.Image` breaks JSX

johnspackman:
fixes a bug where `qx.html.Image` is only suitable for `qx.ui.basic.Image` and should not be part of the JSX / VDOM because it breaks the `` tag


source/class/qx/html/Element.js
```diff
@@ -38,7 +38,8 @@
  *
  * NOTE: Instances of this class must be disposed of after use
  *
- * @require(qx.module.Animation)
+ * NOTE:: This classes used to require `qx.module.Animation` but this brings in a huge
```
goldim, line 41:
```suggestion
 * NOTE:: These classes used to require `qx.module.Animation` but this brings in a huge
```

source/class/qx/html/Element.js
```diff
@@ -1678,17 +1678,13 @@ qx.Class.define("qx.html.Element", {
      * @param direct {Boolean?false} Whether the values should be applied
      *    directly (without queuing)
      * @return {qx.html.Element} this object (for chaining support)
-     */
-    setStyles(map, direct) {
+     */ setStyles(map, direct) {
```
goldim, line 1681: Broken formatting

source/class/qx/html/Node.js
```diff
@@ -31,7 +31,9 @@
  *
  * NOTE: Instances of this class must be disposed of after use
  *
- * @require(qx.module.Animation)
+ * NOTE:: This classes used to require `qx.module.Animation` but this brings in a huge
```
goldim, line 34:
```suggestion
 * NOTE:: These classes used to require `qx.module.Animation` but this brings in a huge
```

goldim (changes requested):
Broken formatting possible somehow to fix?

source/class/qx/application/AbstractGui.js
```diff
@@ -101,9 +102,7 @@ qx.Class.define("qx.application.AbstractGui", {
     // interface method
     close(val) {
       // empty
-    },
-
-    // interface method
+    }, // interface method
```
derrell, line 105: The `// interface method` applies to `terminate`, not to `close`. I think this whitespace was deleted accidentally.

source/class/qx/html/Node.js
```diff
@@ -1261,14 +1263,11 @@ qx.Class.define("qx.html.Node", {
...
-
     /*
     ---------------------------------------------------------------------------
-      PROPERTY SUPPORT
+    PROPERTY SUPPORT
     ---------------------------------------------------------------------------
-    */
-
-    /**
+    */ /**
```
derrell, line 1270: Broken formatting

source/class/qx/html/Node.js
```diff
@@ -1284,8 +1283,7 @@ qx.Class.define("qx.html.Node", {
      * @param getter {Function?} function to read from the DOM
      * @param setter {Function?} function to copy to the DOM
      * @param serialize {Function?} function to serialize the value to HTML
-     */
-    registerProperty(key, get, set, serialize) {
+     */ registerProperty(key, get, set, serialize) {
```
derrell, line 1286: Broken formatting

source/class/qx/html/Node.js
```diff
@@ -1349,8 +1346,7 @@ qx.Class.define("qx.html.Node", {
      * @param direct {Boolean?false} Whether the value should be applied
      *    directly (without queuing)
      * @return {qx.html.Element} this object (for chaining support)
-     */
-    _setProperty(key, value, direct) {
+     */ _setProperty(key, value, direct) {
```
derrell, line 1349: Broken formatting

derrell (changes requested):

source/class/qx/html/Element.js
johnspackman: done

source/class/qx/html/Node.js
johnspackman: done

source/class/qx/html/Node.js
johnspackman: done

johnspackman:
WRT to the formatting issue - prettier keeps breaking things.  I think i've caught them all and put them back, and included turning off the commit hook that was causing it to happen every commit 🙄

goldim:
@johnspackman should this hook be uncommented now after changes?

johnspackman:
> @johnspackman should this hook be uncommented now after changes?

just been trying to make a PR to fix this, but it turns out that it is down to babel mangling the tree before prettier gets involved - and unfortunately, prettier isn't able to fix it.

This means that es6ify should not be used routinely on commits - but prettier should.  I've fixed it, although it won't work on Windows unless there is a bash

goldim (approved):

hkollmann (changes requested):

.github/hooks/pre-commit
```diff
@@ -7,6 +7,10 @@
 #
 #
 
-qx es6ify --git-pre-commit
+#qx es6ify --git-pre-commit
+for file in $(git diff --cached --name-only --diff-filter=ACMR | grep '\.js') ; do
+  npx prettier --write $file
+done
```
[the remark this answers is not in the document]
johnspackman: That would be useful, but I didn't have time and the current implementation breaks code formatting on every commit; I did a quick google, and as I understand it pre-commit hooks don't support windows except as bash scripts - I found an article saying that if you want to run a powershell cmd, you have to write a bash script and get it to run powershell.exe, eg:

```
#!/bin/sh
echo 
exec powershell.exe -ExecutionPolicy RemoteSigned -File '.\pre-commit.ps1'
exit
```

That may be an old article but the previous `pre-commit` had comment lines with `#`, and my Windows says that's not allowed.  Does the hook definitely run on your windows PC?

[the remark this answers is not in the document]
hkollmann: Yes the pre-commit with qx es6ify do work. # is'nt a valid comment sign in windows - that's the reason why you get the error.
What's about building a qx prettier command based on the es6ify command? es6ify just do the git calls with the parameter --git-pre-commit.
But this should be another PR.
So i suggest to remove the pre-commit commands completely in this PR and write an issue about creating a prettier command

hkollmann (approved):

.github/hooks/pre-commit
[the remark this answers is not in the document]
johnspackman: done https://github.com/qooxdoo/qooxdoo/issues/10682

oetiker (approved):

source/class/qx/html/Node.js
```diff
@@ -1264,11 +1267,9 @@ qx.Class.define("qx.html.Node", {
 
     /*
     ---------------------------------------------------------------------------
-      PROPERTY SUPPORT
+    PROPERTY SUPPORT
     ---------------------------------------------------------------------------
-    */
-
-    /**
+    */ /**
```
derrell, line 1272: This is showing as still-broken formatting.

derrell (changes requested):

derrell (approved):

cboulanger (approved):

PR #10681, opened 2024-03-28
adds a `privateArtifacts` option to `compile.json`

johnspackman:
This adds a `privateArtifacts` option to `compile.json` to allow all urls to be locked within an application directory

### DIFF
--- a/docs/development/compiler/configuration/compile.md
+++ b/docs/development/compiler/configuration/compile.md
@@ -579,7 +579,7 @@ It is up to you to implement the mapping inside your web server so that the
 
 ### Keeping all files within the application directory
 
-When the compiler generates an application(s), it creates `transpiled` and `resource`
+When the compiler generates applications, it creates `transpiled` and `resource`
 directories at the same level as the application name, and adds various working
 files, eg:
 
--- a/docs/development/compiler/configuration/compile.md
+++ b/docs/development/compiler/configuration/compile.md
@@ -577,6 +577,29 @@ is the URI.
 It is up to you to implement the mapping inside your web server so that the
 "/some/virtual/uri/path/qooxdoo" URI is able to load the files from `../qooxdoo`
 
+### Keeping all files within the application directory
+
+When the compiler generates an application(s), it creates `transpiled` and `resource`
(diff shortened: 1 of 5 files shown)


cboulanger (approved):

docs/development/compiler/configuration/compile.md
```diff
@@ -577,6 +577,29 @@ is the URI.
 It is up to you to implement the mapping inside your web server so that the
 "/some/virtual/uri/path/qooxdoo" URI is able to load the files from `../qooxdoo`
 
+### Keeping all files within the application directory
+
+When the compiler generates an application(s), it creates `transpiled` and `resource`
```
derrell, line 582: grammar problem with plural. How about "generates one or more applications..."

derrell (changes requested):

derrell (approved):

docs/development/compiler/configuration/compile.md
johnspackman: done

hkollmann (approved):

goldim:
@johnspackman I can't compile your PR bc the compiler complains on new option. Is it relative with `known-good`?

johnspackman:
> Has this to be updated https://qooxdoo.org/schema/compile-1-0-0.json

Yes, I think that gets updated automatically when we do a release, but until then you'll get errors 

> after compiling with option nothing changed: all files on the same with app folder in target folder.

It is intentional that the files are still in the same place - but if you look at your generated `myappname/index.js` you'll see that the path to the files (resources and transpiled directories) has changed; the intention of this change is to allow you to have a URL on your web server where you only expose (or set permissions for) the application directory, and you would need to mount virtual folders within that URL.

EG in order to have an application at the root (eg `http://www.mywebsite.com/`, the `transpiled` and `resource` directory cannot be in `../transpiled`

Where there are multiple apps in a build, the `resource` and `transpiled` directories on disk need to be shared, hence the need to add virtual folders.

goldim (approved):

oetiker (approved):

Issue #10362, opened 2022-02-13
Mixin again: Double Mixin with call to super gives stackoverflow 


hkollmann:
Here is the test:

    testDoubleMixin() {
      qx.Class.define("qx.Double", {
        extend: qx.B,
        members: {
          sayJuhu() {
            return super() + " C";
          }
        }
      });
      qx.Mixin.define("qx.MDA", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " MA";
          },
        }
      });
      qx.Mixin.define("qx.MDB", {
        members: {
          sayJuhu() {
            return super.sayJuhu() + " MB";
          },
        }
      });
      qx.Class.patch(qx.Double, qx.MDA);
      qx.Class.patch(qx.Double, qx.MDB);

      var o = new qx.C();
      this.assertEquals("A MA B MB C MC", o.sayJuhu());
      o.dispose();

    }

hkollmann:
fixed with @7.0.0-beta.10

PR #10689, opened 2024-05-07
improve mixin member super() logic

WillsterJohnsonAtZenesis:
See #10362, #10363

This commit fixes an issue with how the super of a member is determined from within a mixin.

We found that when patching mixins to multiple classes in the same inheritance chain, mixins on parent classes were completely skipped over.

Instead of going `class patched mixin -> class -> superclass' patched mixin -> superclass`, it was skipping over the superclass' patched mixin entirely, which is inconsistent with how an instance of that superclass would behave.

### DIFF
--- a/source/class/qx/Mixin.js
+++ b/source/class/qx/Mixin.js
@@ -371,16 +371,18 @@ qx.Bootstrap.define("qx.Mixin", {
               break;
             }
           }
+
           // Try looking in the class itself
           if (!fn && mixedInAt.prototype[methodName]) {
-            fn = mixedInAt.prototype[methodName].base;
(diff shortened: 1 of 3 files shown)


derrell (approved):
Approved as long as existing mixin tests still pass. I'm a bit worried about BC, but it seems to be an actual bug fix rather than a change in an expected behavior.

oetiker (approved):

johnspackman (approved):

cboulanger (approved):

goldim (approved):

hkollmann (approved):

PR #10702, opened 2024-06-18
qx.data.marshal.Json | ignore non-pojo objects when marshalling

patryk-m-malinowski:
This PR causes non-POJO objects to be ignored from the POJO that we pass into qx.data.marshal.Json.createModel. This behaviour is almost always undesired because we may have third party library objects in the POJO, which we don't want the marshaller to touch. 

Here is a demo qx browser app. Try running it on both this PR branch and master and compare the results.
[pr-demo.zip](https://github.com/user-attachments/files/15889696/pr-demo.zip)

### DIFF
--- a/source/class/qx/Bootstrap.js
+++ b/source/class/qx/Bootstrap.js
@@ -867,8 +867,8 @@ qx.Bootstrap.define("qx.Bootstrap", {
     },
 
     /**
-     * Whether the value is an object. Note that built-in types like Window are
-     * not reported to be objects.
+     * Whether the value is an object i.e. Object.prototype is its prototype or Object.prototype in its prototype chain.
+     * Note that built-in types like Window are not reported to be objects.
      *
      * @param value {var} Value to check.
      * @return {Boolean} Whether the value is an object.
@@ -878,6 +878,20 @@ qx.Bootstrap.define("qx.Bootstrap", {
         value !== undefined &&
         value !== null &&
         qx.Bootstrap.getClass(value) === "Object"
+      );
+    },
+
+    /**
+     * Whether the value is strictly a POJO. It's prototype must not inherit from Object.prototype but be strictly Object.prototype.
+
+     * @param {*} value
+     * @returns {boolean} Whether the value is strictly a POJO. 
+     */
+    isPojo(value) {
+      return (
+        value !== undefined &&
+        value !== null &&
+        Object.getPrototypeOf(value) === Object.prototype
       );
     },
 
--- a/source/class/qx/Bootstrap.js
+++ b/source/class/qx/Bootstrap.js
@@ -867,10 +867,13 @@ qx.Bootstrap.define("qx.Bootstrap", {
     },
 
     /**
-     * Whether the value is an object i.e. Object.prototype is its prototype or Object.prototype in its prototype chain.
-     * Note that built-in types like Window are not reported to be objects.
+     * Whether the value is an POJO (ie {foo: 1})
+     * or an object which is created from a ES6-style class or prototypical-inheritance-based class;
+     * If you need to determine whether something is a POJO and not created from a class, use isPojo instead
      *
-     * @param value {var} Value to check.
+     * Note that built-in types like Window are not deemed to be objects.
+     *
+     * @param {*} value value to check.
      * @return {Boolean} Whether the value is an object.
      */
     isObject(value) {
--- a/source/class/qx/Bootstrap.js
+++ b/source/class/qx/Bootstrap.js
@@ -882,20 +882,6 @@ qx.Bootstrap.define("qx.Bootstrap", {
     },
 
     /**
-     * Whether the value is strictly a POJO. It's prototype must not inherit from Object.prototype but be strictly Object.prototype.
-
-     * @param {*} value
-     * @returns {boolean} Whether the value is strictly a POJO. 
-     */
-    isPojo(value) {
-      return (
-        value !== undefined &&
-        value !== null &&
-        Object.getPrototypeOf(value) === Object.prototype
-      );
-    },
-
-    /**
      * Whether the value is a function.
      *
      * @param value {var} Value to check.
--- a/source/class/qx/lang/Type.js
+++ b/source/class/qx/lang/Type.js
@@ -52,14 +52,22 @@ qx.Bootstrap.define("qx.lang.Type", {
     isArray: qx.Bootstrap.isArray,
 
     /**
-     * Whether the value is an object. Note that built-in types like Window are
-     * not reported to be objects.
+     * Whether the value is an object i.e. Object.prototype is its prototype or Object.prototype is in its prototype chain.
+     * Note that built-in types like Window are not reported to be objects.
      *
      * @signature function(value)
-     * @param value {var} Value to check.
+     * @param {*} value value to check.
      * @return {Boolean} Whether the value is an object.
      */
     isObject: qx.Bootstrap.isObject,
+
+    /**
+     * Whether the value is strictly a POJO. It's prototype must not inherit from Object.prototype but be strictly Object.prototype.
+     * @signature function(value)
+     * @param {*} value
+     * @returns {Boolean} Whether the value is strictly a POJO.
+     */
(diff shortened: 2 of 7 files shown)


cboulanger (commented):
I think the PR makes sense, but it certainly is backward incompatible. There might be users that rely on the expectation that their custom objects will be serialized. I would suggest that the PR is part of the upcoming breaking release and it should include a message in the change log and maybe a note in the documentation.

derrell:
I think this is a useful addition, but I would go a step further than @cboulanger suggested, and add an additional parameter, `strictlyPojo`, for `createModel`, `toModel`, and `toClass` that could be passed when this new behavior is desired. The current behavior should remain the default if no changes are made to callers of these methods.

derrell (changes requested):
see my prior general comment

derrell:
I think we also probably need new tests that (a) test to ensure that the original behavior is retained if the original call arguments are passed (a value for `strictlyPojo` is not passed), and (b) tests of the new functionality showing that objects are marshalled as expected with both settings of `strictlyPojo`.

source/class/qx/Bootstrap.js
```diff
@@ -881,6 +881,20 @@ qx.Bootstrap.define("qx.Bootstrap", {
       );
     },
 
+    /**
```
johnspackman, line 884: `isPojo` does not need to be in `qx.Bootstrap`, which is effectively an internal class that exists in order to work around dependency management issues - IMHO it should move to `qx.lang.Type`

source/class/qx/lang/Type.js
```diff
@@ -52,15 +52,23 @@ qx.Bootstrap.define("qx.lang.Type", {
     isArray: qx.Bootstrap.isArray,
 
     /**
-     * Whether the value is an object. Note that built-in types like Window are
-     * not reported to be objects.
+     * Whether the value is an object i.e. Object.prototype is its prototype or Object.prototype is in its prototype chain.
```
johnspackman, line 55: it would be helpful to explain the net effect - eg:  "Objects can be either a POJO (ie `{}`) or an object which is created from a ES6-style class or prototypical-inheritance-based class; if you need to determine whether something is a POJO and _not_ created from a class, use `isPojo` instead"

source/class/qx/lang/Type.js
```diff
@@ -52,15 +52,23 @@ qx.Bootstrap.define("qx.lang.Type", {
...
+     * Note that built-in types like Window are not reported to be objects.
      *
      * @signature function(value)
-     * @param value {var} Value to check.
+     * @param {*} value value to check.
      * @return {Boolean} Whether the value is an object.
      */
     isObject: qx.Bootstrap.isObject,
 
+    /**
```
johnspackman, line 64: Same as above, the net effect is more useful in documentation than repeating the code, eg "Detects whether an object is just a plain POJO, ie `{}`; note that this excludes objects created via classes - @see `qx.lang.Type.isObject`

johnspackman (changes requested):

johnspackman:
I agree that there should be unit tests for both `isPojo` and the marshalling, but I'm inclined to say that it is actually a bug for the marshalling to try and marshall class instances; the reason that this came up is because we're using `BigNumber.js`, and any kind of instance which has internal class implementation cannot reasonably be marshalled - for example, in the case of `new BigNumber("567.89")` you get an object with `{ s: 1, e: 2, c: [ 567, 890000000 ] }` whereas `JSON.stringify` returns "567.89"

We could use an environment setting to preserve the old behaviour if required (the default being to use the new `isPojo`)?

oetiker:
> I agree that there should be unit tests for both `isPojo` and the marshalling, but I'm inclined to say that it is actually a bug for the marshalling to try and marshall class instances; the reason that this came up is because we're using `BigNumber.js`, and any kind of instance which has internal class implementation cannot reasonably be marshalled - for example, in the case of `new BigNumber("567.89")` you get an object with `{ s: 1, e: 2, c: [ 567, 890000000 ] }` whereas `JSON.stringify` returns "567.89"
> 
> We could use an environment setting to preserve the old behaviour if required (the default being to use the new `isPojo`)?
> 
> Also there should be an explanation & warning in the release notes

how about adding the feature in the next 7.x release but with the default behavior unchanged except that warnings would be emitted when foreign objects are encountered with the suggestion to enable the new mode ?

johnspackman:
> how about adding the feature in the next 7.x release but with the default behavior unchanged except that warnings would be emitted when foreign objects are encountered with the suggestion to enable the new mode ?

may be make it deprecated first, so the new behaviour is enabled by an environment setting and issues a deprecation warning with foreign objects?   And then make it "standard" in future?  

I'm struggling to see when it would be intentional to serialise a properly defined Object (ie instance of an ES6 class) into a JSON structure when marshalling; what happens at the moment is that an actual ES6 Object is _downgraded_ to JSON, when (imho) marshalling is about _upgrading_ to an object model

eg this data:

```
qx.data.marshal.Json.createModel([
  { title: "Lunch", price: new BigNumber(12.34) },
  { title: "Dinner", price: new BigNumber(23.45) }
]);
```

Is kind of equivalent to:

```
qx.Class.define("mypkg.MyClass, {
  properties: {
    title: { check: "String" },
    price: { check: "Object" }
});
new mypkg.MyClass().set({ title: "Lunch", price: { s: 1, e: 2, c: [ 12, 340000000 ] } });
new mypkg.MyClass().set({ title: "Dinner", price: { s: 1, e: 2, c: [ 23, 450000000 ] } });
```

derrell:
The current implementation of `isPojo` prevents object `x` here from being serialized. It is completely reasonable, I believe, to serialize `x`. As I think about this issue, I'm leaning towards John's expectation, but `isPojo` needs a bit of change as it's too limiting.

```
function isPojo(value)
{
  return (
    value !== undefined &&
    value !== null &&
    Object.getPrototypeOf(value) === Object.prototype);
}

const prototype = { num : 42 };
let x = { text : "hello", __proto__ : prototype };

console.log("prototype:", isPojo(prototype));
console.log("x:", isPojo(x));
```

johnspackman:
FYI the latest commit is not intended for review, @p9malino26 is off ill at the moment, we're pushing to his branch from his work PC here so that he can access it at home

p9malino26:
Just pushed some changes addressing the points we mentioned. Hopefully that should do!

source/class/qx/test/lang/Type.js
```diff
@@ -42,6 +42,156 @@ qx.Class.define("qx.test.lang.Type", {
...
+          this.foo = "bar";
+        }
+
+        method() {
+          return "method";
+        }
+      }
+
+      var obj = new ES6Class();
+      this.assertFalse(isPojo(obj));
```
derrell, line 192: Please add one more, as I don't think this is checked in other tests. (Feel free to rename variables to match the style of other tests you've implemented.)

```
const prototype = { num : 42 };
let x = { text : "hello", __proto__ : prototype };
this.assertTrue(isPojo(x);
```

derrell (changes requested):
Nice work on coming up with a full set of tests! Just one more I noticed that should be added.

p9malino26:
I think there is a test case which checks for your case. It's on line 122.
The only difference is that it uses Object.setPrototype instead of setting __proto__ manually because that's safer according to what I've read.

`//Object which has a POJO prototype, which in turn has a POJO prototype
      //Should return true
      var obj = {
        nuts: 345
      };

      var proto1 = {
        myMethod() {
          return "proto1.myMethod";
        }
      };

      Object.setPrototypeOf(obj, proto1);
      var proto2 = {
        myMethod() {
          return "proto2.myMethod";
        },

        method2() {
          return "proto2.method2";
        }
      };`

derrell (approved):

oetiker (approved):

hkollmann:
Any progress here?

p9malino26:
I think everyone's comments have been addressed now in this PR. Is this now ready to merge?

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

PR #10705, opened 2024-06-30
only bind readOnly property for ListItems. 

peuter:
Removes newly introduced requirement that every item put on the list must have the "readOnly" property.

### DIFF
--- a/source/class/qx/ui/form/IListItem.js
+++ b/source/class/qx/ui/form/IListItem.js
@@ -52,7 +52,7 @@ qx.Interface.define("qx.ui.form.IListItem", {
      * @return {Boolean|null}
      */
     setReadOnly(value) {
-      return arguments.length == 1;
+      this.assertArgumentsCount(arguments, 1, 1);
     },
 
     /**
--- a/source/class/qx/ui/form/IListItem.js
+++ b/source/class/qx/ui/form/IListItem.js
@@ -20,4 +20,51 @@
  * This is a marker interface for classes which can act as a children of
  * {@link qx.ui.form.List}
  */
-qx.Interface.define("qx.ui.form.IListItem", {});
+qx.Interface.define("qx.ui.form.IListItem", {
+    /*
+  *****************************************************************************
+     EVENTS
+  *****************************************************************************
+  */
+
+  events: {
+    /** Fired when the value was modified */
+    changeReadOnly: "qx.event.type.Data"
+  },
+
+  /*
+  *****************************************************************************
+     MEMBERS
+  *****************************************************************************
+  */
+
+  members: {
+    /*
+    ---------------------------------------------------------------------------
+      READONLY PROPERTY
+    ---------------------------------------------------------------------------
+    */
+
+    /**
+     * Sets the element's value.
+     *
+     * @param value {Boolean|null} The new value of the element.
+     * @return {Boolean|null}
+     */
+    setReadOnly(value) {
+      return arguments.length == 1;
+    },
+
+    /**
+     * Resets the element's value to its initial value.
+     */
+    resetReadOnly() {},
+
+    /**
+     * The element's user set value.
+     *
+     * @return {Boolean|null} The value.
+     */
+    getReadOnly() {}
+  }
+});
--- a/source/class/qx/ui/form/AbstractSelectBox.js
+++ b/source/class/qx/ui/form/AbstractSelectBox.js
@@ -30,7 +30,7 @@ qx.Class.define("qx.ui.form.AbstractSelectBox", {
   extend: qx.ui.core.Widget,
   include: [qx.ui.core.MRemoteChildrenHandling, qx.ui.form.MForm],
 
-  implement: [qx.ui.form.IForm, qx.ui.form.IListItem],
+  implement: [qx.ui.form.IForm],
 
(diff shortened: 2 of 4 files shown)


hkollmann (approved):

goldim (changes requested):
At first glance this looks like an innocent change but why we need to restrict an item to `ListItem`? In docs said there is `IListItem` interface which `form.List` uses. I've checked it and it is empty but it is another question. Maybe check for `readOnly` property and `toHashCode`?

peuter:
I hadn't seen the interface, that would be indeed the better option to fill and use it instead. I will change my MR

goldim:
@peuter also would it be great if I know what causes you to create this PR, for example I don't see point to add checking binding in remove hanlder. Do you have error in some use-case or crash?

peuter:
We use custom items in Lists (e.g. https://github.com/CometVisu/CometVisu/blob/develop/source/class/cv/ui/manager/Main.js#L1111) and the change introduces with https://github.com/qooxdoo/qooxdoo/pull/10674 causes crashes in our app because the list tried to bind to a property that does not exists in the custom list item. 

This PR adds a check and only creates the binding if the used item implements the IListItem interface to make sure that the property exists. And when you add a condition to create a binding you also have to check if the binding has been created before removing it.

source/class/qx/ui/form/IListItem.js
```diff
@@ -20,4 +20,51 @@
...
+    */
+
+    /**
+     * Sets the element's value.
+     *
+     * @param value {Boolean|null} The new value of the element.
+     * @return {Boolean|null}
+     */
+    setReadOnly(value) {
+      return arguments.length == 1;
```
cboulanger, line 55: I don't understand this line, is this an Interface-specific check?

peuter: Me neither, I just copied it from qx.ui.form.IBooleanForm, seems to be used in some interfaces.

cboulanger (approved):

source/class/qx/ui/form/IListItem.js
goldim: There is `qx.core.Assert.assert` statement for this goal. Not sure returning from interface method will work.

goldim:
Just found this [PR](https://github.com/qooxdoo/qooxdoo/pull/10674). It breaks BC. And we missed it. I think maybe to add separate interface and call it somehow as `IReadOnly`. Just looking at ToggleButton: it has the interface `IListItem` and it raises questions about it.

peuter:
I remove the IListItem interface from ToggleButton (CheckBox inherits from it and uses IListItem) and AbstractSelectBox again. That change was nonsense.

But in general I think using IListItem here and not introducing a new interface for readOnly is ok. It is documentet that usually items that are put on a list should implement this interface and the list binds readOnly to those only now.

goldim (approved):

source/class/qx/ui/form/IListItem.js
johnspackman: > I don't understand this line, is this an Interface-specific check?

in theory interface members provide some validation of the parameters, given that the generator was unable to do checks (the compiled could, but doesn't).  But it's quite random whether its used or not

oetiker (approved):

johnspackman:
Hi,

Thank you for your email - please note that I will be on annual leave and unavailable until July 14th 2024.
  
Regards
John Spackman

source/class/qx/ui/form/IListItem.js
derrell: I recalled that the return value from a function in an Interface could be used for validation, but I can't find that documented anywhere and there are no tests to ensure that method of validation continues to work. There are a couple of tests where the Interface method returns `true`, but none it returns `false`, and the meaning of the return value is entirely undocumented. The only documented way of doing parameter validation, it seems, is with the various `assert*` methods. The use of a boolean return value, as used here (and elsewhere) should really be replaced with a call to one of the `assert` methods.

derrell: And I've now confirmed in the code that the return value from the "precondition" function is ignored:

Interface.js:525
```
          // call precondition
          preCondition.apply(this, arguments);
```

peuter: Seems to have been changed 15 years ago ;-)  https://github.com/qooxdoo/qooxdoo/commit/84c789da40aa3e73919de5f461dd3b20bef75f37

The documentation in interface still describes the old behaviour and needs to be changed too. Also all Interfaces with return values should be changed to use assertions instead. But that sould be done on another PR.

derrell: Ah! They say that memory is the second thing to go -- I can't recall what the first is... but I did recall correctly. :-)

peuter:
I anything else needed here? Because of the BC problem in the current release I would like to get this merged and a new bugfix qx version released.

goldim:
@peuter Hello, I think we merge it soon to master, but about publishing new release I can not promise.

hkollmann:
There is just missing one other approve ...

cboulanger (approved):

johnspackman (approved):

PR #10707, opened 2024-07-15
Property `init` ignored in setter

### LINKED ISSUE #10704: Property `init` ignored in setter
**Describe the bug**
When a property has an init value, and it is called to be set with that same value, the apply method will be called with old and new value as the same value.

**To Reproduce**
```js
qx.Class.define("test.stuff.Test", {
  extend: qx.core.Object,

  properties: {
    flag: {
      init: true,
      apply: "apply",
    },
  },

  members: {
    apply() {
      console.log(arguments);
    },
  },
});

let tester = new test.stuff.Test();

let flagValue = tester.getFlag(); 
console.log(flagValue); // true

tester.setFlag(true);
```

**Expected behavior**
The apply method should not be called, as the value of the property has not changed (set to `true` when already `true` from initial value).
The call to `getFlag` returns the correct value (the init value), so it would be expected that the stored value of the property would be set and be available for comparison in `setFlag`.

**Actual behavior**
The apply method is called with args `(newValue=true, oldValue=true)`, the property is not set to a value internally (despite the prior call to `getFlag`).

**Screenshots**
Screenshot 2024-06-27 at 09 14 40

 - OS: MacOS 13.6.3 
 - Browser: Chrome Dev 
 - Version: 128

**Additional context**

Stepping into the generated setter, `this.$$useinit_[name]` is undefined. This prevents the correct current value, `this.$$init_[name]`, from being used and compared to. This doesn't seem to occur when the first call to set uses a different value to the init value.

WillsterJohnsonAtZenesis:
closes #10704

- rename `useinit` to `initialized`
- change value meaning;
  - `true` -> the value has been initialized (same as `useinit=false`)
  - any falsey value -> the value has not been initialized (same as `useinit=true`)  

Assuming that falsey values, like `undefined`, mean 'we should use the initial value' prevents the issue in #10704.

Previously something like `setFoo` would check if `this.$$useinit_foo`, ignoring `this.$$init_foo` if it's falsey. This is incorrect, as `this.$$useinit_foo` is **always** `undefined` during `setFoo` if it is the first property method to be called for `foo`.

Now, with falsey values as 'yes, use it' instead, `this.$$initialized_foo` being `undefined` is correctly interpreted if `setFoo` is the first property method called for `foo`.

### DIFF
--- a/source/class/qx/core/Property.js
+++ b/source/class/qx/core/Property.js
@@ -1649,7 +1649,7 @@ qx.Bootstrap.define("qx.core.Property", {
         code.push("computed=this.", this.$$store.theme[name], ";");
         code.push("else if(this.", this.$$store.init[name], "!==undefined){");
         code.push("computed=this.", this.$$store.init[name], ";");
-        code.push("this.", this.$$store.initialized[name], "=false;");
+        code.push("delete this.", this.$$store.initialized[name], ";");
         code.push("}");
       } else {
         // Use runtime value as it has higher priority
@@ -1688,7 +1688,7 @@ qx.Bootstrap.define("qx.core.Property", {
         code.push("computed=this.", this.$$store.theme[name], ";");
         code.push("else if(this.", this.$$store.init[name], "!==undefined){");
         code.push("computed=this.", this.$$store.init[name], ";");
-        code.push("this.", this.$$store.initialized[name], "=false;");
+        code.push("delete this.", this.$$store.initialized[name], ";");
         code.push("}");
       } else {
         if (variant === "setRuntime") {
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

WillsterJohnsonAtZenesis:
Changed this to a draft as I still seem to be experiencing some issues

derrell (changes requested):
The renamed `store.useinit` is referenced in `qx.core.Object` but not in this PR

WillsterJohnsonAtZenesis:
It turns out fixing this is a lot more effort than it's worth, particularly considering that the new property system & qooxdoo 8.0.0 are going to come out relatively soon.

The issue with the changes in this PR is that they don't account for `inheritable` properties (eg, `enabled`). accounting for that moves the problem somewhere else, and so on. I can see it taking a long time to figure out exactly what the current behaviors are and exactly how to address the issue in #10704 without causing problems elsewhere.

It's probably best to just leave this unfixed, at least until someone with more time available decides to fix it. I think we should leave the issue open, but I'm closing this PR.

derrell:
These rabbit-hole details are the reason it took a year+ to fully implement the new class/property system of Version 8.0. If someone is reimplementing that work, expect to spend a great amount of time working through these details to maintain BC.

PR #10708, opened 2024-07-19
Allow qx.data.controller.List to have selection that is not in model

patryk-m-malinowski:
The non-breaking PR adds a new property to qx.data.controller.MSelection/qx.data.controller.List/qx.data.controller.Tree, namely `allowSelectionNotInModel`. This allows the selection to include items which are not in the model property. When this is the case, the target widget will only select the items which are in the model. The programmer can set the model and the selection in any order they want.

Also, the controller will not reset its selection when the model is set, and the only source that can change the selection in the controller is a changeSelection fired by the target itself. 

This is very useful when the selection of a controller is bound to a property of a business object, and the model of a controller is dynamic. This is because, if we change the business object being edited, we have to change the model of the controller, and it will fire a changeSelection event, which may overwrite data in the business object. Until now, the programmer would have to remove bindings when they change the model and then add them back in again, or use boolean flags to prevent the change in selection from backfiring and corrupting business objects. This leads us into writing cleaner code.

This PR also implements a method called `syncSelectionWithModel` which removes any items from the selection that are not in the model.

I have attached a demo app for you to see how this PR would be useful in practice: 
[demo.zip](https://github.com/user-attachments/files/16311424/demo.zip)
The app allows the user to customize their car styles (or themes), like in a racing game! Works in this PR branch only.

Note: While MSelection in included in qx.data.controller.List and qx.data.controller.Tree, this only works for controller.List ATM, but it does not break `controller.Tree`. I don't know if we should support the Tree controller because myself and my colleagues at Zenesis are working on qxl.DataGrid, which is due to be the de facto standard tree soon?


hkollmann (approved):

source/class/qx/data/controller/List.js
```diff
@@ -207,14 +207,10 @@ qx.Class.define("qx.data.controller.List", {
...
-  /*
-  *****************************************************************************
-     MEMBERS
-  *****************************************************************************
-  */
-
   members: {
-    // private members
+    /**
+     * Listner ID for the 'change' event of property `model`
```
goldim, line 212:
```suggestion
     * Listener ID for the 'change' event of property `model`
```

goldim (approved):

source/class/qx/data/controller/MSelection.js
```diff
@@ -61,6 +61,22 @@ qx.Mixin.define("qx.data.controller.MSelection", {
       event: "changeSelection",
       apply: "_applySelection",
       init: null
+    },
+
+    /**
+     * If set to true, the selection property will be allowed to have items which are not in the model property.
+     * If the model property changes, this will not cause the selection property to be reset,
+     * If the selection contains items not in the model,
+     * the target widget will only select the items that are in the model.
```
oetiker, line 70: Can you elaborate a bit, I don't understand what will happen here with the 'non-model' item.

oetiker (changes requested):
see inline question

oetiker (approved):

cboulanger (approved):

johnspackman (approved):

hkollmann:
Thanks @p9malino26 !

PR #10715, opened 2024-08-16
Add svg to the qx.io.ImageLoader class data url regexp tag options.

sqville:
This would allow "svg+xml" data uri option for all objects and properties that accept an image URL string as a value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ tmp/
 /tmp
 /bootstrap
 /qx
-/ville
 
 # CSS compiler
 /source/resource/qx/mobile/scss/.sass-cache/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -32,6 +32,7 @@ tmp/
 /tmp
 /bootstrap
 /qx
+/ville
```
goldim, line 35: not sure it should be in framework

goldim (changes requested):

.gitignore
hkollmann: This should be reverted

sqville: Sorry about that. I reverted the commit and made a third commit with just the "svg" addition to ImageLoader

goldim (approved):

cboulanger (approved):

hkollmann (approved):

derrell (approved):

oetiker (approved):

PR #10730, opened 2024-10-09
Fixed null value of `qx.version`

goldim:
Fixed #10726 . There are extracting qooxdoo version from analyser before it was filled with libraries.

PR #10731, opened 2024-10-10
Added checking content on undefined value to fix testRedefineItem

goldim:
Fast fix for unit test `testRedefineItem` which I've spotted in my pull request https://github.com/qooxdoo/qooxdoo/pull/10730.
It is not only one test if I get rid of this test next failed test is also related with `qx.ui.form.Form`. It happens not because of my changes. If I only create new branch from master and try to run tests they are failed too. I've checked requests before and suddenly they were passed. I don't know where to dig in. I think if there is no solution found then I think this request has to be merged.
These failed tests are passed fine locally (if copying them to other qooxdoo project and run) and with different node versions. Is there a way to show debug traces in tests on remote server?

### DIFF
--- a/source/class/qx/ui/core/Widget.js
+++ b/source/class/qx/ui/core/Widget.js
@@ -2128,7 +2128,13 @@ qx.Class.define("qx.ui.core.Widget", {
     }, // property apply
     _applyVisibility(value, old) {
       var content = this.getContentElement();
-      if (value === "visible") {
+      if (content) {
+        if (value === "visible") {
+          content.show();
+        } else {
+          content.hide();
+        } // only force a layout update if visibility change from/to "exclude"
+      }
         content.show();
       } else {
         if (content){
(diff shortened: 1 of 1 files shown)


source/class/qx/ui/core/Widget.js
```diff
@@ -2131,7 +2131,9 @@ qx.Class.define("qx.ui.core.Widget", {
       if (value === "visible") {
```
derrell, line 2131:
```suggestion
      if (content) {
        if (value === "visible") {
          content.show();
        } else {
          content.hide();
        } // only force a layout update if visibility change from/to "exclude"
      }
```

derrell (changes requested):

derrell (approved):

derrell (changes requested):
No, that suggestion was supposed to replace both the .show and .hide code

goldim:
Oh, sorry. Fixed.

derrell (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

johnspackman (approved):

PR #10732, opened 2024-10-10
Label size cache

johnspackman:
This PR adds caching to the label size calculations, which can have a huge effect on performance.  In our app, we had a dashboard with multiple datagrid trees, and on reasonably good i9 laptops could see CPU usage spikes for 20-30 seconds.  My Apple Silicon M2 Max would see spikes to 20%.

### DIFF
--- a/source/class/qx/bom/Label.js
+++ b/source/class/qx/bom/Label.js
@@ -317,8 +317,7 @@ qx.Bootstrap.define("qx.bom.Label", {
      */
     getTextSize(text, styles) {
       let cacheKey = this.__getCacheKey(styles);
-      let cache = qx.bom.Label.__sizeCache[cacheKey];
-      let size = cache?.sizes[text];
+      let size = this.__getCachedSize(cacheKey, text);
       if (size !== undefined) {
         return size;
       }
--- a/source/class/qx/bom/Label.js
+++ b/source/class/qx/bom/Label.js
@@ -39,6 +39,9 @@ qx.Bootstrap.define("qx.bom.Label", {
       letterSpacing: 1
     },
 
+    /** @type{Object<String,Object>} cache of label sizes */
+    __sizeCache: {},
+
     /**
      * Generates the helper DOM element for text measuring
      *
@@ -286,6 +289,13 @@ qx.Bootstrap.define("qx.bom.Label", {
      * @return {Map} A map with preferred <code>width</code> and <code>height</code>.
      */
     getHtmlSize(content, styles, width) {
+      let cacheKey = this.__getCacheKey(styles, width);
+      let cache = qx.bom.Label.__sizeCache[cacheKey];
+      let size = cache?.sizes[content];
+      if (size !== undefined) {
+        return size;
+      }
+
       var element = this._htmlElement || this.__prepareHtml();
 
       // apply width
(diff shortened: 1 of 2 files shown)


source/class/qx/bom/Label.js
```diff
@@ -304,6 +316,13 @@ qx.Bootstrap.define("qx.bom.Label", {
      * @return {Map} A map with preferred <code>width</code> and <code>height</code>.
      */
     getTextSize(text, styles) {
+      let cacheKey = this.__getCacheKey(styles);
+      let cache = qx.bom.Label.__sizeCache[cacheKey];
+      let size = cache?.sizes[text];
```
derrell, line 321: `__sizeCache` is opaque, i.e., manipulated only by its accessor functions, everyplace except here. Suggest replacing this reference to the internal `sizes` with

```
let size = __getCachedSize(cacheKey, text);
```

derrell (changes requested):

hkollmann (approved):

oetiker (approved):

source/class/qx/bom/Label.js
johnspackman: done

goldim:
@johnspackman could you clarify me please for what purpose `lru` array serves? I don't see usage of it and where it is cleared.

goldim:
@johnspackman hello, could you check gitter please?

source/class/qx/bom/Label.js
```diff
@@ -286,14 +289,23 @@ qx.Bootstrap.define("qx.bom.Label", {
      * @return {Map} A map with preferred <code>width</code> and <code>height</code>.
      */
     getHtmlSize(content, styles, width) {
+      let cacheKey = this.__getCacheKey(styles, width);
+      let cache = qx.bom.Label.__sizeCache[cacheKey];
+      let size = cache?.sizes[content];
```
derrell, line 294: Sorry, I missed this one the first time. Lines 293-4 should also probably be replaced with use of the getter:

```
let size = this.__getCachedSize(cacheKey, text);
```

derrell (changes requested):

hkollmann:
@johnspackman : Could you solve the change request please?

hkollmann:
Any news here?

hkollmann:
@johnspackman : Could you resolve the issues?

cboulanger (approved):

hkollmann (approved):

derrell (approved):

goldim (changes requested):
Asked before about the `lru`. I don't see usage of it anywhere and this array just grows. Maybe I don't know something. Is it supposed to use in the next PRs? If yes I gonna approve it.

johnspackman:
sorry folks i will get onto this today

goldim:
@johnspackman Tried to compile my project with this PR. It breaks my app. And the error during the compilation is qx.bom.Label: `[293,48] warning: Unresolved use of symbol text`.  Need to be fixed anyway.

johnspackman:
finally, this should be it :)  I think the only thing not done was the compilation error, everything is now done

goldim (approved):

Issue #10728, opened 2024-10-08
Exclude files in browserify parse process


goldim:
> There is already browserified package which I use in my app and the second browserify process during qooxdoo compilation is not needed for this JS script. But except process time which spend on this library I get a lot of annoying warnings about require()d modules. In my browserified script they are just links and new browserify process can't find them. I can just add option (for target or configuration in general?) which will turn off this warning (missing event fired). Or just search for option in qooxdoo compiler code which turn off parsing this file and just add content to general stream.

Returning to the browserify problem. I digged into the problem about missing dependencies and found out that you can pass via `noParse` option of browserify files which content could be skipped and just added into output of browserify process. There was a bug which already fixed (17.0.1 released, and current version of browserify is 17.0.0). The bug: there only absolute paths have been worked. Now relative paths do work  relatively qooxdoo app root.

Before I wanted just ignore the missing error during parse process of browserify but now I can exclude these huge files from the browserify process. I think this option will be useful for other qooxdoo users.
Question is how to do add this option at configuration. Now browserify library inside qooxdoo compiler only get some hard coded options. For target only? For all targets just in root section?

I know there is an opportunity to change configuration of babel (babelify) maybe to add browserify options the same way?

oetiker:
could this be an option to add into the `compile.json` file ?

goldim:
@oetiker as simple way yes. But I think if there will be more parameters for browserify then adding a section is right decision . But it may not happen and there is only my needed parameter for excluding files.
Another question is to add argument at top level of `compile.json` or include it for a target?

derrell:
There are some configuration variables that may be placed either in top-level or specific to a target. I believe adding it to a target overrides the top-level setting. That's likely a good solution when you aren't sure what section of `compile.json` to add it to. Also, I'd add a `browserify` map which contains `noParse` rather than simply adding `noParse` as the new configuration parameter on its own.

PR #10733, opened 2024-10-17
Browserify options in compile.json

goldim:
Implemented #10728 

Clarify moments:
1) In root of compile.json I did browserify.options like babel.options but there are no so far other section like babel.plugins . Should I make browserifyOptions instead of browserify.options?
2) Should I extend doc and write examples like for babel?
3) Should qx.lang.Object.mergeWith overwrite base browserify options? Now it doesn't.
4) Are there usage examples added like for babel? Or url reference is enough?
5) Take notice I updated version of browserify from 17.0.0 to 17.0.1. There  is change which I have introduced into Browserify repository.

### DIFF
--- a/source/class/qx/tool/compiler/targets/meta/Browserify.js
+++ b/source/class/qx/tool/compiler/targets/meta/Browserify.js
@@ -166,7 +166,7 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
         };
         qx.lang.Object.mergeWith(
           options,
-          this.getAppMeta().getAnalyser().getBrowserifyConfig() || {}
+          this.getAppMeta().getAnalyser().getBrowserifyConfig()?.options || {}
         );
         let b = browserify([], options);
 
--- a/source/class/qx/tool/compiler/targets/meta/Browserify.js
+++ b/source/class/qx/tool/compiler/targets/meta/Browserify.js
@@ -166,7 +166,8 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
         };
         qx.lang.Object.mergeWith(
           options,
-          this.getAppMeta().getAnalyser().getBrowserifyConfig()?.options || {}
+          this.getAppMeta().getAnalyser().getBrowserifyConfig()?.options || {},
+          false
         );
         let b = browserify([], options);
 
--- a/source/class/qx/tool/compiler/targets/meta/Browserify.js
+++ b/source/class/qx/tool/compiler/targets/meta/Browserify.js
@@ -158,12 +158,17 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
       builtins.process = builtins._process;
 
       return new Promise((resolve, reject) => {
-        let b = browserify([], {
+        const options = {
           builtins: builtins,
           ignoreMissing: true,
           insertGlobals: true,
           detectGlobals: true
-        });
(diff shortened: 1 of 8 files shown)


oetiker:
I think the consistent naming is a good thing.

oetiker (approved):

hkollmann:
> Implemented #10728
> 
> Clarify moments:
> 
> 1. In root of compile.json I did browserify.options like babel.options but there are no so far other section like babel.plugins . Should I make browserifyOptions instead of browserify.options?

No. This name is consistent

> 2. Should I extend doc and write examples like for babel?
If you like

> 3. Should qx.lang.Object.mergeWith overwrite base browserify options? Now it doesn't.
The base options are mandatory. So do not overwrite them.

> 4. Are there usage examples added like for babel? Or url reference is enough?
IMHO reference is enough. Use case of this options are rare.

> 5. Take notice I updated version of browserify from 17.0.0 to 17.0.1. There  is change which I have introduced into Browserify repository.
+1

hkollmann (approved):

cboulanger (approved):

source/class/qx/tool/compiler/targets/meta/Browserify.js
```diff
@@ -158,12 +158,17 @@ qx.Class.define("qx.tool.compiler.targets.meta.Browserify", {
...
+        const options = {
           builtins: builtins,
           ignoreMissing: true,
           insertGlobals: true,
           detectGlobals: true
-        });
+        };
+        qx.lang.Object.mergeWith(
+          options,
+          this.getAppMeta().getAnalyser().getBrowserifyConfig()?.options || {}
```
derrell, line 169: This appears to allow the user-provided options to override the default ones. I believe the default ones are mandatory to accomplish the goals of this feature.

source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -318,6 +318,10 @@
             "type": "object",
             "description": "Options given to @babel/preset-env. With this options the output type of babel can be defined. For details see here: <https://babeljs.io/docs/en/babel-preset-env#options>. They can be overridden per target."
           },
+          "browserifyOptions": {
+            "type": "object",
+            "description": "Options given to browserify. For details see here: <https://github.com/browserify/browserify#usage>. They can be overridden per target."
```
derrell, line 323: `https://github.com/browserify/browserify#usage` is not modified in this PR and I haven't seen any other PR that updates it...

derrell (changes requested):

source/resource/qx/tool/schema/compile-1-0-0.json
goldim: Don't understand this notice

source/class/qx/tool/compiler/targets/meta/Browserify.js
goldim: Fixed

source/resource/qx/tool/schema/compile-1-0-0.json
derrell: Disregard. I misread it.

derrell (approved):

goldim:
@johnspackman Hello, could you check this PR please?

johnspackman (approved):

PR #10736, opened 2024-11-06
Use Map instead of Array to register event listeners

### LINKED ISSUE #10620: Slow event registry removal with a large number of objects
As discussed in [Gitter in May 2023](https://matrix.to/#/!PPyLxiIFPbhEqMbDgU:gitter.im/$GPjsL0aw9F1l0SGB1E9HXfCZ_dzvqK_hUu-HXXDEHw8?via=gitter.im&via=matrix.org&via=matrix.freyachat.eu), with dynamic theming enabled and a large number of `LayoutItem`s created. A simple fix was not apparent so I am logging the issue so it is recorded.

The lookup of the events in the event registry becomes exponentially slow. This was noticed for the destruction of a window containing many controls, compounded by the fact that some other elements in the application were not being leaked, adding to the number of event registrations. 

I built a simple repl, although this is not something you would likely find in an application, but it does highlight the inefficiency of the event lookup.
 
**Repl:** https://tinyurl.com/yujzyz2t
This creates 30,000 TextFields, which is already slow to destroy (about 5 seconds), but if you bump the number up to 60,000 then it is very much slower to destroy (around 20 seconds).

The problem seems to be related to looping/manipulation of the event registry array for each event removal in this piece of code:
https://github.com/qooxdoo/qooxdoo/blob/49d5fbcc322ed71a6ab8d998048705766ef2e731/source/class/qx/event/Manager.js#L702-L715

My suggestion to improve was to possibly store the event listeners in a map so they could be removed by key.

Comment from @derrell at the time was:

> Yeah, I can see how that would happen. There may be some O(n^2) looping going on there. The good news is that it should be refactorable in a backward-compatible fashion, I think. I'll look into it. It'll be a bit of work, so not just a 1-day project.

And then:
> Unfortunately, it can't be refactored in a BC fashion. It had appeared that the data structures for maintaining all of the listeners was a purely internal matter, but unfortunately, there is a public interface for retrieving all listeners, which simply returns the existing top-level data structure. Although there are very few uses of that interface within qooxdoo, because it's public, that interface could be used in users' apps, so it can't be changed except with a major version number bump.

rad-pat:
To fix https://github.com/qooxdoo/qooxdoo/issues/10620, change the storage of event listener callbacks from an Array to using a Map, keyed by the unique Id. I think I've made sure the exposed APIs return the same as before.

The real speed-up comes from them storing the unique IDs for the added listeners and then using those for the removal from the Map. This makes a huge difference where there are many listeners in the Array (Map now) for the `qx.theme.manager.Meta` or the `qx.locale.Manager` when dynamic theming or locales are enabled.

### DIFF
--- a/source/class/qx/event/Manager.js
+++ b/source/class/qx/event/Manager.js
@@ -66,7 +66,7 @@ qx.Class.define("qx.event.Manager", {
     }
 
     // Registry for event listeners
-    this.__listeners = {};
+    this.__listeners = new Map();
 
     // The handler and dispatcher instances
     this.__handlers = {};
@@ -264,18 +264,41 @@ qx.Class.define("qx.event.Manager", {
      *       null when no listener were found.
      */
     getListeners(target, type, capture) {
-      var targetKey =
-        target.$$hash || qx.core.ObjectRegistry.toHashCode(target);
-      var targetMap = this.__listeners[targetKey];
+      var targetKey = target.$$hash || qx.core.ObjectRegistry.toHashCode(target);
+      var targetMap = this.__listeners.get(targetKey);
 
       if (!targetMap) {
         return null;
       }
 
       var entryKey = type + (capture ? "|capture" : "|bubble");
-      var entryList = targetMap[entryKey];
+      var entryMap = targetMap.get(entryKey);
 
-      return entryList ? entryList.concat() : null;
+      if (entryMap && entryMap.size > 0) {
+        var listeners = [...entryMap.values()];
+
+        return new Proxy(listeners, {
+          deleteProperty(target, property) {
+            if (property !== "length") {
+              var listener = target[property];
+              entryMap.delete(listener.unique);
+            }
+            delete target[property];
+            return true;
+          },
+          set(target, property, value, receiver) {
+            if (property !== "length") {
+              if (!value.unique) {
+                throw new Error("Cannot store a listener without a unique id. Use addListener()");
+              }
+              entryMap[value.unique] = value;
+            }
+            target[property] = value;
+            return true;
+          }
+        });
+      }
+      return null;
     },
 
     /**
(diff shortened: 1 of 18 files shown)


johnspackman (approved):

hkollmann (approved):

cboulanger (approved):
Seems sensible - I trust that this has been extensively tested not to break anything.

derrell (approved):

rad-pat:
> Seems sensible - I trust that this has been extensively tested not to break anything.

The tests seemed to do a reasonable job of covering the changes, but we can run with this for a while before merging to ensure that nothing is broken.

goldim (changes requested):
I think after removing listener you have to assign null/undefined to private member.

johnspackman:
Hi,

Thank you for your email - I will be on annual leave and unavailable on Tuesday 12th and Wednesday 13th November 2024.
  
Regards
John Spackman

oetiker (dismissed):

rad-pat:
> I think after removing listener you have to assign null/undefined to private member.
When looking at assigning null to the stored listenerIds, it seems that other listener ids are not cleared. Presumably they are garbage collected just fine as they're just a string.

goldim:
> > I think after removing listener you have to assign null/undefined to private member.
> 
> When looking at assigning null to the stored listenerIds, it seems that other listener ids are not cleared. Presumably they are garbage collected just fine as they're just a string.

You want to say that they won't be reusable?

goldim:
Just looked through, yes, most cases - it is cleared in destructor but for example see https://github.com/qooxdoo/qooxdoo/blob/2cab79dc2fa95ab1c6e289c5a01a8a2dcf52c732/source/class/qx/ui/menu/AbstractButton.js . In this file u can twice call `_on changeCommand` and in the second time it also will have string value and `removeListener` will be called.

goldim:
Actually there is a strange logic in this method. There is variable checking on null/undefined and if yes it returns control but if you see the next code of this method there are still checking on null/undefined. I think this is for new issue. I think removing listener should be before the first return otherwise there is no point to remove it.

goldim:
Also you changed API of method `getAllListeners` - returns JS object but now it is Map. It is broken BC. And you have to keep tests untouched. Try to save return value, transform Map into JS object.

goldim:
@rad-pat I've just read the issue conversation and don't know what to say. I didn't participate in it and don't see conclusion. Is there way to save somehow return value of the method?

oetiker (commented):
thanks @goldim for investigating

rad-pat:
> Also you changed API of method `getAllListeners` - returns JS object but now it is Map. It is broken BC. And you have to keep tests untouched. Try to transform Map into JS object when returning value.

The `getAllListeners` was marked as internal, presumably created for the purpose of checking what listeners exist before and after the tests are ran. I've updated all the test instances where it is used to cope with the Map instead, do you want me to revert those changes to tests? I thought because it was internal it would be ok to update that.

oetiker:
> > Also you changed API of method `getAllListeners` - returns JS object but now it is Map. It is broken BC. And you have to keep tests untouched. Try to transform Map into JS object when returning value.
> 
> The `getAllListeners` was marked as internal, presumably created for the purpose of checking what listeners exist before and after the tests are ran. I've updated all the test instances where it is used to cope with the Map instead, do you want me to revert those changes to tests? I thought because it was internal it would be ok to update that.

Unfortunately the `@internal` designation does not mean there is no stable API. The only things that ARE internal are the `__variables` and `__methods`. Unfortunately with this method we are exposing the internal data structure ... 

If you look through the original bug discussion you can see some sample code I created to show how the data interface can be kept ...

rad-pat:
OK, I'll hopefully get back on this at some point next week and see if I can fix it up. 👍

rad-pat:
@oetiker, @goldim I have revised the changes, removed changes to the tests. Added Proxy objects to the results of `getAllListeners` and `getListeners` methods to help monitor any changes made to those arrays.

goldim:
@rad-pat I will check it out on these days or on this weekend.

goldim (approved):

hkollmann:
@oetiker: Is this ok for you now?

oetiker (approved):

Issue #10739, opened 2024-12-02
qx.io.jsonrpc.Client.sendRequest() raises an uncatchable error on server side errors


bjorn-fischer:
I am using ```qx.io.jsonrpc.Client``` for communication with servers:
```
    _callAsync : function(callback, method, data, id) {
      const client = new qx.io.jsonrpc.Client(this.getUrl());
      client.sendRequest(this.getService() + "." + method, data)
      .then(result => callback(result, undefined, id))
      .catch(err => {
        if (err.code == 8) {
          // 8 => stale or uninitialized auth token/rune
          qxcbt.ui.LoginDialog.getInstance().addListenerOnce("runeUpdated", function(e) {
            data.rune = e.getData();
            this._callAsync(callback, method, data, id);
          }, this);
          qxcbt.ui.LoginDialog.getInstance().updateRune();
        } else {
          callback(undefined, err, id);
        }
      });
    }
```
```_callAsync()``` acts like a simple wrapper that automatically acquires an authentication token if the JsonRPC server complains about a missing or stale token and then tries again.
Although this code works I _always_ get an unhandled promise rejection (even before my ```.catch()``` handler is executed) whenever the JsonRPC server responds with an error. In Firefox this works non the less, the token gets updated on error code 8 and the callback is called otherwise. But the unhandled promise rejection is always logged in the console and in debug mode execution always stops.

**Comment by cboulanger** [comment]
Thanks @goldim for taking this on. If the error is hard to track, maybe setting the environment variable `qx.io.jsonrpc.debug` to `true` might help.

**Comment by bjorn-fischer** [comment]
Maybe it is not related, but this looks odd:
```
--- a/source/class/qx/Promise.js
+++ b/source/class/qx/Promise.js
@@ -937,7 +937,7 @@ qx.Class.define("qx.Promise", {
       } else {
         qx.bom.Event.addNativeListener(
           window,
-          "unhandledrejection",
+          "unhandledRejection",
           qx.Promise.__onUnhandledRejection.bind(this)
         );
       }
```

goldim:
@bjorn-fischer I can just remove throw exception because it is not processed (no way to do it).

bjorn-fischer:
maybe depend it on **qx.debug**?

cboulanger:
[...]

Issue #9877, opened 2020-03-03
Switch to native Promises in framework code


cboulanger:
Given universal support of Promises in Browsers and Node (and the availability of polyfills), there isn't a reason to provide our own Promise implementation (`qx.Promise`) any more once qooxdoo switches to support ECMAScript 6, since it introduces backwards-compatibility issues that we don't want to have to deal with, and might also be a performance issue.

johnspackman:
i dont think we should deprecate qx.Promise - it adds a load of features that are useful (eg static methods such as `.all` etc as well as member methods) and which can be reimplemented on top of native Promise; however, ditching bluebird seems reasonable if all supported browsers provide native promises.

goldim:
Tried to cut off blubird promises and use only native promises instead. But got senior trouble:
Native promise can't be cancellable. There is some work around but not sure about it.

johnspackman:
I think it's possible, eg I've just read this: https://medium.com/@masnun/creating-cancellable-promises-33bf4b9da39c

cboulanger:
Let's try to do this for qx. 8.0 - there is no need for an underlying Bluebird implementation in 2025...

cboulanger:
@johnspackman I just looked at the [source for qx.Promise](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/Promise.js) and it looks like you already set it up so that a different implementation could be used - however, there is no API for that, is there? Do you have any clue where the ["unhandled promise rejection" error](https://github.com/qooxdoo/qooxdoo/issues/10739) could originate from that [only shows up in Firefox](https://github.com/qooxdoo/qooxdoo/pull/10742#issuecomment-2566984707)?

cboulanger:
As you know the code best, do you think we could have a BC fix that would allow me to use a native promise instead of Bluebird when creating a new `qx.Promise` instance? Could Bluebird be prevented from overriding the `Promise` symbol?

cboulanger:
It seems like a lot of the binding stuff has become obsolete since we have `() => {}` - maybe the wrapping could be omitted for when the promise resolver is a lamda?

johnspackman:
> As you know the code best, do you think we could have a BC fix that would allow me to use a native promise instead of Bluebird when creating a new `qx.Promise` instance?

[...]

PR #10742, opened 2024-12-27
Fix for #10739

cboulanger:
This rewrites part of `qx.io.jsonrpc.(Client|Request)` in order to fix some incorrect Promise handling. ~~However, the bug cannot be fully fixed with the current use of `qx.Promise`, which has a bug/incompatibility that produces the "unhandled rejection" error. To make current code work, it is necessary to add `qx.io.jsonrpc.protocol.Request.useNativePromise();` to your code, which has been introduced to be able to switch to the native Promise implementation without breaking compatibility (although I very much doubt that anyone has been relying on this dependency). In qooxdoo 8.0, the implementation will use the native `Promise` implementation by default and this method will be deprecated.~~  See https://github.com/qooxdoo/qooxdoo/issues/10739#issuecomment-2581947780 for an update on the causes of the bug.

### DIFF
--- a/source/class/qx/io/jsonrpc/Client.js
+++ b/source/class/qx/io/jsonrpc/Client.js
@@ -126,7 +126,7 @@ qx.Class.define("qx.io.jsonrpc.Client", {
 
     /**
      * Fires "error" event and rejects the pending requests' promises.
-     * The method will be renamed and made private in v8.  
+     * The method will be renamed and made private in v8.
      * @param exception
      * @private
      */
@@ -147,15 +147,23 @@ qx.Class.define("qx.io.jsonrpc.Client", {
      * Send the given JSON-RPC message object using the configured transport
      *
      * @param {qx.io.jsonrpc.protocol.Message|qx.io.jsonrpc.protocol.Batch} message
-     * @return {qx.Promise} Promise that resolves (with no data) when the message has been successfully 
+     * @return {qx.Promise} Promise that resolves (with no data) when the message has been successfully
      * sent out. As this means different things depending on the transport implementation, it is best
      * not to base any kind of business logic on the fulfillment of that promise.
-     * 
-     *  Rejections are forwarded 
-     * to the request's promise, which also fulfills with the jsonrpc response. Thus, there is no  need to 
-     * await the returned promise. Instead, await the request's promise, which and rejects either with a 
-     * {@link qx.io.exception.Transport} in case of a transport error or with {@link qx.io.protocol.Error} 
-     * in case of a jsonrpc error. The returned promise will never be rejected.
+     *
+     * The current behavior is to return the promise from the transport `send()` implementation, which
+     * might be rejected with a {@link qx.io.exception.Transport} in case of a transport error.
+     * This has caused problems with "unhandled promise rejection" errors, so a new behaviour will be
+     * the default in v8: The returned promise is already resolved, and any rejection of the transport
+     * promise will only be forwarded to the promise(s) of the request(s) contained in the the `message`
+     * argument. The returned promise will never be rejected. This behavior can be enabled by setting
+     * the environment variable `qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest` to `true` in v7,
+     * which will be deprecated in v8.
+     *
+     * In any case, the result of the jsonrpc request is retrieved by awaiting {@link qx.io.jsonrpc.protocol.Request}'s
+     * promise, which is resolved with the jsonrpc server's response or is rejected either  with a
+     * {@link qx.io.exception.Transport} in case of a transport error or with {@link qx.io.protocol.Error}
+     * in case of a jsonrpc error.
      */
     async send(message) {
       if (
@@ -205,6 +213,7 @@ qx.Class.define("qx.io.jsonrpc.Client", {
       if (qx.core.Environment.get("qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest")) {
         // forward rejections to the requests' promises, which will be standard behavior in v8
         transportPromise.catch(error => {
+          // wrap error in transport exception
           if (!(error instanceof qx.io.exception.Transport)) {
             error = new qx.io.exception.Transport(
               error.toString(),
@@ -214,10 +223,10 @@ qx.Class.define("qx.io.jsonrpc.Client", {
           }
           this._throwTransportException(error)
         })
-        // return a resolved promise so that the actual completion of the transport is not awaited 
+        // return a resolved promise so that the actual completion of the transport is not awaited
         return qx.Promise.resolve()
       } else  {
-        // default behavior: return promise from transport
+        // default behavior in v7: return promise from transport
         return transportPromise;
       }
     },
@@ -388,7 +397,7 @@ qx.Class.define("qx.io.jsonrpc.Client", {
      * If true, log detailed information on the jsonrpc traffic in the console
      */
     "qx.io.jsonrpc.debug": false,
-    
+
     /**
      * If true, forward transport errors to the running jsonrpc requests' promise instead of rejecting
      * the promise returned by {@link #send}. Default is `false`. Behavior will change in v8 and this
--- a/source/class/qx/io/jsonrpc/Client.js
+++ b/source/class/qx/io/jsonrpc/Client.js
@@ -157,8 +157,9 @@ qx.Class.define("qx.io.jsonrpc.Client", {
      * the default in v8: The returned promise is already resolved, and any rejection of the transport
      * promise will only be forwarded to the promise(s) of the request(s) contained in the the `message`
      * argument. The returned promise will never be rejected. This behavior can be enabled by setting
-     * the environment variable `qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest` to `true` in v7,
-     * which will be deprecated in v8.
+     * the environment variable `qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest` to `true` in v7.
+     * In v8, the default of `qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest` will become `true`,
+     * and that environment variable will become deprecated.
      *
      * In any case, the result of the jsonrpc request is retrieved by awaiting {@link qx.io.jsonrpc.protocol.Request}'s
      * promise, which is resolved with the jsonrpc server's response or is rejected either  with a
@@ -400,8 +401,10 @@ qx.Class.define("qx.io.jsonrpc.Client", {
 
     /**
      * If true, forward transport errors to the running jsonrpc requests' promise instead of rejecting
-     * the promise returned by {@link #send}. Default is `false`. Behavior will change in v8 and this
-     * setting will be deprecated.
+     * the promise returned by {@link #send}. Default is `false`.
+
+     * @deprecated
+     * Behavior in v8 will be as if the environment variable value is `true`, but the environment variable will no longer be available.
      */
     "qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest": false
   }
--- a/source/class/qx/io/jsonrpc/Client.js
+++ b/source/class/qx/io/jsonrpc/Client.js
@@ -262,7 +262,7 @@ qx.Class.define("qx.io.jsonrpc.Client", {
         // rethrow
         throw error;
       }
-      return 
+      return  result;
     },
 
     /**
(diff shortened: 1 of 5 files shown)


goldim (approved):

johnspackman:
Hi,

Thank you for your email - I will be on annual leave from 16th December and unavailable until January 6th.  I am available prior to 16th December
  
Regards
John Spackman

hkollmann (approved):

johnspackman:
@cboulanger is there a reproducable sample I can use to look at this?  The `qx.Promise` API should be compatible, and removing it in 8.0 will mean introducing a breaking change, so IMHO it could be useful to identify exactly what's going on (going wrong) with qx.Promise now.  Maybe add some deprecated traps or something to qx.Promise?

cboulanger:
@johnspackman Sorry if I was unclear - I didn't mean to remove `qx.Promise` from the framework, just from the implementation of `qx.io.jsonrpc.Request`, where I also don't need the additional API that `qx.Promise` has over the native Promise implementation.

johnspackman:
In my mind, qx.Promise is advertised as being backward compatible with native Promise - actually it kind of has to be, because Promises are how the compiler (or the native JS interpretter) implements async/await.

I agree we should reduce the qx.Promise code to not include Bluebird, ie use native Promise implementation, and keep the additional API that qx.Promise provides in v8.0

But I'd like to know why qx.Promise is causing a problem here, because I'm wondering if this is showcasing a bug in qx.Promise?

cboulanger:
I'll write a playground example showing the incompatibility later, @goldim has something in the original bug that showed it, if I remember correctly.

cboulanger:
@johnspackman Thanks for insisting - interestingly, I cannot reproduce the problem now and don't get the error when commenting out `qx.io.jsonrpc.protocol.Request.useNativePromise();` in the tests. This indicates that the problem was not with `qx.Promise` after all but with my original code. I still would prefer the native Promise but that does not justify introducing additional API methods if there is no real need. It would make more sense to replace Bluebird with a native Promise in `qx.Promise` which would probably result in a slight speed-up of qooxdoo on the whole. But that's outside the scope of this issue. 

What I'll do:
- run the CI framework tests using `qx.Promise`
- if the tests pass, remove `qx.io.jsonrpc.protocol.Request.useNativePromise();`

For qx 8, I can then still switch to native Promises or simply use an updated version of `qx.Promise`.

CC @bjorn-fischer

cboulanger:
Ok, everything works without changing the Promise implementation, so I think this PR can be merged as-is now.

Ha, I even opened an [issue about switching to native promises four years ago](https://github.com/qooxdoo/qooxdoo/issues/9877) that I have totally forgot about.

source/class/qx/io/jsonrpc/protocol/Request.js
```diff
@@ -62,20 +69,41 @@ qx.Class.define("qx.io.jsonrpc.protocol.Request", {
       id = ++qx.io.jsonrpc.protocol.Request.__current_request_id;
     }
     this.set({ id });
-    this.__promise = new qx.Promise();
```
johnspackman, line 65: you shouldnt need to do this - the only reason that `new qx.Promise()` exists is to support an externally accessible `resolve()` and `reject()`.

IE the previous code is correct, and your Request.resolve method becomes

```
resolve(value) {
      this.__promise.resolve(value);
    },
```

Although it seems strange that the `Request`'s promise is resolved/rejected  by `Client`?

My guess is that the reason there was a problem in the first place is because `Request.handleTransportException` rejected the promise and so does `Client.handleMessage`.  

That makes it sound like the two classes are arguing over who is responsible for the promise?  EG maybe only `Client` should resolve/reject?

cboulanger: > you shouldnt need to do this - the only reason that new qx.Promise() exists is to support an externally accessible resolve() and reject().

You are right. This will make the fix even simpler. Local tests pass, let's see if the CI tests do.

> That makes it sound like the two classes are arguing over who is responsible for the promise? EG maybe only Client should resolve/reject?

Good point - I don't remember why I thought that a transport error should be handled at the level of the Request instead of the Client - maybe it should throw an exception and the Client should then handle it by rejecting the Request's promise. As this is a change of behavior, we would need to postpone it to qx 8.

cboulanger: Problem is: The test I added doesn't seem to catch the problem raised in #10739. I just ran [testIssue10739()](https://github.com/cboulanger/qooxdoo/blob/Issue10739/source/class/qx/test/io/jsonrpc/Client.js#L342) with current master and the unhandled promise exception was not raised - so unfortunately, we are not smarter than before! :-( We need @bjorn-fischer to test whether the PR in its current form (i.e. by not using the qx.Promise API) helps in any way.

goldim:
@cboulanger Hello, I've checked your last changes and they show up unhanded error. But with useNativePromise there is no the error in console. I know it is workaround to skip/suppress the error.
Environment which I use:
```javascript
// Application js
        const url = "http://localhost:8081";
        qx.io.jsonrpc.protocol.Request.useNativePromise();
        const transport = new qx.io.transport.Xhr(url);
        transport.getTransportImpl().setRequestHeader("Access-Control-Allow-Origin", "*");
        const client = new qx.io.jsonrpc.Client(transport);
       
        try {
          let result = await client.sendRequest("hello_world", {});
          console.log("result", result);
        }
        catch(ex) {
          console.log("error", ex);
        }
```

Server which sends error (Run it with node command):
```javascript
const http = require("http");

const server = http.createServer((req, res) => {
  res.statusCode = 200;
  res.setHeader("Content-Type", "text/plain");
  const anwser = {"error": {code: 333}, "id": 1};
  res.end(`${JSON.stringify(anwser)}\n`);
});

const PORT = 8081;
server.listen(PORT, () => {
  console.log(`Server running at http://localhost:${PORT}/`);
});
``
And the main thing - CORS Everywhere pluggin for you browser. I use firefox.

**Review by goldim — CHANGES_REQUESTED** [review]
PR now doesn't fix the issue.

**Comment by cboulanger** [comment]
Ok, now I understand: this is a Firefox issue - it shows up in the [Issue10739 test](https://github.com/cboulanger/qooxdoo/blob/Issue10739/source/class/qx/test/io/jsonrpc/Client.js#L342), but only in Firefox, without making the test fail. I have been running test in Chrome so I didn't see it. An it might be uncatchable. I wonder if that's a qooxdoo or a Bluebird problem.

**Comment by cboulanger** [comment]
I can confirm that reverting this PR to 2ccce613dd6263bce54081756ec6e922720970df removes the "unhandled promise rejection error" in Firefox.

File: source/class/qx/io/jsonrpc/protocol/Request.js (hunk 1 above)
Comment (cboulanger) [reply source/class/qx/io/jsonrpc/protocol/Request.js:65]: Reverted to use qx.Promise.

File: source/class/qx/io/jsonrpc/Client.js [hunk 2]
```diff
@@ -208,8 +208,17 @@ qx.Class.define("qx.io.jsonrpc.Client", {
         params
       );
 
-      await this.send(request);
-      return await request.getPromise();
+      // sending request asychnonously without waiting for it completing, 
+      // since we're awaiting the fulfillment of the request promise below
+      this.send(request);
```
derrell, line 213: This change appears to be simply masking the problem, not fixing it. `this.send()` is, in fact, an async call, so not awaiting it (or .then/.catch it) is simply hiding the rejection that was previously trapped by the unhandled rejection handler. No?

derrell (changes requested):

source/class/qx/io/jsonrpc/Client.js
cboulanger: This might be the case but it is not relevant because what we're waiting for is the fulfilment/rejection of the request's promise, not the promise of the `send()` method - which does not have any data in any case. The promise comes from the transport implementation's `send()` method; however, you are right that this masks an error that is raised by this method. I add a catch for that.

cboulanger: The problem is that this "double await" - awaiting the transport to finish sending out the request and then awaiting thre request's promise to fulfil, seems to be creating the "unhandled promise rejection", since they seem to get in the way of each other. And we really do not need to await the completion of `send()`, we just need to handle if it rejects by rejecting the requests promise with a "Transport Error".

cboulanger: It don't think this can happen - depending on the transport implementation the `send()` promise either fullfills when the request goes out and the answer has come in (XHR, fetch), and only afterwards is the request's promise settled, or it fulfills when the request has gone out (e.g. when using the PostMessage transport when communicating with a WeWorker). In any case, this promise is rejected when anything goes wrong during transport. The request's promise is settled afterwards in every case.

cboulanger: we just need to handle `send()`-Promise rejections, I am just working on that.

cboulanger: I now implemented it so the transport promise can be awaited as well, and using an environment variable to keep the old behaviour fully unless configured otherwise - thanks for the suggestions. I haven't updated the tests fully so they don't pass with the future behaviour, but will do so when I get to it.

source/class/qx/io/jsonrpc/Client.js
```diff
@@ -125,28 +125,37 @@ qx.Class.define("qx.io.jsonrpc.Client", {
...
-     * @return {qx.Promise} Promise that resolves (with no data)
-     * when the message has been successfully sent out, and rejects
-     * when there is an error or a cancellation up to that point.
+     * @return {qx.Promise} Promise that resolves (with no data) when the message has been successfully 
+     * sent out. As this means different things depending on the transport implementation, it is best
+     * not to base any kind of business logic on the fulfillment of that promise.
+     * 
+     *  Rejections are forwarded 
+     * to the request's promise, which also fulfills with the jsonrpc response. Thus, there is no  need to 
+     * await the returned promise. Instead, await the request's promise, which and rejects either with a
```
derrell, line 156: "which and rejects"... Should that be "which fulfills and rejects" or "which rejects"?

derrell (changes requested):

source/class/qx/io/jsonrpc/Client.js
cboulanger: Oh darn I forgot to update the docs here, sorry. Will do when I fix the tests.

cboulanger: Updated the docs. Also, we should now have a fully BC fix.

source/class/qx/io/jsonrpc/Client.js
```diff
@@ -350,6 +393,16 @@ qx.Class.define("qx.io.jsonrpc.Client", {
...
-    "qx.io.jsonrpc.debug": false
+    /**
+     * If true, log detailed information on the jsonrpc traffic in the console
+     */
+    "qx.io.jsonrpc.debug": false,
+
+    /**
+     * If true, forward transport errors to the running jsonrpc requests' promise instead of rejecting
+     * the promise returned by {@link #send}. Default is `false`. Behavior will change in v8 and this
+     * setting will be deprecated.
```
derrell, line 404:
```suggestion
     * the promise returned by {@link #send}. Default is `false`.
     
     * @deprecated
     * Behavior in v8 will be as if the environment variable value is `true`, but the environment variable will no longer be available.
```

derrell (changes requested):

source/class/qx/io/jsonrpc/Client.js
```diff
@@ -125,28 +125,45 @@ qx.Class.define("qx.io.jsonrpc.Client", {
...
+     * not to base any kind of business logic on the fulfillment of that promise.
+     *
+     * The current behavior is to return the promise from the transport `send()` implementation, which
+     * might be rejected with a {@link qx.io.exception.Transport} in case of a transport error.
+     * This has caused problems with "unhandled promise rejection" errors, so a new behaviour will be
+     * the default in v8: The returned promise is already resolved, and any rejection of the transport
+     * promise will only be forwarded to the promise(s) of the request(s) contained in the the `message`
+     * argument. The returned promise will never be rejected. This behavior can be enabled by setting
+     * the environment variable `qx.io.jsonrpc.forwardTransportPromiseRejectionToRequest` to `true` in v7,
+     * which will be deprecated in v8.
```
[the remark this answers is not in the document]
cboulanger: Corrected, thanks!

source/class/qx/io/jsonrpc/Client.js
cboulanger: Thanks!

derrell (approved):

goldim:
@cboulanger Hello, sorry. I can't run test example today. I've only checked the code so far. Gonna run examples tomorrow.

goldim (approved):

oetiker (approved):

hkollmann:
Just a question: Is the change of the promise after fixing the protocol really necessary?

johnspackman (approved):

PR #10744, opened 2024-12-28
Test client with postMessage transport receiving out-of-order responses 

cboulanger:
This is an additional test suite for the `qx.io.jsonrpc.Client` using a `postMessage` communication transport, making sure that the communication architecture works as intended. I have also added some assertions for Promises which might or might not be worth adding to the the other assertions in `qx.dev.unit`.


cboulanger:
Depends on #10742

hkollmann (approved):

johnspackman (approved):

oetiker (approved):

source/class/qx/io/jsonrpc/protocol/Request.js
```diff
@@ -29,12 +29,35 @@ qx.Class.define("qx.io.jsonrpc.protocol.Request", {
...
+
+    /**
+     * The default Promise constructor used in qx 7 is qx.Promise. Unfortunately,
+     * it has a bug/incompatibiity that leads to uncaught promise rejection errors
+     * Use call this static method to use the native Promise object instead. This
+     * will be the default in qooxdoo 8.0, when this function will we deprecated.
+     */
+    useNativePromise() {
+      qx.io.jsonrpc.protocol.Request.__Promise_constructor = Promise;
     }
```
derrell, line 61: Since this is a global change (not sometimes in an app qx.Promise while other times, Promise), wouldn't this best be implemented as an environment setting rather than requiring a code change to call the `useNativePromise` method?

derrell (changes requested):

source/class/qx/io/jsonrpc/protocol/Request.js
cboulanger: Thatnk you, that is a good idea. The best solution, however, is if we can fix the underlying problem rather than the symptom - I hope @johnspackman can have a look soon and we can fix `qx.Promise` to not raise the "uncaught promise rejection" error in Firefox in a BC way, then this PR won't be necessary at all.

derrell: I'm not convinced there's a problem with `qx.Promise`. In Chrome, in the playground, I get the unhandled rejection error whether using qx promises or native promises. Try pasting this code into the qooxdoo playground. Run it as is, then toggle the value of `native` and run it again. I get the same results either way, and this is in Chrome. 

```javascript
let native = true;
const promiseImpl = native ? Promise : qx.Promise;

(async () => {
	try {
		const promise = new promiseImpl((resolve, reject) => {
			reject(new Error("Rejecting first promise"));
		});
		await new promiseImpl((resolve) => setTimeout(() => {
		        console.log("Resolving second promise")
		        resolve();
		}, 1000));
		await promise;
	} catch (reason) {
		console.log("Caught the first promise rejection");
	}
})();
```

I tried trapping the `unhandledrejection` error as follows, but it is trapped too late:

```javascript
let native = true;
const promiseImpl = native ? Promise : qx.Promise;

globalThis.addEventListener("unhandledrejection", (event) => {
  console.log("Trapped unhandled rejection with error=", event);
  event.preventDefault();
});

(async () => {
	try {
		const promise = new promiseImpl((resolve, reject) => {
			reject(new Error("Rejecting first promise"));
		});
		await new promiseImpl((resolve) => setTimeout(() => {
		        console.log("Resolving second promise")
		        resolve();
		}, 1000));
		await promise;
	} catch (reason) {
		console.log("Caught the first promise rejection");
	}
})();
```

The closest I came to something desirable, was to attach a `catch` handler at promise creation time. This eliminates the error in the console, but also prevents the `catch` of the try/catch from being called. If I rethrow the error, we are, alas, back to the original problem of unhandled rejection.

```javascript
let native = true;
const promiseImpl = native ? Promise : qx.Promise;

(async () => {
	try {
		const promise = new promiseImpl((resolve, reject) => {
			reject(new Error("Rejecting first promise"));
		}).catch((e) => {
		  console.log("Trapped the first promise rejection");
		});
		await new promiseImpl((resolve) => setTimeout(() => {
		        console.log("Resolving second promise")
		        resolve();
		}, 1000));
		await promise;
	} catch (reason) {
		console.log("Caught the first promise rejection");
	}
})();
```

In all cases, this has nothing to do with `qx.Promise`. The results are the same whether using it or native promises.

cboulanger: There is something wierd going on, though. If you look at the console output in Chrome using the native Promise, the error messages are from `qx.Promise`:

grafik

cboulanger: In contrast, in an `about:blank` window with no qooxdoo present, the console output hides the error message when the promise resolves:

grafik

This suggests that this is a qooxdoo issue probably related to `qx.Promise`.

cboulanger: but we're in a wrong issue, this should be discussed in #9877 or #10739.

johnspackman: I agree with @derrell, Im getting the same in Chrome and Firefox and teh same with `qx.Promise` and `Promise`.

I think I have an explanation - I've updated #10739 comment here: https://github.com/qooxdoo/qooxdoo/issues/10739#issuecomment-2571361288

cboulanger: But did you try this outside the qooxdoo context - because then the behavior is different (see above).

cboulanger: In contrast to Chrome, in a `about:blank` window in Firefox, we get the promise rejection error:

grafik

cboulanger:
@derrell PR no longer uses native promises, tests pass locally in Firefox and Chrome

cboulanger:
@johnspackman I provided a [special mixin with asserts to check the state of promises](https://github.com/qooxdoo/qooxdoo/blob/791e8b9f3dd4857c51a7c3b66ceae603f8c0221b/source/class/qx/test/io/MAssert.js) which works with native promises and qx.Promise. I am not familiar enough with the Promises/A+ specification. Is this something that is already possible with qx.Promise, i.e. am I duplicating features, or would it make sense to add this to our assertion library?

cboulanger:
@hkollmann Still depends on #10739, so only merge afterwards.

johnspackman (approved):

derrell (approved):

goldim (approved):

PR #10749, opened 2025-01-12
Allow runtime environment mutations for testing (issue 10747)

### LINKED ISSUE #10747: Allow qx.core.Environment entries be changed at runtime for testing
**Is your feature request related to a problem? Please describe.**
At the moment, the values managed by `qx.core.Environment` are set at compile time and are immutable at runtime. This makes sense because of the idea of a fixed "environment" that is controlled by the developer and determined by the actual environment such as browser etc. However, this makes unit testing of classes difficult when the behavior of these classes differs based on the environment settings. The values of the environment variables cannot be manipulated as they are stored in a private class member. 

**Describe the solution you'd like**
We should add a method `qx.core.Environment.set(key, value)` which is only enabled in a testing environment. Ideally, this would mean that the method is only available depending on whether it is purposely enabled by an compile-time environment variable setting.

cboulanger:
Fixes #10747

@johnspackman In the compiler, I have added an exception for environment value optimization if "qx.environment.allowRuntimeMutations" is true. This is enough to implement the feature. If you want to change the behaviour more generally, I can wait for that or you'll later modify my addition.

### DIFF
--- a/source/class/qx/core/Environment.js
+++ b/source/class/qx/core/Environment.js
@@ -913,7 +913,8 @@ qx.Bootstrap.define("qx.core.Environment", {
       "qx.promise.warnings": true,
       "qx.promise.longStackTraces": true,
       "qx.command.bindEnabled": false,
-      "qx.headless": false
+      "qx.headless": false,
+      "qx.environment.allowRuntimeMutations": false
     },
 
     /**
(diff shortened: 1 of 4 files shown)


johnspackman (approved):

hkollmann (approved):

oetiker (approved):

cboulanger:
Tests pass now, @derrell, @goldim, can you have a look?

derrell (approved):

goldim (approved):

Issue #10752, opened 2025-01-14
Test run to completion when linter fails, then tests fail


cboulanger:
In our framework tests, if a lint error occurs, the tests suite fully runs to completion to then fail. See, for example, [here](https://github.com/qooxdoo/qooxdoo/actions/runs/12735608759/job/35496171357?pr=10749#step:5:100).

This is wasteful and unnecessary. The test should immediately abort and report the linting error.

cboulanger:
@hkollmann do you know if there is a reason why we don't use the [`--fail-fast`](https://github.com/qooxdoo/qooxdoo/blob/master/source/class/qx/tool/cli/commands/Test.js#L49) parameter for [`qx test`](https://github.com/qooxdoo/qooxdoo/blob/master/package.json#L41)?

hkollmann:
It was my decision to see the result of all tests. With fail-fast execution stops after the first failure - if you have more than one this needs to much turn arounds.
You can use fail-fast in local tests.

But leaving tests after lint fails is quite an good idea and simple to implement. I can prepare an PR

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo/pull/10753

cboulanger:
Thank you for your work on this, much appreciated!

Issue #10751, opened 2025-01-14
Change test matrix


cboulanger:
Our [current framework testing strategy](https://github.com/qooxdoo/qooxdoo/blob/master/.github/workflows/test-framework.yml#L27C1-L31C28) is the following:

```yaml
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest]
        node-version: [16.x, 18.x, 19.x]
        browser: [chromium]
```

Node v16 and v19 are EOL and should not be supported anymore.

It should instead consist only of actively supported node versions. Also, Firefox should be tested as well.

```yaml
    strategy:
      matrix:
        os: [ubuntu-latest, windows-latest]
        node-version: [18.x, 20.x, 22.x]
        browser: [chromium, firefox]
```

hkollmann:
Last time we test against firefox tests are not working.
You can test locally: qx test --browsers=firefox --terse --headless --set-env qx.test.delay.scale=10

cboulanger:
(moving this from #10753)

I think we need to test other browsers than Chrome but to my dismay I just saw that [Firefox has only 2.9% market share left](https://en.wikipedia.org/wiki/Usage_share_of_web_browsers), so even though it is my favorite browser (out of habit) there is probably no very good reason why we test for Firefox and not for, say, Safari, Edge, or Opera. 

But thinking about it more systematically, it would probably make sense to have a fast-track test where we only test for Chrome, and then a complete one that we only run sporadically (maybe once a week?) and which would test a combination of browser engines and javascript engines:

| Render/JS engines  | V8 (Chromium-based)       | SpiderMonkey (Mozilla) | JavaScriptCore (WebKit-based) |
|------------------|---------------------------|-------------------------|-------------------------------|
| Blink (Chromium) | Chrome, Edge, Opera, Brave |                         |                               |
| Gecko            |                           | Firefox, Tor Browser    |                               |
| WebKit           |                           |                         | Safari                        |

So testing Firefox and Safari (as far as that's workable given that [MacOS workflows are much more expensive (10x) in terms of free workflow time](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes)) could be an additional quality assurance check that we run independent of a particular PR and which would alert us to problems in the master branch. 

[...]

PR #10753, opened 2025-01-14
fastFail for lint

hkollmann:
fix for https://github.com/qooxdoo/qooxdoo/issues/10752
fix for https://github.com/qooxdoo/qooxdoo/issues/10751

I added a fastFail property for tests to exit the test process when this special test fails. For qooxdoo testing i add this to the lint test.

In this PR i also changed node test versions to the last 3 lts versions.

The new testapper app is testing chromium,firefox,webkit in  a parallel task. For this it was necessary to increase some timeouts.

### DIFF
--- a/.github/workflows/test-framework.yml
+++ b/.github/workflows/test-framework.yml
@@ -28,7 +28,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         node-version: [18.x, 20.x, 22.x]
-        browser: [chromium, firefox]
 
     steps:
       - name: checkout
@@ -43,4 +42,4 @@ jobs:
         run:  npm  ci
 
       - name: install and run tests with ${{ matrix.browser }}
-        run:  npm test -- --browsers=${{ matrix.browser }} --terse --headless --set-env qx.test.delay.scale=10
+        run:  npm test -- --browsers=chromium,firefox --terse --headless --set-env qx.test.delay.scale=10
--- a/.github/workflows/test-framework.yml
+++ b/.github/workflows/test-framework.yml
@@ -14,16 +14,12 @@ on:
     types: 
       - opened
       - reopened
-    paths:
-      - 'compile.*'
-      - 'source/**'
-      - 'test/**'
 
   workflow_dispatch:
 
 jobs:
   test:
-    name: ${{ matrix.os }}, node ${{ matrix.node-version }}, ${{ matrix.browser }}
+    name: ${{ matrix.os }}, node ${{ matrix.node-version }}
     runs-on: ${{ matrix.os }}
     if: "!contains(github.event.head_commit.message, 'skip ci')"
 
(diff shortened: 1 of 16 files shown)


cboulanger (approved):

goldim:
@hkollmann does firefox option slow overall test time? And why was it included?

hkollmann:
see https://github.com/qooxdoo/qooxdoo/issues/10751
It's just a test if firefox is going through the tests.
Yes it will double the test time.

goldim (approved):

goldim:
Actually I doubt if firefox is good idea to add. Just I will test locally and result of local tests is different. I will wait more time to see/test what's wrong on remote autobuild server.

cboulanger:
I think we need to test other browsers than Chrome but to my dismay I just saw that Firefox has only 2.9 market share left, so even though it is my favorite browser (out of habit) there is probably no very good reason why we test for firefox and not for, say, Safari, Edge, or Opera. 

But thinking about it more systematically, it would probably make sense to have a fast-track test where we only test for Chrome, and then a complete one that we only run sporadically (maybe once a week?) and which would test a combination of browser engines and javascript engines:

| Render/JS engines  | V8 (Chromium-based)       | SpiderMonkey (Mozilla) | JavaScriptCore (WebKit-based) |
|------------------|---------------------------|-------------------------|-------------------------------|
| Blink (Chromium) | Chrome, Edge, Opera, Brave |                         |                               |
| Gecko            |                           | Firefox, Tor Browser    |                               |
| WebKit           |                           |                         | Safari                        |

So testing Firefox and Safari (as far as that's workable given that [MacOS workflows are much more expensive (10x) in terms of free workflow time](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes)) could be an additional quality assurance check that we run independent of a particular PR and which would alert us to problems in the master branch.

cboulanger:
Thinking of this even further, wouldn't it make sense to have a standard combination of Node LTS + Chrome for PRs, and a more thorough nightly test, [by re-triggering the PR check](https://github.com/marketplace/actions/rerun-checks). Since most PRs won't be approved immediately, that would assure that the nightly test would catch any problems in the PR. At the same time, the fast-track PR check would run much faster.

cboulanger:
Or like this:
1. A fast test with Chrome + Node LTS on Linux & Windows on all pushes to PRs 
2. A re-check with different stable Node versions, Firefox & Edge on the PR branch with label "Ready for merge", which is made a requirement for merge
3. A weekly (?) thorough test including the current unstable Node version and Safari on **master**

johnspackman (approved):

johnspackman (approved):

cboulanger (approved):

derrell (approved):

oetiker (approved):

PR #10760, opened 2025-02-19
Consider size when caching widget placement.

ljelinkova:
After the update to Qooxdoo 7.8 I have encountered a problem with the right menu. It is not positioned correctly and part of it seems to be "outside" of the page.

This is what it Looked like in 7.7.2:
Screenshot from 2025-02-18 11-46-47

This is how it looks now in 7.8.0:
Screenshot from 2025-02-18 11-47-42

I pinpointed the cause to the following commit: https://github.com/qooxdoo/qooxdoo/pull/10722

After some debugging I found out that the problem is that when placing the menu, not only coordinates of `MenuButton`, but also the bounds of `Menu` are considered. The methods are called in this order:

`MPlacement.placeToWidget -> MPlacement._place -> MPlacement.__getPlacementSize -> Menu._computePlacementSize -> Menu._getMenuBounds -> Menu.getBounds()`

I think that when caching the `__lastKnownCoords` in the  `MPlacement.placeToWidget`, also the placement size should be cached. I've created a pull request that does that.

### DIFF
--- a/source/class/qx/ui/core/MPlacement.js
+++ b/source/class/qx/ui/core/MPlacement.js
@@ -205,6 +205,7 @@ qx.Mixin.define("qx.ui.core.MPlacement", {
     __ptwLiveUpdateDisappearListener: null,
     /**@type {Record<"top" | "right" | "bottom" | "left", number> | null}*/
     __lastKnownCoords: null,
+    __lastKnownSize: null,
 
     /**
      * Returns the location data like {qx.bom.element.Location#get} does,
@@ -378,17 +379,23 @@ qx.Mixin.define("qx.ui.core.MPlacement", {
 
       var coords =
         target.getContentLocation() || this.getLayoutLocation(target);
+      var size = this.__getPlacementSize();
 
       if (coords != null) {
-        if (
-          coords.top === this.__lastKnownCoords?.top &&
-          coords.right === this.__lastKnownCoords?.right &&
-          coords.bottom === this.__lastKnownCoords?.bottom &&
-          coords.left === this.__lastKnownCoords?.left
-        ) {
+
+        const boundsAreEqual = (bound1, bound2) => bound1 && bound2 &&
+          bound1.top === bound2.top &&
+          bound1.right === bound2.right &&
+          bound1.bottom === bound2.bottom &&
+          bound1.left === bound2.left;
+
+        if (boundsAreEqual(coords, this.__lastKnownCoords) &&
+          boundsAreEqual(size, this.__lastKnownSize)) {
           return true;
         }
+
         this.__lastKnownCoords = coords;
+        this.__lastKnownSize = size;
         this._place(coords);
         return true;
       } else {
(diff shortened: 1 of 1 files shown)


source/class/qx/ui/core/MPlacement.js
```diff
@@ -376,26 +377,32 @@ qx.Mixin.define("qx.ui.core.MPlacement", {
         );
       }
 
-      var coords =
+      const coords =
```
johnspackman, line 380: please do not use `const` in this instance - firstly, changing `let` to `const` is not relevant to the change that this PR is about, and secondly the value referred to is not "constant" - it is a variable.

Even if it was a constant, the coding standards would expect the variable name to show that eg `const SECONDS_PER_HOUR = 60 * 60`

source/class/qx/ui/core/MPlacement.js
```diff
@@ -376,26 +377,32 @@ qx.Mixin.define("qx.ui.core.MPlacement", {
...
         target.getContentLocation() || this.getLayoutLocation(target);
+      const size = this.__getPlacementSize();
```
johnspackman, line 382: This is a particularly good example of why this use of `const` is inappropriate - the object in `size` is not constant at all.

source/class/qx/ui/core/MPlacement.js
```diff
@@ -376,26 +377,32 @@ qx.Mixin.define("qx.ui.core.MPlacement", {
...
         this.__lastKnownCoords = coords;
+        this.__lastKnownSize = size;
         this._place(coords);
         return true;
       } else {
         return false;
       }
     },
 
+    _boundsAreEqual(bound1, bound2) {
```
johnspackman, line 398: As `_boundsAreEqual` is a special purpose (and quite concise) function required solely for the workings of `placeToWidget`, IMHO this should be as an inner function inside `placeToWidget` - alternatively, if its placed in the members it should be private (prefixed with double `_`) and documented.

If this was an inner function, then it would be appropriate to use the `const` keyword, because Functions are immutable.  EG:

```
const boundsAreEqual = (bound1, bound2) => {
  return bound1 && bound2 &&
        bound1.top === bound2.top &&
        bound1.right === bound2.right &&
        bound1.bottom === bound2.bottom &&
        bound1.left === bound2.left
};
/* ... snip ... */
if (boundsAreEqual(coords, this.__lastKnownCoords) &&
          boundsAreEqual(size, this.__lastKnownSize)) {
```

johnspackman (changes requested):

source/class/qx/ui/core/MPlacement.js
ljelinkova: I've changed it back to `var`.

source/class/qx/ui/core/MPlacement.js
ljelinkova: As suggested, I've changed it to `var`.

source/class/qx/ui/core/MPlacement.js
ljelinkova: Nice, I've changed it accordingly.

goldim (approved):
Great job! Thank you!

johnspackman (approved):

cboulanger (approved):

hkollmann (approved):

oetiker (approved):

derrell (approved):

ljelinkova:
Thank you all

goldim:
@ljelinkova I am not sure that we will publish this near soon as new version. You may use master branch in your projects or downgrade qooxdoo.

PR #10766, opened 2025-03-31
Add option to exclude Bluebird from qx.Promise

patryk-m-malinowski:
This allows one to make qx.Promise use native promises under the hood instead of Bluebird by setting the environment setting `qx.promise.useNativePromises` to `true`. This can take away the heavyweight Bluebird dependency, making this useful for thin, lightweight applications.

This also includes the commits made by [this PR](https://github.com/qooxdoo/qooxdoo/compare/master...patryk-m-malinowski:qooxdoo:promise-unit-tests?expand=1), which complete the unit tests for qx.Promise.

Important notes:
- Due to the complexity in implementing cancellation using native promises, cancellation does not work when `qx.promise.useNativePromises` is `true`. 
- `qx.Promise.some` now throws the native AggregateError instead of Bluebirds AggregateError. Unless there is code out there which looks at the details of the AggregateError object, this shouldn't be too much of the problem and fixing the code is trivial.


cboulanger (approved):

hkollmann (approved):

johnspackman (changes requested):
The only real difficulty (AIUI) of removing Bluebird support is that its provides the `cancel` method for promises, which is not ratified by ES6 and (also AIUI) not going to be added any time soon.  The logistics of `cancel` is complicated, with lots of difficult and complicated edge cases, and needs the support of the underlying implementation of Promise in order to be implemented.

This means that we cannot switch to native implementation and fabricate own own polyfill for `cancel`.

Removing Bluebird and switching to native is therefore a BC issue, although hopefully unlikely to cause many issues in practice.

@patryk-m-malinowski when we are _not_ using native promises, please can you add a warning message to the implementation of `cancel` that this is deprecated?  Make this warning only print once and only if `qx.debug` is `true`

oetiker (approved):

johnspackman:
@patryk-m-malinowski also the environment setting needs to be `qx.Promise.useNativePromises` and declared in an `environment` block in `qx.Promise`, so that the default can be set in code

johnspackman (approved):

hkollmann:
Any idea why the checks are not triggered?

patryk-m-malinowski:
> @patryk-m-malinowski also the environment setting needs to be `qx.Promise.useNativePromises` and declared in an `environment` block in `qx.Promise`, so that the default can be set in code

This will not be in line with the naming convention that was used before, where the environment settings `qx.promise`, `qx.promise.warnings` and `qx.promise.longStackTraces` all use the lowercase `qx.promise`.

Also, there is an environment section in qx.Promise (around line 1067) which includes `qx.promise.useNativePromise`. Do you reckon it should be defined elsewhere in the file? Also, did you mean you want the environment setting to be called `qx.promise.useNativePromises` instead of `qx.promise.useNativePromise`?

johnspackman:
> This will not be in line with the naming convention that was used before, where the environment settings `qx.promise`, `qx.promise.warnings` and `qx.promise.longStackTraces` all use the lowercase `qx.promise`.

Then those other settings are wrong; the requirement is that environment settings are prefixed with the case sensitive class name, and this is what allows the compiler to find the default values.  There is a mapping of values in the compiler, but it is not really extensible (eg this PR should not require changes to the compiler code)

> Also, there is an environment section in qx.Promise (around line 1067) which includes `qx.promise.useNativePromise`. Do you reckon it should be defined elsewhere in the file? 

Ah, sorry I missed that section - but it has to be `qx.Promise.useNativePromise`, otherwise the compiler cannot find it if it is references in other classes.

> Also, did you mean you want the environment setting to be called `qx.promise.useNativePromises` instead of `qx.promise.useNativePromise`?

Not especially, singular or plural would be fine by me

patryk-m-malinowski:
Okay. This means that some users may need to do some renames in their code for the environment variables, but this will be trivial.

derrell:
> Okay. This means that some users may need to do some renames in their code for the environment variables, but this will be trivial.

Trivial, but not BC, so it'll have to be a major version bump.

johnspackman:
> > Okay. This means that some users may need to do some renames in their code for the environment variables, but this will be trivial.
> 
> Trivial, but not BC, so it'll have to be a major version bump.

Yes, @derrell is right - perhaps we can have something ugly for now like:

```
if (qx.core.Environment.get("qx.promise.warnings") || qx.core.Environment.get("qx.Promise.warnings")) {...
```
which is ugly, but also have in `qx.Promise.defer`:

```
if (qx.core.Environment.get("qx.promise.warnings") !== true) { 
   console.warn("Deprecated use of ....
```

and then in 8.x delete the support for the lower case (or throw an exception).  We could also consider changing the default of `qx.Promise.useNativePromises` to `true`, but thats a whole other discussion.

johnspackman (approved):

source/class/qx/Promise.js
```diff
@@ -287,25 +299,24 @@ qx.Class.define("qx.Promise", {
      * function returns a promise or a thenable, then the result of the promise is awaited, before continuing with
      * next iteration.
      *
-     * @param iterable {Iterable} An iterable object, such as an Array
      * @param iterator {Function} the callback, with <code>(value, index, length)</code>
      * @return {qx.Promise}
      */
-    forEach(iterable, iterator) {
+    forEach(iterator) {
```
derrell, line 305: This looks like it's not BC

source/class/qx/Promise.js
```diff
@@ -287,25 +299,24 @@ qx.Class.define("qx.Promise", {
...
      * @param iterator {Function} the callback, with <code>(value, index, length)</code>
-     * @param options {Object?} options; can be:
-     *  <code>concurrency</code> max nuber of simultaneous filters, default is <code>Infinity</code>
+     * @param options {Object?} Either:
+     *   A native object with one key: <code>concurrency</code>: max number of simultaneous filters, default is <code>Infinity</code>
+     * Or: any other object (subclass of qx.core.Object), in which case this will be the context for the iterator
      * @return {qx.Promise}
      */
-    filter(iterable, iterator, options) {
+    filter(iterator, options) {
```
derrell, line 319: ditto

source/class/qx/Promise.js
```diff
@@ -322,13 +333,13 @@ qx.Class.define("qx.Promise", {
...
      * @param iterator {Function} the callback, with <code>(value, index, length)</code>
-     * @param options {Object?} options; can be:
-     *  <code>concurrency</code> max nuber of simultaneous filters, default is <code>Infinity</code>
+     * @param options {Object?} Either:
+     *   A native object with one key: <code>concurrency</code>: max number of simultaneous filters, default is <code>Infinity</code>
+     * Or: any other object (subclass of qx.core.Object), in which case this will be the context for the iterator
      * @return {qx.Promise}
      */
-    map(iterable, iterator, options) {
+    map(iterator, options) {
```
derrell, line 342: It looks like there are lots of these. I won't point out each one, but all of these methods must be BC if not a major version bump... and even if it is a major version bump, this could mean major changes to users' apps.

derrell (changes requested):
It appears the changes are not backward compatible, right? See the few places I marked, but it seems all functions removed the first parameter, breaking BC...

This leads me to wonder whether there's even a reason to make this change. All of my own apps use native promises directly, without any `qx.Promise` wrapper. Maybe we simply deprecate `qx.Promise` and suggest native promises in its stead? Or is there some significant benefit to using `qx.Promise` over native promises?

source/class/qx/Promise.js
johnspackman: also here

source/class/qx/Promise.js
johnspackman: i think that this deserves a proper check over on all the methods, but AIUI none of the member methods need an iterable passed in

derrell: They may not require an iterable to be passed in, but the argument was there so any users of `qx.Promise` having seen that would have passed _something_ as that iterable; the iterator would be the second argument they passed, not the first. Am I missing something?

johnspackman: if they did, then tjheir code crashes because a function (ie `iterator`) is expected as the first argument.  I just tried this code in the playground:

```
let p = new qx.Promise(resolve => resolve([1,2,3]));
p.forEach(x => console.log(x));
```
which outputs:
```
1
2
3
```

then:

```
let p = new qx.Promise(resolve => resolve([1,2,3]));
p.forEach([4,5,6], x => console.log(x));
```
which outputs:
```
qx.Promise: Unhandled promise rejection: TypeError: expecting a function but got [object Array]
```

derrell: I don't understand, but empirical evidence seems adequate proof.

johnspackman:
> This leads me to wonder whether there's even a reason to make this change. All of my own apps use native promises directly, without any `qx.Promise` wrapper. Maybe we simply deprecate `qx.Promise` and suggest native promises in its stead? Or is there some significant benefit to using `qx.Promise` over native promises?

PS sorry missed your question earlier - qx.Promise adds various helper methods that can be useful, eg `qx.Promise.all` is used inside Qooxdoo, and I've used it a bit in my own code; the most useful one I personally find is the `qx.Promise` constructor, which allows the promise to be resolved or rejected outside of the constructor, eg:

```
someInitFunction() {
   this.__whenSomethingIsDone = new qx.Promise();
},
// ... snip ...
__doSomeLongTask() {
  this.__whenSomethingIsDone.resolve();
},
async waitForReady() {
   await this.__whenSomethingIsDone();
}
```

derrell:
Hmmm... [Promise.all()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) available with native promises. The resolver example is just done very slightly differently with native promises, using [Promise.withResolvers()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers):

```
let whenSomethingIsDone;

function someInitFunction()
{
  console.log("Creating promise");
  whenSomethingIsDone = Promise.withResolvers();
}

function doSomeLongTask()
{
  console.log("setting timer");
  setTimeout(
    () =>
    {
      console.log("resolving");
      whenSomethingIsDone.resolve();
    },
    2000);
}

async function waitForReady()
{
  console.log("awaiting resolution...");
  await whenSomethingIsDone.promise;
  console.log("now resolved");
}

async function main()
{
  someInitFunction();
  doSomeLongTask();
  await waitForReady();
}

main();
```

johnspackman:
> Hmmm... [Promise.all()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) available with native promises. 

fair enough, i wasn't aware, I've been looking at the Promises/A+ standard so I guess I'm a little of out date with how browsers have caught up.  There are other functions in `qx.Promise` though, and IMHO it would not be appropriate to simply delete `qx.Promise` because there is a new way of doing things - although that kind of review would lead to a lot of thinning down of the entire framework, it would also lead to some significant backwards compatibility headaches.

Your review still shows as requesting changes - is that out of date now, ie are you happy for this PR to be merged?

patryk-m-malinowski:
> Any idea why the checks are not triggered?

Sorry for not noticing your reply. I've tried to run the checks locally on my machine on a clean version on master and it also hung! I ran the command `npm test -- --browsers=chromium,firefox,webkit --terse --headless --set-env qx.test.delay.scale=10 > testout.txt` in the root folder of Qooxdoo. Here is my
[standard output](https://github.com/user-attachments/files/19633914/testout.txt). Perhaps one or more of the libraries/programs which are downloaded for the tests have changed and caused compatibility issues?

derrell (approved):

hkollmann:
@patryk-m-malinowski : @patryk-m-malinowski : Could you have a look at the new  qx.test.Promise:testMapSeriesPromiseArray test in master? It is failing...

hkollmann:
Is this BC nd can be merged? Or should we hold it for v8?

hkollmann:
@patryk-m-malinowski : Could you resolve the conflicts please?

patryk-m-malinowski:
> @patryk-m-malinowski : @patryk-m-malinowski : Could you have a look at the new qx.test.Promise:testMapSeriesPromiseArray test in master? It is failing...

As far as I can tell the tests for this PR were cancelled [https://github.com/qooxdoo/qooxdoo/actions/runs/14177090624/job/39753762648](url). I also ran the tests locally both for this branch and master and got no problems. Was I not looking in the right places?

patryk-m-malinowski:
A ok got it!

patryk-m-malinowski:
This PR is intended to be BC and intended to be merged into the current master. The unit tests were extended, not modified and should pass if they would work on Github.

PR #10783, opened 2025-08-27
Move 'blur' handling for 'focusout'.

### LINKED ISSUE #10756: [question] blur listeners work differently after upgrade to qooxdoo 7
We are trying to upgrade from qooxdoo 6.0.2 to 7.8.0 and testing if our application behavior is still the same. It looks like the behavior of blur listeners is different starting with version 7.0.0. 

We have a `TextArea` widget deriving from `qx.ui.form.TextArea`. If it is used in isolation, the blur listener seems to be called the same way as before the upgrade. However, when this widget is used inside our grid widgets (derived from `qx.ui.table.Table`), the behavior is different. Before the upgrade, the blur listener was called when the user switched from one grid cell to another one. After the upgrade, the blur listener is not called in this case anymore.

We thought that this may have to do with our grid cells not having the `tabIndex` property. Adding the `tabIndex` property helps partially. What I mean is, clicking on a different grid cell does invoke the blur listener. However, pressing the Tab key does not and we need to handle the Tab key separately (with a "keydown" listener). Another difference is that clicking on a different cell does not select the new cell yet, and one needs to click on the new cell again for the cell to be selected.

This may be a very specific problem related only to our code. However, since we only started observing the problem when we started upgrading to qooxdoo 7, I would like to ask if you know anything about the change in the behavior of the blur functionality and if you have any recommendations how we can work with it correctly.

simozzy:
This may fix https://github.com/qooxdoo/qooxdoo/issues/10756

### DIFF
--- a/source/class/qx/ui/table/pane/Scroller.js
+++ b/source/class/qx/ui/table/pane/Scroller.js
@@ -1973,16 +1973,14 @@ qx.Class.define("qx.ui.table.pane.Scroller", {
     _onFocusinCellEditorAddBlurListener(e) {
       this.debug("executed FOCUSIN event listener for hash: " + e.getTarget().$$hash);
       qx.event.Timer.once(function() {
-        this._cellEditor.addListenerOnce('blur', this._onBlurCellEditorStopEditing, this);
-        this.debug('added BLUR listener to hash: ' + this._cellEditor.$$hash);
+        this._cellEditor.addListener('focusout', this._onFocusoutCellEditorStopEditing, this);
+        this.debug('added FOCUSOUT listener to hash: ' + this._cellEditor.$$hash);
       }, this, 1);
     },
 
-    /**
-     * Stop editing whenever the cell editor blurs.
-     */
-    _onBlurCellEditorStopEditing(e) {
-      this.debug("executed BLUR listener for hash " + e.getTarget().$$hash);
+
+    _onFocusoutCellEditorStopEditing(e) {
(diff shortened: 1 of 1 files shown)


johnspackman (approved):

source/class/qx/ui/table/pane/Scroller.js
```diff
@@ -1973,16 +1973,14 @@ qx.Class.define("qx.ui.table.pane.Scroller", {
     _onFocusinCellEditorAddBlurListener(e) {
       this.debug("executed FOCUSIN event listener for hash: " + e.getTarget().$$hash);
       qx.event.Timer.once(function() {
-        this._cellEditor.addListenerOnce('blur', this._onBlurCellEditorStopEditing, this);
-        this.debug('added BLUR listener to hash: ' + this._cellEditor.$$hash);
+        this._cellEditor.addListener('focusout', this._onFocusoutCellEditorStopEditing, this);
+        this.debug('added FOCUSOUT listener to hash: ' + this._cellEditor.$$hash);
```
derrell, line 1977: What does this do when the cell editor is a modal window instead of inline editor? Blur wouldn't happen with a modal window, but focusout could, and that shouldn't cause the modal window to close just because the mouse pointer moved out of that window...

derrell (changes requested):

derrell (approved):
Thanks for the confirmation.

oetiker (approved):

hkollmann (approved):

cboulanger (approved):

PR #10786, opened 2025-09-12
Fixed command misses execution after disable/enable cycle

### LINKED ISSUE #10785: Command misses execution after disable/enable cycle
Hello,

I am encountering behavior similar to the one described in issue #8911, with qooxdoo 7.9.1 (browser is Firefox or Chromium).

Code to reproduce in https://tinyurl.com/2666p6tp

Steps:
1. (Button is initially disabled)
2. Select a table row
3. (Button is now enabled)
4. Click the button
5. (Command is executed: debug message is printed, button is again disabled)
6. Select another table row
7. (Button is again enabled)
8. Click the button
9. (Command is not executed - this is the issue)
10. Click the button
11. (Command is executed)

TIA

goldim:
Fixed #10785
Do I get it right that the semaphore is just protection from double execution?
The idea of my fix is to just allow the semaphore to work every time but checking just state of a widget on disable/enable to execute the widget.

### DIFF
--- a/source/class/qx/ui/core/MExecutable.js
+++ b/source/class/qx/ui/core/MExecutable.js
@@ -61,7 +61,11 @@ qx.Mixin.define("qx.ui.core.MExecutable", {
 
   members: {
     __executableBindingIds: null,
-    __semaphore: false,
+    
+    /**
+     * @type {Boolean} Whether the command already executed. The protection flag to stop the second command execution.
+     */
+    __commandExecuted: false,
     __executeListenerId: null,
 
     /**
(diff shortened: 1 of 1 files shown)


oetiker:
no idea about the intend :(

johnspackman (changes requested):

goldim:
@oetiker the indend to fix bug described in issue mentioned here. Also keeping resolved another problem related with disabled widgets.

oetiker:
maybe I missunderstand commands, but it is not allowed to execute them multiple times ? the code essentially prevents every second execution ? what is this good for ?

goldim:
@oetiker No, not the second execution but double execution I would say. It is not allowed to trigger `execute` both in the same time from command and widget (button for example).
In the first version of this code semaphore worked always but it was allowed to run command which triggered "execute" event. Then there were PRs which fixed this problem but revelead new one which is described in the issue attached to this PR.
This PR code allows semaphore work always but if the widget is DISABLED `execute` event is not triggered. It is just roll back to original behaviour.

johnspackman (approved):

hkollmann (approved):

derrell (approved):

cboulanger (approved):

oetiker (approved):

PR #10801, opened 2025-11-05
Fix compiler warnings for ES6 shorthand methods in delegate objects

### LINKED ISSUE #10591: compiler generates warning on map entry `get(x)` but not `get : function(x)`
The following code generates a compiler warning when the newer form of a function declaration in a map is used, but not when the traditional form of function declaration is used. See the comment in the code below.

```javascript
qx.Class.define("xxx.Application",
{

  // These generate a compiler warning that `property` and `value` are
  // unresolved. The warnings can be eliminated by changing the
  // function declaration type from
  //
  //   get(property) {
  //
  // to
  //
  //   get : function(property) {
  //
  // This code here demonstrates the error, which also occurs in
  // qx.data.Array in the new-class-property-system branch
  //
  delegate: {
    get(property) {
      return property === "0";
    },

    set(property, value) {
      this[property] = !!value;
    }
  },

  members :
  {
    main()
    {
      super.main();
    }
  }
});
```

The generated warnings are:
```
xxx.Application: [20,13] warning: Unresolved use of symbol property
xxx.Application: [24,25] warning: Unresolved use of symbol value
```

Although the released compiler also complains about the `delegate` key being disallowed, it is allowed in the new class/property system version of the code.

It is impossible to avoid this warning by simply using the traditional function declaration form, because the hooks invoked with a commit to qooxdoo reformats the code to use the new form.

hkollmann:
## Summary

Fixes compiler generating spurious "Unresolved use of symbol" warnings when using ES6 shorthand method syntax in delegate objects.

## Problem

The compiler was generating false warnings like:
- "Unresolved use of symbol property"
- "Unresolved use of symbol value"

When using ES6 shorthand method syntax in delegate objects:

```javascript
delegate: {
  get(property) {  // ❌ Warning: Unresolved use of symbol property
    return this[property];
  },
  set(property, value) {  // ❌ Warnings for property and value
    this[property] = value;
  }
}
While traditional function syntax worked without warnings:

delegate: {
  get: function(property) {  // ✅ No warning
    return this[property];
  }
}
This issue affected classes like qx.data.Array and made it impossible to use modern JavaScript syntax with auto-formatting tools.

Root Cause
The delegate property in class definitions didn't have explicit handling in the ObjectProperty visitor of CLASS_DEF_VISITOR. This meant its children (ObjectMethod nodes with ES6 shorthand syntax) weren't being properly traversed with the VISITOR, which is responsible for calling enterFunction() to register method parameters as declarations in the scope.

Solution
Added explicit handling for the delegate property in source/class/qx/tool/compiler/ClassFile.js to ensure proper traversal of ES6 shorthand methods:

} else if (keyName == "delegate") {
  path.skip();
  path.traverse(VISITOR);
This follows the same pattern as other class definition sections like events, properties, etc.

Changes
Fix: Added delegate property handling in ClassFile.js (3 lines)
Test: Created integration test in test/tool/integrationtest/test-issues/issue10591/ with:
TestArray class demonstrating ES6 shorthand methods in delegate
Application exercising the delegate functionality
Documentation explaining the issue and expected behavior
Testing
✅ Bootstrap compiler completes successfully ✅ Full compilation produces no "Unresolved use of symbol" warnings ✅ Integration test compiles without warnings ✅ Verified with qx.data.Array and custom delegate objects

Fixes
Closes #10591

### DIFF
--- a/source/class/qx/tool/compiler/ClassFile.js
+++ b/source/class/qx/tool/compiler/ClassFile.js
@@ -1529,6 +1529,9 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
               });
             }
             path.traverse(VISITOR);
+          } else if (keyName == "delegate") {
+            path.skip();
+            path.traverse(VISITOR);
           } else if (keyName == "aliases") {
             path.skip();
             if (!prop.value.properties) {
--- a/test/tool/integrationtest/test-issues/issue10591/Manifest.json
+++ b/test/tool/integrationtest/test-issues/issue10591/Manifest.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "https://qooxdoo.org/schema/Manifest-2-0-0.json",
+  "info": {
+    "name": "issue10591",
+    "summary": "Test for ES6 shorthand methods in delegate objects",
+    "description": "Tests that ES6 shorthand method syntax in delegate objects compiles without spurious unresolved symbol warnings",
+    "homepage": "https://github.com/qooxdoo/qooxdoo/issues/10591",
+    "license": "MIT license",
+    "authors": [],
+    "version": "1.0.0"
+  },
+  "provides": {
+    "namespace": "issue10591",
+    "encoding": "utf-8",
+    "class": "source/class",
+    "resource": "source/resource",
+    "translation": "source/translation"
+  },
+  "requires": {
+    "@qooxdoo/framework": "^8.0.0"
+  }
+}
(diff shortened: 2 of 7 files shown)


johnspackman (approved):

goldim (approved):

cboulanger (approved):

goldim:
I tried to compile my simple test app with `branch_8_0_beta` and no warning at all.

derrell (approved):

oetiker (approved):

PR #10802, opened 2025-11-05
Compiler incorrectly flagging JavaScript labels in continue statements (Issue #10727)

### LINKED ISSUE #10727: `Unresolved use of symbol` when using labels
**Describe the bug**
When using [labels](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label), the compiler complains of unresolved symbols.

This doesn't have any real impact on the app at runtime, but it would be nice to correctly identify labels to prevent the warning.

**To Reproduce**
Attempt to compile the following code;

```js
qx.Class.define('appropriate.Name', {
  extend: qx.core.Object,

  members: {
    labeledLoop() {
      $outer: for (let x = 0; x < 10; x++)
        for (let y = 0; y < 10; y++)
          if (x ** y > 100_000)
            continue $outer;
    }
  }
})
```

**Expected behavior**
The symbol `$outer` should be resolved as it is declared in an enclosing scope.

**Actual behavior**
```
appropriate.Name: [9,21] [9,27] warning: Unresolved use of symbol $outer
```
The compiler is unaware that the label was defined, and reports it's use as an unresolved symbol.

hkollmann:
## Summary
Fixes #10727 - Resolves false "Unresolved use of symbol" warnings for JavaScript labels in continue statements.

## Problem
Valid code like this incorrectly generates warnings:
```javascript
$outer: for (let x = 0; x < 10; x++) {
  for (let y = 0; y < 10; y++) {
    if (x ** y > 100000) {
      continue $outer;  // Warning: Unresolved use of symbol $outer
    }
  }
}
Root Cause
ContinueStatement was missing from the CHECK_FOR_UNDEFINED list in ClassFile.js, while BreakStatement and LabeledStatement were already included. This inconsistency caused label identifiers to be incorrectly checked as variable references.

### DIFF
--- a/source/class/qx/tool/compiler/ClassFile.js
+++ b/source/class/qx/tool/compiler/ClassFile.js
@@ -1716,7 +1716,8 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
             ClassDeclaration: 1,
             ClassMethod: 1,
             LabeledStatement: 1,
-            BreakStatement: 1
+            BreakStatement: 1,
+            ContinueStatement: 1
           };
 
           // These are AST node types we expect to find at the root of the identifier, and which will
--- a/test/tool/unittest/compiler/classIssue10727.js
+++ b/test/tool/unittest/compiler/classIssue10727.js
@@ -1,0 +1,42 @@
+qx.Class.define("classIssue10727",
+{
+  extend: qx.core.Object,
+
+  members: {
+    // Test case from issue #10727: JavaScript labels should not generate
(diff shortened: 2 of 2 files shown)


johnspackman (approved):

goldim (approved):

cboulanger (approved):

goldim:
This one I will move on master too.

derrell (approved):

oetiker (approved):

PR #10804, opened 2025-11-05
Fix #10693: Add property field to metadata for property accessor methods

hkollmann:
## Summary

Fixes #10693 - The Properties button in the API viewer now works correctly by adding the missing `property` field to property accessor methods in metadata.

## Problem

The API viewer (qxl.apiviewer) expects a `property` field in method metadata to identify methods that were automatically generated from properties (like `getSomeProperty()`, `setSomeProperty()`, etc.). This field was missing from the qooxdoo compiler's metadata generation, causing the Properties button toggle to malfunction.

## Solution

Added a new method `__addPropertyAccessors()` to the `MetaDatabase` class that generates metadata entries for all property accessor methods:

- **Getter methods**: `getSomeProperty()`
- **Boolean getters**: `isSomeProperty()` for boolean properties
- **Setter methods**: `setSomeProperty(value)`
- **Reset methods**: `resetSomeProperty()`
- **Async variants**: All above with `Async` suffix if property is async

Each generated method includes:
- A `property` field pointing to the source property name
- Proper JSDoc documentation
- Correct function signature and parameters
- Location information from the property definition
- Public access level

### Modified Files
- `source/class/qx/tool/compiler/MetaDatabase.js` (+203 lines)
  - Added `__addPropertyAccessors()` method
  - Integrated into metadata generation pipeline

### Test Files
- `test/tool/integrationtest/testapp/source/class/testapp/Issue10693.js` (new)
  - Test class with various property types
- `test/tool/integrationtest/test-deps.js` (+81 lines)
  - 42 comprehensive test assertions

## Test Results

All tests passing: **89/89 tests** ✅

Test coverage includes:
- ✅ String property accessors (get, set, reset)
- ✅ Boolean property accessors (get, set, reset, is)
- ✅ Number property accessors (get, set, reset)
- ✅ Async property accessors (all variants with Async suffix)
- ✅ Verification that explicit methods are not overwritten
- ✅ Verification that each accessor has the `property` field

## Impact

- **API Viewer**: Properties button will now correctly toggle property-related methods
- **Backward Compatible**: Existing functionality is preserved
- **No Breaking Changes**: Only adds missing metadata, doesn't modify existing behavior

## Related Issue

Closes #10693

### DIFF
--- a/source/class/qx/tool/compiler/MetaDatabase.js
+++ b/source/class/qx/tool/compiler/MetaDatabase.js
@@ -199,6 +199,8 @@ qx.Class.define("qx.tool.compiler.MetaDatabase", {
 
         meta.fixupJsDoc(typeResolver);
 
+        this.__addPropertyAccessors(metaData);
+
         this.__fixupMembers(metaData);
 
         this.__fixupEntries(metaData, "members");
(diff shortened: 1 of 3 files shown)


cboulanger (approved):

goldim:
This one I will move on master too.

derrell (approved):

oetiker (approved):

johnspackman (changes requested):
Sorry but I do not agree with this PR - these methods were deliberately left out of the meta database, because they are all implied by the specification of properties.  

The API viewer should be able to understand the property definition already in meta and know how to interpret it - this would allows the API viewer to know (for example) that some functions are "virtual" and display them differently ... eg personally I find the pages of properties get/set/reset/etc methods listed in a class definition very distracting.

hkollmann:
The API Viewer does not seem to be working at the moment. Therefore, I am closing this for now.

PR #10810, opened 2025-11-06
Fix issue #10808: Auto-convert Form.add() name parameter to camelCase

### LINKED ISSUE #10808: In method add of class qx.ui.form.Form third argument is name and it has to be LOWERCASE
In method add of class qx.ui.form.Form third argument is name and it has to be LOWERCASE else I've get an error that property (seems like it creates property with this name) is missing during binding process.

This is different behavior than in v7. 

From discussion https://github.com/qooxdoo/qooxdoo/discussions/10798:

I'm not aware of this, but looking at the docs it looks like its for binding to a property; property names have always had a convention of camel case, but that convention is effectively relied on (aka enforced) by parts of the v7 framework. In v8, those conventions have been formalised, so maybe that's having a side effect here. Please can you provide an example and I'll take a look

point 4: Whats about a toLowercase in add?

_Originally posted by @hkollmann in https://github.com/qooxdoo/qooxdoo/discussions/10798#discussioncomment-14883987_

> In method add of class qx.ui.form.Form third argument is name and it has to be LOWERCASE else I've get an error that property (seems like it creates property with this name) is missing during binding process.

@johnspackman  example:
```javascript
qx.Class.define("test.Form", {
    extend: qx.ui.form.Form,
    
    construct(){
        super();
        this.__createUserNameField();
    },    

    members: {
        __createUserNameField(){
            const host = new qx.ui.form.TextField();
            this.add(host, "Username", null, "Username");
        }
    }
});

let form = new test.Form();//qx.ui.form.Form();
const controller = new qx.data.controller.Form(null, form);
// const field = new qx.ui.form.PasswordField();
// form.add(field, "test", null, "test");
controller.createModel();
```

> Did you try setting the environment variables to control this? I'll have to check but I thought that the v7 behaviour was enabled by default (you're not supposed to see that effect)

No, I didn't change environment settings.

> Definitely a bug - I don't know why this is happening, but the error message is an indication that qxWeb is broken

I try to create minimal example.

_Originally posted by @goldim in https://github.com/qooxdoo/qooxdoo/discussions/10798#discussioncomment-14884901_

hkollmann:
## Summary

Fixes #10808 

In qooxdoo v8, the third parameter (name) of `qx.ui.form.Form.add()` must follow camelCase conventions (first letter lowercase) for data binding with `qx.data.controller.Form.createModel()` to work properly. This broke v7 compatibility where capitalized names like "Username" or "EmailAddress" worked fine.

This PR automatically converts capitalized names to camelCase using `qx.lang.String.firstLow()`, maintaining backward compatibility with v7 code.

## Example

Before this fix:
```javascript
var form = new qx.ui.form.Form();
form.add(new qx.ui.form.TextField(), "Username", null, "Username");
var controller = new qx.data.controller.Form(null, form);
var model = controller.createModel(); // ❌ FAILS - binding error
After this fix:

var form = new qx.ui.form.Form();
form.add(new qx.ui.form.TextField(), "Username", null, "Username");
var controller = new qx.data.controller.Form(null, form);
var model = controller.createModel(); // ✅ WORKS - "Username" → "username"
model.getUsername(); // Property is correctly camelCased

Backward Compatibility
✅ This change is fully backward compatible:

Code using lowercase names continues to work unchanged
Code using capitalized names (which failed in v8) now works correctly
The conversion only affects the first character, preserving camelCase for multi-word names (e.g., "EmailAddress" → "emailAddress")

### CODE REVIEW

**Review by cboulanger — APPROVED** [review]

**Review by johnspackman — APPROVED** [review]

**Comment by hkollmann** [comment]
https://github.com/hkollmann/qooxdoo/actions/runs/19166267029

**Review by goldim — CHANGES_REQUESTED** [review]
The method `getItem` will returned lowercase and what if user code checks it. It will fail. I think it is also BC. So broken BC wasn't resolved. I think.

**Comment by hkollmann** [comment]
> The method `getItem` will returned lowercase and what if user code checks it. It will fail. I think it is also BC. So broken BC wasn't resolved. I think.

Could you explain what you mean please?

**Comment by goldim** [comment]
Yes. After we call `form.add` with a name passed via the third argument. We can get this name via form.getItem(). And if user call it with uppercase he will get it in lowercase. Is it fine?

**Comment by hkollmann** [comment]
mplement proper solution for issue https://github.com/qooxdoo/qooxdoo/issues/10808: Preserve form names, conv…
…ert in data controller

This commit implements the correct solution addressing goldim's concerns:

**Problem:**
The previous approach converted names to camelCase in Form.add(), which broke
getItem() - users couldn't retrieve items using the name they provided.

**Solution:**
1. Form.js: Store names exactly as provided (no conversion)
   - getItem() now works with the original name
   - Maintains backward compatibility
   - Separation of concerns

2. Data Controller: Convert to camelCase when working with the model
   - New helper method: __convertNameToCamelCase()
   - Converts names to camelCase for v8 property naming conventions
   - Applied in all model-related operations:
     * createModel() - when creating the model
     * __setUpBinding() - when setting up bindings
     * _applyModel() - when removing bindings
     * __tearDownBinding() - when tearing down bindings
     * addBindingOptions() - when adding binding options
     * updateModel() - when manually updating the model

**Benefits:**
- ✅ getItem() works with the name provided to add()
- ✅ Data binding uses camelCase for v8 compatibility
- ✅ Clean separation: Form stores, Controller converts
- ✅ No breaking changes for existing code

**Review by goldim — APPROVED** [review]

**Comment by oetiker** [comment]
The automated fix proposed will cause problems when the v7 code had a form with the fields "Username" and "username".

I see the benefits of backward compatibility it has highest prio for me. But WHY do these keys HAVE to be camelCase in the first place? If there is technical reason for camelCase, why not just internally prefix everything with 'k' ?

If they must be camelCase, we should not fix this silently at runtime, but cry bloody murder and fail with a bang. In the error message we could then offer a hint what env switch can be set to turn the error into a warning for backward compatibility.

**Review by derrell — CHANGES_REQUESTED** [review]
I agree with @oetiker's comment https://github.com/qooxdoo/qooxdoo/pull/10810#issuecomment-3521353118

**Comment by hkollmann** [comment]
Then let's create a test case for this

**Comment by hkollmann** [comment]
Now we have an detailed analysis with aon solution wich thows an error when it detects problems.

**Comment by hkollmann** [comment]
# v8 Binding System Analysis - Issue #10808

## The v8 Change: Automatic Lowercase Conversion

### Found in: `source/class/qx/data/binding/PropNameSegment.js:31-37`

```javascript
construct(binding, propName) {
  super(binding);
  let lower = qx.lang.String.firstLow(propName);
    if (lower !== propName) {
      this.warn(`Binding: property name "${propName}" should be lower case, using "${lower}" instead`);
    }
  }
  this.__propName = lower;
}
```

## What Happens in v8

1. **New Binding System** (2022-2023 by John Spackman/Zenesis)
   - Completely rewritten Class/Property System
   - New binding with `SingleValueBinding` and `PropNameSegment`

2. **Automatic Conversion**
   - EVERY property name in binding is converted to lowercase
   - `"Username"` → automatically becomes `"username"`
   - `"EmailAddress"` → automatically becomes `"emailAddress"`

3. **In Debug Mode**
   - Warning: `"Binding: property name "Username" should be lower case, using "username" instead"`
   - But the lowercase version is ALWAYS used

4. **Binding Flow:**
   ```
   SingleValueBinding.__parseSegments(binding, path)
     → Line 785: new PropNameSegment(binding, seg)
       → PropNameSegment constructor
         → Line 31: let lower = qx.lang.String.firstLow(propName);
         → Line 37: this.__propName = lower;
   ```

## Why Issue #10808 Occurs

### Before (v7):
```javascript
form.add(field, "Username Label", null, "Username");
// Binding tries to bind property "Username" → works!
```

### Now (v8):
```javascript
form.add(field, "Username Label", null, "Username");
// Binding system automatically converts: "Username" → "username"
// Looks for property "username" (lowercase)
// → ERROR if property is named "Username"!
```

## The Technical Necessity

**v8 ENFORCES camelCase convention in the binding system!**

This is NOT just a warning or recommendation - it's an **automatic conversion**.

### Consequences

1. **All property names in binding MUST be lowercase-first**
   - This is now hard-coded in v8
   - No option to bypass it

2. **Our solution MUST use camelCase**
   - We CANNOT use "Username" as a property name
   - We MUST use "username"

3. **Collision Detection is ESSENTIAL**
   - If code has both "Username" and "username"
   - Both are converted to "username"
   - → Conflict!

## Why v8 Converts Only in Binding, Not in Model

**Critical understanding:** v8 converts property names **only during binding**, not during model creation!

1. **Form.js** - stores names as provided
2. **Form Controller createModel()** - creates model with property names
3. **v8 Binding** - converts during binding

**If we DON'T convert:**
```javascript
// User calls:
form.add(field, "Label", null, "Username");

// createModel() WITHOUT conversion creates:
model = { Username: value }  // Property named "Username"

// v8 Binding tries to bind:
// → looks for "username" (automatic firstLow!)
// → does NOT find it, because property is named "Username"
// → ERROR!
```

**With our conversion:**
```javascript
// User calls:
form.add(field, "Label", null, "Username");

// createModel() WITH conversion creates:
model = { username: value }  // Property named "username"

// v8 Binding tries to bind:
// → looks for "username" (automatic firstLow!)
// → finds it! ✓
// → WORKS!
```

## Solution B: Implementation Details

### Overview

Solution B implements camelCase conversion in the Form Controller while preserving backward compatibility for Form.getItem().

### Architecture

**Separation of Concerns:**
- **Form** (`qx.ui.form.Form`): Stores original field names
- **Form Controller** (`qx.data.controller.Form`): Converts to camelCase for model creation
- **Result**: `form.getItem("Username")` works, `model.getUsername()` works

### Implementation

#### 1. Helper Method: `__convertNameToCamelCase()`

Located in `source/class/qx/data/controller/Form.js:107-119`

```javascript
__convertNameToCamelCase(name) {
  if (!name) {
    return name;
  }

  // For deep binding, split by dot and convert each part
  var parts = name.split(".");
  var convertedParts = parts.map(function (part) {
    return qx.lang.String.firstLow(part);
  });

  return convertedParts.join(".");
}
```

**Features:**
- Handles simple names: `"Username"` → `"username"`
- Handles deep binding: `"User.Name"` → `"user.name"`
- Handles nested paths: `"Company.Owner.FirstName"` → `"company.owner.firstName"`

#### 2. Model Creation with Collision Detection

Located in `source/class/qx/data/controller/Form.js:176-248`

```javascript
createModel(includeBubbleEvents) {
  var target = this.getTarget();
  if (target == null) {
    throw new Error("No target is set.");
  }

  var items = target.getItems();
  var data = {};
  var nameMapping = {}; // For collision detection

  for (var name in items) {
    // Convert to camelCase for v8 compatibility
    var camelCaseName = this.__convertNameToCamelCase(name);

    // COLLISION DETECTION
    if (nameMapping[camelCaseName] && nameMapping[camelCaseName] !== name) {
      throw new Error(
        "Form field naming collision detected (issue #10808): " +
        "Fields '" + nameMapping[camelCaseName] + "' and '" + name + "' " +
        "both convert to the same camelCase property name '" + camelCaseName + "'.\n\n" +
        "This happens when field names differ only in capitalization.\n" +
        "qooxdoo v8's binding system automatically converts property names to lowercase first letter,\n" +
        "which causes both field names to map to the same model property.\n\n" +
        "To fix this issue:\n" +
        "  1. Rename one of the conflicting fields to have a distinct name\n" +
        "  2. Ensure all field names are unique when converted to camelCase\n\n" +
        "Conflicting fields:\n" +
        "  - '" + nameMapping[camelCaseName] + "'\n" +
        "  - '" + name + "'\n" +
        "Both map to model property: '" + camelCaseName + "'"
      );
    }
    nameMapping[camelCaseName] = name;

    // Create model with camelCase property names...
    var names = camelCaseName.split(".");
    // ... (property creation logic)
  }

  var model = qx.data.marshal.Json.createModel(data, includeBubbleEvents);
  this.setModel(model);
  return model;
}
```

**Collision Detection Features:**
- **Fail-Fast**: Error thrown immediately at model creation time
- **Detailed Error Message**: Shows exactly which fields collide
- **Helpful Guidance**: Suggests how to fix the issue
- **Issue Reference**: Links to #10808 for context

#### 3. Binding Methods

All binding-related methods updated to use `__convertNameToCamelCase()`:

**a) `addBindingOptions()` - Lines 136-164**
```javascript
addBindingOptions(name, model2target, target2model) {
  // ... setup code ...
  var modelPropertyName = this.__convertNameToCamelCase(name);
  this.__objectController.removeTarget(item, targetProperty, modelPropertyName);
  this.__objectController.addTarget(/*...*/, modelPropertyName, /*...*/);
}
```

**b) `updateModel()` - Lines 233-284**
```javascript
updateModel() {
  // ... validation code ...
  for (var name in items) {
    var modelPropertyName = this.__convertNameToCamelCase(name);
    qx.data.SingleValueBinding.updateTarget(/*...*/, modelPropertyName, /*...*/);
  }
}
```

**c) `_applyModel()` - Lines 281-346**
```javascript
_applyModel(value, old) {
  // ... cleanup code ...
  for (var name in items) {
    var modelPropertyName = this.__convertNameToCamelCase(name);
    this.__objectController.removeTarget(/*...*/, modelPropertyName);
  }
  // ... setup code ...
}
```

**d) `__setUpBinding()` - Lines 330-387**
```javascript
__setUpBinding() {
  // ... initialization ...
  for (var name in items) {
    var modelPropertyName = this.__convertNameToCamelCase(name);
    this.__objectController.addTarget(/*...*/, modelPropertyName, /*...*/);
  }
}
```

**e) `__tearDownBinding()` - Lines 395-418**
```javascript
__tearDownBinding(oldTarget) {
  // ... setup code ...
  for (var name in items) {
    var modelPropertyName = this.__convertNameToCamelCase(name);
    this.__objectController.removeTarget(/*...*/, modelPropertyName);
  }
}
```

### Backward Compatibility

#### Form.getItem() Preservation

**File:** `source/class/qx/ui/form/Form.js`
**Change:** NONE - Form stores original names exactly as provided

```javascript
// User code:
form.add(usernameField, "Username Label", null, "Username");

// Form stores "Username" as the key
// This still works:
var field = form.getItem("Username"); // ✓ Returns usernameField

// This does NOT work (but never did in Form):
var field = form.getItem("username"); // ✗ Returns null
```

#### Model Property Access

```javascript
// After createModel():
var model = controller.createModel();

// Model has camelCase properties:
model.setUsername("john");      // ✓ Works
var value = model.getUsername(); // ✓ Works

// Original capitalized name does NOT work in model:
model.setUsername("john");      // ✗ Error - property doesn't exist
```

### Error Messages

#### Collision Error Example

```javascript
// User code with collision:
form.add(field1, "User 1", null, "Username");
form.add(field2, "User 2", null, "username");

// Error thrown by createModel():
Error: Form field naming collision detected (issue #10808):
Fields 'Username' and 'username' both convert to the same camelCase
property name 'username'.

This happens when field names differ only in capitalization.
qooxdoo v8's binding system automatically converts property names to
lowercase first letter, which causes both field names to map to the
same model property.

To fix this issue:
  1. Rename one of the conflicting fields to have a distinct name
  2. Ensure all field names are unique when converted to camelCase

Conflicting fields:
  - 'Username'
  - 'username'
Both map to model property: 'username'
```

### Testing Strategy

All tests located in `source/class/qx/test/data/controller/Form.js`:

#### 1. `testCamelCaseConversion()` - Lines 952-1002
Tests that capitalized field names are correctly converted to camelCase model properties:
- Adds fields: "Username", "EmailAddress", "PassWord"
- Verifies model has: `getUsername()`, `getEmailAddress()`, `getPassWord()`
- Tests bidirectional binding

#### 2. `testGetItemAfterCamelCaseConversion()` - Lines 1004-1046
Tests backward compatibility - Form.getItem() uses original names:
- Adds field with name "Username"
- Verifies `form.getItem("Username")` works
- Verifies model has `getUsername()` (camelCase)
- Tests binding works correctly

#### 3. `testCollisionDetection()` - Lines 1156-1187
Tests fail-fast collision detection:
- Adds two fields: "Username" and "username"
- Verifies `createModel()` throws error
- Verifies error message contains "issue #10808"

### Files Modified

**Implementation:**
- `source/class/qx/data/controller/Form.js` - Main implementation
  - Added `__convertNameToCamelCase()` helper method
  - Modified `createModel()` with collision detection
  - Updated all binding methods to use conversion

**Tests:**
- `source/class/qx/test/data/controller/Form.js` - Test coverage
  - `testCamelCaseConversion()` - Conversion test
  - `testGetItemAfterCamelCaseConversion()` - Backward compatibility test
  - `testCollisionDetection()` - Collision detection test

**Documentation:**
- `V8_BINDING_ANALYSIS.md` - This file

**No Changes:**
- `source/class/qx/ui/form/Form.js` - Unchanged (stores original names)

### Benefits of Solution B

✅ **v8 Compatible**: Works with v8's automatic lowercase conversion
✅ **Backward Compatible**: `form.getItem("Username")` still works
✅ **Fail-Fast**: Collisions detected immediately with clear error message
✅ **Future-Proof**: Follows qooxdoo's camelCase convention
✅ **Well-Tested**: Comprehensive test coverage
✅ **Clear Separation**: Form stores original, Model uses camelCase

### Migration Guide for Users

#### No Action Needed

If your v7 code used lowercase-first field names:
```javascript
// This continues to work without changes:
form.add(field, "Username", null, "username");
```

#### Action Required

If your v7 code used capitalized field names:

**Before (v7):**
```javascript
form.add(usernameField, "Username", null, "Username");
form.add(emailField, "Email", null, "Email");

// Accessing the model:
model.setUsername("john");  // This was Username in v7
model.setEmail("john@example.com");  // This was Email in v7
```

**After (v8 with fix):**
```javascript
// Form.add() call stays the same:
form.add(usernameField, "Username", null, "Username");
form.add(emailField, "Email", null, "Email");

// Model properties are now camelCase:
model.setUsername("john");  // Property is now username (lowercase)
model.setEmail("john@example.com");  // Property is now email (lowercase)

// form.getItem() still uses original name:
form.getItem("Username");  // ✓ Still works
```

**Collision Detection:**
```javascript
// This will now throw an error:
form.add(field1, "User", null, "Username");
form.add(field2, "User", null, "username");  // ✗ Collision!

// Fix: Rename one field:
form.add(field1, "User", null, "Username");
form.add(field2, "User", null, "userLogin");  // ✓ No collision
```

## Conclusion

### ✅ v8 Has Stricter Enforcement

- Automatic lowercase conversion in binding system
- Hard-coded in `PropNameSegment.js`
- No way to bypass it

### ✅ Solution B is the Correct Approach

1. camelCase conversion in createModel()
2. Collision detection with fail-fast
3. Detailed error messages
4. Backward compatibility for getItem()

### ✅ Testing Confirms

- With Solution B: Everything works ✓
- Without conversion: Model has "Username", binding looks for "username" → ERROR ✗

## Technical References

**v8 Binding System:**
- `source/class/qx/data/binding/PropNameSegment.js:31-37` - Automatic conversion
- `source/class/qx/data/SingleValueBinding.js:785` - Creates PropNameSegment

**Our Solution:**
- `source/class/qx/data/controller/Form.js` - createModel() with conversion + collision detection
- `source/class/qx/ui/form/Form.js` - Stores original names for getItem()

**Copyright:**
- PropNameSegment: 2022-2023 Zenesis Limited (John Spackman)
- Part of v8.0 Class/Property System Rewrite

derrell:
@johnspackman Why is there now a limitation on property names that they must be camelCase beginning with a lower-case letter? That seems an unreasonable restriction, and I can't see the benefit of it...???

johnspackman:
I havn't looked in detail at how `Form` works so in apologies in advance, but from what I can see the idea is that `form.add` creates a class with properties based on the new fields, and uses `qx.data.marshal.Json` under the hood.

I imagine that the issue is not an issue with binding per se, but how the new property system works - and that properties are required to be camelCase.

The binding in v7 was implemented by `SingleValueBinding` which would "sniff" the properties and make a set of assumptions about them, including that the name was camel case and the event name followed the name of the property.

While it was possible to define properties in a non-standard way, this was not reliable or really supported by Qooxdoo, and ultimately would be considered a "bad practice".  If we look at all the thousands of properties in use they all follow this one standard.

For example, if the `SingleValueBinding` sees a binding path of "username" it will add events for `changeUsername` and call `setUsername`, `getUsername` etc - i.e. it will act as though there is a property (actual property or pseudo property) called `username`.


### Fix
In the examples above, the code specifies `form.add` with (eg) `"Username"` and then accesses the model with `setUsername` - i.e. the example code expects the property to be called `"username"`

Surely a simple (and BC) fix would be for the `form.add` to lowercase *the first character only*?  This would apply to `form.getItem` also?

oetiker:
I know I have quite a lot of code which programmatically creates forms, field names will often not show in the clear in the code but are generated, so an automatic migrator will most likely not pick them up. The code assumes that there are no restrictions on how to name the fields as it was in v7 (regardless if it was cool or not). 

This change is bound to provide lots of wonderful bug hunting opportunities in large applications ... NOT looking forward to it.

johnspackman:
@oetiker said:
> I know I have quite a lot of code which programmatically creates forms, field names will often not show in the clear in the code but are generated, so an automatic migrator will most likely not pick them up. The code assumes that there are no restrictions on how to name the fields as it was in v7 (regardless if it was cool or not).

I'm guessing a bit at what `form.add` actually did, but if it created a property on an object called "Username" then the result would have been an object that supported `setUsername()`, `getUsername()`, `resetUsername()` and `"changeUsername"` - and according to every other part of Qooxdoo, the property name was `username`.

Now that we support native properties, `myobject.username` and `myobject.Username` *could* refer to different things, but `myobject.getUsername()` can only have ever referred to *one* thing.

IMHO sounds like:

(a) the issue here is that validation in v8 is correctly saying that trying to distinguish between `Username` and `username` would be impossible (in v7 and in v8); 

(b) if we change the `form.add` to lowercase the first letter of the property name, we fix it for everyone, without a migration step

derrell (approved):

oetiker (approved):

cboulanger (approved):

PR #10811, opened 2025-11-06
Implement getPropertyDescriptor for issue #10805

### LINKED ISSUE #10805: Missing  getPropertyDescriptor for class instance in v8
**Describe the bug**
In branch branch_8_0_beta method `getPropertyDescriptor ` is missing but in docs said it exists.

**Expected behavior**
Method exists or removed any mention from docs about it.

hkollmann:
## Summary

Implements the missing `getPropertyDescriptor` method for accessing property metadata in qooxdoo v8, resolving issue #10805.

This PR adds two variants of the method:
- **Static method**: `qx.Class.getPropertyDescriptor(clazz, propertyName)` - Access property descriptor for any class
- **Instance method**: `instance.getPropertyDescriptor(propertyName)` - Convenient access from object instances

Both methods provide access to the first-class Property object in qooxdoo v8, returning the property configuration/metadata or `null` if the property doesn't exist.

### Usage Example

```javascript
// Static method
var descriptor = qx.Class.getPropertyDescriptor(MyClass, "myProperty");

// Instance method  
var instance = new MyClass();
var descriptor = instance.getPropertyDescriptor("myProperty");

// Access property configuration
console.log(descriptor.name);     // Property name
console.log(descriptor.init);     // Initial value
console.log(descriptor.check);    // Type check
console.log(descriptor.nullable); // Nullable flag

Fixes #10805

### DIFF
--- a/docs/core/understanding_properties.md
+++ b/docs/core/understanding_properties.md
@@ -1106,51 +1106,31 @@ retrieve values from a property.
 
 ### Getting a Property Descriptor
 
-There are two ways to obtain a property descriptor:
-
-**1. Via an instance method:**
+You can obtain a property descriptor using the static `qx.Class.getPropertyDescriptor()` method:
 
 ```javascript
 let instance = new MyClass();
-let descriptor = instance.getPropertyDescriptor("myProp");
-```
-
-**2. Via the static `qx.Class.getPropertyDescriptor()` method:**
-
-```javascript
 let descriptor = qx.Class.getPropertyDescriptor(MyClass, "myProp");
 ```
 
-Both methods return the same property descriptor object, or `null` if the
-property doesn't exist.
+The method returns a property descriptor object, or `null` if the property doesn't exist.
 
 ### Using Property Descriptors
 
-When obtained via an instance, the descriptor's `set()` and `get()` methods are
-already bound to that instance, allowing direct calls:
-
-```javascript
-let instance = new MyClass();
-let descriptor = instance.getPropertyDescriptor("myProp");
-
-// Set a value using the descriptor - simple and direct!
-descriptor.set(42);
-
-// Get the value using the descriptor
-let value = descriptor.get();
-```
-
-When using the static method, you can optionally provide an instance to get a
-bound descriptor, or use `.call()` for unbound descriptors:
+You can optionally provide an instance as the third parameter to get a descriptor
+with `set()` and `get()` methods bound to that instance:
 
 ```javascript
 // Get a bound descriptor (direct calls)
+let instance = new MyClass();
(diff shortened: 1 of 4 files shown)


johnspackman (changes requested):

hkollmann:
Test: https://github.com/hkollmann/qooxdoo/actions/runs/19167097650

derrell (changes requested):

hkollmann:
Reverted to instance.getPropertyDescriptor

johnspackman:
As I understand it, ‘myInstance.constructor’ is the only way to obtain the class of an instance - I agree that it’s not an intuitive name, but it is essential and we have to support it / recommend it. I entirely agree that it would be very poor practice to use ‘myInstance.constructor.SOMESTATIC’.I have a vague feeling that I learnt this from the Qooxdoo docs many years agoIt’s not a massive issue to have an instance method, I’m just conscious of the namespace pollutionSent from my iPhoneOn 8 Nov 2025, at 10:20, Henner Kollmann ***@***.***> wrote:﻿
@hkollmann commented on this pull request.



In docs/core/understanding_properties.md:
>  ```
 
-With that property descriptor in hand, you can now manipulate the
+#### Via Instance Constructor
+
+When you only have an instance, you can access the method through the instance's constructor:
+
+```javascript
+// Recommended: Direct call on constructor
+let propDesc = myInstance.constructor.getPropertyDescriptor("myProp");
+
+// Alternative: Passing constructor to qx.Class
+let propDesc = qx.Class.getPropertyDescriptor(myInstance.constructor, "myProp");
+```

After giving it some thought, I would agree with @derrel. myInstance.constructor is not very intuitive and is not used anywhere else in qooxdoo. That's why I would actually prefer myInstance.getPropertyDescriptor.
@johnspackman: Could you live with that?
@ALL: What do you think?

hkollmann:
So you agree?

oetiker (approved):

johnspackman:
> So you agree?

No, I think that `myInstance.constructor` is a valid and normal (and essential) expression when you need to get the class of an object, and that getting a property definition is something that is asked from a class.

For example, we do not have `myInstance.hasInterface`, `myInstance.supportsEvent`, ... or even `myInstance.hasProperty` and `myInstance.getProperties`

If `myInstance.getPropertyDefinition` was _significantly_ more useful on a daily basis to an instance that could be different, but I dont think that it is justified in this instance.

hkollmann:
You are right, QooxDoo way is qx.Class,.hasInterface. 
So i remove the instance method

hkollmann:
Done

johnspackman (approved):

cboulanger (approved):

goldim (approved):

derrell (approved):

PR #10812, opened 2025-11-07
Fix #10194: Auto-detect package changes without --clean flag

### LINKED ISSUE #10194: Adding a package requires "clean" for `npx qx serve` to notice it
After installing a package with `npx qx pkg add <package>`, simply restarting `npx qx serve` isn't enough for the package to be found and used. It is necessary to do `npx qx serve --clean` for it to locate the newly added package. I found a random message about "clean" in something unrelated. So the questions are, (a) should it be necessary to add --clean in order to get the packages to start being used, and (b) if so, is there a good place to document that?

@johnspackman responded to a query about this in gitter:
> the compiler checks for changes to the environment, it could easily check for changes to the list of available packages

hkollmann:
# Fix #10194: Auto-detect package changes without --clean flag

## Summary

Fixes #10194 - The compiler now automatically detects when packages are added or removed via `qx pkg add` without requiring the `--clean` flag.

## Problem

Previously, after installing a new package using `qx pkg add <package>`, users had to explicitly run `qx serve --clean` or `qx compile --clean` for the compiler to recognize the newly added package. Without the `--clean` flag, the compiler would use cached data and fail to detect the new library.

## Solution

Extended the `isContextChanged()` method in `Analyser.js` to automatically check if the list of available libraries has changed. The method now:

- Compares current libraries with those stored in the database
- Detects when libraries are added or removed
- Detects when library versions change

When a change is detected, the compiler automatically triggers a clean rebuild without requiring the user to specify `--clean`.

## Changes

- **File**: `source/class/qx/tool/compiler/Analyser.js`
- **Method**: `isContextChanged()`
- **Lines Added**: 26 new lines for library change detection

## Testing

Automated integration tests have been added to `test/tool/integrationtest/test-qx-package.js` that verify:

✅ Adding a package and compiling without `--clean` - package is automatically detected  
✅ Adding a second package and compiling without `--clean` - both packages detected  
✅ Removing a package and compiling without `--clean` - removal is automatically detected

Run the tests with:
```bash
npm test
Or run just the package tests:

node test/tool/integrationtest/test-qx-package.js
Impact
User Experience: Significantly improved - users no longer need to remember the --clean flag
Performance: No negative impact - only triggers rebuild when actually needed
Compatibility: Fully backward compatible - existing workflows continue to work
Related
Issue: #10194
Discussion: As noted by @johnspackman, "the compiler checks for changes to the environment, it could easily check for changes to the list of available packages"

### DIFF
--- a/source/class/qx/tool/compiler/Analyser.js
+++ b/source/class/qx/tool/compiler/Analyser.js
@@ -1088,7 +1088,7 @@ qx.Class.define("qx.tool.compiler.Analyser", {
 
     /**
      * Whether the compilation context has changed since last analysis
-     * e.g. compiler version, environment variables
+     * e.g. compiler version, environment variables, or available libraries
      *
      * @return {Boolean}
      */
@@ -1114,6 +1114,31 @@ qx.Class.define("qx.tool.compiler.Analyser", {
       // And Qooxdoo version (this can differ from the compiler version when cross compiling)
       if (db.libraries.qx !== this.findLibrary("qx").getVersion()) {
         return true;
+      }
+
+      // Check if the list of available libraries has changed (issue #10194)
+      // This ensures that newly added packages are detected without requiring --clean
+      if (db.libraries) {
+        const currentLibraries = this.getLibraries().reduce((acc, library) => {
+          acc[library.getNamespace()] = library.getVersion();
+          return acc;
+        }, {});
+
+        const dbLibraryKeys = Object.keys(db.libraries).sort();
+        const currentLibraryKeys = Object.keys(currentLibraries).sort();
+
+        // Check if a library was added or removed
+        if (dbLibraryKeys.length !== currentLibraryKeys.length ||
+            !dbLibraryKeys.every((key, index) => key === currentLibraryKeys[index])) {
+          return true;
+        }
+
+        // Check if any library version changed
+        for (let ns in currentLibraries) {
+          if (db.libraries[ns] !== currentLibraries[ns]) {
+            return true;
+          }
+        }
       }
 
       return false;
--- a/test/tool/integrationtest/test-qx-package.js
+++ b/test/tool/integrationtest/test-qx-package.js
@@ -125,3 +125,116 @@ test("Install qxl.test1@commit", async assert => {
   }
 });
 
+/**
+ * Tests for Issue #10194: Adding a package requires "clean" for `npx qx serve` to notice it
+ *
+ * These tests verify that the compiler automatically detects newly added/removed packages
+ * without requiring the --clean flag.
+ */
+
+test("Issue #10194: Setup - Initial compile for baseline", async assert => {
+  try {
+    // Clean up from previous tests
+    let result = await testUtils.runCommand(appDir, qxCmdPath, "package", "remove", "qooxdoo/qxl.test1");
+    // Ignore exit code - package may not be installed
+
+    // Initial compile to establish baseline
+    result = await testUtils.runCompiler(appDir);
+    assert.ok(result.exitCode === 0, testUtils.reportError(result));
+
+    // Verify db.json exists
+    const dbPath = path.join(appDir, "compiled/source/db.json");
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by derrell — APPROVED** [review]

**Review by oetiker — APPROVED** [review]

**Review by johnspackman — APPROVED** [review]

**Review by goldim — CHANGES_REQUESTED** [review]
The command `qx pkg` doesn't have subcommand `add`.
```
/var/www/html/test_beta$ npx qx --version
(node:47399) [DEP0176] DeprecationWarning: fs.F_OK is deprecated, use fs.constants.F_OK instead
(Use `node --trace-deprecation ...` to show where the warning was created)
7.9.2
/var/www/html/test_beta$ npx qx pkg --help
(node:47441) [DEP0176] DeprecationWarning: fs.F_OK is deprecated, use fs.constants.F_OK instead
(Use `node --trace-deprecation ...` to show where the warning was created)
qx pkg <command> [options]

alias for 'qx package'

Commands:
  qx pkg install [uri[@release_tag]]  installs the latest compatible release of
                                      package (as per Manifest.json). Use "-r
                                      <release tag>" or @<release tag> to
                                      install a particular release.
                                      examples:
                                      * qx package install name: Install latest
                                      published version
                                      * qx package install name@v0.0.2: Install
                                      version 0.0.2,
                                      * qx package install name@master: Install
                                      current master branch from github
  qx pkg list [repository]            if no repository name is given, lists all
                                      available packages that are compatible
                                      with the project's qooxdoo version
                                      ("--all" lists incompatible ones as well).
                                      Otherwise, list all compatible packages.
  qx pkg publish                      publishes a new release of the package on
                                      GitHub. Requires a GitHub access token. By
                                      default, makes a patch release.
  qx pkg remove [uri]                 removes a package from the configuration.
  qx pkg update [repository]          updates information on packages from
                                      github. Has to be called before the other
                                      commands. If a package URI is supplied,
                                      only update information on that package
  qx pkg upgrade [library_uri]        if no library URI is given, upgrades all
                                      available libraries to the latest
                                      compatible version, otherwise upgrade only
                                      the package identified by the URI.
  qx pkg migrate                      migrate a library/application to a newer
                                      qooxdoo version

Options:
  --force, -F, -F        Override warnings            [boolean] [default: false]
  --config-file, -c, -c  Specify the config file to use                 [string]
  --verbose, -v, -v      enables additional progress output to console
                                                      [boolean] [default: false]
  --debug                enables debug output         [boolean] [default: false]
  --quiet, -q, -q        suppresses normal progress output to console  [boolean]
  --colorize             colorize log output to the console using ANSI color
                         codes                         [boolean] [default: true]
  --version              Show version number                           [boolean]
  --set                  sets an environment value for the compiler      [array]
  --set-env              sets an environment value for the application   [array]
  --help                 Show help                                     [boolean]
```

hkollmann:
It's wrong in the issue. https://github.com/qooxdoo/qooxdoo/issues/10194
Means install. Sorry.
Implementation is correct and uses install

goldim:
@hkollmann I tried with `qx pkg install` but couldn't reproduce the issue. I thought that I missed something and reread the issue and found `add`. I have to try repoduce it. What I did:
1) created my package 
2) install it in my project
3) run it (check trace shown)
4) changed the package: changed version and changed text of trace
5) installed it again
6) run it (new trace shown)

This scenario works for v7 and v8. Is it correct scenario?

hkollmann:
qx pkg upgrade could do the job

goldim:
@hkollmann uh, I try to install another package then.

hkollmann:
you can try the test withou the fix

goldim (approved):
I did try and it works. For `qx serve` only after stop and rerun the command.  For PR it is enough but it would be great if `qx serve` applied changes after installing new package or removing on fly.

hkollmann:
that should be another issue

goldim:
Agree, I will open it.

cboulanger (approved):

PR #10815, opened 2025-11-09
Fix #10411: Correct blocker positioning in blockContent()

### LINKED ISSUE #10411: qx.ui.core.Blocker.blockContent() misplaces the blocker element.

hkollmann:
## Summary

Fixes #10411 - `qx.ui.core.Blocker.blockContent()` was misplacing the blocker element due to incorrect positioning calculation.

### Problem
When `blockContent()` was called, the blocker element became a child of the target widget rather than its layout parent. The blocker was positioned using the widget's bounds (left, top coordinates), which caused a double offset since the blocker was already positioned relative to the widget.

### Solution
- Added `__blockingContent` member variable to track when the blocker is in content blocking mode
- Modified `_updateBlockerBounds()` to use `left: 0, top: 0` when `__blockingContent` is true (blocker is child of widget)
- Fixed `_block()` to properly pass the `blockContent` parameter to the appear listener binding
- Reset `__blockingContent` to false in `__unblock()`

## Test Plan

Added three new test cases:
- [x] `testBlockContent()` - Verifies basic blockContent functionality and correct 0,0 positioning
- [x] `testBlockContentPositioning()` - Tests with container at non-zero position (150, 100) to ensure blocker is still positioned at 0,0
- [x] `testNormalBlockVsBlockContent()` - Compares normal `block()` vs `blockContent()` to verify different positioning behaviors

### DIFF
--- a/source/class/qx/test/ui/core/Blocker.js
+++ b/source/class/qx/test/ui/core/Blocker.js
@@ -236,6 +236,127 @@ qx.Class.define("qx.test.ui.core.Blocker", {
 
       txt.destroy();
       this.flush();
+    },
+
+    testBlockContent() {
+      // Create a container widget with specific position
+      var container = new qx.ui.container.Composite(new qx.ui.layout.Canvas());
+      this.getRoot().add(container, { left: 100, top: 50 });
+      container.setWidth(200);
+      container.setHeight(150);
+      this.flush();
+
+      // Create blocker for the container
+      var blocker = new qx.ui.core.Blocker(container);
+      var blockerElement = blocker.getBlockerElement();
+
+      // Block content
+      blocker.blockContent(10);
+      this.flush();
+
+      this.assertTrue(blocker.isBlocked(), "blocker should be blocked");
+      this.assertTrue(blockerElement.isIncluded(), "blocker element should be included");
+
+      // Verify blocker dimensions match container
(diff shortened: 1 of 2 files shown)


cboulanger (approved):

oetiker (approved):

johnspackman (approved):

derrell (approved):

goldim (approved):
Could be moved to v7

PR #10820, opened 2025-11-12
Fix issue #10816: Wrap assert calls with debug checks

### LINKED ISSUE #10816: Missing asserts in build target cause error
**Describe the bug**
Some code has `this.assert` calls.

**To Reproduce**
1. Compile build version with `qx.ui.form.renderer.Single` for example
2. Open console

**Expected behavior**
No asserts via this.

 - OS: Ubuntu 22.04
 - Browser any
 - Version qooxdoo 8

**Additional context**
Noticed in `qx.data.SingleValueBinding` code.
Solution is to frame with `if (qx.core.Environment.get("qx.debug")) {`.

hkollmann:
This commit addresses issue #10816 where assertion statements were appearing in production builds. All `this.assert*` calls have been wrapped with `if (qx.core.Environment.get("qx.debug"))` checks to ensure they are stripped out in production builds.

Changes include:
- qx/Interface.js: Wrapped assertObject and assert method bodies
- qx/data/SingleValueBinding.js: Wrapped assertNotNull call
- qx/data/binding/ArrayIndexSegment.js: Wrapped assertTrue call
- 9 interface files: Wrapped all assertion calls in interface method stubs
  * qx/ui/tree/provider/IVirtualTreeProvider.js
  * qx/ui/tree/core/IVirtualTree.js
  * qx/ui/virtual/core/ILayer.js
  * qx/ui/form/IListItem.js
  * qx/ui/form/IRadioItem.js
  * qx/ui/window/IDesktop.js
  * qx/ui/window/IWindowManager.js
  * qx/ui/core/scroll/IScrollBar.js
  * qx/event/IEventDispatcher.js

This ensures that assertions do not appear in production builds, reducing bundle size and preventing potential runtime errors.

Fixes #10816

### DIFF
--- a/source/class/qx/Interface.js
+++ b/source/class/qx/Interface.js
@@ -359,16 +359,18 @@ qx.Bootstrap.define("qx.Interface", {
      *  @param iface {Interface} The interface to verify
      */
     assertObject(object, iface) {
-      var clazz = object.constructor;
-      this.__checkMembers(object, clazz, iface, false, true);
-      this.__checkProperties(clazz, iface, true);
-      this.__checkEvents(clazz, iface, true);
+      if (qx.core.Environment.get("qx.debug")) {
+        var clazz = object.constructor;
+        this.__checkMembers(object, clazz, iface, false, true);
+        this.__checkProperties(clazz, iface, true);
+        this.__checkEvents(clazz, iface, true);
 
-      // Validate extends, recursive
-      var extend = iface.$$extends;
-      if (extend) {
-        for (var i = 0, l = extend.length; i < l; i++) {
-          this.assertObject(object, extend[i]);
+        // Validate extends, recursive
+        var extend = iface.$$extends;
+        if (extend) {
+          for (var i = 0, l = extend.length; i < l; i++) {
+            this.assertObject(object, extend[i]);
+          }
         }
       }
     },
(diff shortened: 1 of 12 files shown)


derrell (changes requested):
Is there ever a time that `assert()` calls should be logged to the console in a `build` environment? The code, in general, would be far cleaner if we added the environment check in `assert` itself, and revert this PR's changes...

goldim:
@derrell what about to make it viable in other PR? This PR required to fix some forgotten asserts in props. I am not sure that is good idea to use env checks in interfaces.

goldim (approved):
I am not sure that there is a point to check env in interface classes because it is framed when they are used but I think there is no negative effect.

hkollmann:
This pattern is used in the whole framework. This PR just fixes the forgotten ones.
If we put it in the assert method
```
        this.__checkMembers(clazz.prototype, clazz, iface, wrap, true);
        this.__checkProperties(clazz, iface, true);
        this.__checkEvents(clazz, iface, true);

        // Validate extends, recursive
        var extend = iface.$$extends;
        if (extend) {
          for (var i = 0, l = extend.length; i < l; i++) {
            this.assert(clazz, extend[i], wrap);
          }
        }
```
is not thrown out.

derrell (approved):

johnspackman (approved):

oetiker (approved):

cboulanger (approved):

PR #10825, opened 2025-11-16
Fix issue #9142: Allow classes to override mixin methods

### LINKED ISSUE #9142: Mixins: allow mixed in methods to be overridden
Methods which are included via a Mixin cannot be overridden in the class that mixed them in; in conceptual OO terms, there is no reason why this restriction should exist because a derived class can easily override the method.

It would be useful to see this restriction lifted

hkollmann:
## Summary

This PR fixes issue #9142 by allowing classes to override methods defined in mixins, aligning qooxdoo's behavior with standard object-oriented principles.

## Problem

Previously, qooxdoo threw an error "Overwriting member or property of Class is not allowed!" when a class attempted to override a method defined in a mixin. This restriction was inconsistent with OOP principles where derived classes can override parent methods.

## Solution

The fix introduces a mechanism to track class-defined members and allows them to take precedence over mixin members:

1. **Track class-defined members**: Added `$$ownMembers` property to track which members are defined by the class itself (not from mixins)
2. **Modified member addition logic**: Updated `addMembers()` to check if a member is class-defined before throwing overwrite errors
3. **Class members take precedence**: When a mixin tries to add a member that the class has already defined, the mixin member is skipped

### Core Implementation (`source/class/qx/Class.js`)
- Track class-defined members in `$$ownMembers` during class definition
- Modified `addMembers()` to allow class members to override mixin members
- Preserved existing behavior for mixin-to-mixin conflicts (still throws errors)
- No changes to `qx.Class.patch()` behavior (still overwrites as before)

### Tests (`source/class/qx/test/Mixin.js`)
- Added comprehensive `testMixinMethodOverride()` test covering:
  - Simple override without calling super
  - Override with super call to extend mixin functionality
  - Inheritance of overridden methods
  - Multiple mixin scenarios
- Updated existing tests that expected errors when overriding mixin methods
- Added verification that class members take precedence

## Behavior Changes

| Scenario | Before | After |
|----------|--------|-------|
| Class overrides mixin method | ❌ Error thrown | ✅ Allowed, class version used |
| `qx.Class.include()` with conflicts | ❌ Error thrown | ✅ Allowed, class version preserved |
| Mixin-to-mixin conflicts | ❌ Error thrown | ❌ Error thrown (unchanged) |
| `qx.Class.patch()` | ✅ Overwrites | ✅ Overwrites (unchanged) |

## Test Plan

- [x] Added new test `testMixinMethodOverride()` demonstrating the desired behavior
- [x] Updated existing tests to reflect new behavior
- [x] All scenarios covered:
  - Class overrides mixin method without super
  - Class overrides mixin method with super call
  - Inheritance of overridden methods
  - Post-definition mixin inclusion with `qx.Class.include()`

## Files Changed

- `source/class/qx/Class.js` - Core implementation
- `source/class/qx/test/Mixin.js` - Tests

## Related Issue

Fixes #9142

### DIFF
--- a/source/class/qx/test/Mixin.js
+++ b/source/class/qx/test/Mixin.js
@@ -959,18 +959,21 @@ qx.Class.define("qx.test.Mixin", {
       this.assertEquals("Hello from Class", obj1.greet());
       obj1.dispose();
 
-      // Test 2: Class overrides mixin method and calls super
+      // Test 2: Class overrides mixin method and calls base
+      // Note: We use this.base(arguments) instead of super because
+      // JavaScript's super keyword looks in the superclass, not in mixins
       qx.Class.define("qx.OverrideClass2", {
         extend: qx.core.Object,
         include: qx.MOverridable,
 
         members: {
           greet() {
-            return super.greet() + " and Class";
+            var mixinGreeting = this.base(arguments);
+            return mixinGreeting + " and Class";
           },
 
           calculate(a, b) {
-            var mixinResult = super.calculate(a, b);
+            var mixinResult = this.base(arguments, a, b);
             return mixinResult * 2;
           }
         }
@@ -1005,7 +1008,7 @@ qx.Class.define("qx.test.Mixin", {
 
         members: {
           getValue() {
-            return "Overridden: " + super.getValue();
+            return "Overridden: " + this.base(arguments);
           }
         }
       });
--- a/source/class/qx/test/Mixin.js
+++ b/source/class/qx/test/Mixin.js
@@ -21,6 +21,7 @@
  * @ignore(qx.MMix2, qx.MPatch, qx.Patch1, qx.Patch2, qx.UseLog1, qx.UseLog2)
  * @ignore(qx.UseLog3, qx.MOverridable, qx.MOverridable2, qx.OverrideClass1)
  * @ignore(qx.OverrideClass2, qx.OverrideClass3, qx.OverrideClass4)
+ * @ignore(qx.PatchIncludeTest)
  */
 
 qx.Class.define("qx.test.Mixin", {
@@ -67,6 +68,7 @@ qx.Class.define("qx.test.Mixin", {
       this.assertEquals("red", mix.getColor());
 
       if (this.isDebugOn()) {
+        // Test that including multiple mixins with conflicting members still throws an error
         this.assertException(
           function () {
             qx.Class.define("qx.Mix1", {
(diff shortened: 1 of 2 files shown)


derrell (changes requested):
I'm not convinced that I agree with the premise of this issue. I do agree that a subclass should be able to override member methods of its superclass that were included into the superclass via mixin. I disagree, however, that the class including the mixin should be able to override the mixin. My expectation is that the class is fully created before the mixin is included, and thus it would be an error for the mixin to attempt to override a member of the class (unless done using `patch`).

source/class/qx/test/Mixin.js
```diff
@@ -959,18 +959,21 @@ qx.Class.define("qx.test.Mixin", {
...
         members: {
           greet() {
-            return super.greet() + " and Class";
+            var mixinGreeting = this.base(arguments);
+            return mixinGreeting + " and Class";
           },
 
           calculate(a, b) {
-            var mixinResult = super.calculate(a, b);
+            var mixinResult = this.base(arguments, a, b);
```
johnspackman: Please can you explain this?  `super()` is recommended and `this.base` could be deprecated

johnspackman (changes requested):

johnspackman:
> I disagree, however, that the class including the mixin should be able to override the mixin. My expectation is that the class is fully created before the mixin is included, and thus it would be an error for the mixin to attempt to override a member of the class (unless done using `patch`).

IMHO when a mixin is included into a class, the class knows what is being mixed in; the developer makes the choice to mix it in, and I agree that the methods in the mixin can override methods in the super class.

Because the class knows what is being mixed in, it seems reasonable that the class can choose to override and augment those methods - it has to be done with a certain amount of thought and care, just like overriding a method in a super class.

I.e. it would not make sense for a mixin's methods to replace those in the class because the mixin is written independently with no knowledge of the class, but the opposite is not true - the class knows exactly what the mixin is and does, so if it chooses to override the mixin it can.

At the moment, overriding the mixin's methods can be done by defining a separate class for the sole purpose of mixing in the mixin, and then deriving from that class.

The effect (of writing a special purpose class for including mixins) is the same as this proposal, and I can't see the benefit of making the developer jump through that extra hoop.

derrell:
@johnspackman Point made. Thanks.

derrell (approved):

cboulanger (approved):

oetiker (approved):

goldim (approved):

goldim:
@johnspackman Hello, could you check this PR please?

johnspackman (approved):

johnspackman:
aagh!! sorry, i merged on automatic pilot here (I somehow forgot that this is a Qooxdoo repo and not one of our internal ones)

I think i was the last needed review though, apologies if not...

PR #10844, opened 2026-02-21
fix: replace browserify/babelify with esbuild for CommonJS bundling

hkollmann:
## Summary

- Replaces `browserify@17` and `babelify@10` with `esbuild@0.27` + `esbuild-plugin-polyfill-node`
- Removes 118 transitive packages, eliminates the `glob@7` → `inflight` deprecation chain
- `esbuild` handles ESM and CJS natively — no Babel transform needed for bundling
- `globalThis.require()` shim maintains full runtime compatibility with existing Qooxdoo code using `require('module-name')`
- Bundle shrinks from ~1.4MB to ~17KB for uuid (esbuild uses browser-native APIs instead of Node.js polyfills)
- All 25 integration tests pass


johnspackman (approved):

oetiker (approved):

cboulanger (approved):

derrell (approved):

goldim (changes requested):
Can't approve so far. Have troubles with including some node packages in my app. Also 'noParse' option should be removed or replaced/adopted.

goldim:
I think my problem is that I used global.<<some exported>> stuff in my node package. And now has to replace it with globalThis. global is undefined. globalThis isn't. Is it fine? Is it not broken BC? Could someone clarify me this moment please?

goldim:
I used `global` and browserify/babel processed this case. Should this utility do the same? I can make changes in my package but maybe do this on framework level?

patryk-m-malinowski (changes requested):

patryk-m-malinowski (approved):

goldim:
@hkollmann I have an error and debugging. Your last changes improved situation a lot and I get running application but still there is an error related with parsing already browserified package/dependency by esbuilder. I found place and commenting/skipping this place gives me good working app. Esbuilder creates new nest classes and calling it like parent class but with next index. For example "ClientSubscriptionImpl" I get @ClientSubscriptionImpl2" class. But it is fine except asserts which checks against parent name. I try figure out why it happens and how to avoid it.

goldim:
In browserify even without `noParse` the code of included already browserifed code is untouched. The option `noParse` option gives optiomization no parsing the second time. In current esbuilder it does some changes to this code.

hkollmann:
how do you include browserified code?

goldim:
@hkollmann with require(). It is just a node package.
```javascript
const impl = require("nodespace/node-opcua-bundle");
return impl.OPCUA;
```

goldim:
@hkollmann It may help! Thanks. Gonna check it.

goldim (approved):
Awesome. Thank you.

PR #10847, opened 2026-02-26
fix: resolve mixin this.base() clobbering when mixin is included in multiple classes

hkollmann:
## Problem

When the same mixin is included in **two different classes**, `qx.Class` sets `member.base` on the **shared mixin function object** each time. The second class overwrites `.base` for both, so `super` calls from the mixin resolve to the wrong base method in the first class.

**Example:**

```javascript
qx.Mixin.define("qx.MGreet", {
  members: {
    greet() { return super.greet() + " [mixed]"; }
  }
});

qx.Class.define("qx.ClassA", {
  extend: qx.core.Object,
  include: [qx.MGreet],
  members: { greet() { return "Hello from A"; } }
});

// ClassB is defined later — overwrites MGreet.$$members.greet.base
qx.Class.define("qx.ClassB", {
  extend: qx.core.Object,
  include: [qx.MGreet],
  members: { greet() { return "Hello from B"; } }
});

new qx.ClassA().greet(); // Expected: "Hello from A [mixed]"
                          // Actual:   "Hello from B [mixed]"  ← wrong base!
```

The same problem occurs when a class is patched with multiple mixins that each override the same method: the shared `.base` pointer causes infinite recursion.

## Fix

Store base pointers per-mixin-per-class in `clazz.$mixinBases: Map<mixin, {methodName: base}>` instead of on the shared function object. `qx.Mixin.baseClassMethod` looks up the correct base via `mixedInAt.$mixinBases.get(mixin)[methodName]`, ensuring each class/mixin combination resolves to its own base independently.

**`Class.js`** — per-mixin-per-class storage in `addMembers` (called with the mixin as new 5th arg from `addMixin`):
```javascript
member.base = clazz.prototype[key];
if (mixin) {
  if (!clazz.$mixinBases) clazz.$mixinBases = new Map();
  if (!clazz.$mixinBases.has(mixin)) clazz.$mixinBases.set(mixin, {});
  clazz.$mixinBases.get(mixin)[key] = clazz.prototype[key];
}
```

**`Mixin.js`** — look up `$mixinBases` in `baseClassMethod`, with fallback to `.base` chain:
```javascript
if (mixedInAt.$mixinBases && mixedInAt.$mixinBases.has(mixin)) {
  fn = mixedInAt.$mixinBases.get(mixin)[methodName] || null;
}
if (!fn) {
  fn = mixedInAt.prototype[methodName];
  for (let i = 0; i < mixedInAt.$$flatIncludes.length; i++) {
    if (!mixedInAt.$$flatIncludes[i].$$members[methodName]) continue;
    fn = fn.base;
  }
}
```

### DIFF
--- a/source/class/qx/Class.js
+++ b/source/class/qx/Class.js
@@ -803,10 +803,18 @@ qx.Bootstrap.define("qx.Class", {
      * @param members {Map}
      *   The map of members to attach
      *
+     * @param events {Map?}
+     *   The map of events declared by the mixin (used for member validation)
+     *
      * @param patch {Boolean ? false}
      *   Enable patching
+     *
+     * @param mixin {qx.Mixin?}
+     *   The mixin being applied, if any. Used to store per-mixin-per-class base
+     *   method references so that super calls resolve correctly when the same
+     *   mixin is included in multiple classes.
      */
-    addMembers(clazz, members, events, patch) {
+    addMembers(clazz, members, events, patch, mixin) {
       let proto = clazz.prototype;
       let classOwnMembers = {}; // Track class members to restore after mixin addition
 
@@ -906,11 +914,14 @@ qx.Bootstrap.define("qx.Class", {
               member.self = clazz;
             }
             member.base = clazz.prototype[key];
-            // Fix: per-class storage so multiple class-applications don't clobber the shared fn.base
-            if (!clazz.$memberBases) {
-              clazz.$memberBases = {};
+            // Fix: per-mixin-per-class storage to prevent clobbering when the same mixin
+            // is included/patched into multiple classes, or when multiple mixins override
+            // the same method in the same class.
+            if (mixin) {
+              if (!clazz.$mixinBases) clazz.$mixinBases = new Map();
+              if (!clazz.$mixinBases.has(mixin)) clazz.$mixinBases.set(mixin, {});
+              clazz.$mixinBases.get(mixin)[key] = clazz.prototype[key];
             }
-            clazz.$memberBases[key] = clazz.prototype[key];
           }
 
           // Create the storage for this member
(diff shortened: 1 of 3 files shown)


source/class/qx/Class.js
```diff
@@ -906,6 +906,11 @@ qx.Bootstrap.define("qx.Class", {
               member.self = clazz;
             }
             member.base = clazz.prototype[key];
+            // Fix: per-class storage so multiple class-applications don't clobber the shared fn.base
+            if (!clazz.$memberBases) {
+              clazz.$memberBases = {};
+            }
+            clazz.$memberBases[key] = clazz.prototype[key];
```
derrell: At line 905, it checks whether `member` is a function. This implies that `member` could also be a member variable. Is it appropriate to store anything in `$memberBases` for member variables? The modified code in Mixin.js already checks whether there is an entry in `$memberBases`. I think it would be better to only add an entry to `$memberBases` for member _functions_ but not for member _variables_.

source/class/qx/Mixin.js
```diff
@@ -361,7 +361,10 @@ qx.Bootstrap.define("qx.Mixin", {
               if (!mixedInAt.$$flatIncludes[i].$$members[methodName]) {
                 continue;
               }
-              fn = fn.base;
+              // Fix: use per-class $memberBases instead of shared mutable fn.base
+              fn = (mixedInAt.$memberBases && mixedInAt.$memberBases[methodName] !== undefined)
+                ? mixedInAt.$memberBases[methodName]
+                : fn.base;
```
derrell: I believe the conditional will be false when the member is a variable rather than a function. Is that correct? This should have a comment explaining when `fn.base` will be used.

derrell (changes requested):

patryk-m-malinowski:
There are a few things which don't make sense to me. You've mentioned the classes `RadioButtonGroupContainer` and `RadioButtonGroupBox` but I can't see them in the Qooxdoo project. Also, do you have example code which can reproduce the issue on this commit?

hkollmann:
I added a new test

patryk-m-malinowski (approved):

johnspackman (approved):

oetiker (approved):

goldim (approved):

hkollmann:
@derrell, @cboulanger: Could you please review this? Many thanks.

cboulanger (approved):

PR #10851, opened 2026-03-09
docs: document data payload for making/made events in Compile

hkollmann:
## Summary
- `making` and `made` events were changed to `qx.event.type.Data` and now fire `{maker}` as data payload
- Updated JSDoc in `Compile.js` to document the `maker` property in the event data
- Updated `docs/development/compiler/internals/Events.md` (CLI Commands section) accordingly

### DIFF
--- a/docs/development/compiler/internals/Events.md
+++ b/docs/development/compiler/internals/Events.md
@@ -26,11 +26,11 @@ which your code can listen. Here is a list of events with a brief explanation:
 ## CLI Commands
 
 Instances of `qx.tool.cli.commands.Compile` and its subclasses fire the following events:
-- `checkEnvironment`: Fired after all environment data is collected. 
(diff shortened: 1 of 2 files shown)


johnspackman (approved):

derrell (approved):

cboulanger (approved):

goldim:
@hkollmann maker is needed for future goals? I don't know what a point to fire it, sorry.

patryk-m-malinowski:
What help will it give us when the maker is returned by the making/made events? If you have a Qooxdoo project with multiple targets (hence multiple makers), then assuming that the order the makers will make the applications is undefined, this behaviour will be non-deterministic and dependent on which maker starts making first or stops making last.

patryk-m-malinowski (changes requested):
Please see my comment

hkollmann:
is not working as expected

PR #10852, opened 2026-03-26
Eliminate global registries for events and bindings

patryk-m-malinowski:
This PR removes global registries used for storing event listeners and bindings, meaning that objects are no longer being prevented from being garbage collected without manually disposing them. Instead, the metadata is being stored in the objects themselves using properties with `Symbol` keys, which guarantees no naming collisions.

For bindings, the static properties `__bindingsBySource` and `__bindingsByTarget` which stored the bindings for source and target objects were removed from `qx.data.SingleValueBinding`. The source/target object now stores its bindings under `qx.data.SingleValueBinding.__SymbolBindingsInfo`.
This means that the following code would create a memory leak when the function returns (unless the objects are disposed), but it doesn't now:
```js
function foo() {
  let person1 = new app.Person();
  let person2 = new app.Person();
  person1.bind("name", person2, "name");
}
```

Similarly, for events, `qx.event.Manager` no longer uses a possibly-large `Map` to store listeners for targets but stores the info in the target object with key `qx.event.Manager.__SymbolListeners`. `qx.event.Registration` does something similar for its window objects. This solves the problem where if a listener callback has a dependent object in its closure then that dependent object is prevented from being garbage-collected.

The code below would create a memory leak before but doesn't anymore:
```js
function foo() {
  let person1 = new app.Person();
  let person2 = new app.Person();
  person1.addListener("changeName", (evt) => person2.setName(evt.getData()));
}
```

This PR introduces the breaking changes where the functions `qx.data.SingleValueBinding.getAllBindings` and `qx.event.Manager.getAllListeners` were removed, but I don't think those functions have any practical use cases and they can return enormous amounts of data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 # 8.0.0-beta.2
 
 ## Breaking changes
+- Bindings (made using class qx.data.SingleValueBinding) are now no longer stored in a global registry,
+because this prevented their source and target objects from being garbage-collected without manually disposing them first.
+- Removed `qx.data.SingleValueBinding.getAllBindings` as a consequence of the above,
+and also because there is no practical use case for it.
+- Event listener callbacks are now stored in their target objects themselves instead of a global registry,
+because this prevented objects with event listeners from being garbage collected unless they were manually disposed,
+and also prevented any objects in the closures of the listeners' callbacks from being garbage collected as well.
+- Removed method `qx.event.Mananger.getAllListeners` as a consequence of the above,
+and also because there is no practical use case for it.
 - Removed `async: true` key from property definition because now all properties have `setAsync` methods,
 which can be used to await the apply function's return value and the event handlers when a property is set. 
 The meaning of this setting wasn't very clear, also given that we now have async property storage.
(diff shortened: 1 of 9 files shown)


hkollmann (approved):

derrell (approved):

cboulanger (approved):

oetiker:
why did this get merged with only 3 votes ?

oetiker:
a breaking change must at least be added to the migration assistant!

better still having dummy placeholders for the removed functions and having them issue warnings to the console and returning an empty list would provide a better chance for non-fixed code to continue working than simply erroring-out.

johnspackman:
> This PR introduces the breaking changes where the functions qx.data.SingleValueBinding.getAllBindings and qx.event.Manager.getAllListeners were removed, but I don't think those functions have any practical use cases and they can return enormous amounts of data.

@patryk-m-malinowski what we discussed was that these APIs were only used by the unit testing framework (AFAIK); and that while it is flawed to have that use because the testing framework is not able to run tests 100% perfectly isolated, to support backwards compatibility (and the unit test framework), you would create a plug-in API which allows all bindings to be recorded.

EG you would create an interface which has methods along the lines of `registerBinding` and `unregisterBinding`, and create a static method which allowed an instance of that interface to be registered.

PR #10853, opened 2026-03-31
async pseudo property fixes

patryk-m-malinowski:
Problem: `qx.core.property.Property.hasLocalValue` fails on pseudo properties, which can cause errors in things like asynchronous bindings.

Solution: Introduce a way of getting a pseudo property safely. Do this by adding a boolean parameter called `safe` which when set to `true`, make the getter return `undefined` if the property is not ready.

To make things more consistent, getting a property safely now requires passing `true` as an argument into the injected getter, and methods `getSafe` from `qx.core.property.Property` and `MProperty` were removed, which is a minor breaking change for new v8 features only.

### DIFF
--- a/docs/core/advanced_properties.md
+++ b/docs/core/advanced_properties.md
@@ -195,6 +195,9 @@ It will first attempt to get the property synchronously and then attempt asynchr
 NOTE: It is possible but discouraged to call `getAsync` on a property that doesn't support an async getter.
 Doing so will print out a warning.
 
+The property object (`qx.core.property.Property`) will support async getting,
+meaning that calling `supportsGetAsync` on it will return `true`.
+
 We can `get` the property value synchronously after the initial `getAsync`, like in the following example:
 ```js
 let myObject = myObjectFactory.getByUuid("asdfasdfk");
@@ -260,3 +263,50 @@ you can add `get`, `set`, and optionally `getAsync` functions to your definition
 
 This is known as inline or explicit property storage. This internally uses `qx.core.property.ExplicitPropertyStorage`,
 which simply defers to the `get`, `set`, `setAsync` functions in your definition.
+
+## Pseudo properties with async getter
+
+If a class contains a pseudo property (i.e. a hand-written property with at least get/event)
+and a `getPropertyAsync` method, the property object will treat the property as one supporting an async getter,
+meaning `property.supportsGetAsync` will return `true` and calling `property.getAsync(obj)` will straight defer to the `getPropertyAsync()` method.
+
+Example code:
+```js
+let Clazz = qx.Class.define("app.AsyncPseudoProperty", {
+  extend: qx.core.Object,
+
+  events: {
+    changePseudoProp: "qx.event.type.Data"
+  },
+
+  members: {
+    __pseudoProp: undefined,          
+
+    async getPseudoPropAsync() {
+      if (this.__pseudoProp !== undefined) {
+        return this.__pseudoProp;
+      }
+      return (this.__pseudoProp = 24);
+    },
+
+    getPseudoProp(safe) {            
+      if (!safe && !this.__pseudoProp) {
+        throw new Error("Property " + this.classname + ".pseudoProp is not yet ready!");
+      }
+      return this.__pseudoProp;
+    },
+
+    setPseudoProp(value) {
+      let oldValue = this.__pseudoProp;
+      this.__pseudoProp = value;
+      this.fireDataEvent("changePseudoProp", value, oldValue);
+    }
+  }
+});
+
+let obj = new app.AsyncPseudoProperty();
+let prop = qx.Class.getByProperty(obj.contructor, "pseudoProp");
+prop.supportsGetAsync() === true
+prop.getAsync(obj) === 24
+```
+
--- a/source/class/qx/core/MProperty.js
+++ b/source/class/qx/core/MProperty.js
@@ -146,13 +146,13 @@ qx.Mixin.define("qx.core.MProperty", {
       if (safe) {
         let property = qx.Class.getByProperty(this.constructor, prop);
         if (property) {
-          return property.getSafe(this);
+          return property.get(this, safe);
         }
       }
 
       // Otherwise, see if there's a hand-written getter method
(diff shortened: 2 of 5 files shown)


hkollmann (approved):

goldim (changes requested):
изображение

Doesn't work. Maybe to do left old behaviour and give warning or just ignore it?

goldim:
Also what about revert all changes which were commited directly into 8 beta and last merged request?
Update: I removed "async" key but it wasn't a problem. The problem is now that everything doesn't work. I am still for reverting any changes related with it. Working point for me is patch 2.

goldim:
I am debugging right now. Maybe the issue is related with mixin changes.

patryk-m-malinowski:
Reverting all the changes it going to be difficult and risky because they were made over multiple commits and it's gonna introduce a lot of merge conflicts. I think it would be better if we leave it as it is and then change things which are absolutely necessary. I have created a [discussion](https://github.com/qooxdoo/qooxdoo/discussions/10854) which summarises all my breaking changes which you should be aware of and where we should discuss those changes. I am open to any suggestions to change what I have done.

goldim:
@patryk-m-malinowski I try to find out what is wrong. This is definitely some type of BC. It happens by another reasons not because of removal async: true. I used it once and it was no reason to use. I just removed. Maybe bc of mixins changes but I don't remember I tested on my applications. Maybe because of global events. I try give feed back near soon.

goldim:
@patryk-m-malinowski Maybe you help me. I had the next code which works for beta 2 package.
```javascript
        loaded: {
            init: false,
            event: "changeSoundOn",
            check: "Boolean",
            apply: "_applyLoaded"
        },
        
        members: {
                _applyLoaded() {
                      this.getView().load();
                },

        }
```
There is no anywhere `this.getView().load` but it is called and with an error that there is no method load for result of getView.
I removed it because it is not used anywhere butI guess somewhere the same error happens. What changed?
Don't look at event name. It is also a mistake.

goldim:
@patryk-m-malinowski @hkollmann I spotted error and fixex it. It is related with `init` keyward triggers `apply`. I didn't check for value and false value gave wrong sequance of actions. I thought it was introduced with new property system but this is my error has not been appealed till this moment. I wonder why.

patryk-m-malinowski:
I have enabled the autoapply behaviour by default because that new behaviour is encouraged and the old behaviour is deprecated. You can either disable that behaviour via the environment settings or by setting autoApply to false in the property definition. My bad for not letting you know.

goldim (approved):

cboulanger (approved):
Can't really judge since I have nothing to test it on, so I am trusting those who can.

johnspackman (changes requested):
I agree with @oetiker's concerns [in Review: Async Property Changes](https://github.com/qooxdoo/qooxdoo/discussions/10854#discussioncomment-16420813) that _"edge cases surface → add more concepts to handle them (safe getters, isInitializing, applyReturnedPromiseWarning)"_ and I don't think that it is appropriate to add a "safe" getter for psuedo properties

The issue there is a pseudo property is just a property - it is one which has been implied due to historical code emulating properties, and also other historical code infering behaviour from those emulations.  We cannot simply define new ways of working, and more importantly pseudo-properties have to work like normal properties otherwise they are not properties.

It is 100% the intention that this is backwards compatible; although this PR refers to solving a problem, I cannot see what that problem is - I think that it is on the chat somewhere?  I'll try and find it and add it here.

johnspackman:
I've had a proper catch up with @patryk-m-malinowski and I think we tracked down what is really going on here - the underlying issue is that "pseudo properties" is something which has been very loosely defined (if at all) where code can be written that makes a class act as though it supports a given property.  

However, this is very loose and in general it means that those properties are not properly implemented - this lead Qooxdoo core (most notably qx.data.SingleValueBinding) to infer properties, and really it means that a subset of functionality can be relied on, with edge cases being ignored or not working.

The problem that this PR was trying to solve was actually a bug in the new implementation, but only because the new implementation had inferred something that the prevous Qooxdoo v7 (and earlier) would have actually ignored, even though it looked important at first glance.

In this code:
```
const Test = qx.Class.define("com.ieatsquirrels.Test", {
    extend: qx.core.Object,
    properties: {
      target: {
        check: "String",
        apply: "_applyTarget"
      }
    },
    events: {
      changeName: "qx.event.type.Data"
    },
    members: {
      _applyTarget(value) {
        console.log("in _applyTarget: value=" + value);
      },
      async getNameAsync() {
        console.log("in getNameAsync");
        return "Patryk";
      },
      getName() {
        console.log("in getName");
        return this._name;
      },
      setName(value) {
        console.log("in setName");
        if (this._name !== value) {
          let oldValue = this._value;
          this._name = value;
          this.fireDataEvent("changeName", value, oldValue);
        }
      }
    }
  });

let test = new Test();
test.bind("name", test, "target");
test.setName("hello");
```

In the above example, the property `name` is a pseudo property - it is not defined in the `properties` section, but it exists and `qx.data.SingleValueBinding` will bind to it.  But look again at `getName` and `getNameAsync` - they return different things.  Here's the console output to prove it:

```
839517 qx.application.Standalone[411-0]: Starting application 'Hello World (modified)' ...
VM2216:24 in getName
VM2216:24 in getName
VM2216:28 in setName
VM2216:17 in _applyTarget: value=hello
index.js:105 839521 qx.application.Standalone[411-0]: Successfully started.
```

The fact is that the `getNameAsync` is completely ignored - it is not part of the pseudo property convention (noting of course that the convention is not written down anywhere, except in `SingleValueBinding` implementation code).

The new implementation was looking at `getNameAsync`, but it should not have done so.

This PR was also passing `safe:Boolean` to the pseudo property getter, but that is also a distraction and not appropriate - (a) there is no previous inference for that, (b) pseudo properties are a hack at the best of times so adding functionality is not really helping the situation and (c) pseudo properties are deprecated by the new property system, because the new system will make sure that *all* the functionality of properties are supported by *all* properties

patryk-m-malinowski:
Closing this PR. A better, alternative fix is coming soon.

patryk-m-malinowski:
I've just created https://github.com/qooxdoo/qooxdoo/pull/10857 which is a better fix for the problem.

PR #10856, opened 2026-04-08
Revert "Eliminate global registries for events and bindings"

oetiker:
I think the PR is really cool, but merging like that (with only 3 votes) is against our policies. Lets get this into shape first.

Reverts qooxdoo/qooxdoo#10852

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,15 +3,6 @@
 # 8.0.0-beta.2
 
 ## Breaking changes
-- Bindings (made using class qx.data.SingleValueBinding) are now no longer stored in a global registry,
-because this prevented their source and target objects from being garbage-collected without manually disposing them first.
-- Removed `qx.data.SingleValueBinding.getAllBindings` as a consequence of the above,
-and also because there is no practical use case for it.
-- Event listener callbacks are now stored in their target objects themselves instead of a global registry,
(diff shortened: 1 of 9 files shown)


goldim (approved):

cboulanger:
We should update the automated merge rules. I see that I could merge this revert without respecting the policy.

cboulanger (approved):

goldim:
@cboulanger Agree but only in PRs. Possibility to return merge politics (e.g. emergency merging) for all has to exist in settings of repository because of a bus factor.

hkollmann:
We should reopen #10856 after meeging

hkollmann (approved):

patryk-m-malinowski (approved):

johnspackman (approved):

hkollmann:
@patryk-m-malinowski : Could you reopen https://github.com/qooxdoo/qooxdoo/pull/10852 please?

patryk-m-malinowski:
I might reopen that PR but later. My reason for creating it in the first place was that I was writing some additions for the QX compiler which made it support Svelte files and convert them into Qooxdoo classes and this PR made things a lot easier for me. However, there are now more important things for the business which I need to get on with so I will create this PR again with addressing others' concerns in the later future. It's also better if we don't distract ourselves for now while we discuss the changes to async properties.

PR #10862, opened 2026-05-06
feat: inherit statics via prototype chain in qx.Class.define

hkollmann:
## Problem

Subclasses defined with `qx.Class.define` did not inherit static methods and properties from their superclass. A call like `Sub.staticMethod()` would fail with `is not a function` even though the static existed on the parent. Callers had to work around this with `.call(Sub, ...)` syntax.

## Root Cause

`qx.Class._extend` correctly sets up the **instance** prototype chain via `qx.Bootstrap.extendClass`, but never calls `Object.setPrototypeOf(subclass, superclass)` for the constructor function itself. Without that link the JS engine cannot traverse to the superclass when looking up a static on the constructor.

### `source/class/qx/Class.js`

**`_extend` (line ~710)** — add after `qx.Bootstrap.extendClass(...)`:
```js
if (superclass !== Object) {
  Object.setPrototypeOf(subclass, superclass);
}
```
Guard `superclass !== Object` leaves root classes (no `extend`, or `extend: Object`) untouched.

**`addEvents`, `addMixin`, `addInterface`, `_attachAnno`** — switch prototype-traversing reads of `clazz.$$events`, `clazz.$$includes`, `clazz.$$implements`, `clazz.$$annotations` to `Object.prototype.hasOwnProperty.call(clazz, "$$...")`.

Without this second fix the new prototype chain caused those methods to find — and mutate — the **parent's** metadata objects instead of initialising a fresh one on the subclass.

### `source/class/qx/test/Class.js`

New test `testStaticInheritance` verifies:
- inherited static is callable on the subclass
- `this` inside the inherited static refers to the **calling** class, not the defining class
- an own static on the subclass shadows the inherited one without affecting the base class

## Behaviour after this change

```js
qx.Class.define("my.Base", {
  statics: { create() { return new this(); } }
});
qx.Class.define("my.Sub", { extend: my.Base });

my.Sub.create(); // works — returns a my.Sub instance
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Upcoming
 
+## Breaking changes
+- Subclasses now inherit static methods and properties from their superclass via
+  the constructor prototype chain, matching native JavaScript `class extends`
+  semantics: reading `Sub.STATIC` resolves to the superclass value, assigning
+  `Sub.STATIC = ...` shadows it on the subclass, and mutations stay isolated.
+  Previously statics were strictly per-class and not visible on subclasses.
+
 # 8.0.0-beta.2
 
 ## Breaking changes
--- a/source/class/qx/Class.js
+++ b/source/class/qx/Class.js
@@ -618,7 +618,7 @@ qx.Bootstrap.define("qx.Class", {
 
           // Call any mixins' constructors and those mixins'
           // dependency mixins' constructors
-          if (subclass.constructor.$$flatIncludes) {
+          if (Object.prototype.hasOwnProperty.call(subclass.constructor, "$$flatIncludes")) {
             subclass.constructor.$$flatIncludes.forEach(mixin => {
               if (mixin.$$constructor) {
                 mixin.$$constructor.apply(this, args);
@@ -1268,7 +1268,7 @@ qx.Bootstrap.define("qx.Class", {
           break;
         }
 
-        var interfaces = superclass.$$implements;
+        var interfaces = Object.prototype.hasOwnProperty.call(superclass, "$$implements") ? superclass.$$implements : null;
         if (interfaces) {
           for (let i = 0; i < interfaces.length; i++) {
             qx.Interface.assert(clazz, interfaces[i], true);
@@ -1490,7 +1490,7 @@ qx.Bootstrap.define("qx.Class", {
      */
     getByMixin(clazz, mixin) {
       while (clazz) {
-        if (clazz.$$includes) {
+        if (Object.prototype.hasOwnProperty.call(clazz, "$$includes")) {
           let list = clazz.$$flatIncludes;
 
           for (let i = 0, l = list.length; i < l; i++) {
@@ -1539,7 +1539,7 @@ qx.Bootstrap.define("qx.Class", {
      *   Whether the class includes the mixin directly.
      */
     hasOwnInterface(clazz, iface) {
-      return clazz.$$implements && clazz.$$implements.includes(iface);
+      return Object.prototype.hasOwnProperty.call(clazz, "$$implements") && clazz.$$implements.includes(iface);
     },
 
     /**
@@ -1573,7 +1573,7 @@ qx.Bootstrap.define("qx.Class", {
       let list = [];
 
       while (clazz) {
-        if (clazz.$$implements) {
+        if (Object.prototype.hasOwnProperty.call(clazz, "$$implements")) {
           list.push.apply(list, clazz.$$flatImplements);
         }
 
--- a/source/class/qx/Class.js
+++ b/source/class/qx/Class.js
@@ -709,6 +709,10 @@ qx.Bootstrap.define("qx.Class", {
 
       qx.Bootstrap.extendClass(subclass, config.construct, superclass, classname);
 
+      if (superclass !== Object) {
+        Object.setPrototypeOf(subclass, superclass);
+      }
+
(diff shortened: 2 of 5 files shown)


derrell (changes requested):
This changes 20 years of qooxdoo historical expectations. It probably warrants some discussion of benefits vs (if any) problems it may cause, rather than simply a code review.

goldim:
@hkollmann Tests can't stop. Should we stop them?

cboulanger:
I remember that I ran into the current behavior, finding it unexpected, unintuitive and annoying. FWIW, here's a claude-generated breakdown about how this is handled in different languages:

| Language      | Subclass sees static?   | Per-subclass copy?   | Mutation isolated? |
|---------------|-------------------------|----------------------|--------------------|
| Python        | ✅ Yes (MRO)             | On write (shadowing) | ✅ Yes              |
| JavaScript/TS | ✅ Yes (prototype chain) | On write (shadowing) | ✅ Yes              |
| Ruby (@@)     | ✅ Yes                   | ❌ No (shared)        | ❌ No               |
| C#            | ✅ Yes (alias)           | ❌ No                 | N/A                |
| Java          | ✅ Yes (alias)           | ❌ No                 | N/A                |
| C++           | ✅ Yes (alias)           | ❌ No                 | N/A                |
| PHP (default) | ✅ Yes                   | ❌ No (shared)        | ❌ No               |
| Kotlin        | ❌ No                    | N/A                  | N/A                |

The implementation differs (inheritance vs. visibility) but no other language except Kotlin shares the non-visibility of static properties of parent classes. Since we are breaking things anyways, this seems a sensible change if we flag it in the changelog.

hkollmann:
@derrell: we can discuss it here.

cboulanger:
Maybe also interesting excerpt from Claude's answer:

### What's the Norm?

**Visibility: inheritance is near-universal.** Almost every language lets you write `Child.staticProp` and get the value — the question is *what's happening underneath*.

**True per-subclass semantics (copy-on-write / shadowing) is the minority**, with Python and JavaScript being the notable examples. Most compiled languages (Java, C++, C#) just have one storage location and let you reference it via the subclass name as a convenience.

The conceptual split is really **"class objects as first-class objects with their own inheritance chain"** (Python, JS) vs. **"static = global-scoped-to-a-namespace"** (Java, C++, C#), where the subclass name is just an alternate path to the same thing.

patryk-m-malinowski (changes requested):
I am getting an exception in a basic Qooxdoo browser app caused by your changes, during the class definition stage before the main method starts running. Just create a clean Qooxdoo app using `qx create myapp --type desktop --noninteractive --out=.`: 

Image

The reason is that `qx.ui.decoration.Decorator` implements that interface, but its superclass `qx.ui.decoration.Abstract` also implements that interface as well. Because $$implements is now inherited, Qooxdoo thinks that the interface is listed in the class twice.

Also, I think that this will cause subclasses to inject data into their superclass' $$ properties. For instance, if a subclass defines some events, those events will be injected into its superclass' `$$events` so the superclass' `$$events` will have the events from its subclasses as well.

Prior to your changes, all the $$ properties on class were relevant to that class only (not superclasses), but following your changes they can now leak from the superclass, which I guess is why you added those `Object.prototype.hasOwnProperty` checks? Your change could break other people's existing code if it references the $$ properties on user-defined classes.

I think a safer approach would be to explicitly set the $$ properties to null when a class is defined so they don't leek though from the superclass and only define them if the class defines it own events, includes, etc.

derrell:
I favor qooxdoo's behavior tracking native JavaScript's behavior. From the table that Christian presented:

> Language | Subclass sees static? | Per-subclass copy? | Mutation isolated?
> JavaScript/TS	✅ Yes (prototype chain)	On write (shadowing)	✅ Yes

JavaScript does map statics to subclasses. It isolates mutations, though, through write shadowing of the prototype chain. Can we mimic that behavior, or would we simply map `my.SubClass.StaticVariable = my.BaseClass.StaticVariable`? Ideally, the qooxdoo behavior (modified) would exactly match the native JavaScript behavior.

oetiker:
@hkollmann can you address @patryk-m-malinowski s input?

hkollmann:
@derrell To your question — the PR mimics native JavaScript behavior exactly, it does *not* copy `SubClass.X = BaseClass.X`. `Object.setPrototypeOf(subclass, superclass)` is the same mechanism native `class Sub extends Base` uses (`Object.getPrototypeOf(Sub) === Base`):

| Action | Behavior | Native JS? |
|---|---|---|
| Read `Sub.STATIC` | resolves to `Base.STATIC` via the chain | identical |
| Write `Sub.STATIC = x` | creates an own property on `Sub`, shadows `Base.STATIC` | identical (write shadowing) |
| Read `Base.STATIC` afterwards | unchanged — mutation isolated | identical |

The one caveat is also native: for object/array statics, shadowing only isolates the *assignment* `Sub.CONFIG = ...`, not the *mutation* `Sub.CONFIG.foo = 1` (which hits the shared object). qooxdoo does not deviate from native JS here.

derrell:
Thanks @hkollmann.

derrell (approved):

patryk-m-malinowski (approved):

cboulanger (approved):

oetiker (approved):

goldim (approved):

PR #10867, opened 2026-05-27
Static inheritance fixes

patryk-m-malinowski:
This PR fixes some bugs introduced by the recent PR (https://github.com/qooxdoo/qooxdoo/pull/10862) which introduced inheritance of statics in classes.

- Added environment setting "qx.Class.staticInheritance" which turns on/off this behaviour. This is useful for cases where the JavaScript engine does not support Object.setPrototypeOf, such as my Rhino, or when we simply don't want this breaking behaviour to occur.
- Fixed a bug with QxObjects by adding the `subclass.$$objects = null;`
- Fixed a bug with annotations by changing the line to `tmp.$$annotations != null`

### DIFF
--- a/source/class/qx/Annotation.js
+++ b/source/class/qx/Annotation.js
@@ -116,7 +116,7 @@ qx.Bootstrap.define("qx.Annotation", {
     __getAnnos(clazz, name, group, annoClass) {
       var result = [];
       for (var tmp = clazz; tmp; tmp = tmp.superclass) {
-        if (tmp.$$annotations !== undefined) {
+        if (tmp.$$annotations != null) {
           var annos = group ? tmp.$$annotations[group] : tmp.$$annotations;
           var src = annos && annos.hasOwnProperty(name) ? annos[name] : null;
           if (src) {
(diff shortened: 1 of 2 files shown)


goldim (approved):

cboulanger (approved):

hkollmann (changes requested):
- Should we test if setPrototypeOf exists?
- the enviroment variable should disable the tests for staticinheritance

patryk-m-malinowski:
> * Should we test if setPrototypeOf exists?

But this will mean that the code will have different behaviour depending on the JavaScript engine, and the user might not be aware of this. I think it's better if the user explicitly enables/disables the behaviour.

> the enviroment variable should disable the tests for staticinheritance

You mean have something like this in the unit tests?
```
if (!qx.core.Environment.get("qx.Class.staticInheritance")) {
  this.skip();
}
```
But is anyone going to be changing the environment settings for the unit tests?

hkollmann (approved):

oetiker (approved):

hkollmann:
@johnspackman: Could you have a look at this please?

johnspackman (approved):

PR #10873, opened 2026-06-18
fix(MPlacement): stop liveupdate placement loop when target is disposed

rad-pat:
### Problem

`qx.ui.core.MPlacement#placeToWidget(target, true)` (the `liveupdate` variant) registers a listener on the global `qx.event.Idle` `interval` event that re-places the widget against `target` on every tick. That listener is only released when the **placing** widget fires `disappear` or is destructed — never when the **target** is disposed.

Popups/menus attach to the application root rather than as children of the widget they point at. So if the target's containing window/tab/panel is closed while the popup is still open, the target gets disposed but the popup (and its idle listener) lives on. The next idle tick calls `target.getContentLocation()`, whose `getContentElement()` is now `null`, throwing:

```
TypeError: Cannot read properties of null (reading 'getDomElement')
    at getContentLocation (qx/ui/core/Widget.js)
    at placeToWidget (qx/ui/core/MPlacement.js)
    at <idle interval listener>
```

Because `qx.event.Idle` keeps firing, this repeats on every tick — we saw hundreds of identical errors logged from a single leaked popup.

### Fix

Bail out of `placeToWidget` and release the idle listener (`__cleanupFromLastPlaceToWidgetLiveUpdate`) when the target is disposed.

### Tests

Two regression tests added to `qx.test.ui.core.Placement`:

- `testPlaceToDisposedWidgetReturnsFalse` — `placeToWidget` on a disposed target returns `false` instead of throwing.
- `testLiveUpdateStopsWhenTargetDisposed` — after a live-update placement whose target is then disposed, firing the `qx.event.Idle` `interval` no longer throws.

Verified that both tests **fail without the fix** with the exact production error (`Cannot read properties of null (reading 'getDomElement')` at `MPlacement.js:350`) and pass with it; full framework suite green (`chromium: 3325 ok, 0 not ok`).

### DIFF
--- a/source/class/qx/test/ui/core/Placement.js
+++ b/source/class/qx/test/ui/core/Placement.js
@@ -173,6 +173,54 @@ qx.Class.define("qx.test.ui.core.Placement", {
       var w = new qx.ui.menu.Menu();
       this.__testAlwaysVisibleElementBelow(w);
       w.destroy();
+    },
+
+    testPlaceToDisposedWidgetReturnsFalse() {
+      var target = new qx.ui.core.Widget();
+      this.getRoot().add(target, { left: 10, top: 10 });
+      this.flush();
+
+      target.destroy();
+      this.flush();
+
+      // getContentLocation() on a disposed target would throw on its null
+      // content element; placeToWidget must bail and report failure instead.
+      var popup = new qx.ui.popup.Popup();
+      this.assertFalse(popup.placeToWidget(target));
+      popup.destroy();
+    },
+
+    testLiveUpdateStopsWhenTargetDisposed() {
+      var target = new qx.ui.core.Widget();
+      this.getRoot().add(target, { left: 10, top: 10 });
+      this.flush();
+
+      var idle = qx.event.Idle.getInstance();
+      var mgr = qx.event.Registration.getManager(idle);
+      var intervalListeners = function () {
+        return (mgr.getListeners(idle, "interval", false) || []).length;
+      };
+      var baseline = intervalListeners();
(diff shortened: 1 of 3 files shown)


johnspackman (approved):

cboulanger (approved):

source/class/qx/test/ui/core/Placement.js
```diff
@@ -173,6 +173,43 @@ qx.Class.define("qx.test.ui.core.Placement", {
...
+      this.flush();
+
+      // getContentLocation() on a disposed target would throw on its null
+      // content element; placeToWidget must bail and report failure instead.
+      var popup = new qx.ui.popup.Popup();
+      this.assertFalse(popup.placeToWidget(target));
+      popup.destroy();
+    },
+
+    testLiveUpdateStopsWhenTargetDisposed() {
```
goldim, line 193: any check/assert in this test?

rad-pat: Good catch — the test relied on an implicit "didn't throw" pass. I've amended the commit to add an explicit assertion: it now snapshots the `qx.event.Idle` "interval" listener count, asserts it goes baseline → +1 when the liveupdate placement subscribes, and asserts it drops back to baseline after the post-disposal tick releases the leaked listener.

derrell (approved):

hkollmann (approved):

hkollmann:
@goldim: Is it ok for you know?

goldim (approved):

PR #10877, opened 2026-08-07
fix(treevirtual): tolerate a row index the model no longer has

rad-pat:
Follow-up to #10874, at the row lookups that fix didn't cover.

`SimpleTreeDataModel.getNode(rowIndex)` throws on an out-of-range index **by design**:

```js
getNode(rowIndex) {
  if (rowIndex < 0 || rowIndex >= this._rowArr.length) {
    throw new Error("this._rowArr row (" + rowIndex + ") out of bounds: …");
  }
```

Three places hand it an index they never bounds check.

## The focused row, which is reachable by construction

Collapsing a branch rebuilds the row array shorter, but the focus doesn't move with it. `SimpleTreeDataModel.__render()` fires `dataChanged` with only `firstRow`/`lastRow` — no `removeStart`/`removeCount` — so `qx.ui.table.Table._updateTableData`'s

```js
// remove focus if the focused row has been removed
if (this.__focusedRow >= removeStart && this.__focusedRow < removeStart + removeCount) {
  this.setFocusedCell();
}
```

never runs for a tree, though it correctly protects a flat table (`qx.test.ui.table.Table#testFocusAfterRemove`). `setState()`'s `bOpened` handling resets the *selection model* but not the focused row. So after any collapse, `getFocusedRow()` can point past the end — and both consumers throw:

- **`SelectionManager.handleButtonClick()`** already meant to handle it. It bails on `if (!node) { return false; }`, which has never been reachable, because `getNode()` throws rather than returning nothing.
- **`TreeVirtual._onKeyDown()`** dereferences the node in five places (Enter, and the Shift/Ctrl arrow pairs), so any of those keys after a collapse throws out of the key handler.

Both now resolve through one `_getFocusedNode()` that answers null when the row is gone, and each site treats a missing node the way it already treats a node that doesn't qualify.

## `_calculateSelectedNodes()`

Same shape, for selected ranges. This is where it was first reported from production (a tap, on mobile Safari). The state is representable — the selection model isn't validated against the row count — but I want to be straight that **the exact route that report took is not established**: `qx.ui.table.pane.Scroller._getRowForPagePos` clamps against the live row count, so a merely stale row is not tappable, and the in-tap collapse-then-select path needs `openCloseClickSelectsRow`, which defaults false. It's guarded on the same principle as #10874's siblings — a selection pointing at a row that no longer exists selects nothing — rather than on a reproduction.

## Test

`qx.test.ui.treevirtual.TreeVirtual` covers the focused-row case end to end: it asserts the focus really is left behind by a collapse (the mechanism above), then that the lookup answers null instead of throwing, plus the selection helper's skip and a live row still resolving. Remove either guard and they fail with `this._rowArr row (2) out of bounds`.

The bound is `getRowCount()`, which is `this._rowArr.length` — exactly what `getNode()` checks against. It's read per iteration in `_calculateSelectedNodes` rather than hoisted, since `setState()` re-enters through the selection model and a `changeSelection` listener may reload the tree.

### DIFF
--- a/source/class/qx/test/ui/treevirtual/TreeVirtual.js
+++ b/source/class/qx/test/ui/treevirtual/TreeVirtual.js
@@ -1,0 +1,174 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2026 qooxdoo contributors
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+************************************************************************ */
+
+/**
+ * Tests for the row lookups that have to tolerate a row index outliving the row
+ * it points at. Collapsing a branch rebuilds the row array shorter, and
+ * SimpleTreeDataModel.getNode() throws on an index past the end rather than
+ * returning nothing.
+ *
+ * The focused row is the case that is reachable by construction: the tree's
+ * dataChanged event carries no removeStart/removeCount, so Table's "remove
+ * focus if the focused row has been removed" branch never runs for a tree.
+ */
+qx.Class.define("qx.test.ui.treevirtual.TreeVirtual", {
+  extend: qx.dev.unit.TestCase,
+
+  members: {
+    __tree: null,
+    __root: null,
+
+    setUp() {
+      this.__tree = new qx.ui.treevirtual.TreeVirtual(["Tree"]);
+      var dataModel = this.__tree.getDataModel();
+      // Node 0 is the model's own hidden root, so the visible rows start here.
+      this.__root = dataModel.addBranch(null, "root", true);
+      dataModel.addLeaf(this.__root, "one");
+      dataModel.addLeaf(this.__root, "two");
+      dataModel.setData();
+    },
+
+    tearDown() {
+      this.__tree.destroy();
+      this.__tree = null;
+    },
+
+    testFocusedNodeIsNullOnceItsRowIsGone() {
+      this.__tree.setFocusedCell(0, 2, false);
+      this.assertNotNull(
+        this.__tree._getFocusedNode(),
+        "the leaf resolves while it exists"
+      );
+
+      this.__shrinkToRoot();
+
+      this.assertEquals(
+        2,
+        this.__tree.getFocusedRow(),
+        "a collapse leaves the focus where it was"
+      );
+      this.assertNull(
+        this.__tree._getFocusedNode(),
+        "so the focused row no longer resolves to a node"
+      );
+    },
+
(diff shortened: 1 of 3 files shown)


cboulanger (approved):

derrell (approved):

rad-pat:
@cboulanger @derrell — my apologies. This PR was closed yesterday **in error**: I closed it by mistake while tidying up after merging the corresponding change in our own repo, one second after that merge. It was never merged upstream, and nothing here was intended to be withdrawn. It is reopened and the branch is intact.

Sorry in particular for spending your review twice. You had both already approved it (@cboulanger on the 7th, @derrell yesterday afternoon), and I've since force-pushed an update, so the approvals no longer apply to what's on the branch. That update is genuine work rather than churn, and it's worth being explicit about what changed since you looked, so you can decide how much of it needs re-reading:

**1. Dead locals my own change created.** In `_onKeyDown`'s two SHIFT branches, `focusedRow` and `treeCol` were left assigned but unread once `_getFocusedNode()` stopped taking an argument — while the same commit correctly deleted that line in the Enter branch and kept it in the Ctrl pair, where `setFocusedCell()` still reads it. Inconsistent with itself, and neither lint nor prettier catches an unused local. Removed.

**2. A null-focus hole in the guard itself.** `_getFocusedNode()` checked `focusedRow >= 0`, and `null >= 0` is `true` in JS — so an unset focus (which `Table.__focusedRow` starts as) went through to `getNode(null)` and died on a missing row-array entry rather than the bounds error. Now `typeof focusedRow === "number" && …`, the same shape as `moveFocusedCell()`'s existing BUG #4676 guard.

**3. Test coverage went from one guard to all six.** Previously a single test pinned Ctrl+Left. Now every key combination that resolves the focused node is driven (Enter, both Ctrl arrows, both Shift arrows) plus the `SelectionManager` path, and I verified each guard individually: deleting any one of the six turns a test red. There's also a comment on the focus assertion noting it records current behaviour, not a requirement — the guard stops the throw, it does not reconcile the focus, so whoever fixes reconciliation later will know that red test is expected.

**4. Non-ASCII removed.** My comments had used em dashes; no other file under `source/class/qx/ui/` does (0 of 511), so the diff is now pure ASCII to match the codebase.

The substance you approved is unchanged: the guards, where they sit, and the honest scoping of what is and isn't established about the original production report. Happy to split (1)–(4) into a separate follow-up PR if you'd rather re-review the original diff as you first saw it — just say so and I'll rebase it out.

oetiker (approved):

derrell (approved):

johnspackman (approved):

hkollmann (approved):

PR #10880, opened 2026-09-02
fix(html.Image): keep a clipped image's offset when its padding changes

rad-pat:
## Summary

`qx.html.Image#setPadding` maps padding onto `background-position`:

```js
setPadding(paddingLeft, paddingTop) {
  this.__paddingLeft = paddingLeft;
  this.__paddingTop = paddingTop;

  if (this.getNodeName() == "div") {
    this.setStyle("backgroundPosition", paddingLeft + "px " + paddingTop + "px");
  }
}
```

That is right for an unclipped image, but a **clipped (combined) image is painted by scrolling the combined image with `background-position`** so the wanted glyph lands in the box — see the offset handling in `qx.bom.element.Decoration.__processRepeats`. Writing the padding straight into that property throws the offset away, so the widget renders whichever glyph happens to sit at `[0, 0]` of the combined image.

## Reproduction

Any widget whose icon comes from a clipped resource, as soon as its padding is applied — in practice from an appearance during a theme change. The included test covers it directly.

It has survived because something else usually repaints the icon first: property jobs run before style jobs in `qx.html.Element#_syncData`, so an image that happens to be re-sourced in the same flush overwrites the bad value on its way past, while everything else keeps the wrong glyph. That makes it look like only *some* icons are affected, which is a good way to spend a while chasing the wrong thing.

We ship an icon set as a single sprite strip, so this was very visible: comparing every icon's rendered `background-position` against its resource offset plus padding, one light→dark→light toggle left **16 icons wrong in dark and 34 back in light**, versus **0 and 0** with this change, repeatably.

## Fix

Re-apply the source instead of computing the position here. `qx.bom.element.Decoration` already folds the padding into the offset (and adds the matching clip) when it resolves a combined resource, and produces a plain padding offset for an unclipped one, so both cases stay correct and the unclipped path is unchanged in effect.

Adds `qx.test.ui.basic.Image#testPaddingKeepsTheClippedOffset`, which uses an existing combined asset (`qx/decoration/Classic/form/checkbox-checked-disabled.png`, offset `-336, 0`) so it needs no new fixture. It fails before the change and passes after.


hkollmann (approved):

cboulanger (approved):

source/class/qx/html/Image.js
```diff
@@ -70,10 +70,22 @@ qx.Class.define("qx.html.Image", {
...
+        var source = this._getProperty("source");
+        if (source) {
+          // A clipped image paints its glyph by scrolling the combined image
+          // with background-position, so writing the padding straight into
+          // that property would drop the offset and show whichever glyph sits
+          // at [0, 0]. Re-apply the source instead - qx.bom.element.Decoration
+          // folds the padding into the offset (and the clip) when it resolves
+          // a combined resource, and is a plain offset for an unclipped one.
+          this.setSource(null);
+          this.setSource(source);
```
derrell, line 88: I suspect this full reset of the source is much slower than simply setting the `backgroundPosition` style -- probably noticeably slower if a large number of images are having their padding adjusted. Have you by chance evaluated the speed difference? (I'm not arguing for keeping it as it is, as clearly there's a bug if the wrong image gets presented, but I wonder if only the necessary operations could be done rather than the full reset...)

derrell (changes requested):

rad-pat:
Fair question — I hadn't measured it, so I have now.

Headless Chromium, 200 `qx.ui.basic.Image` widgets in the document, timing `setPadding` over all of them plus one `qx.html.Element.flush()`, 20 passes alternating padding 4/0, median of three runs:

| image | direct `background-position` write | re-apply the source |
|---|---|---|
| clipped | 3.33 µs each — but the wrong glyph | 10.50 µs each |
| unclipped | 2.77 µs each | 8.50 µs each |

So the re-apply costs about 7 µs per image on this machine — 1.4 ms for all 200. For scale, a theme switch in the app where we hit this calls `setPadding` on 49 images, so ~0.35 ms once per switch. Real, but not something you'd see.

You're right that the unclipped case is paying for nothing, though, so I've pushed a second commit that only takes the slow path where there is an offset to preserve: gate on `qx.util.ResourceManager#getCombinedFormat`, which is the same thing `qx.bom.element.Decoration.__processRepeats` branches on. Measured, an unclipped image goes back to **2.65 µs**, i.e. the original cost; a clipped one is unchanged.

That also fixes something the blanket version got wrong. For a base64-combined resource `__processRepeats` uses `toDataUri(sourceid)` — the individual image, so offset `[0, 0]` — and ignores the padding entirely, which meant re-applying the source *dropped* the padding. Those now keep the direct write, as before.

I did look at computing the position here instead, which would be the cheapest of all, but the clipped branch also derives a **clip rect** from the padding (and has to remove it again when the padding goes back to zero), so it would mean keeping a second copy of that half of `__processRepeats` in `qx.html.Image` — which is how the original bug got in. Re-applying the source keeps one owner for the rule.

Added `testPaddingOnAnUnclippedImage` to pin the fast path.

patryk-m-malinowski:
Could you please upload a zip file with a simple example project that reproduces this issue? This save the developers' time and helps them understand the issue. Thanks.

rad-pat:
Of course — standalone project here, nothing to install beyond the compiler:

- **Zip:** https://github.com/rad-pat/qooxdoo/archive/refs/heads/repro/html-image-padding.zip
- **Browse:** https://github.com/rad-pat/qooxdoo/tree/repro/html-image-padding

```sh
qx compile
qx serve
```

then open http://localhost:8080/imagepaddingrepro/ and press **Add padding (4px)**.

It is an unmodified `qx create` skeleton with one file changed, and it ships no images of its own. The eight glyphs are `qx/decoration/Classic/form/*`, which the framework already provides pre-combined into `checkbox-radiobutton-combined.png` — so they are clipped resources out of the box, and the setup needs nothing bespoke to be believable.

They are shown magnified 5x next to a reference of `checkbox-checked-pressed`, the glyph that happens to sit at offset `[0, 0]` of that strip. On 7.9.3, adding the padding turns all eight into that reference glyph.

The page doesn't ask you to judge it by eye. For each image it reads `backgroundPosition` back off the DOM node and compares it with the offset `qx.util.ResourceManager` holds for that resource, plus the padding. On **7.9.3**:

```
FAIL - 8 of 8 glyphs lost their offset.
 FAIL checkbox.png                         expected -486px 0px     got 4px 0px
 FAIL checkbox-checked.png                 expected -262px 0px     got 4px 0px
 FAIL checkbox-checked-disabled.png        expected -332px 0px     got 4px 0px
 FAIL checkbox-invalid.png                 expected -346px 0px     got 4px 0px
 FAIL radiobutton.png                      expected -220px 0px     got 4px 0px
 FAIL radiobutton-checked.png              expected -318px 0px     got 4px 0px
 FAIL radiobutton-checked-disabled.png     expected -206px 0px     got 4px 0px
 FAIL radiobutton-invalid.png              expected -276px 0px     got 4px 0px
```

and with this PR's branch added to `compile.json` as `"libraries": ["/path/to/qooxdoo", "."]`:

```
PASS - all 8 glyphs kept their offset.
```

One wrinkle worth flagging, because it can make a repro of this look like a non-repro: the combined image has to be asseted alongside the individual glyphs, i.e. `@asset(qx/decoration/Classic/checkbox-radiobutton-combined.png)` as well as the `form/*` glob. Without it the strip is absent from the resource manager's registry, the glyphs load individually, nothing is clipped, and the bug doesn't appear at all.

oetiker (approved):

johnspackman (approved):

patryk-m-malinowski (changes requested):
Thanks for the zip file, that really helped me understand what's going on. I think it's better if we use the more modern `let` instead of `var` for variables though when appropriate when writing new code. The `var source` in qx.html.Image.setPadding can be changed to `let` as well.

rad-pat:
Glad the example helped, and agreed on `let` — done.

That is `source` and `clipped` in `qx.html.Image#setPadding`, plus the locals in both new tests. I folded it into the two existing commits rather than adding a style commit on top, so each one still reads cleanly on its own and the intermediate commit never introduces a `var` for the next one to change. The surrounding pre-existing `var`s in those two files are untouched.

Prettier is happy, and I re-ran the example project against the branch to confirm it still reports `PASS - all 8 glyphs kept their offset.`

johnspackman:
> Thanks for the zip file, that really helped me understand what's going on. I think it's better if we use the more modern `let` instead of `var` for variables though when appropriate when writing new code. The `var source` in qx.html.Image.setPadding can be changed to `let` as well.

While that is an understandable comment, we do not have any kind of coding standards for this and as a general rule we try and stay away from micro-change requests.

We definitely do not want to get bogged down in debates about "better" ways to write code, because it very quickly becomes a debate about personal preference and distracts from the task being achieved.

patryk-m-malinowski (approved):

derrell (approved):

PR #10881, opened 2026-09-02
fix(Scroller): make cellEditorBlurAction reach the editor, and not steal focus

rad-pat:
`Table#cellEditorBlurAction` is wired to the cell editor's own focus events, in
a way that is racy, silently inert for whole classes of editor, and that takes
the focus off whatever the user just moved to.

## What's wrong today

Measured on `master`, on a mounted one-cell table with
`cellEditorBlurAction: "save"`:

1. **A non-focusable cell editor throws.** `startEditing` calls
   `_cellEditor.focus()` unconditionally, and `Widget#focus` throws
   `"Widget is not focusable!"` in a debug build. A factory is entitled to
   return a widget that takes no focus of its own — one whose UI is a popup it
   manages itself, say — and `getFocusTarget()` exists precisely because such
   widgets are expected.

2. **…and never gets the blur action anyway.** `qx.ui.core.EventHandler`
   resolves focus-event targets through `Widget#getFocusTarget()`, the nearest
   *focusable* ancestor. For a non-focusable editor that is the `Table`, so the
   editor never fires `focusin`, the `focusout` listener is never attached, and
   the property silently does nothing.

3. **The wiring is a race.** The `focusout` listener is attached from a
   `focusin` handler via a 1 ms `qx.event.Timer`. An edit blurred inside that
   window is dropped, and if editing stops inside it the deferred callback
   dereferences a nulled `_cellEditor` and throws from a timer. (That crash is
   what #10876 guards; this supersedes it by removing the deferral.)

4. **The shipped CheckBox editor never commits, at any timing.**
   `celleditor.CheckBox` returns a focusable `Composite` that hands the focus to
   its child checkbox on `focus`. So the editor reports a `focusout` of its own
   the instant the edit starts, and the *child* — not the editor — is what
   reports one when the focus really leaves. `this._cellEditor === e.getTarget()`
   rejects both. Verified: edit, wait 100 ms, blur → `isEditing() === true`,
   model unchanged.

5. **A blur-save steals the focus back.** `flushEditor` calls `table.focus()`
   even when the flush was *caused by* the focus leaving, so clicking from a
   cell editor into another field lands the focus back on the table. Verified:
   after the blur, `FocusHandler.getActiveWidget()` is the `Table` and the
   clicked field does not have the `focused` state.

## The change

- focus the editor only when `isFocusable()`;
- register `focusout` in `startEditing` and drop it in `cancelEditing` — no
  timer, so no race and nothing to null-deref;
- decide on the editor's **subtree**, not the editor widget: the focus leaving
  is a `focusout` from inside the editor to somewhere outside it. Both ends are
  checked, which fixes CheckBox and any other editor that manages its own focus;
- the blur path marks the flush so it does not take the focus back. Enter is
  unchanged and still returns the focus to the table.

That last one is deliberately **not** an argument to `stopEditing`. That method
is widely overridden, and an override declaring no parameters of its own
compiles `this.base(arguments)` / `super.stopEditing()` down to a call that
forwards none — so the argument was dropped before it ever reached
`flushEditor`, and the save went on stealing the focus. I hit exactly that in
our own codebase before spotting it; there is a regression test for that shape
of subclass.

An editor that takes no focus does not participate in the blur action and must
commit its own value. That was already true by accident; it is now stated at the
handler.

## Tests

`qx.test.ui.table.CellEditorLifecycle` is new: it drives every editor the
framework ships — TextField, PasswordField, ComboBox, SelectBox, CheckBox,
Dynamic — through the real lifecycle on a mounted table, under each of the three
blur actions. There was no such coverage: the existing
`qx.test.ui.table.celleditor.*` tests call the factories directly with a
hand-built `cellInfo` and say nothing about behaviour once the `Scroller` is
driving. Against unmodified `master` the `"save"` and `"cancel"` sweeps both
fail; the `"nothing"` sweep passes on both and is a regression guard rather than
a demonstration.

`qx.test.ui.table.Table` gains the unfocusable editor, a factory declining the
cell by returning `null`, the focus outcome of `stopEditing`, the
overridden-`stopEditing` case above, and a focusout carrying no related target.

Not covered, and unchanged by this PR: modal-window cell editors
(`_cellEditor instanceof qx.ui.window.Window`) get no blur handling before or
after, and `qx.ui.treevirtual.celleditor.NodeEditor` is exercised downstream
rather than here.

## Compatibility

No public signature changes. Tables that set `"save"` or `"cancel"` will now see
the action fire in cases where it previously did not — CheckBox and other
composite editors, and blurs inside the first millisecond — and a blur-save no
longer returns the focus to the table. Tables on the default `"nothing"` are
unaffected.

A focusout carrying **no** related target still does not commit, deliberately.
An earlier revision of this PR treated it as the focus leaving, which destroyed
a ComboBox editor as it opened (see the comment below); every real way out of an
edit moves the focus onto something. The consequence is that a window losing
focus entirely — alt-tabbing away — does not commit, and cannot without
reintroducing that.

Supersedes #10876.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AZ7Ey5mhSN4ifkJPeuxeKm

### DIFF
--- a/source/class/qx/test/ui/table/CellEditorLifecycle.js
+++ b/source/class/qx/test/ui/table/CellEditorLifecycle.js
@@ -1,0 +1,238 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2004-2026 1&1 Internet AG, Germany, http://www.1und1.de
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+************************************************************************ */
+
+/**
+ * Drives every cell editor the framework ships through the real editing
+ * lifecycle on a mounted table: start the edit, change the value, then let the
+ * focus leave, under each {@link qx.ui.table.Table#cellEditorBlurAction} in
+ * turn.
+ *
+ * The per-editor tests under qx.test.ui.table.celleditor.* call the factories
+ * directly with a hand-built cellInfo, so they say nothing about how an editor
+ * behaves once qx.ui.table.pane.Scroller is driving it — which is where the
+ * focus handling lives, and where an editor that manages focus of its own (the
+ * CheckBox editor hands focus to its child) parts company with a plain field.
+ *
+ * The focus is moved as soon as the edit starts, deliberately: the blur action
+ * has to hold from the first moment, not once some deferred wiring has caught
+ * up.
+ */
+qx.Class.define("qx.test.ui.table.CellEditorLifecycle", {
+  extend: qx.test.ui.LayoutTestCase,
+
+  members: {
+    __fixtures: null,
+
+    /**
+     * One spec per cell editor the framework ships. `edit` makes a change the
+     * editor would report, and `edited` is what committing it must write.
+     *
+     * @return {Array} the specs.
+     */
+    editorSpecs() {
+      return [
+        {
+          name: "TextField",
+          create: () => new qx.ui.table.celleditor.TextField(),
+          value: "before",
+          edit: editor => editor.setValue("after"),
+          edited: "after"
+        },
+
+        {
+          name: "PasswordField",
+          create: () => new qx.ui.table.celleditor.PasswordField(),
+          value: "before",
+          edit: editor => editor.setValue("after"),
+          edited: "after"
+        },
+
+        {
+          name: "ComboBox",
+          create() {
+            var factory = new qx.ui.table.celleditor.ComboBox();
+            factory.setListData(["before", "after"]);
+            return factory;
+          },
+          value: "before",
+          edit: editor => editor.setValue("after"),
+          edited: "after"
+        },
+
+        {
+          name: "SelectBox",
+          create() {
+            var factory = new qx.ui.table.celleditor.SelectBox();
+            factory.setListData([
+              ["Before", null, "before"],
+              ["After", null, "after"]
+            ]);
+
+            return factory;
+          },
+          value: "before",
+          edit(editor) {
+            editor.setSelection([
+              editor.getChildrenContainer().findItem("After")
+            ]);
+          },
+          edited: "after"
+        },
+
+        {
+          // A focusable composite that hands the focus straight to its child
+          // checkbox, so the editor reports a focusout of its own the moment
+          // the edit starts, and the child — not the editor — is what reports
+          // one when the focus really leaves. Neither may be read at face value.
+          name: "CheckBox",
+          create: () => new qx.ui.table.celleditor.CheckBox(),
+          value: false,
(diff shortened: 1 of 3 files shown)


derrell (approved):

rad-pat:
Apologies for the churn — I've force-pushed a correction, and I'm sorry to anyone who has already read the previous version. Thankfully there is only the one review so far.

**What was wrong.** A colleague testing the branch downstream found that double-clicking a cell whose editor is a `ComboBox` opened no editor at all, and dropped the focus. They were right, and it was this PR.

Instrumenting a real double-click (Playwright driving actual pointer input, blur action disabled so nothing interferes) shows the ComboBox editor churning focus three times in the six milliseconds it takes to open:

```
+3ms focusin   related=Table    active=DIV
+8ms focusout  related=NULL     active=BODY     <- transient
+8ms focusin   related=NULL     active=DIV      <- focus comes straight back
+9ms focusout  related=NULL     active=BODY     <- transient again
+9ms focusin   related=NULL     active=INPUT    <- settles on the real input
```

The handler acted on that first transient `focusout` and destroyed the editor five milliseconds after creating it — and since the blur path deliberately does not return the focus to the table, the focus was left nowhere.

**The bit I got wrong.** I had characterised the 1 ms `qx.event.Timer` purely as a race to be removed. It was also, incidentally, suppressing an editor's own focus churn while it opens — a protection I removed without recognising it as one. My earlier claim in this PR that acting on a focus loss with no related target was a wanted improvement ("alt-tabbing away now commits") was simply wrong: a null related target also occurs transiently during ordinary pointer interaction.

**The fix.** A `focusout` carrying no related target is no longer treated as the focus leaving the edit. Every real way out of an edit moves the focus onto something — another field, a toolbar button, a window's own close button (which reports the `Window` as the related target, even though the button itself is not focusable). Only the transient churn moves it to nothing.

`qx.test.ui.table.Table#testFocusGoingNowhereDoesNotEndTheEdit` covers it. I mutation-tested that one rather than trusting it: with the guard removed it fails, and so does the real double-click.

The CheckBox subtree fix, the unfocusable-editor fix, the timer removal and the focus-stealing fix are all unchanged. The full editor × blur-action matrix in `CellEditorLifecycle` still passes, as do the pre-existing table tests.

**Worth recording for reviewers:** this class of bug — "the gesture now does nothing" — was invisible to the unit suite until I added a test aimed straight at it. Nothing in `qx.test.*` drives real pointer input, so it took someone exercising the branch with a real double-click to find it. Any future change to the focus handling here deserves the same treatment.

cboulanger (approved):

johnspackman (approved):

derrell (approved):

oetiker (approved):

derrell (approved):

hkollmann (approved):

PR #10884, opened 2026-09-04
fix(table): dispose a row renderer when it is replaced

### LINKED ISSUE #10883: qx.ui.table.Table leaks the row renderer it allocates when one is set in its place (every TreeVirtual leaks one)
`qx.ui.table.Table`'s constructor allocates a row renderer:

```js
// Table.js:178
this.setDataRowRenderer(new qx.ui.table.rowrenderer.Default(this));
```

`dataRowRenderer` is an ordinary property with no `apply` and no ownership semantics, so replacing it simply drops the previous value. `Table`'s destructor disposes only whichever renderer is *current*:

```js
// Table.js:2180
var dataRowRenderer = this.getDataRowRenderer();
if (dataRowRenderer) {
  dataRowRenderer.dispose();
}
```

So every replacement orphans a renderer for the life of the page.

## Where the framework does this to itself

`qx.ui.treevirtual.TreeVirtual` replaces it during construction:

```js
// TreeVirtual.js:200
this.setDataRowRenderer(custom.dataRowRenderer);
```

That is the only in-framework replacement besides `Table`'s own, so **every `TreeVirtual` ever constructed leaks one `qx.ui.table.rowrenderer.Default`**, even when the tree is disposed correctly.

Application code that swaps in its own renderer hits the same thing — which is how we noticed; we had to add a "dispose the one you displaced" helper on our own `Table` subclasses.

## Why it costs more than memory

`qx.ui.table.rowrenderer.Default`'s constructor registers a `changeTheme` listener on the `qx.theme.manager.Meta` singleton and removes it in its destructor (guarded by `qx.dyntheme`). A leaked renderer therefore stays on the list that `qx.theme.manager.Meta.setTheme()` walks, so **live theme switching gets slower in proportion to every table/tree ever created**, not to what is on screen.

Measured in a ~750-class application, at the end of a unit-test run: **468 orphaned `qx.ui.table.rowrenderer.Default`**, against only 106 tables still alive.

This is the same shape as #9488 (`qx.ui.form.List._applyOrientation` leaking the previous layout instance) and the general point raised in #3295 (properties do not dispose an overwritten value).

## Question before proposing a patch

Two ways to fix it, and I would rather match your preference than guess:

1. **Narrow** — have `TreeVirtual` dispose the renderer it displaces. Fixes the framework's own leak, leaves application replacements to the application.
2. **General** — give `dataRowRenderer` an `apply` that disposes the previous value, so `Table` owns its renderer throughout, not just at destruct. Fixes application code too, but it is a behaviour change for anyone deliberately reusing a renderer instance across tables, and I appreciate that "the caller disposes what it created" has been the stated rule elsewhere (#5351, #9375).

Happy to raise a PR for either. My inclination is (1) as obviously safe, with (2) as a separate discussion.

rad-pat:
Fixes #10883.

## The leak

`qx.ui.table.Table`'s constructor allocates a row renderer:

```js
// Table.js:178
this.setDataRowRenderer(new qx.ui.table.rowrenderer.Default(this));
```

`dataRowRenderer` had no `apply`, and the destructor disposes only whichever renderer is *current* (`Table.js:2180`), so a renderer that was replaced was never disposed at all.

`qx.ui.treevirtual.TreeVirtual` replaces it in every instance (`TreeVirtual.js:200`), so **every `TreeVirtual` ever constructed leaked one `rowrenderer.Default`**, even when the tree itself was disposed correctly.

## Why it costs more than memory

`rowrenderer.Default` registers a `changeTheme` listener on the `qx.theme.manager.Meta` singleton and removes it in its destructor. A leaked one stays on the list `qx.theme.manager.Meta.setTheme()` walks, so live theme switching gets slower in proportion to every tree ever created rather than to what is on screen. In an application with ~750 widget classes, that was 468 orphaned renderers alive at the end of a unit-test run.

## The fix

`_applyDataRowRenderer` disposes the renderer it replaces, so the table owns its row renderer throughout, not only at the end. `TreeVirtual` needs no change, and any other subclass that swaps the default is covered.

This is the model `qx.ui.table.columnmodel.Resize` already uses for `behavior`: a property of the same shape, whose default its constructor allocates, with an `apply` that disposes `old`.

Setting the current renderer again (`setDataRowRenderer(getDataRowRenderer())`) is unaffected: the generated setter returns on `===` before calling `apply`, so the live renderer is kept.

**Behaviour change:** code that keeps using a renderer after replacing it — to put it back later, or to share one between tables — now holds a disposed object. Sharing already broke when a table was destroyed, and nothing in the framework reuses a renderer (`Pane` only reads the current one), but it may merit a release note.

## Tests

The framework had no tests covering row renderers or tree disposal. Added to `qx.test.ui.table.Dispose`:

| Test | master | this PR |
|---|---|---|
| `testReplacedDefaultRowRenderer` | fails | passes |
| `testReplacedSuppliedRowRenderer` | fails | passes |
| `testTreeVirtual` | fails | passes |
| `testSettingTheCurrentRowRendererAgainKeepsIt` | passes | passes |

The first three go through `assertDestroy`, and on master each fails on the leaked `changeTheme` listener. The last one guards the one case where disposing `old` could take out the live renderer. `testTreeVirtual` uses a `Basic` column model, because the default `Resize` model has a leak of its own (#10882).

All 60 `qx.test.ui.table` and 6 `qx.test.ui.treevirtual` tests pass.

## Related

- #10883 — the issue this fixes.
- #10882 — a sibling leak in `qx.ui.table.columnmodel.resizebehavior.Default`, same "dispose what you dropped" shape.
- #9488, #3295 — prior art for this class of fix.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LXEJuDGLkQ365twMtUQ8yD


cboulanger (approved):

johnspackman (approved):

oetiker (approved):

source/class/qx/ui/treevirtual/TreeVirtual.js
```diff
@@ -196,8 +196,16 @@ qx.Class.define("qx.ui.treevirtual.TreeVirtual", {
...
+    // Set the data row renderer, disposing the one qx.ui.table.Table's
+    // constructor allocated. The dataRowRenderer property does not own its
+    // value, and Table's destructor disposes only whichever renderer is
+    // current, so replacing it here would otherwise orphan that instance for
+    // the life of the page.
+    var previousRowRenderer = this.getDataRowRenderer();
     this.setDataRowRenderer(custom.dataRowRenderer);
+    if (previousRowRenderer && previousRowRenderer !== custom.dataRowRenderer) {
+      previousRowRenderer.dispose();
+    }
```
derrell: I wonder if a more general fix may be more appropriate? The more general fix would be adding an `apply` handler to the `dataRowRenderer` property in `Table` that stores the original `Default` renderer and checks on each apply whether the old renderer is that original one, and properly disposes of it. This would correct for extending classes other than TreeVirtual that did something similar; and would also keep the ownership relationship more proper by having `Table`, which instantiates the original renderer, dispose of it, rather than special-casing it in TreeVirtual and ignoring the ownership relationship?

I'm not wedded to this. If you see any downside to what I suggest, I could be easily convinced to approve this PR. I do lean towards the more general fix, however.

derrell (changes requested):

source/class/qx/ui/treevirtual/TreeVirtual.js
rad-pat: Agreed — `Table` should own this rather than `TreeVirtual` special-casing it. I've reworked the PR that way and rebased it onto master.

`_applyDataRowRenderer` now disposes the renderer it replaces, so `TreeVirtual` is back to a plain `setDataRowRenderer`, and any other subclass that swaps the default is covered. That fits how `Table` already behaves, since its destructor disposes whichever renderer is current. It is also what `qx.ui.table.columnmodel.Resize` does for `behavior`: a property of the same shape, whose default its constructor allocates.

One behaviour change worth a release note: code that keeps using a renderer after replacing it — to put it back later, or to share one between tables — now holds a disposed object. Nothing in the framework does that (`Pane` only reads the current renderer).

Setting the current renderer again is safe: the generated setter returns on `===` before calling `apply`. A test pins that, since it's the one case where disposing `old` would take out the live renderer.

Tests are in `qx.test.ui.table.Dispose`; the framework had none covering row renderers or tree disposal. Replacing the default renderer, replacing a supplied one, and constructing a `TreeVirtual` each fail on master through `assertDestroy`, and pass with the change. The tree uses a `Basic` column model because `Resize` has a leak of its own (#10882). All 60 `qx.test.ui.table` and 6 `qx.test.ui.treevirtual` tests pass.

hkollmann (approved):

derrell (approved):

